Issue #523, opened 2016-03-29
20.5.0+ breaks existing setup.py files - UndefinedEnvironmentName


lnielsen:
The latest 20.6.4 PyPI is breaking our [setup.py](https://github.com/inveniosoftware/invenio-records-rest/blob/master/setup.py) files with the following exception:

```
Traceback (most recent call last):
  File "setup.py", line 163, in <module>
    cmdclass={'test': PyTest},
  File "/opt/python/2.7.9/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/opt/python/2.7.9/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/opt/python/2.7.9/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/setuptools/command/test.py", line 152, in run
    self.distribution.fetch_build_eggs(self.distribution.tests_require)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/setuptools/dist.py", line 313, in fetch_build_eggs
    replace_conflicting=True,
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources/__init__.py", line 814, in resolve
    if req.marker and not req.marker.evaluate():
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources/_vendor/packaging/markers.py", line 273, in evaluate
    return _evaluate_markers(self._markers, current_environment)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources/_vendor/packaging/markers.py", line 198, in _evaluate_markers
    lhs_value = _get_env(environment, lhs.value)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pkg_resources/_vendor/packaging/markers.py", line 180, in _get_env
    "{0!r} does not exist in evaluation environment.".format(name)
pkg_resources._vendor.packaging.markers.UndefinedEnvironmentName: 'extra' does not exist in evaluation environment.
```

lnielsen:
Seems to be related to this commit d25e2a585c0e80dc89d72a91f8da0ba8198a4f7c

lnielsen:
Looks pretty serve. Just upgrading setuptools in a virtualenv breaks the code.

[...]

PR #530, opened 2016-04-04
Restore evaluating environment markers in WorkingSet

s-t-e-v-e-n-k:
Correctly deal with parsed requirements that include extras as a
marker inside WorkingSet that are populated from METADATA inside
wheels, like we get from pip>=7.

This partially reverts commit 04d10ff025e1cbef7ec93a2008c930e856045c8a.

Closes: #523

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -791,6 +791,8 @@ class WorkingSet(object):
         # key -> dist
         best = {}
         to_activate = []
+        # Map requirement to the extras that require it
+        extra_req_mapping = {}
 
         # Mapping of requirement to set of distributions that required it;
         # useful for reporting info about conflicts.
@@ -804,9 +806,14 @@ class WorkingSet(object):
                 continue
             # If the req has a marker, evaluate it -- skipping the req if
             # it evaluates to False.
-            # https://github.com/pypa/setuptools/issues/523
-            _issue_523_bypass = True
-            if not _issue_523_bypass and req.marker and not req.marker.evaluate():
+            if req.marker:
+                result = []
+                if req in extra_req_mapping:
+                    for extra in extra_req_mapping[req] or ['']:
+                        result.append(req.marker.evaluate({'extra': extra}))
+                else:
+                    result.append(req.marker.evaluate())
+                if not any(result):
                     continue
             dist = best.get(req.key)
             if dist is None:
@@ -840,6 +847,7 @@ class WorkingSet(object):
             # Register the new requirements needed by req
             for new_requirement in new_requirements:
                 required_by[new_requirement].add(req.project_name)
+                extra_req_mapping[new_requirement] = req.extras
 
             processed[req] = True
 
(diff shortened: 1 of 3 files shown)


pkg_resources/__init__.py
```diff
@@ -840,6 +847,7 @@ def resolve(self, requirements, env=None, installer=None,
             # Register the new requirements needed by req
             for new_requirement in new_requirements:
                 required_by[new_requirement].add(req.project_name)
+                extra_req_mapping[new_requirement] = req.extras
```
rbtcollins: This is buggy vs the sketch I put up in my fork - you're not unioning together all the extras. Consider A and B both depend on C - one on C[a] and one on C[b]. We want to activate C[a,b].

pkg_resources/tests/test_resources.py
```diff
@@ -181,6 +180,52 @@ def test_environment_markers(self):
...
+        ws = WorkingSet([])
+        Foo = Distribution.from_filename("/foo_dir/Foo-1.2.dist-info")
+        ad.add(Foo)
+        res = ws.resolve(parse_requirements("Foo;python_version>='2'"), ad)
+        assert list(res) == [Foo]
+
+    def test_marker_evaluation_with_extras(self):
+        """
+        Extras are also evaluated as markers at resolution time.
+        """
```
rbtcollins: nit: Single line docstrings should be like

 """line""" per the PEP.

pkg_resources/tests/test_resources.py
```diff
@@ -181,6 +180,52 @@ def test_environment_markers(self):
...
+        ad = pkg_resources.Environment([])
+        ws = WorkingSet([])
+        # Metadata needs to be native strings, so use str().
```
rbtcollins: Worth noting the cause - cStringIO behaviour in email.parser on 2.6.

pkg_resources/__init__.py
rbtcollins: See https://github.com/rbtcollins/setuptools/commit/8bd6f6fa260424315e0606a1d5b2ece8a006d149

rbtcollins: Looks good now, since that test is covering it ;)

rbtcollins:
+!

jaraco:
Thanks @s-t-e-v-e-n-k and @rbtcollins for the patch with tests and review.

pkg_resources/__init__.py
```diff
@@ -804,9 +806,14 @@ def resolve(self, requirements, env=None, installer=None,
                 continue
             # If the req has a marker, evaluate it -- skipping the req if
             # it evaluates to False.
-            # https://github.com/pypa/setuptools/issues/523
-            _issue_523_bypass = True
-            if not _issue_523_bypass and req.marker and not req.marker.evaluate():
+            if req.marker:
+                result = []
+                if req in extra_req_mapping:
+                    for extra in extra_req_mapping[req] or ['']:
```
jaraco: @s-t-e-v-e-n-k, @rbtcollins What is the `or ['']` for here? The tests pass without it. Can I remove it? If not, can you create a test to capture the use case that drives it?

s-t-e-v-e-n-k: The tests do pass, but it actually ends up not calling marker.evaluate() in that case at all. The only thing I could think of to deal with that was asserting len(result) >= 1, but that concerned me that it might cause some head-scratching further down the line.

I'll see if I can come up with a test case.

PR #532, opened 2016-04-06
Reinstate the or guard in WorkingSet._markers_pass

s-t-e-v-e-n-k:
Thanks to @jaraco refactoring the marker evaluation in WorkingSet.resolve(), we can test it directly to show that the [''] guard is required.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -861,7 +861,7 @@ class WorkingSet(object):
 
         result = []
         if req in extra_req_mapping:
-            for extra in extra_req_mapping[req]:
+            for extra in extra_req_mapping[req] or ['']:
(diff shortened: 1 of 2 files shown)


jaraco:
I'm still a little uneasy about substituting an empty tuple for a single item list of empty string... but at least now the expectation that the False value will be an empty tuple is captured. I can work with that. Thanks.

jaraco:
In 8ce73d4450e, I've finished up another refactor which attempts to simplify the implementation and clarify the expectation when no extras were indicated (not weirdly allowing `extra=''` to pass), but I'm not sure this is right. Thoughts?

jaraco:
A better perspective is [these two commits together](https://github.com/pypa/setuptools/compare/be663b8...8ce73d).

Issue #526, opened 2016-03-31
Intermittent test failures on Python 3


jaraco:
These two failures were reported in [this job](https://travis-ci.org/pypa/setuptools/jobs/119702714) even though another job on the same Python version and code passed in the same build.

```
=================================== FAILURES ===================================

________________________ TestEntryPoints.testParseList _________________________

self = <pkg_resources.tests.test_resources.TestEntryPoints object at 0x7f290b6cd7b8>
    def testParseList(self):
>       self.checkSubMap(EntryPoint.parse_group("xyz", self.submap_str))
pkg_resources/tests/test_resources.py:337: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pkg_resources.tests.test_resources.TestEntryPoints object at 0x7f290b6cd7b8>
m = {'feature1': EntryPoint.parse('feature1 = somemodule:somefunction'), 'feature2': EntryPoint.parse('feature2 = another.module:SomeClass [extra1,extra2]'), 'feature3': EntryPoint.parse('feature3 = this.module [something]')}
    def checkSubMap(self, m):
        assert len(m) == len(self.submap_expect)
        for key, ep in self.submap_expect.items():
>           assert repr(m.get(key)) == repr(ep)
E           assert "EntryPoint.p...ra1,extra2]')" == "EntryPoint.pa...ra2,extra1]')"
E             Skipping 50 identical leading characters in diff, use -v to show
E             - ass [extra1,extra2]')
E             ?           ^      ^
E             + ass [extra2,extra1]')
E             ?           ^      ^
pkg_resources/tests/test_resources.py:322: AssertionError

_________________________ TestEntryPoints.testParseMap _________________________

self = <pkg_resources.tests.test_resources.TestEntryPoints object at 0x7f290cbfd978>
    def testParseMap(self):
        m = EntryPoint.parse_map({'xyz':self.submap_str})
>       self.checkSubMap(m['xyz'])
pkg_resources/tests/test_resources.py:345: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pkg_resources.tests.test_resources.TestEntryPoints object at 0x7f290cbfd978>
[...]

PR #533, opened 2016-04-06
Stop comparing repr()'s in TestEntryPoint

s-t-e-v-e-n-k:
In Python 3, the default order of iterables can not determined,
so comparing the repr of objects that include tuples is not static
like it is under Python 2. Compare the attributes of EntryPoint
instead, making sure to sort .attrs and .extras.

Closes: #526

### DIFF
--- a/pkg_resources/tests/test_resources.py
+++ b/pkg_resources/tests/test_resources.py
@@ -382,7 +382,10 @@ class TestEntryPoints:
     def checkSubMap(self, m):
         assert len(m) == len(self.submap_expect)
         for key, ep in self.submap_expect.items():
-            assert repr(m.get(key)) == repr(ep)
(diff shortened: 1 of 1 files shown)


jaraco:
Fantastic. Thanks!

One clarification - tuples and lists are ordered and will always have consistent order, but on late versions of Python, it's possible that ordering of items in identical sets is not stable, depending on the PYTHONHASHSEED:

```
$ PYTHONHASHSEED=3 python3.5 -c "print(list({'extra1', 'extra2'}) == list({'extra2', 'extra1'}))"
False
$ PYTHONHASHSEED=0 python3.5 -c "print(list({'extra1', 'extra2'}) == list({'extra2', 'extra1'}))"
True
```

PR #537, opened 2016-04-07
This helps finding packages with bad utf

ri0t:
Checking hundreds of possibly installed packages manually should NOT be expected of the user ;)


jaraco:
Excellent. Thanks for the contrib.

One concern I have - what if there are clients relying on the UnicodeDecodeError to trap such exceptions? Perhaps that's unlikely and this would be safe. Let's roll it out and see.

Also, I feel a little dirty evaluating `str(exc)` and then raising a new exception. Given that Python 3 has built-in exception chaining, I'm more inclined to go with a simple `raise Exception("Invalid encoding on package metadata %s" % self.path)` and let the user re-run the decoding on Python 2 if they need to trace the invalid characters.

PR #552, opened 2016-04-17
Feature/msvc discovery

JGoutin:
It's finished !! Now, it fully support MSVC from 9.0 to 14.0, and all theses standalone distributions:
- Microsoft Visual C++ Compiler for Python 2.7 (MSVC++9.0, x86, amd64)
- Microsoft Windows SDK 6.1 (MSVC++9.0, x86, x64, ia64)
- Microsoft Windows SDK 7.0 (MSVC++9.0, x86, x64, ia64)
- Microsoft Windows SDK 7.1 (MSVC++10.0, x86, x64, ia64)
- Microsoft Visual C++ Build Tools 2015 (MSVC++14.0, x86, x64, arm)

Next step:
- Code review.
- Test it with some MSVC versions/Architectures/Python versions combinations.

I have also some comments on the global implementation:
- I think `msvc9_find_vcvarsall` can now be deleted.
- Maybe rename the file, by example "msvc_support.py".

### DIFF
--- a/setuptools/msvc9_support.py
+++ b/setuptools/msvc9_support.py
@@ -1131,5 +1131,6 @@ class EnvironmentInfo:
             name = os.listdir(path)
             if name:
                 return '%s\\' % name[0]
+            return ''
         except FileNotFoundError:
             return ''
--- a/setuptools/msvc9_support.py
+++ b/setuptools/msvc9_support.py
@@ -380,11 +380,32 @@ class RegistryInfo:
         return os.path.join(self.microsoft, path)
 
     @property
+    def microsoft_sdk(self):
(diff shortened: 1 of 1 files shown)


jaraco:
Looks like tests are failing due to `import six`. Since Setuptools doesn't have dependencies, you have to import six from the extern package. Search for six in the package for other examples.

JGoutin:
Fixed Six import. But still fail, but with "No environment variables set' error, because tests are on Linux.

jaraco:
We can work out the issues with the tests on Linux later. Thanks for sharing this. I will review as time permits.

PR #553, opened 2016-04-18
Preserve order of egg_info section in setup.cfg

dims:
egg_info is the dictionary with information that is injected
into setup.cfg. edit_config uses RawConfigParser which uses
collections.OrderedDict for all the data. Since we use a
simple dict(), when we loop through items in edit_config, we
see random behavior as a result the fields
tag_svn_revision/tag_date/tag_build are added to the setup.cfg
randomly. Let's add a OrderedDict to make it more predictable.

### DIFF
--- a/setuptools/command/setopt.py
+++ b/setuptools/command/setopt.py
@@ -2,6 +2,7 @@ from distutils.util import convert_path
 from distutils import log
 from distutils.errors import DistutilsOptionError
 import distutils
+import operator
 import os
 
 from setuptools.extern.six.moves import configparser
@@ -42,7 +43,8 @@ def edit_config(filename, settings, dry_run=False):
     log.debug("Reading configuration from %s", filename)
     opts = configparser.RawConfigParser()
     opts.read([filename])
-    for section, options in settings.items():
+    for section, options in sorted(settings.items(),
+                                   key=operator.itemgetter(0)):
         if options is None:
             log.info("Deleting section [%s] from %s", section, filename)
             opts.remove_section(section)
@@ -50,7 +52,8 @@ def edit_config(filename, settings, dry_run=False):
             if not opts.has_section(section):
                 log.debug("Adding new section [%s] to %s", section, filename)
                 opts.add_section(section)
-            for option, value in options.items():
+            for option, value in sorted(options.items(),
+                                        key=operator.itemgetter(0)):
                 if value is None:
                     log.debug(
                         "Deleting %s.%s from %s",
--- a/setuptools/tests/test_egg_info.py
+++ b/setuptools/tests/test_egg_info.py
@@ -1,7 +1,16 @@
 import os
 import glob
+import re
 import stat
+import sys
 
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from setuptools.command.egg_info import egg_info
+from setuptools.dist import Distribution
 from setuptools.extern.six.moves import map
 
 import pytest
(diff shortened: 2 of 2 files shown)


rbtcollins:
If we're going to make this something folk depend on, it really needs a test so that future developers don't break it by mistake.

jaraco:
Yes. I also think a test is called for. I'm not sure the implementation as written will work as I believe the keyword arguments will be ordered according to the PYTHONHASHSEED.

dims:
@rbtcollins @jaraco i've added a test and sorted the items by key. Seems to work. Please take a look.

jaraco:
It seems that the tests aren't capturing the new expectation in egg_info.py. If I apply the following patch to your branch, the tests still pass:

```
diff -r f08b20844fbd setuptools/command/egg_info.py
--- a/setuptools/command/egg_info.py	Mon Apr 18 15:08:49 2016 -0400
+++ b/setuptools/command/egg_info.py	Tue Apr 19 12:20:13 2016 -0400
@@ -2,11 +2,6 @@
 
 Create a distribution's .egg-info directory and contents"""
 
-try:
-    from collections import OrderedDict
-except ImportError:
-    OrderedDict = dict
-
 from distutils.filelist import FileList as _FileList
 from distutils.util import convert_path
 from distutils import log
@@ -72,10 +67,10 @@
 
     def save_version_info(self, filename):
         values = dict(
-            egg_info=OrderedDict(
+            egg_info=dict(
+                tag_svn_revision=0,
+                tag_date=0,
                 tag_build=self.tags(),
-                tag_date=0,
-                tag_svn_revision=0,
             )
         )
         edit_config(filename, values)
```

That leads me to believe the OrderedDict isn't required at all.

Furthermore, this new pull request alters the way that all config files are generated, which may have implications for files other than setup.cfg and sections other than egg_info. Rather than retaining the current order, it specifies a new order (sorted lexicographically).

In my experimentation, RawConfigParser will retain order of any existing options. It's only when new items are added that order is relevant (and they're always appended).

From the perspective of 'edit_config', I think I'd prefer to give control to the caller as to what order the settings are applied, rather than forcing lexicographic ordering.


Could you elaborate on what is the issue you're encountering that's inspiring this change?

dims:
Jason, Here's the background. We were trying to verify that old tarball(s) were not tampered with by generating fresh tar.gz(s) and comparing the two. Example output : https://etherpad.openstack.org/p/lJgI00FaLw

dims:
in the end, i'd like to be able to run md5sum on generated tar.gz(s) so we need the files generated to be deterministic.

dims:
@jaraco I don't know enough to say this fix is the best way to get what i want :) Any suggestions i'd be happy to explore

rbtcollins:
So I think that determinism is worth aspiring to. Note though that .gz
itself is not deterministic. In particular the mtime is that of the tarball
being compressed - E.g. now - and this we are going to have to address that.

I suggest a functional test that sdist of a reference project, done twice,
is byte identical - since after all, that is what you want to achieve.
Sorry if that is what your current test does - I'm responding to the email
thread.

jaraco:
> We were trying to verify that old tarball(s) were not tampered with by generating fresh tar.gz(s) and comparing the two.

So updating setuptools won't support this use case for existing old tarballs, as they were already potentially generated with the nondeterministic ordering. This will only allow new tarballs, generated by users using this new functionality, to support this use case.

> So I think that determinism is worth aspiring to.

Agreed, especially if the determinism doesn't also preclude giving control to the user over what fields appear in which order.


I'm slightly annoyed by how Setuptools is injecting default values into all of these tarballs. I wonder if instead Setuptools should just stop injecting these values, if that would solve the issue at hand.

jaraco:
> Setuptools is injecting default values

That characterization isn't correct. These aren't default values, but overrides. Setuptools is _replacing_ the user-supplied values, disabling 'tag_svn_revision' and 'tag_date', and instead materializing those values in 'tag_build'.

That makes sense, now that I think about it. You don't want those values being recalculated when the sdist is installed, but you want them materialized at build time so they're fixed to the build.

jaraco:
@dims: Does this latest implementation suit your needs?

dims:
@jaraco i'll test it and get back to you later today. thanks!

Issue #557, opened 2016-04-25
Bootstrap script unable to locate releases following 20.10.0 release


idgserpro:
How to test:

Download

https://bootstrap.pypa.io/bootstrap-buildout.py

At the time of this ticket, `__version__` of this file is [**2015-07-01**](https://github.com/buildout/buildout/blob/9ba1e3170a474e6139c9a7eadd32a498c22cb3bb/bootstrap/bootstrap.py).

Run

`python bootstrap-buildout.py`

You get

```
Downloading https://pypi.python.org/packages/source/s/setuptools/setuptools-20.10.1.zip
Extracting in /tmp/tmpIKldPD
Traceback (most recent call last):
  File "bootstrap-buildout.py", line 117, in <module>
    ez['use_setuptools'](**setup_args)
  File "<string>", line 173, in use_setuptools
  File "<string>", line 124, in _do_download
  File "<string>", line 64, in _build_egg
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "<string>", line 103, in archive_context
  File "/usr/lib/python2.7/zipfile.py", line 714, in __init__
    self._GetContents()
  File "/usr/lib/python2.7/zipfile.py", line 748, in _GetContents
    self._RealGetContents()
  File "/usr/lib/python2.7/zipfile.py", line 763, in _RealGetContents
    raise BadZipfile, "File is not a zip file"
zipfile.BadZipfile: File is not a zip file
```

You get the same error with 20.10.0.

`python bootstrap-buildout.py --setuptools-version=20.10.0`

```
Downloading https://pypi.python.org/packages/source/s/setuptools/setuptools-20.10.0.zip
Extracting in /tmp/tmp6G6ZVT
Traceback (most recent call last):
  File "bootstrap-buildout.py", line 117, in <module>
    ez['use_setuptools'](**setup_args)
  File "<string>", line 173, in use_setuptools
  File "<string>", line 124, in _do_download
  File "<string>", line 64, in _build_egg
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "<string>", line 103, in archive_context
  File "/usr/lib/python2.7/zipfile.py", line 714, in __init__
    self._GetContents()
  File "/usr/lib/python2.7/zipfile.py", line 748, in _GetContents
    self._RealGetContents()
  File "/usr/lib/python2.7/zipfile.py", line 763, in _RealGetContents
[...]

Issue #519, opened 2016-03-17
ez_setup.use_setuptools() fails if it needs to download a newer setuptools


ghost:
Originally reported by: **agronholm (Bitbucket: [agronholm](http://bitbucket.org/agronholm), GitHub: [agronholm](http://github.com/agronholm))**

---

In setup.py:

```
#!python

import ez_setup
ez_setup.use_setuptools('20.3')

```

Then, with setuptools 20.1.1 installed:

```
$ python setup.py
Traceback (most recent call last):
  File "setup.py", line 38, in <module>
    ez_setup.use_setuptools('20.3')
  File "/home/alex/checkout/crossbar/ez_setup.py", line 173, in use_setuptools
    return _do_download(version, download_base, to_dir, download_delay)
  File "/home/alex/checkout/crossbar/ez_setup.py", line 132, in _do_download
    import setuptools
  File "build/bdist.linux-x86_64/egg/setuptools/__init__.py", line 11, in <module>
  File "build/bdist.linux-x86_64/egg/setuptools/extern/__init__.py", line 1, in <module>
  File "build/bdist.linux-x86_64/egg/pkg_resources/__init__.py", line 48, in <module>
  File "/home/alex/virtualenv/crossbar/local/lib/python2.7/site-packages/pkg_resources/extern/__init__.py", line 43, in load_module
    mod = sys.modules[extant]
AttributeError: 'NoneType' object has no attribute 'modules'
```

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/519

PR #559, opened 2016-04-25
Fix setuptools url and curl wants --location to follow redirects

jimfulton:
This change also removes:

    sys.meta_path = [importer for importer in sys.meta_path if
                     importer.__class__.__module__ != 'pkg_resources.extern']

from _unload_pkg_resources.

These lines were removed from the version at: https://bootstrap.pypa.io/ez_setup.py

I don't know why the code there and here differs.

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -33,7 +33,7 @@ except ImportError:
 
 LATEST = object()
 DEFAULT_VERSION = LATEST
-DEFAULT_URL = "https://pypi.python.org/packages/source/s/setuptools/"
+DEFAULT_URL = "https://pypi.io/packages/source/s/setuptools/"
 DEFAULT_SAVE_DIR = os.curdir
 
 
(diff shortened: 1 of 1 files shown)


jimfulton:
Fixes 557, I think.

jimfulton:
Fixes #557, I think.

agronholm:
That meta path manipulation unloads the old import hook when upgrading setuptools. Why did you remove it? It was necessary to fix #519.

jimfulton:
@agronholm I was just reflecting what was at https://bootstrap.pypa.io/ez_setup.py

I'll put them back.

jimfulton:
@agronholm I restored your fix and the PR makes more sense now.

jimfulton:
Thanks @jaraco .

Can you also merge: https://github.com/pypa/setuptools/pull/560 other otherwise get the ez_setup.py changes on the bootstrap branch so the show up at: https://bootstrap.pypa.io/ez_setup.py ?

Issue #571, opened 2016-05-03
[build_ext] bulding shared libraries: sysconfig._CONFIG_VARS vs distutils.sysconfig._config_vars


anthrotype:
Hello,

I was browsing the setuptools sources, and I discovered that it has (experimental?) support for building native shared libraries (.dylib, .dll, etc.) to be linked with Python extension modules.
I'm referring to `setuptools.extension.Library` class and the `setup_shlib_compiler` method of `build_ext` command.

I also noticed there was a `tests/shlib_test` (which somehow was removed with commit  https://github.com/pypa/setuptools/commit/654c26f78a3031c34b81f338d3b6d1a3a87c34da). I only had to modify the "shlib_test/setup.py" script a bit to make it run with the current setuptools (21.0):

```diff
-from setuptools import setup, Extension, Library
+from setuptools import setup, Extension
+from setuptools.extension import Library
```

I encountered an issue in the `setup_shlib_compiler` method of `build_ext` command, where the compiler is configured for building shared libraries. When the platform is OSX ("darwin"), the global `_CONFIG_VARS` are temporarily modified (the flags `-dynamiclib -undefined dynamic_lookup` are required for .dylib); then the `distutils.sysconfig.customize_compiler` function is called in order to set these flags.
The problem is the distutils' `customize_compiler` function is reading from a different dictionary than then one setuptools is modifying, namely `distutils.sysconfig._config_vars`; therefore, the default values are used instead of the temporarily modified ones, and a `-bundle` (which can't be linked at compile time) instead of a `-dynamiclib` gets created.

The issue was introduced with this old commit https://github.com/pypa/setuptools/commit/c04abca6 which replaced all uses of `distutils.sysconfig` in setuptools, including `build_ext` module, with equivalent imports from the new `sysconfig` module, available since Python >= 2.7/3.2.

The issue can be fixed by importing `_config_vars` from `distutils.sysconfig` instead of the top-level `sysconfig`. Then the `customize_compiler` function can use the temporarily overridden variables as originally intended.

I hope this makes sense... :)

PR #572, opened 2016-05-04
build_ext: always import _CONFIG_VARS from distutils.sysconfig instead of sysconfig

anthrotype:
otherwise `distutils.sysconfig.customize_compiler` does not configure the OS X compiler for -dynamiclib

See https://github.com/pypa/setuptools/issues/571

### DIFF
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -22,6 +22,27 @@ get_config_var("LDSHARED")  # make sure _config_vars is initialized
 del get_config_var
 from distutils.sysconfig import _config_vars as _CONFIG_VARS
 
+
+def _customize_compiler_for_shlib(compiler):
+    if sys.platform == "darwin":
+        # building .dylib requires additional compiler flags on OSX; here we
+        # temporarily substitute the pyconfig.h variables so that distutils'
+        # 'customize_compiler' uses them before we build the shared libraries.
+        tmp = _CONFIG_VARS.copy()
+        try:
+            # XXX Help!  I don't have any idea whether these are right...
+            _CONFIG_VARS['LDSHARED'] = (
+                "gcc -Wl,-x -dynamiclib -undefined dynamic_lookup")
+            _CONFIG_VARS['CCSHARED'] = " -dynamiclib"
+            _CONFIG_VARS['SO'] = ".dylib"
+            customize_compiler(compiler)
+        finally:
+            _CONFIG_VARS.clear()
+            _CONFIG_VARS.update(tmp)
+    else:
+        customize_compiler(compiler)
+
+
 have_rtld = False
 use_stubs = False
 libtype = 'shared'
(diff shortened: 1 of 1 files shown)


setuptools/command/build_ext.py
```diff
@@ -16,15 +16,11 @@
 except ImportError:
     _build_ext = _du_build_ext
 
-try:
-    # Python 2.7 or >=3.2
```
jaraco: This change seems to be moving in the wrong direction, conflicting with the intention of the author to support the newer, preferred module. I don't know what the implications are, but what would prevent someone from making the reverse change again?

I think this behavior needs a unit test in the test suite to capture the desired behavior.

Also, is it possible to allow the newer, preferred module to still be used, but always invoke the `get_config_var`? What is the relationship between `sysconfig._CONFIG_VARS` and `distutils.sysconfig._config_vars`?

anthrotype: Thanks for having a look.
Yes, you're right about the need for a unit test in the test suite.

Here the intent of the "original" author (before https://github.com/pypa/setuptools/commit/c04abca6 broke it by importing the newer module) was to import the global `distutils.sysconfig._config_vars` dictionary, not to get the values, but to temporarily set them before calling `distutils.sysconfig.customize_compiler`. This is only for darwin, where building dynamic load libraries requires special compiler flags.

The problem is the `customize_compiler` function uses the `_config_vars` which are defined in the `distutils.sysconfig` module, not the one from the new `sysconfig` module. Therefor, importing `_CONFIG_VARS` from the latter, and modifying them, does not have any effect on `customize_compiler`.

I'm not sure about the relationship between `sysconfig._CONFIG_VARS` and `distutils.sysconfig._config_vars`.

jaraco: Thanks for the clarification. After re-reading the ticket, I see you also detailed much of this there. Sorry to ask you to repeat yourself.

An alternative to a unit test would be to have a function in build_ext which documents the intention, such that a subsequent change when reviewed would be clearly in violation of the stated intention.

anthrotype:
I moved the block customizing the compiler for dylib to a separate function, preceded with an underscore to make it private. I also added a comment explaining the intention.
Please take a look and let me know if you require further changes.
Thank you.

setuptools/command/build_ext.py
```diff
@@ -16,15 +16,32 @@
...
+def _customize_compiler_for_shlib(compiler):
+    if sys.platform == "darwin":
+        # building .dylib requires additional compiler flags on OSX; here we
+        # temporarily substitute the pyconfig.h variables so that distutils'
+        # 'customize_compiler' uses them before we build the shared libraries.
+        tmp = _CONFIG_VARS.copy()
+        try:
+            # XXX Help!  I don't have any idea whether these are right...
+            _CONFIG_VARS['LDSHARED'] = (
+                "gcc -Wl,-x -dynamiclib -undefined dynamic_lookup")
```
anthrotype: As the comment above shows, also the original author wasn't 100% sure whether these are the correct flags.

I found that I need to remove the `-Wl,-x` flags so that `clang` can compile a C++ library I'm working on, otherwise it throws a cryptic error like this:
```
ld: internal error: atom not found in symbolIndex(__ZNKSt3__115basic_stringbufIcNS_11char_traitsIcEENS_9allocatorIcEEE3strEv) for architecture x86_64`
```

Others have encountered similar issues: e.g. see https://bitbucket.org/premake/premake-dev/issues/263/os-x-gcc-clang-and-the-infamous-wl-x

clang also gives me a warning about the `-dynamiclib` flag which is passed here via `_CONFIG_VARS['CCSHARED']`:
```
clang: warning: argument unused during compilation: '-dynamiclib'
```

Hopefully some else in the setuptools community knows more about how to compile dynamic libraries for OS X.

jaraco:
Sounds good to me. Let's get it out and see how it fares.

Issue #575, opened 2016-05-05
please include the CHANGES.rst file in the source distribution


doko42:
please re-add the CHANGES.rst file.

jaraco:
Fine by me. Probably when it was renamed the manifest.in needed to be updated.

PR #576, opened 2016-05-06
Add CHANGES.rst

gforcada:
Fixes https://github.com/pypa/setuptools/issues/575


jaraco:
Thanks for this, but I'm going to decline. It only gets CHANGES.rst, but not README.rst.

gforcada:
Well that's what #575 asked for...

jaraco:
Indeed. I made that suggestion without looking at the code. It certainly would have fixed the issue. Thanks again.

PR #582, opened 2016-05-10
Feature/msvc-discovery

JGoutin:
Some fixes after testing VC Builds Tools 2015 with current = Amd64 and target = Amd64, x86, Arm

### DIFF
--- a/setuptools/msvc9_support.py
+++ b/setuptools/msvc9_support.py
@@ -324,11 +324,11 @@ class PlatformInfo:
             "\current" if target architecture is current architecture,
             "\current_target" if not.
         """
-        path = self.target_dir(True)
-        if self.target_cpu != self.current_cpu:
-            current = 'x86' if forcex86 else self.current_cpu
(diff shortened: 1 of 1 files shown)


setuptools/msvc9_support.py
```diff
@@ -324,11 +324,11 @@ def cross_dir(self, forcex86=False):
...
-        path = self.target_dir(True)
-        if self.target_cpu != self.current_cpu:
-            current = 'x86' if forcex86 else self.current_cpu
-            path = path.replace('\\', '\\%s_' % current)
-        return path
+        current = 'x86' if forcex86 else self.current_cpu
+        if self.target_cpu != current:
+            return self.target_dir().replace('\\', '\\%s_' % current)
+        else:
+            return ''
```
jaraco: This return value no longer matches the doc string.

jaraco:
I've also merged this feature branch with master in order to fix test failures (c98abcb08a).

PR #587, opened 2016-05-19
Update msvc9_support.py

JGoutin:
Hello,

- Fix and improve docstrings (Thank you for reporting error).
- Add "is Windows" check before patching.
- Avoid `except:` with no specified exception.
- Standardize `cross_dir` function to be more like `current/target_dir` functions.

_Note: I'll be away during 3 next weeks, I'll finish testing this after my return._

### DIFF
--- a/setuptools/msvc9_support.py
+++ b/setuptools/msvc9_support.py
@@ -4,6 +4,7 @@ This module improve support for Microsoft Visual C++ compilers. (Windows Only)
 import os
 import collections
 import itertools
+import platform
 import distutils.errors
 from setuptools.extern.six.moves import filterfalse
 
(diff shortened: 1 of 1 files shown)


setuptools/msvc9_support.py
```diff
@@ -57,6 +58,10 @@ def patch_for_specialized_compiler():
     Microsoft Visual C++ 14.0:
         Microsoft Visual C++ Build Tools 2015 (x86, x64, arm)
     """
+    if platform.system() != 'Windows':
```
jaraco: This check feels a bit redundant, or maybe an over-optimization. What's the motivation behind it? It seems to me if a platform has the relevant distutils module, it should be patched.

JGoutin: I did this to help pass test on Linux.
If related files are missing on other platform, I am OK it is redundant. If files are present, patched them if they are not used is not useful.

PR #597, opened 2016-05-27
Ensure package name isn't unicode in Python2 distutils

### LINKED ISSUE #190: Unfriendly error message when unicode is passed to package_dir or packages
Originally reported by: **jaraco (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

Adding `__future__.unicode_literals` causes setup.py scripts to fail. In the mailman project, here is an example error when running python2 setup.py build:

```
running build
running build_py
  File "setup.py", line 112, in <module>
    test_suite = 'nose2.collector.collector',
  File "C:\Program Files\Python27\lib\distutils\core.py", line 152, in setup
    dist.run_commands()
  File "C:\Program Files\Python27\lib\distutils\dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "C:\Program Files\Python27\lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "C:\Program Files\Python27\lib\distutils\command\build.py", line 127, in run
    self.run_command(cmd_name)
  File "C:\Program Files\Python27\lib\distutils\cmd.py", line 326, in run_command
  File "C:\Program Files\Python27\lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "build\bdist.win-amd64\egg\setuptools\command\build_py.py", line 42, in run
  File "C:\Program Files\Python27\lib\distutils\command\build_py.py", line 372, in build_packages
    self.build_module(module, module_file, package)
  File "build\bdist.win-amd64\egg\setuptools\command\build_py.py", line 60, in build_module
  File "C:\Program Files\Python27\lib\distutils\command\build_py.py", line 333, in build_module
    "'package' must be a string (dot-separated), list, or tuple")
TypeError: 'package' must be a string (dot-separated), list, or tuple
```

A different error occurs when using 'develop':

```
running develop
error: 'egg_base' must be a directory name (got `src`)
```

Setuptools could make this error message nicer.

- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/190

benjaoming:
Problem encountered on Ubuntu 14.04

Fixes #190

If people need to fix this without this patch, they have to wrap their package names in `str()` like so:

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-
from __future__ import absolute_import, print_function, unicode_literals

setup(
    # ....
    package_dir=[
        str('package_name'),
    ]
)
```

### DIFF
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -8,6 +8,7 @@ import io
 import distutils.errors
 import itertools
 
+from setuptools.extern import six
 from setuptools.extern.six.moves import map, filter, filterfalse
 
 try:
@@ -66,6 +67,9 @@ class build_py(orig.build_py, Mixin2to3):
         return orig.build_py.__getattr__(self, attr)
 
     def build_module(self, module, module_file, package):
+        if six.PY2 and isinstance(package, six.string_types):
+            # avoid errors on Python 2 when unicode is passed (#190)
+            package = package.split('.')
         outfile, copied = orig.build_py.build_module(self, module, module_file,
                                                      package)
         if copied:


jaraco:
Since the issue only affects Python 2, I'd expect the fix to only affect Python 2, which would also provide a clearer indication of the scope of the error. As written, I wouldn't expect my future self to know that this code can be removed when Python 2.7 support is removed. Also, why select on `string_types`? Can package_name ever be anything other than a string type?

benjaoming:
@jaraco -- you are right, and I didn't research if later Python 2 distutils versions have fixed this. I think it would be okay to not care which distutils versions exactly are causing the issue and simplify this as a Python 2 fix, expressing that in the comments...?

benjaoming:
> Also, why select on `string_types`? Can package_name ever be anything other than a string type?

Yes. Read #190:

> TypeError: 'package' must be a string (dot-separated), list, or tuple

jaraco:
I guess what I'm getting at is that since the Python 3 implementation does the right thing currently, this code shouldn't affect that, but this code does. In particular, the proposed implementation will convert bytes into the repr of their value, masking the underlying error.

How about instead, something like:

```
if six.PY2 and isinstance(package, six.text_type):
  # avoid errors on Python 2 when unicode is passed (#190)
  package = package.split('.')
```

That implementation (a) captures the PY2 specific behavior, (b) only affects if unicode is passed, and (c) provides forward-compatibility, allowing for unicode values to be passed and treated just like unicode values are handled on Python 3.

benjaoming:
Hi @jaraco 

> In particular, the proposed implementation will convert bytes into the repr of their value, masking the underlying error.

Thanks for dedicating interest and putting thought into a better solution! I agree, it's a better fix, cleaner and doesn't affect Python 3 behaviour.

Squashed into 1 commit and `-f`'ed.

jaraco:
Thanks for bringing the suggestions to the table and working through my concerns.

Issue #445, opened 2015-10-01
"Resource Management API" links broken in docs


bb-migration:
Originally reported by: **lukeyeager (Bitbucket: [lukeyeager](http://bitbucket.org/lukeyeager), GitHub: [lukeyeager](http://github.com/lukeyeager))**

---

These sections each contain a link to "Resource Management API".

http://pythonhosted.org/setuptools/setuptools.html#accessing-data-files-at-runtime
http://pythonhosted.org/setuptools/setuptools.html#non-package-data-files

That links to:

https://bitbucket.org/pypa/setuptools/src/11264de04d44e4cb82e1e797e91549a6a71fb99b/docs/setuptools.txt?at=default&fileviewer=file-view-default#setuptools.txt-919

But the link no longer exists.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/445

benoit-pierre:
No longer applicable.

PR #601, opened 2016-06-02
Update link to Resource Management API

inirudebwoy:
Hi,

I believe this could fix issue #445.


jaraco:
Seems reasonable to me.

PR #606, opened 2016-06-06
Setuptools scm

Geetika7:


jaraco:
I'm really not sure what to do with this. The branch that's proposed merging causes the continuous integration tests to fail because they can't bootstrap. I suspect it will also cause issues in other bootstrap scenarios. This work if merged will likely cause disruption. And at least until the CI issues are sorted, it's not acceptable. I do invite you to grab the code and investigate the issues and devise solutions.

PR #609, opened 2016-06-08
check if a download is successfull before deciding not to try the nex…

JensTimmerman:
…t possible dist

### DIFF
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -602,9 +602,8 @@ class PackageIndex(Environment):
                     continue
 
                 if dist in req and (dist.precedence <= SOURCE_DIST or not source):
-                    mylocation = self.download(dist.location, tmpdir)
-                    if os.path.exists(mylocation):
-                        dist._location = mylocation
+                    dist.download_location = self.download(dist.location, tmpdir)
+                    if os.path.exists(dist.download_location):
                         return dist
 
         if force_scan:
@@ -632,7 +631,7 @@ class PackageIndex(Environment):
             )
         else:
             self.info("Best match: %s", dist)
-            return dist.clone(location=dist._location)
+            return dist.clone(location=dist.download_location)
 
     def fetch(self, requirement, tmpdir, force_scan=False, source=False):
         """Obtain a file suitable for fulfilling `requirement`
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -605,14 +605,15 @@ class PackageIndex(Environment):
                     mylocation = self.download(dist.location, tmpdir)
                     if os.path.exists(mylocation):
                         return dist, mylocation
+            return None, None
 
         if force_scan:
             self.prescan()
             self.find_packages(requirement)
             dist, mylocation = find(requirement)
 
-        if local_index is not None:
-            dist, mylocation = dist, mylocation if dist else  find(requirement, local_index)
+        if not dist and local_index is not None:
+            dist, mylocation = find(requirement, local_index)
 
         if dist is None:
             if self.to_scan is not None:
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -602,24 +602,26 @@ class PackageIndex(Environment):
                     continue
 
                 if dist in req and (dist.precedence <= SOURCE_DIST or not source):
-                    return dist
+                    mylocation = self.download(dist.location, tmpdir)
+                    if os.path.exists(mylocation):
+                        return dist, mylocation
 
         if force_scan:
             self.prescan()
             self.find_packages(requirement)
-            dist = find(requirement)
+            dist, mylocation = find(requirement)
 
         if local_index is not None:
-            dist = dist or find(requirement, local_index)
+            dist, mylocation = dist, mylocation if dist else  find(requirement, local_index)
 
         if dist is None:
             if self.to_scan is not None:
                 self.prescan()
-            dist = find(requirement)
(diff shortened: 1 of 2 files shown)


setuptools/package_index.py
```diff
@@ -588,24 +588,26 @@ def find(req, env=None):
...
 
         if force_scan:
             self.prescan()
             self.find_packages(requirement)
-            dist = find(requirement)
+            dist, mylocation = find(requirement)
 
         if local_index is not None:
-            dist = dist or find(requirement, local_index)
+            dist, mylocation = dist, mylocation if dist else  find(requirement, local_index)
```
jayvdb: Two consecutive spaces here

setuptools/package_index.py
```diff
@@ -588,24 +588,26 @@ def find(req, env=None):
...
-                    return dist
+                    mylocation = self.download(dist.location, tmpdir)
+                    if os.path.exists(mylocation):
+                        return dist, mylocation
 
         if force_scan:
             self.prescan()
             self.find_packages(requirement)
-            dist = find(requirement)
+            dist, mylocation = find(requirement)
```
jayvdb: Since you are modifying this chunk...maybe?

   if not dist and local_index is not None:

JensTimmerman: indeed, that would be nicer, fixed.

jaraco:
Can you supply a justification for this change? What behavior led you to make the change? How does the change improve the behavior? Are there any backward compatible considerations?

setuptools/package_index.py
```diff
@@ -588,24 +588,27 @@ def find(req, env=None):
                     continue
 
                 if dist in req and (dist.precedence<=SOURCE_DIST or not source):
-                    return dist
+                    mylocation = self.download(dist.location, tmpdir)
+                    if os.path.exists(mylocation):
+                        return dist, mylocation
+            return None, None
```
jaraco: The presence of `None, None` is a bad smell to me. Better would be to return above a single value which wraps the necessary components, maybe a `namedtuple('DistLocation', 'dist location')`.

jaraco: Or maybe augment the resulting dist with a `._found_location` attribute or similar.

JensTimmerman: sure, I can clean this up, 
 thought this would be ok, since  `process_index` inner  `scan(link)` function already does the same ;-)

jaraco: No worries. I acknowledge - there are some old and undesirable patterns in setuptools, given its 11 year history.

JensTimmerman:
the problem this fixes for me is: when I specify `dependency_links` as something like
```
 dependency_links=[
        "git+https://server1.example.com/mydependency.git@1.5.5#egg=mydependnecy-1.5.5",
        "git+ssh://server2.example.com/mydependency.git@1.5.5#egg=mydependnecy-1.5.5",
],
```
where server1 and server2 could e..g be a public github.com and a private github enterprise, 
and public network is not available, then setuptools will currently fail to locate this dependency.
(changing the order of the links doesn't seem to matter)
However, with this change in place, after the first download fails, it tries the second download, and succeeds.

I could see more cases where it would be beneficial to try a second location if it exists, if the first download fails. Since, we have a second location, why not?

JensTimmerman:
backwards compatibility: I don't believe a lot of work-flows would rely on the fact that an installation won't happen if the first tried link doesn't work, because this first tried link seems to be random. Depending on a few factors the first or second link will be tried first or last.

setuptools/package_index.py
```diff
@@ -588,15 +588,17 @@ def find(req, env=None):
...
 
         if force_scan:
             self.prescan()
             self.find_packages(requirement)
             dist = find(requirement)
 
-        if local_index is not None:
-            dist = dist or find(requirement, local_index)
+        if not dist and local_index is not None:
+            dist = find(requirement, local_index)
```
JensTimmerman: this change is not relevant to this pr, but this happens to be a nicer way to check, so it's in here.

JensTimmerman:
old behaviour:
```
Searching for vsc-base>=2.4.16
Best match: vsc-base 2.4.16
Doing git clone from ssh://github.ugent.be/hpcugent/vsc-base.git to /tmp/easy_install-qVryB5/vsc-base.git
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Processing vsc-base.git
error: Couldn't find a setup script in /tmp/easy_install-qVryB5/vsc-base.git
```
new behaviour:
```
Searching for vsc-base>=2.4.16
Doing git clone from ssh://github.ugent.be/hpcugent/vsc-base.git to /tmp/easy_install-A9wgLC/vsc-base.git
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Doing git clone from https://github.com/hpcugent/vsc-base.git to /tmp/easy_install-A9wgLC/vsc-base.git
Best match: vsc-base 2.4.16
Processing vsc-base.git
Writing /tmp/easy_install-A9wgLC/vsc-base.git/setup.cfg
Running setup.py -q bdist_egg --dist-dir /tmp/easy_install-A9wgLC/vsc-base.git/egg-dist-tmp-djmMaJ
...
```

jaraco:
Can you please add an entry to the changelog (0.1 bump) to describe to the users what they should expect from the change?

JensTimmerman:
@jaraco  added changes and rebased on top of master

Issue #450, opened 2015-10-16
egg_info command is very slow if there are lots of non source files in directory


ghost:
Originally reported by: **spookylukey (Bitbucket: [spookylukey](http://bitbucket.org/spookylukey), GitHub: [spookylukey](http://github.com/spookylukey))**

---

The approach taken by egg_info is to list all the files in the source directory, recursively, and then apply rules. (via `manifest_maker` and `FileList.findall`)

If you have a lot of files, this can take a really long time. It combines especially badly with tools like 'tox', which put lots of virtualenvs in the current dir, putting many thousands of files there.

On one small project I maintain, running tox from a cold start has a 5 minute start up time (because it runs setup.py, which runs egg_info or something similar), which is a real pain when the tests themselves run far more quickly.

Would you consider a patch to improve this? Attempting a full fix is very difficult, due to the way that MANIFEST rules are essentially an imperative language in which order matters, as files are added and removed.

One strategy that might work would be to attempt to optimise very few MANIFEST commands - for example, only 'prune' would be considered. That is, a custom `FileList.find_all` command would understand the 'prune' command, and avoid recursing into that directory, but only if the prune command comes after any commands that would add files. From my understanding of how MANIFEST is processed https://docs.python.org/2/distutils/sourcedist.html, I believe this would preserve correctness.

This would not be much of an optimisation out-of-the box, but it would allow you to add "prune .tox" and "prune .git" etc. at the end of your MANIFEST.in to get big speedups.

Would this approach be accepted if I worked on it?

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/450

Issue #249, opened 2014-08-26
Have a way to ingore specific dirs when building list of files in egg_info


bb-migration:
Originally reported by: **ionelmc (Bitbucket: [ionelmc](http://bitbucket.org/ionelmc), GitHub: [ionelmc](http://github.com/ionelmc))**

----------------------------------------

I have a very large `.tox` dir, i'd like to have a way so that setuptools ignores it completely. Now it recurses inside it and takes few minutes to run egg_info and other commands that rely on distutils.filelist.listall

----------------------------------------
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/249

spookylukey:
I'm thinking that a totally new approach to handling the MANIFEST.in file is the answer.

The current approach (inherited from distutils) is to recursively list all the files, bringing the whole lot into memory, and then apply MANIFEST.in commands to that list. This is the problem, and the fix is "don't do that".

It is just the wrong approach. It gives you a small speedup in one case - when you have very few files, and you end up processing MANIFEST.in directives that would cause multiple walks over the directory structure. However, in this case, OS caching means that you the second walk is usually going to be very fast.

A far better approach is to just handle MANIFEST.in directives directly - for example, a `recursive-include foo *.py` would just cause the foo directory to be recursively walked looking for `*.py` files. This sounds 'dumb', but in reality is going to beat the current approach 99.99% of the time, and by a massive factor in many cases.

There are two ways to go about this:

1) Create a class that inherits from `distutils.filelist.FileList` and then change the implementation to not use `allfiles` at all, but have `include_pattern` and `exclude_pattern` search the filesystem directly. This will break whenever other code assumes the current implementation of `FileList` e.g. the existence of the `allfiles` attribute. setuptools does this in one place, it would also have to be fixed, I think it should be possible relatively easily. (The offending code is setuptools.command.egg_info.manifest_maker._add_egg_info, and I think the fix is actually to remove the whole of the function and the one place it is called. It seems that the entire function is actually a workaround for the fact that `include_pattern` scans not the filesystem but the internally buffered version of the filesystem). However, there might be other issues that crop up.

2) Start from scratch, with a test-driven approach to implementing MANIFEST.in directive parsing.It should be possible to do this relatively easily - the language is not that complicated. Further, it shouldn't be that hard to do it with plenty of unit tests etc., especially if the code which actually walks the file system is done using a separate class for which a mock version can be supplied.

[...]

PR #610, opened 2016-06-09
Much faster implementation of FileList, for big egg_info speedups

spookylukey:
Refs #249 and #450 

This is not a complete PR, but request for the approach to be confirmed before I do more work.

There are some notes:

# Code

The approach, as outlined before, is to reduce `findall` to a no-op, so the directory tree is not scanned, and have `include_pattern` (which handles all 'include' type directives) execute an `os.walk` directly, but trimming the search tree according to the different types of rules.

It currently doesn't handle all cases correctly (specific `graft` doesn't work correctly), but is a proof of concept. If the approach is accepted, I would add more tests (since we can no longer rely on distutils have tests since we are overriding the implementation) and fix this bug. All the tests in setuptools pass with the patch.

Some testing on Windows will be necessary, the directory path separator thing gets quite involved, I don't have a Windows box I could easily use.

Performance tests are hard - we really want to test that other parts of the filesystem are *not* being scanned, which is hard to do.

## Removal of `_add_egg_info` function, called from `add_defaults`

I believe this is correct. The behaviour of `add_defaults` seems to be tested by `test_egg_base_installed_egg_info` and `test_install_requires_with_markers` which both fail if you break it, along with a few other tests that fail.

AFAICS, the code is only working around the fact that `include_pattern` queries the `allfiles` attribute and not the filesystem, and is therefore out of date with respect to files that have been created by setuptools itself. Since we've changed the behaviour of include_pattern, I don't think this is necessary, and the tests seem to verify this.

# Metrics

Obviously this depends on how many files you've got in your working folder, and on whether the directory listings are in the OS cache.

I used this project as a not-too-extreme example, and because I actually work on this project:

https://github.com/django-money/django-money

It uses tox to test against about 80 environments, which are stored as virtualenvs in .tox

## Without patch

From a 'cold' state, but with virtualenvs created from previous usage, I ran the tests on a single environment:

    tox -e django18-py27

This took 647 seconds (10 minutes), of which just 2.8 seconds were the actual tests i.e. 645 seconds of overhead from setuptools and tox.

Doing it a second time, this reduced to 37 seconds of overhead, due to OS caching of filesystem.

## With this patch

With the patch, the setuptools + tox overhead of the same command reduced to just 2 seconds, whether from warm or cold i.e. a speed up of 18 times to 320 times (and more importantly, down to "a few seconds", an acceptable time to wait for your tests to start running). This would make a BIG difference to workflow on quite a few projects I use.

To reproduce this speedup, however, you will need to install the dev version of setuptools in both the outer project (in which tox is installed) and in the virtualenvs it created.

There are obviously cases and projects where this strategy will be less efficient than the previous one, because it requires multiple `os.walk` walks over the file system, instead of just one, but for normal projects, and for any that contain large numbers of non source files, it is likely to be a very big
win.

## Notes

There will be no speedups if MANIFEST.in contains a global-include directive, and in fact probably severe slow-downs if you have lots of files or lots of global-include directives. This is because you cannot trim the search tree at all for this directive - a global-include includes all subdirectories, including hidden ones such as ".tox", ".my_IDE_like_to_put_stuff_here" etc. This means that in general the global-include directive should be avoided, because most tools expect to be able to put all kinds of things in hidden folders and have them ignored. This warning should probably be submitted to distutils as a documentation patch.

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -2,7 +2,7 @@
 
 Create a distribution's .egg-info directory and contents"""
 
-from distutils.filelist import FileList as _FileList
+from distutils.filelist import FileList as _FileList, translate_pattern
 from distutils.util import convert_path
 from distutils import log
 import distutils.errors
@@ -235,6 +235,24 @@ class egg_info(Command):
 class FileList(_FileList):
     """File list that accepts only existing, platform-independent paths"""
 
+    def __init__(self):
+        _FileList.__init__(self)  # old-style class, can't use super
+
+        # Our version fixes the performance problems which stem from
+        # scanning filesystema and creating 'allfiles'.
+        #
+        # To ensure we didn't break anything, we want to fail loudly if any
+        # code is depending on the inherited behaviour of FileList in which it
+        # generates an 'allfiles' attribute as a side effect. So we remove the
+        # attribute so calling code will get AttributeError.
+        try:
+            del self.allfiles
+        except AttributeError:
+            pass
+
+    def findall(self, dir=None):
+        pass  # Don't scan the filesystem or create 'allfiles' attribute
+
     def append(self, item):
         if item.endswith('\r'):  # Fix older sdists built on Windows
             item = item[:-1]
@@ -245,6 +263,56 @@ class FileList(_FileList):
 
     def extend(self, paths):
         self.files.extend(filter(self._safe_path, paths))
+
+    # Because we don't have 'allfiles', we have to re-implement include_pattern
+    # completely.
+    def include_pattern(self, pattern, anchor=1, prefix=None, is_regex=0):
+        files_found = 0
+        pattern_re = translate_pattern(pattern, anchor=anchor, prefix=prefix,
+                                       is_regex=is_regex)
+
+        dir = os.curdir
+        for (dirpath, dirnames, filenames) in os.walk(dir, topdown=True):
+            if dirpath == ".":
+                start = "."
+            else:
+                start = "." + os.path.sep
+            if dirpath.startswith(start):
+                dirpath_rel = dirpath[len(start):]
+
+            for filename in filenames:
+                fullname = os.path.join(dirpath_rel, filename)
+                if pattern_re.match(fullname):
+                    self.files.append(fullname)
+                    files_found = 1
+
+            if not is_regex:
+                # If we have a regex, we can't easily reverse engineer it.
+
+                if anchor:
+                    # Not 'global-include', which we can't do anything
+                    # about.
+                    if prefix is None:
+                        # 'include' - can't match subdirs
+                        dirnames[:] = []
+                    else:
+                        # 'recursive-include'
+                        def should_recurse(d):
+                            # dirpath will have OS specific separators:
+                            parts = dirpath_rel.split(os.path.sep)
+                            if parts[0] == '':
+                                parts.pop(0)
+                            parts.append(d)
+
+                            # MANIFEST.in paths should have /
+                            prefix_parts = prefix.split('/')
+
+                            # TODO handle 'graft' which can have patterns in
+                            # prefix
+                            return parts[0:len(prefix_parts)] == prefix_parts
+
+                        dirnames[:] = filter(should_recurse, dirnames)
+        return files_found
 
     def _repair(self):
         """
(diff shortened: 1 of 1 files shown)


jaraco:
The main reluctance I have with this approach is it increases the reliance on `MANIFEST.in`, which is a duplicative, redundant functionality with source-code control systems (as [I mentioned in this comment](https://github.com/pypa/setuptools/issues/450#issuecomment-202921921)). The SCM systems keep a manifest of files in the project and have mechanisms to ignore whole trees (e.g. `.gitignore`). Furthermore, these systems provide a robust, layered mechanism for ignoring files and trees at the system or user level in addition to the project level. The more we can leverage this functionality, which is almost universally adopted, the more we can rely on these advances. I imagine developing from non-SCM tarballs is rare and that issues like #249 can be disregarded in those cases.

I'd really prefer a solution that relies on SCM file finders and limit the reliance on the (soft-) deprecated MANIFEST.in. If a SCM file finder solution is nonviable, I'd at least like to have robustly determined such before adopting another approach.

spookylukey:
In what way is this an increased reliance on MANIFEST.in? The situation is exactly the same as before, my patch is simply a change to the implementation of the MANIFEST.in processing to be (typically) much faster than the one inherited from distutils.

It would be nice if this wasn't necessary, but at the moment, using an SCM file finder will not improve the performance problems with egg_info, because `FileList.findall()` runs as well as the SCM file finder. AFAICS, this behaviour can't change, because:

1) Sometimes you don't want to package everything under source control
2) Sometimes you want to add something not under source control (e.g. a file generated from one under SCM).

Number 1 is explicitly talked about in the setuptools-git docs - https://pypi.python.org/pypi/setuptools-git

Is there actually a plan to change this and a way forward to solve these?

With this in mind, I can't see how it is possible to rely on an SCM file finder to 'provide' optimisation. The only thing I can think of is if we use '.gitignore' etc. files to avoid searching into directories. However, this is full of problems:

1) `.gitignore` files will probably list generated files/directories, but we might actually want these in distributed tarballs.
2) It could easily get complicated by SCM 'global' ignore files (e.g. I have `.tox` in my global `.gitignore`, not local)
3) It would require a lot of additional functionality in setuptools, interfacing with different SCMs and all the ways they might specify files to ignore.

It also seems a very strange dependency - why do we need an SCM tool to tell us not to bother looking in certain directories, when there wasn't a reason to look in them in the first place? With hindsight, "first get a list all the files in all the folders" was simply not a very good strategy for implementing the MANIFEST.in language, and the solution is simply "don't do that", which is what this patch implements.

jaraco:
> In what way is this an increased reliance on MANIFEST.in? 

If someone wants to gain the performance benefits of this patch, they will be required to create a MANIFEST.in, even if the project is relying on SCM info, and adds confounding factors to its purpose.

What I'd really like to see is a logic flow where file discovery techniques are clearly described and selected... something like a Discovery class with three implementations, one for Manifest.in, another for SCM, and another for the most simple heuristic (traversal including *.py files). And exactly one of those is chosen.

What worries me most is this logic feels interwoven with an already complicated implementation. Every time I look at it, I feel like I don't fully grasp all the implications. I worry the added complication will lead to more problems with the design down the road, and it won't be obvious where this functionality exists or why.

But I'm probably just overcautious. Do feel free to continue with the effort. We can roll it out and see what reception it gets.

timheap:
Many projects I work on contain files that are tracked in git that should not be packaged (such as docs, tests, etc), and files that are not tracked but should be packaged (such as compiled JS, CSS assets). Integrating with git or any other SCM would be counter productive for the projects I work on, and I suspect many others as well.

While the overhead in the projects I work on is not quite as extreme as for @spookylukey, it is still noticeable. A fix would be nice. I've also submitted a patch for pip, and another for setuptools (#733) for similar speed improvements. I was very happy to see a fix for this issue already submitted!

@jaraco I understand that you desire a more flexible solution to this problem, with multiple ways of specifying files to include. I've also struggled with the limitations of the MANIFEST file format. This PR solves a problem that people are experiencing right now though. Merging this PR would not prevent you from implementing your Discovery class in the future, and when the Discovery class is implemented this solution could be made more generic so more classes could reuse it.

Is there anything I can do to help move this PR along?

jaraco:
I believe this PR is a work in progress. My last message indicated that we should roll it out, so I'm waiting on a follow-up to say that it's ready to go.

@timheap, you can help by reconciling which of these two patches are preferable (or justifying how they're both desirable).

timheap:
@jaraco This patch and #733 address different performance problems in different areas, so they are both desirable. They affect unrelated sections of setuptools, the only similarity is the performance fix.

I'll have a look to see if I can help finish this PR as well. I installed the combination of this PR and mine and it made setuptools very speedy!

spookylukey:
@timheap thanks for the offer of help.

I'm unlikely to have time to look more at this in coming months. In terms of implementation, I don't think there was a huge amount more to do on this PR, although there is at least one TODO in there.

However, more tests would be nice. I'm afraid I can't remember how much of this code had test coverage. I think it probably needs more testing on other platforms.

spookylukey:
@jaraco thanks for your continued help/patience with this PR, by the way!

jaraco:
I'm going to close this PR, but to be sure, I'm welcoming further work on it. Just open a new one when you have more code for consideration.

Issue #398, opened 2015-06-19
Removal of escaping in shebang in pip install causes "failed to create process" on Windows


ghost:
Originally reported by: **joeyespo (Bitbucket: [joeyespo](http://bitbucket.org/joeyespo), GitHub: [joeyespo](http://github.com/joeyespo))**

---

I have Python installed in `C:\Program Files (x86)\Python` (for legacy reasons--it's kind of late to move it now), and any `console_script` I install from PyPI breaks now. Running the script causes a "failed to create process" error.

I noticed that older console_scripts were working, and turns out it's because they were quoted. Ex: `#!"C:\Program Files (x86)\Python\python.exe"`

Manually adding quotes around the newly installed `projectname-script.py` fixes it.

It looks like the culprit may have been the fix in #188. From that issue:

> If it is possible to detect being on a posix platform (probably not that easy?), we could just skip the escaping, as it is never beneficial. If this is not possible, maybe just skip the escaping all together? The benefit seems very platform specific.

Turns out skipping escaping all together just threw the problem back over the fence to Windows. Perhaps it'd be best to detect the platform? Escaping is indeed a platform-specific thing.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/398

tomduck:
See [`pip` Issue #2783][1] for a discussion of the same problem, including a workaround that can be used in `setup.py`.

[1]: https://github.com/pypa/pip/issues/2783

fkrull:
I did some digging into this issue; it's either a setuptools or distutils problem. I'll stick to cpplint, but any package using setuptools and entry points should show this behaviour. Things I have observed:

1) `pip install cpplint` installs the wheel and works fine.
2) `pip install --no-binary cpplint cpplint` **doesn't** quote the shebang properly. It installs from sdist by running `setup.py install --single-version-externally-managed` locally.
3) `setup.py install` from the source installs fine.
4) `setup.py install --single-version-externally-managed` from the source **doesn't** quote the shebang line properly.

I don't know what the egg installation code does different, but in [setuptools/command/install_scripts.py line 33](https://github.com/pypa/setuptools/blob/master/setuptools/command/install_scripts.py#L33) `exec_param` is retrieved from the build_scripts command which in turn gets it from the build command (`distutils/command/build_scripts.py` line 38) which, if unspecified, sets it to this (`distutils/command/build.py line 120):

    self.executable = os.path.normpath(sys.executable)

[...]

Issue #188, opened 2014-04-15
NT escaping in shebang in easy_install


ghost:
Originally reported by: **Anonymous**

---

Hi.
As far as I can see, there is still escaping of spaces in the shebang in the current version:
https://bitbucket.org/pypa/setuptools/src/0e53d50b76aabfdee7b6166593e41d41a16e6a25/setuptools/command/easy_install.py?at=default#cl-1546
I find this quite unfortunate, as this is of very limited use, but prevents some valid use cases.

The current code quotes all shebang strings that contain spaces. On all POSIX systems that I tried (current Ubuntu, ancient redhat, current OS X) this is illegal and simply doesn't work. Given the name of the function, it seems this was intended for windows NT (?).

However, this escaping prevents valid use cases, for example it doesn't allow the specification of "/bin/env /bin/python". I think there is a workaround for this specific case. However I have a use case where I need a different command to produce the correct python executable. This would be easy, except for the escaping.

If it is possible to detect being on a posix platform (probably not that easy?), we could just skip the escaping, as it is never beneficial. If this is not possible, maybe just skip the escaping all together? The benefit seems very platform specific.

Thanks,
Andy

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/188

PR #616, opened 2016-06-22
Fix "failed to create process" issue on Windows

fkrull:
Fix for issue #398. As I explained there (https://github.com/pypa/setuptools/issues/398#issuecomment-225444249), this is actually a setuptools issue. I added a test to check for the problem as well as the fix itself.

This seems to be only relevant to Windows; according to https://lists.gnu.org/archive/html/bug-bash/2008-05/msg00052.html, "Unix" (whatever that means in the context) doesn't have a way to escape spaces in the interpreter name so I didn't test for that.

### DIFF
--- a/setuptools/command/install_scripts.py
+++ b/setuptools/command/install_scripts.py
@@ -1,6 +1,7 @@
 from distutils import log
 import distutils.command.install_scripts as orig
 import os
+import sys
 
 from pkg_resources import Distribution, PathMetadata, ensure_directory
 
@@ -37,6 +38,10 @@ class install_scripts(orig.install_scripts):
         if is_wininst:
             exec_param = "python.exe"
             writer = ei.WindowsScriptWriter
+        if exec_param == sys.executable:
+            # In case the path to the Python executable contains a space, wrap
+            # it so it's not split up.
+            exec_param = [exec_param]
         # resolve the writer to the environment
         writer = writer.best()
         cmd = writer.command_spec_class.best().from_param(exec_param)
--- a/setuptools/tests/test_install_scripts.py
+++ b/setuptools/tests/test_install_scripts.py
@@ -1,0 +1,88 @@
+"""install_scripts tests
+"""
+
+import io
+import sys
+
+import pytest
+
+from setuptools.command.install_scripts import install_scripts
+from setuptools.dist import Distribution
+from . import contexts
+
+
+class TestInstallScripts:
+    settings = dict(
+        name='foo',
+        entry_points={'console_scripts': ['foo=foo:foo']},
+        version='0.0',
+    )
+    unix_exe = '/usr/dummy-test-path/local/bin/python'
+    unix_spaces_exe = '/usr/bin/env dummy-test-python'
+    win32_exe = 'C:\\Dummy Test Path\\Program Files\\Python 3.3\\python.exe'
+
+    def _run_install_scripts(self, install_dir, executable=None):
+        dist = Distribution(self.settings)
+        dist.script_name = 'setup.py'
+        cmd = install_scripts(dist)
+        cmd.install_dir = install_dir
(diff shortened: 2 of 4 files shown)


jaraco:
I very much appreciate the test and the implementation. How confident can we be that this fix doesn't cause a regression w.r.t. #188?

fkrull:
...that's a good point, actually; it does regress on that.

It's tricky because the 'executable' argument that install_scripts pulls from build_scripts can be either
1. a single command with spaces in its path (reasonably likely on Windows, not supported on Unix)
2. or a command plus arguments (possible on both).
And there's no way to tell which it is. The current behaviour always assumes 2, this PR right now always assumes 1 which breaks issue #188 again.

Something has to give. My preference would be 1 on Windows, 2 everywhere else; that makes it impossible to specify extra arguments on Windows, but Python installed in Program Files is more useful IMHO.

fkrull:
Ok, I went back and forth on it a bit and I think I came up with a better solution by only special-casing sys.executable. I added some more tests that check shebang quoting for Windows and Unix. I also ended up doing some simple test maintenance, so now all tests succeed on Windows with Python installed in Program Files (they didn't before).

fkrull:
Bump?

jaraco:
Sorry it's taken so long to get to this. And thanks for the ping.

Overall, this seems like a sound solution (much better than just vacillating between two broken implementations).

I'll make some comments inline to the commits.

jaraco:
Also, I appreciate the extra effort that you've put into this, given that the status quo is that the last change broke the tests on Windows. It's a shame that the tests haven't been run on Windows frequently enough to catch the former regressions.

setuptools/command/install_scripts.py
```diff
@@ -37,6 +38,10 @@ def run(self):
         if is_wininst:
             exec_param = "python.exe"
             writer = ei.WindowsScriptWriter
+        if exec_param == sys.executable:
```
jaraco: The whole point of creating the CommandSpec class was to try to encapsulate as much of these decision points as possible, so it's sad to see this one re-introduced. What's worse is that it seems to be special-casing a special case, inferring that if the string is `sys.executable`, then it was meant to be escaped, but in any other case it's not. This expectation would do the wrong thing if someone passed an executable with spaces that wasn't sys.executable (e.g. "C:\My Python Env\python.exe" where sys.executable is "C:\Python35\python.exe"). I'm not sure if there's something better do be done other than grin and bear it.

setuptools/tests/test_install_scripts.py
```diff
@@ -0,0 +1,88 @@
...
+        cmd = install_scripts(dist)
+        cmd.install_dir = install_dir
+        if executable is not None:
+            bs = cmd.get_finalized_command('build_scripts')
+            bs.executable = executable
+        cmd.ensure_finalized()
+        with contexts.quiet():
+            cmd.run()
+
+    @pytest.mark.skipif(sys.platform == 'win32', reason='non-Windows only')
```
jaraco: I dislike seeing all of these platform-specific tests. The apparent dichotomy here suggests that there are actually two behaviors of setuptools, one for Windows and another for other. Perhaps that makes sense, but as much as possible, I'd like for the behavior to be universal and tested on both platforms. That way, even though the behavior wouldn't be invoked in that form on that platform in the wild, it'll be tested by someone developing setuptools on that platform. Still, something that we can iterate on subsequently.

fkrull:
Regarding the `sys.executable` special-case: I couldn't come up with a way that didn't break in *some* situation. The optimal solution would probably be to extend build_scripts and handle the --executable argument in such a way that the user can specify a string with spaces or a list of strings (possibly with spaces). I'm not familiar enough with distutils to know if its argument parsing allows that or if that can be done in a backwards-compatible way.

PR #620, opened 2016-06-25
readme: update links to documentation

benoit-pierre:
Apparently the documentation as moved to [Read the Docs](http://setuptools.readthedocs.io/en/latest/).


jaraco:
I'm half-waiting to see if the docs get a final home at setuptools.pypa.io, but for now, best to have the code match reality.

PR #624, opened 2016-07-02
Improved support for MSVC discovery

jaraco:


jaraco:
@JGoutin I'm rolling this out as v24.0.0. Please watch for the release, test it in your environments, and let's resolve any emergent issues quickly.

JGoutin:
OK. I'll watch this.

PR #631, opened 2016-07-09
Add python_requires keywords to setup

xavfernandez:
This should allow setuptools to write the metadata Requires-Python and
Requires-External from PEP345 to the PKGINFO file

Looking for validation on the idea before improving it (docs and tests).
Are `python_requires` and `external_requires` the right keywords, or should they be `requires_python` and `requires_external` ?

This would enable me to revive https://github.com/pypa/pip/pull/3640 using `Requires-Python` info.

cc @dstufft


dstufft:
I don't think I would add the external stuff, I don't think it's generally useful and I think that we would want to avoid shipping that to people since it doesn't have much use.

Totally on board with adding ``Requires-Python`` though.

xavfernandez:
external_requires removed and 2 basic tests added.
Feedback welcome :-)

xavfernandez:
@jaraco any feedback ?

jayvdb:
Why is setuptools rejecting implementation of parts of an accepted PEP ?  Should the PEP be edited?

jaraco:
I haven't had a chance to review. I'll try to get to it soon. Ping me in a week if you haven't heard anything.

dstufft:
@jayvdb Because it was never implemented, will likely be replaced in the future, and is a needless footgun.

jaraco:
Looks good to me. Let's give it a go.

xavfernandez:
Oh
Happy to see it merged, but it looks like I did not push my latest change for the Metadata-Version update, I'll push it via an other PR in an hour.

PR #632, opened 2016-07-12
Fix quantity of blank lines after code object.

stepshal:


jaraco:
This seems like a reasonable change. Consistency is good. The changes to files in _vendor while acceptable are discouraged as they will be overridden when that package is refreshed and also, it makes the file checksums not match any version in the upstream package. The best thing to do for those would be to request these changes in the relevant upstream package and then they'll eventually get refreshed here.

setuptools/unicode_utils.py
```diff
@@ -4,6 +4,8 @@
 from setuptools.extern import six
 
 # HFS Plus uses decomposed UTF-8
+
```
jaraco: I think this change is wrong. In my experience, the comments should be adjacent to their relevant code. These additional new lines make the comment appear as a standalone construct.

setuptools/tests/test_bdist_egg.py
```diff
@@ -26,6 +27,7 @@ def setup_context(tmpdir):
 
 
 class Test:
+
```
jaraco: Is the double new line after a class decoration something that's part of the standard? I'm not familiar with it. Can you provide a reference?

setuptools/package_index.py
```diff
@@ -160,6 +163,8 @@ def interpret_distro_name(
         )
 
 # From Python 2.7 docs
+
```
jaraco: Here again, the whitespace betrays the meaning.

pkg_resources/_vendor/pyparsing.py
```diff
@@ -3162,6 +3233,8 @@ def z(*paArgs):
 #
 # global helpers
 #
+
```
jaraco: Here's a case where maybe the extra space makes sense. The comment isn't pertinent to the immediate following block but to the rest of the file.

pkg_resources/_vendor/pyparsing.py
```diff
@@ -766,11 +794,14 @@ def nullDebugAction(*args):
 
 # this version is Python 2.x-3.x cross-compatible
 'decorator to trim function calls to match the arity of the target'
+
```
jaraco: I'm pretty sure this string should be the do string for the function. Perhaps also the comment should be moved inside.

setuptools/tests/test_bdist_egg.py
stepshal: it's E309, but it is not included in standard of pep8.

stepshal:
@jaraco Done.

stepshal:
@jaraco Done.

stepshal:
@jaraco In this (https://github.com/pypa/setuptools/pull/648) PR those two line adjacent to their relevant code.

PR #638, opened 2016-07-13
Fix non-membership check.

stepshal:


jaraco:
I'm going to decline this one. The commit message doesn't seem to match the mostly trivial change... and furthermore, I don't even see the missing trailing newline in my checkout (though it does appear to be missing in Github).

stepshal:
@jaraco i mess up, change in commit was already merge, and in this PR supposed to be change like:
from
```py
assert (not 'without-baz' in dist.feature_negopt)
```
to 
```py
assert ('without-baz' not in dist.feature_negopt)
```
Make PR?

PR #643, opened 2016-07-13
Fix under-indented continuation line for visual indent.

stepshal:


jaraco:
I'm disinclined to accept this change for three reasons. 

The first is that these hanging indents rely on fixed width fonts and make the code look strangely indented with proportional width fonts. Text, including code, looks better in proportional width font, so I'd like not to lose support for that. 

Second, fonts aside, this form of indentation has limited editor support, often requiring the author to manually set and adjust the indentation, often over several lines. Some sophisticated editors can be configured to apply some form of this styling, but even then it often requires editor intervention.

Third, hanging indents often make code highly indented and strangely wrapped and sometimes necessarily lead to inconsistent forms. 

I've specifically been working to move away from this style, and these changes almost surely undo that work.

Thanks anyhow for the suggestion.

PR #650, opened 2016-07-14
Match closing bracket identation of opening bracket's line.

stepshal:


jaraco:
Those first three changes look funny to me. Is that what PEP-8 recommends for multiline function signature definitions?

stepshal:
@jaraco Seems so. I deleted them.

PR #654, opened 2016-07-14
Fix anomalous backslash in string.

stepshal:


jaraco:
This change feels like it's adding a lot of unnecessary syntax. I'm also unsure that the change to the PYPI regex won't break it, as making it a raw string affects the meaning:

```
>>> r'\\'
'\\\\'
>>> '\\'
'\\'
```

I've added a test to the suite that likely would have captured this expectation, but I don't see any way to ask Travis to re-run the pull request against the latest master.

In any case, overall, I feel this change detracts more than it adds.

PR #663, opened 2016-07-21
Fix continuation line unaligned for hanging indent.

stepshal:


setuptools/command/install_egg_info.py
```diff
@@ -85,7 +85,7 @@ def install_namespaces(self):
         "p = os.path.join(sys._getframe(1).f_locals['sitedir'], *%(pth)r)",
         "ie = os.path.exists(os.path.join(p,'__init__.py'))",
         "m = not ie and "
-            "sys.modules.setdefault(%(pkg)r, types.ModuleType(%(pkg)r))",
+        "sys.modules.setdefault(%(pkg)r, types.ModuleType(%(pkg)r))",
```
jaraco: This line was intentionally indented to give a visual cue that this string is a continuation of the previous one, and not just another string in the list. Without that cue, it would be an easy mistake to miss the lack of a comma.

setuptools/dist.py
```diff
@@ -823,7 +823,7 @@ class Feature:
     def warn_deprecated():
         warnings.warn(
             "Features are deprecated and will be removed in a future "
-                "version. See https://github.com/pypa/setuptools/issues/65.",
+            "version. See https://github.com/pypa/setuptools/issues/65.",
```
jaraco: Same issue here.

stepshal:
@jaraco Fixed.

PR #665, opened 2016-07-21
Also update the Metadata-Version when adding Requires-Python

xavfernandez:
Follow-up on #631 to also update the Metadata-Version to 1.2 (PEP 345).

This code seems cumbersome, I'm wondering if completely rewriting `write_pkg_file` would not be cleaner cc @jaraco @dstufft


xavfernandez:
Cf #667 for an alternate implementation

PR #666, opened 2016-07-21
Sort imports in same block alphabeticaly.

stepshal:


jaraco:
I'm somewhat reluctant on this one. While I see some value in sorting imports alphabetically, there's also value in imports being presented differently: in the order they were created or in an order that's more meaningful (such as grouping `import distutils.*` and `from distutils import ...`) or in an order that presents a different aesthetic (e.g. sorted by length).

I also worry that if some preference is adopted for import order that it will cause additional flux in the codebase subsequently, when imports are added "out of order" and then later re-ordered to comply.

How about we just leave these as they are?

Issue #671, opened 2016-07-22
Test failure in Appveyor around MSVC support


jaraco:
In the [new Appveyor builds](https://ci.appveyor.com/project/jaraco/setuptools), this test failure is seen:

```
________ TestModulePatch.test_no_registry_entryies_means_nothing_found ________
self = <setuptools.tests.test_msvc.TestModulePatch object at 0x0000000C9C29BBA8>
    def test_no_registry_entryies_means_nothing_found(self):
        """
            No registry entries or environment variable should lead to an error
            directing the user to download vcpython27.
            """
        find_vcvarsall = distutils.msvc9compiler.find_vcvarsall
        query_vcvarsall = distutils.msvc9compiler.query_vcvarsall
    
        with contexts.environment(VS90COMNTOOLS=None):
            with mock_reg():
                assert find_vcvarsall(9.0) is None
    
                expected = distutils.errors.DistutilsPlatformError
                with pytest.raises(expected) as exc:
>                   query_vcvarsall(9.0)
E                   Failed: DID NOT RAISE
setuptools\tests\test_msvc.py:88: Failed
```

@JGoutin Could you investigate?

JGoutin:
I will look at this. Theses tests were not updated for the new improved MSVC support.

The test wait a fail that was fixed by new behavior  at `setuptools/msvc.py` line 171.

PR #675, opened 2016-07-23
Fix test failure in Appveyor around MSVC support

JGoutin:
Fix for #671.

Exception will not raise because MSVC9 is find by new behavior in `setuptools/msvc.py` line 171. 

MSVC9 was found because [there is "Microsoft Windows SDK for Windows 7 and .NET Framework 3.5 SP1" on AppVeyor VM](http://www.appveyor.com/docs/installed-software).

I skipped the raise test in this case. But, it still work if there is no MSVC9 version installed.


jaraco:
Excellent. Thanks.

PR #680, opened 2016-07-24
docs/conf.py: format block comments.

stepshal:


jaraco:
This is all boilerplate that comes from sphinx. Seems silly to be tweaking it here in setuptools. Perhaps it makes sense to just remove all the unused boilerplate instead.

PR #682, opened 2016-07-24
Replace assert statement with self.assertTrue/False.

stepshal:


jaraco:
I much prefer the simple asserts (with the pytest 'rewrite asserts' functionality). It means that tests need not be derived from any base class and that the tracebacks are much clearer about the location of the issue (often when using a framework with `assertFalse`, the traceback appears rooted in that method and not where the assertion was made).

A better change would be to remove all of the `.assertSomething` methods in favor of direct asserts.

PR #695, opened 2016-07-27
unpin flake8<3dev

minrk:
pytest-flake8 0.6 has been released, fixing the bug that prompted the pin.

The pin currently makes setuptools master fail to install because pytest-flake8 requires flake8 >= 3, while setuptools requires flake8 < 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,9 +180,7 @@ setup_params = dict(
     scripts=[],
     tests_require=[
         'setuptools[ssl]',
-        'pytest-flake8',
-        # workaround for pytest-flake8 #7
-        'flake8<3dev',
+        'pytest-flake8>=0.6',
         'pytest>=2.8',
     ] + (['mock'] if sys.version_info[:2] < (3, 3) else []),
     setup_requires=[


jayvdb:
Not everyone can upgrade to flake8 v3 due to many bugs, and flake8 v3 explicitly breaks on py26.
IMO pytest-flake8 needs to support flake8 v2 also, otherwise it isnt usable for any projects that continue to support py26.

minrk:
Gotcha, in that case this would need to pin pytest-flake8 to < 0.6 as well as flake8 to < 3. However, it's odd that this runs successfully with flake8 3 on Python 2.6 on Travis, isn't it?

jayvdb:
It is quite odd.  Maybe this project is avoiding the single function which uses them `flake8.utils.parameters_for`, which is about plugins.

```
flake8 $ git grep OrderedDict
src/flake8/utils.py:        parameters = collections.OrderedDict([
src/flake8/utils.py:        parameters = collections.OrderedDict([
```

jaraco:
Sorry I missed this before committing my own fix at 93c69277ef35.

PR #698, opened 2016-07-27
only insert eggs ahead of parent if not already on path

### LINKED ISSUE #686: Difference in sys.path order, depending on how run
In the coverage.py test suite, I have tests that check various environmental values to see that they are they same for `coverage run myprog.py` as they are for `python myprog.py`.  One of the values is sys.path.  Recently, these tests started failing because a .egg file I installed with easy_install would appear in the middle of sys.path for coverage, but at the end for python.

I'm guessing this has something to do with #674, but I don't understand why.

The actual diff of sys.path for the two environments is:
```
      "path": [
          "/private/var/folders/j2/gr3cj3jn63s5q8g3bjvw57hm0000gp/T/test_cover_52005401",
          "/Users/ned/coverage/trunk/tests/modules",
          "/Users/ned/coverage/trunk/tests/zipmods.zip",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python27.zip",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/plat-darwin",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/plat-mac",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/plat-mac/lib-scriptpackages",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/lib-tk",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/lib-old",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/lib-dynload",
          "/usr/local/pythonz/pythons/CPython-2.7.11/lib/python2.7",
          "/usr/local/pythonz/pythons/CPython-2.7.11/lib/python2.7/plat-darwin",
          "/usr/local/pythonz/pythons/CPython-2.7.11/lib/python2.7/lib-tk",
          "/usr/local/pythonz/pythons/CPython-2.7.11/lib/python2.7/plat-mac",
          "/usr/local/pythonz/pythons/CPython-2.7.11/lib/python2.7/plat-mac/lib-scriptpackages",
-         "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/site-packages/covtestegg1-0.0.0-py2.7.egg",
          "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/site-packages",
          "/Users/ned/coverage/trunk",
+         "/Users/ned/coverage/trunk/.tox/py27/lib/python2.7/site-packages/covtestegg1-0.0.0-py2.7.egg",
      ]
```

Is this expected? Is there something I can do to get the identical behavior in the two environments?

minrk:
in `insert_on(replace=False)`

and set `replace=False` by default in `activate()`

Fixes pkg_resources modifying sys.path for all eggs with SETUPTOOLS_SYS_PATH_TECHINQUE=raw

There's one failing test where setup_requires should be loading a package at higher priority, but the `activate(replace=False)` change means it is showing up at lower priority, rather than higher.

The setup_requires code triggered in this test does indeed have several explicit `replace=True`, `remove_conflicting` flags, but the only one that seems to matter with respect to sys.path is the *implicit* one passed to `.activate()`, which is registered as a zero-argument callback in `self.callbacks`. I'll see if I can figure out a way to pass the same `replace` argument in `.add` to `.activate`.

closes #686

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2503,11 +2503,11 @@ class Distribution(object):
             for line in self.get_metadata_lines(name):
                 yield line
 
-    def activate(self, path=None):
+    def activate(self, path=None, replace=False):
         """Ensure distribution is importable on `path` (default=sys.path)"""
         if path is None:
             path = sys.path
-        self.insert_on(path, replace=True)
+        self.insert_on(path, replace=replace)
         if path is sys.path:
             fixup_namespace_packages(self.location)
             for pkg in self._get_metadata('namespace_packages.txt'):
@@ -2585,7 +2585,24 @@ class Distribution(object):
         return self.get_entry_map(group).get(name)
 
     def insert_on(self, path, loc=None, replace=False):
-        """Insert self.location in path before its nearest parent directory"""
+        """Ensure self.location is on path
+
+        If replace=False (default):
+            - If location is already in path anywhere, do nothing.
+            - Else:
+              - If it's an egg and its parent directory is on path,
+                insert just ahead of the parent.
+              - Else: add to the end of path.
+        If replace=True:
+            - If location is already on path anywhere (not eggs)
+              or higher priority than its parent (eggs)
+              do nothing.
+            - Else:
+              - If it's an egg and its parent directory is on path,
+                insert just ahead of the parent,
+                removing any lower-priority entries.
+              - Else: add it to the front of path.
+        """
 
         loc = loc or self.location
         if not loc:
@@ -2600,6 +2617,9 @@ class Distribution(object):
                 break
             elif item == bdir and self.precedence == EGG_DIST:
                 # if it's an .egg, give it precedence over its directory
+                # UNLESS it's already been added to sys.path and replace=False
+                if (not replace) and nloc in npath[p:]:
+                    return
                 if path is sys.path:
                     self.check_version_conflict()
                 path.insert(p, loc)
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2620,7 +2620,11 @@ class Distribution(object):
 
         for p, item in enumerate(npath):
             if item == nloc:
-                break
+                if replace:
+                    break
+                else:
+                    # don't modify path (even removing duplicates) if found and not replace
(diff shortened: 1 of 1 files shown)


minrk:
I've added code to passthrough the `replace` argument from `.add` to callbacks, which fixes the setup_requires test but I'm not sure it's the right thing to do. It depends, to some degree, on what `.activate()` is meant to do, and when it is meant to be called. In master, it's called for every distribution found on sys.path and added with `dist.add()`, and its behavior is to ensure that every package added has a high import priority. The first commit on this PR is to call it on the same packages, but never modify import priority if packages are found.

Another behavior would be to only activate 'new' packages with high priority, and leave those found by default as lower priority

minrk:
This now calls `dist.activate(replace=True)` for runtime-added dists (same as previous behavior), only modifying the activation of the initial set, loading with `replace=False`, so that sys.path is left completely unmodified for normal behavior.

I'm still not confident about what circumstances there are in which packages are activated beyond the initial set, so this may not be doing the right thing in every case. It seems to me that `sys.path` should not be modified in any way, unless:

1. the package is certainly not on sys.path
2. there are known conflicts with other packages already on sys.path

This PR does not go that far, but it would be nice if ultimately pkg_resources modification of sys.path could be confined to clearly defined, explicit, rare cases.

jaraco:
Thanks @minrk for looking at this. Also thanks to @embray for creating the tests to protect the `setup_requires` expectation. It does feel like we're slowly refining to the behavior that's desirable, though I do worry about boolean flags proliferating. As you say, we can leave a refactor to a more explicit, clear implementation as a later exercise.

embray:
> This PR does not go that far, but it would be nice if ultimately pkg_resources modification of sys.path could be confined to clearly defined, explicit, rare cases.

Agreed. I think most of the use case for this is packages that are loaded via the `setup_requires` mechanism later in the setup process.  And in *principle* this could happen any time but maybe we should just put our foot down and limit the contexts where that can actually happen.

PR #699, opened 2016-07-27
Fix regression in commit e404a4a

JGoutin:
Missing `,`.


jaraco:
Thanks for catching and correcting my mistake.

PR #708, opened 2016-07-30
WIP: Add new long_description_content_type kwarg

msabramo:
**Work in progress**

This is used to populate the newly proposed `Description-Content-Type` distribution metadata field, that is proposed in https://github.com/pypa/python-packaging-user-guide/pull/258.

**We can discuss and evolve the implementation, but I don't think this should be merged until https://github.com/pypa/python-packaging-user-guide/pull/258 is accepted.**
### Related:
- https://github.com/pypa/python-packaging-user-guide/pull/258 (Add `Description-Content-Type` field)
- https://github.com/pypa/readme_renderer/pull/3 (Add Markdown support)
- https://mail.python.org/pipermail/distutils-sig/2016-May/028689.html
### Branch

https://github.com/msabramo/setuptools/tree/long_description_content_type

Cc: @jaraco, @dstufft, @ncoghlan, @pfmoore, @nicktimko

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,6 +94,7 @@ setup_params = dict(
     author="Python Packaging Authority",
     author_email="distutils-sig@python.org",
     long_description=long_description,
+    long_description_content_type='text/x-rst; charset=UTF-8',
     keywords="CPAN PyPI distutils eggs package management",
     url="https://github.com/pypa/setuptools",
     src_root=None,
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -397,6 +397,10 @@ def write_pkg_info(cmd, basename, filename):
         metadata = cmd.distribution.metadata
         metadata.version, oldver = cmd.egg_version, metadata.version
         metadata.name, oldname = cmd.egg_name, metadata.name
+        metadata.long_description_content_type = getattr(
+            cmd.distribution,
+            'long_description_content_type'
+        )
         try:
             # write unescaped data to PKG-INFO, so older pkg_resources
             # can still parse it
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -67,6 +67,13 @@ def _patch_distribution_metadata_write_pkg_file():
         if self.download_url:
             file.write('Download-URL: %s\n' % self.download_url)
 
+        long_desc_content_type = getattr(
+            self,
+            'long_description_content_type',
+            None
+        ) or 'UNKNOWN'
+        file.write('Description-Content-Type: %s\n' % long_desc_content_type)
+
         long_desc = rfc822_escape(self.get_long_description())
         file.write('Description: %s\n' % long_desc)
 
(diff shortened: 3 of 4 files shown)


jaraco:
Given that the code is in `dist.py`, I would expect the test to be in test_dist.py, but I see no such test. It may be the case that you'll be writing the first tests for this. Of course, dist.py is a patch of distutils.dist, so maybe there are tests in the stdlib that apply.

It's an old package, so I'd say there's lots of cruft here - don't assume because you find it that it's recommended. Rely on pytest docs and your best instincts to put together the best tests you can. Thanks for all the work on this.

xavfernandez:
For `Requires-Python` and `Metadata-Version` I put the tests in https://github.com/pypa/setuptools/blob/d079163/setuptools/tests/test_egg_info.py#L213-L237

Which makes me think: is there a PEP defining this new field ? Should the `Metadata-Version` be bumped also ?

msabramo:
There is no PEP, but there is a PR to update the "Python Packaging Guide" (https://github.com/pypa/python-packaging-user-guide/pull/258). I think @ncoghlan said on the distutils-sig mailing list that this was the proper place now to propose new metadata fields. I'm not sure if that's a precursor to creating a PEP or replaces it? Also I'm not sure whether adding one optional field requires the metadata version to be bumped or not. I'll let someone else chime in here on the proper procedure and I'll follow whatever it is. 

Thanks for the test pointers!

msabramo:
Added a simple test in bd5403e, borrowing from the tests that @xavfernandez and then simplifying a bit.

ncoghlan:
Yep, the PR-to-PyPUG is sufficient to add the new field (once accepted) - my view is that bumping the spec version mainly matters if we change the meaning of an *existing* field, since new fields are trivially detected by checking for them directly and having a defined fallback assumption to use if the field isn't found. 

More generally, https://github.com/pypa/pypa.io/issues/11 covers the status of the (currently stalled) efforts to both streamline and properly document the way we use the PEP process as part of PyPA interoperability discussions.

msabramo:
It doesn't look like the AppVeyor failure is related to my change, but not 100% sure.

jaraco:
Indeed - tests are [now passing](https://ci.appveyor.com/project/jaraco/setuptools/build/1.0.94).

msabramo:
Cool, thanks @jaraco!

jaraco:
@msabramo Is this still a work in progress?

msabramo:
Possibly. I need to check the latest discussions on this idea which might be spread out on a few PRs and emails. And I'll be AFK for a few days so I won't have time to do it right away but it's on my list to do.

nicktimko:
As mentioned, this would follow
https://github.com/pypa/python-packaging-user-guide/pull/258 which also
looks lonely

msabramo:
Yep. I need to check all the things and won't be able to for a bit. 

If anyone is itching to help, they can send PRs to my branches for any changes folks suggested.

msabramo:
Yep. I need to check all the things and won't be able to for a bit. 

If anyone is itching to help, they can send PRs to my branches for any changes folks suggested.

jaraco:
I'm going to close this to avoid re-reading it, though I gladly welcome anyone to revive the effort in a new PR.

nicktimko:
Is there something wrong with *this* PR or just that it is dependent on a spec change that's still TBD? Just trying to improve your open ticket stats? :wink:

jaraco:
Nothing specifically wrong with it, but it does appear to be in need of some attention. Instead, it's just periodically distracting me. I like to reserve open PRs for work that's being actively developed or ready for review.

PR #717, opened 2016-08-03
Fix logging using arguments instead of formatter

vallsv:
Hi. While i was working around setuptools, here is very small clean up about code quality.
In case you take care of it.
Regards

### DIFF
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -1038,8 +1038,7 @@ def open_with_auth(url, opener=urllib.request.urlopen):
         cred = PyPIConfig().find_credential(url)
         if cred:
             auth = str(cred)
-            info = cred.username, url
-            log.info('Authenticating as %s for %s (from .pypirc)' % info)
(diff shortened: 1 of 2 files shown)


setuptools/package_index.py
```diff
@@ -1038,8 +1038,7 @@ def open_with_auth(url, opener=urllib.request.urlopen):
         cred = PyPIConfig().find_credential(url)
         if cred:
             auth = str(cred)
-            info = cred.username, url
-            log.info('Authenticating as %s for %s (from .pypirc)' % info)
+            log.info('Authenticating as %s for %s (from .pypirc)', cred.username, url)
```
jaraco: I think this line is now too long.

jaraco:
The rest of the changes are great, so I'll take the good with the bad. Thanks.

vallsv:
you can use `*info` if you prefer. I can send a patch.

vallsv:
Oh it is already done :-D
Thanks for the merge.

PR #718, opened 2016-08-04
use abi3 extension if Extension().is_abi3

dholth:
Add an Extension(is_abi3=True) property that can be used to set its filename to something that will be loadable by more than one version of Python 3. An '.abi3.so' extension on Linux, just .pyd on Windows. Tested on Linux so far.

It will still be necessary to `-DPy_LIMITED_API=0x03030000` somewhere for the extension to truly be abi3.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,14 @@ v25.1.3
 * #714 and #704: Revert fix as it breaks other components
   downstream that can't handle unicode. See #709, #710,
   and #712.
+* Add Extension(py_limited_api=True). When set to a truthy value,
+  that extension gets a filename apropriate for code using Py_LIMITED_API.
+  When used correctly this allows a single compiled extension to work on
+  all future versions of CPython 3.
+  The py_limited_api argument only controls the filename. To be
+  compatible with multiple versions of Python 3, the C extension
+  will also need to set -DPy_LIMITED_API=... and be modified to use
+  only the functions in the limited API.
 
 v25.1.2
 -------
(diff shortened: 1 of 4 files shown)


jaraco:
This patch appears to be failing on Python 3.

jaraco:
Also, before adding a boolean parameter, is there a more descriptive parameter that could be added? Something like `binary_interface='abi3'` instead of `is_abi3=True`?

dholth:
How about py_limited_api like the #ifdef, it could even be set to a version. py_limited_api = (3,3,0) or py_limited_api = 0x03030000.

jaraco:
`py_limited_api` sounds fine. I think I like `(3,3,0)` better as it's more pythonic.

dholth:
What if we let truthy set the extension, then reserve the tuple for future use (smart choice of wheel tags?) Similar to how 1 sets the basic feature in C.

jaraco:
It feels to me like `abi3` appears to much in the implementation. Can we remove the `3` so that the implementation would work for future Pythons with a new ABI version?

dholth:
I think abi3 is correct. The idea is that abi3 with Python 3.3 features works on all later versions of Python (it was introduced in py3.2, and 3.3 specifically adds some functions that were forgotten). (The extension doesn't say anything about the minimum Python ver

dholth:
The other default abi is of course the "exactly this version of Python" default.

jaraco:
Hi dholth. Sorry I missed your IRC messages when you sent them. Is this ready to go? Would you add a changelog entry for it?

dholth:
It's ready to go if you don't have any better ideas for the name (which I think is a reasonable name). Would be happy to add a changelog entry.

Issue #668, opened 2016-07-21
Option to skip extras requires


jakirkham:
Currently we find ourselves needing to patch the `setup.py` file of `setuptools` as it has `extras_require` and `dependency_links`, which don't agree with our package manager. We can provide these during the build. However, we find that `setuptools` tries to download them anyways. It would be nice if we had an option to disable these settings during the install or at least recognize that they are already accounted for.

cc @msarahan @pelson @ocefpaf

jakirkham:
See this PR ( https://github.com/pypa/setuptools/pull/721 ) for a potential solution to this problem.

jaraco:
Setuptools doesn't try to download these when I build/install it. Can you share more about what use case is triggering the undesirable behavior?

jaraco:
I'm not sure I follow. I understand you want to disable setuptools from doing any downloads or installs. I'm not sure you can achieve this goal in the general case, because many projects rely on features like `setup_requires` to load requirements at build time.

I still don't understand under what circumstances the extras_requires declared in setuptools are being invoked in the conda environment.

jakirkham:
So, we actually don't find too many issues with our current strategy of disabling downloads (excepting this case with `setuptools`). As we are able to specify [`build` requirements]( http://conda.pydata.org/docs/building/meta-yaml.html#build ) to `conda`, we can satisfy `setup_requires` before any build step has occurred. Again it would be nice if we can agree on some better way to suppress downloads by `setuptools` (perhaps some option provided in a config file). However, this is a separate discussion and probably merits its own issue.

> I still don't understand under what circumstances the extras_requires declared in setuptools are being invoked in the conda environment.

When I started writing this, I was trying to demonstrate and thoroughly explore the problem. However, it is now clear that adding the arguments `--single-version-externally-managed --record=record.txt` makes this unnecessary. I decided to keep this in the post here as it may be of historic value.

[...]

PR #721, opened 2016-08-05
WIP: Optionally skip extra requires

jakirkham:
Fixes https://github.com/pypa/setuptools/issues/668

Allows one to optionally leave out the `extras_require` and `dependency_links` options by setting `SETUPTOOLS_DISABLE_EXTRAS_REQUIRE` or `DISTRIBUTE_DISABLE_EXTRAS_REQUIRE` to `1`. This should make it easier for package maintainers to properly separate `setuptools` from these dependencies and maintain them separately.

cc @msarahan @pelson @ocefpaf

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,44 @@ def read_commands():
     return command_ns['__all__']
 
 
+def _extras_require():
+    # Gentoo distributions manage the python-version-specific scripts
+    # themselves, so those platforms define an environment variable to
+    # suppress the creation of the version-specific scripts.
+    var_names = (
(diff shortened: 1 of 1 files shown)


jaraco:
I'm reluctant to accept this because it provides a mechanism to create alternate builds of setuptools with different metadata. Extras are meant to be optional, so I wouldn't expect them to be installed unless opted into. Still, I'll await more details in the bug report to better understand what this PR is trying to achieve.

jakirkham:
Closing as further exploration of the underlying issue has demonstrated the problem can be resolved without this or any other patch for that matter.

PR #715, opened 2016-08-02
#694 Patch distutils._msvccompiler.library_dir_option

JGoutin:
Try to fix #694.

Wait result from issue before merging.

PR #723, opened 2016-08-05
quote library_dir_option after calling unpatched version

### LINKED ISSUE #722: setuptools 25.1.4 adds extra quotes on LIBPATH
#715's quoting has caused my appveyor builds to start failing with 25.1.4, due to some extra quoting.

With setuptools 25.1.3, builds [succeed](https://ci.appveyor.com/project/minrk/pyzmq/build/master+97/job/b5xoge86se10eh3a), and with 25.1.4 they [fail](https://ci.appveyor.com/project/minrk/pyzmq/build/master+98/job/80o5mu31egb6s946).

The key difference:

on 25.1.3, the link command had args like:

    "/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\LIB"

on 25.1.4, they look like:

    "/LIBPATH:"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\LIB""

(note the extra quotes).

These are CPython 3.5 x86 builds on appveyor; I don't know what other versions are affected.

minrk:
rather than before (#715), which should avoid double-quotes if the unpatched function does its own quoting.

I *believe* this closes #722

### DIFF
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -233,10 +233,11 @@ def msvc14_library_dir_option(self, dir):
     ------
     "\LIBPATH" argument: str
     """
-    if ' ' in dir and '"' not in dir:
+    opt = unpatched['msvc14_library_dir_option'](self, dir)
+    if ' ' in opt and '"' not in opt:
         # Quote if space and not already quoted
-        dir = '"%s"' % dir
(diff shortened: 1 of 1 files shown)


anthrotype:
I confirm that works for me, thanks! 👍

Issue #694, opened 2016-07-26
Linking ext on Windows Python 3.5 do not quote libpath params


vallsv:
Hi.

I have a problem to link an extension from a project.

```LINK : fatal error LNK1181: cannot open input file 'Files.obj'```

It fail with python 3.5 and setuptools > 23 (all versions of branch 24 and 25 fail).

Here it fail
```
(venv) >pip list
cycler (0.10.0)
Cython (0.24.1)
fabio (0.4.0)
h5py (2.6.0)
lxml (3.6.0)
matplotlib (1.5.1)
numpy (1.11.1+mkl)
Pillow (3.3.0)
pip (8.1.2)
pyparsing (2.1.5)
python-dateutil (2.5.3)
pytz (2016.6.1)
scipy (0.17.1)
six (1.10.0)
wheel (0.29.0)

(venv) >pip install setuptools
Collecting setuptools
  Using cached setuptools-25.1.0-py2.py3-none-any.whl
Installing collected packages: setuptools
Successfully installed setuptools-25.1.0

(venv) >python setup.py build
running build
running build_py
running build_ext
No module named 'numpy.distutils._msvccompiler' in numpy.distutils; trying from
distutils
building 'pyFAI.ext.splitPixelFullCSR' extension
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /
nologo /Ox /W3 /GL /DNDEBUG /MD -IpyFAI\ext\include -IpyFAI\ext -IC:\Users\valls
\pyFAI_venv\lib\site-packages\numpy\core\include -Ic:\python35-amd64\include -Ic
:\python35-amd64\include -I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\
VC\INCLUDE" -I"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INC
LUDE" -I"C:\Program Files (x86)\Windows Kits\10\include\10.0.10586.0\ucrt" -I"C:
\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" -I"C:\Program Files
 (x86)\Windows Kits\10\include\10.0.10586.0\shared" -I"C:\Program Files (x86)\Wi
ndows Kits\10\include\10.0.10586.0\um" -I"C:\Program Files (x86)\Windows Kits\10
[...]

PR #725, opened 2016-08-05
revert `library_dir_option` patch.

JGoutin:
Revert patch on `distutils._msvccompiler.MSVCCompiler.library_dir_option`.

See comments on #694.


anthrotype:
Thanks. I tried again to build brotli from my local Windows machine (Python35-x86) using the latest setuptools tag `v25.1.6`, and it finally works.
This time, I made sure the correct setuptools was imported before I can say "it works". Sorry for the confusion.

Issue #612, opened 2016-06-09
License clarification


jakirkham:
Sorry, I think I'm missing something really obvious and could use some help understanding. I think `setuptools` use to be licensed with PSF or ZPL. However, I do not actually see a license file here. All I see is a [classifier]( https://github.com/pypa/setuptools/blob/cb0a2aeb86953d6b3cf6f039bb887fb8c7e8d1e9/setup.py#L132 ) that suggests this is MIT. Is this MIT or something else? Also could we please have a license file here to make that a bit clearer? Thanks.

jaraco:
See #132. The classifier isn't a suggestion but a declaration and follows the distutils guide for declaring a license. I consider inclusion of a license file redundant and error prone.

jakirkham:
Thanks for clarifying.

I'm sure that you guys do a good job of keeping the classifiers up-to-date, not all Python software does though. So that's why I wanted to double check.

Ultimately, it is your call what you do in terms of a license file. IMHO it makes this information easier to find and clearer. For example, who has the copyright on `setuptools`? While I can take a decent guess, this is not something I want to be doing when it comes to licensing.

-------------------------------------------------------------------------------------------------------------------------------------------
As you are higher up in the community and have a chance of influencing this, I figure I should raise this somewhat orthogonal point. Trove classifiers for licenses can be a bit vague in some case. While this doesn't affect MIT so much, it does matter when someone adds the BSD trove classifier, which appears to be totally unversioned AFAICT. Someone could be using this for a BSD 4-Clause license and we would have no way of knowing. While this could be specified with the `license` field, this is sadly rarely the case. It might be nice to clear these up a bit.

jakirkham:
Also would like to point out issue ( https://github.com/pypa/setuptools/issues/357 ). Technically our packaging at conda-forge is non-compliant with the MIT license terms, but I really don't know what to do other than create our own copy of the MIT license and assign the copyright to...PyPA?

glyph:
One of the other issues with licensing under the MIT license but leaving out the MIT license itself is that the _main point_ of the MIT license is this provision:

```
The above copyright notice and this permission notice shall be
included in all copies or substantial portions of the Software.
```

If you don't actually include the license file, then it's not clear how anyone distributing it is supposed to comply.

sigmavirus24:
[...]

PR #730, opened 2016-08-12
Add LICENSE file for expectant auditors

sigmavirus24:
The classifier in setup.py should be enough for most auditors, but some
have automated tooling that checks for LICENSE files and reads those.
This helps all auditors (downstream redistributors and users) who can
not just rely on the classifier.

Related to #612

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 CHANGES
 =======
 
+v25.2.0
+-------
+
+* #612 via #730: Add a LICENSE file which needs to be provided by the terms of
+  the MIT license.
+
 v25.1.6
 -------
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,19 @@
+Copyright (C) 2016 Jason R Coombs <jaraco@jaraco.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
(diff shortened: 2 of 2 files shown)


sigmavirus24:
Ah, forgot the changelog entry.

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ v25.1.6
 -------
 
 * #725
+* #612 via #730: Add a LICENSE file which neds to be provided by the terms of
```
nicktimko: needs?

jaraco: Also, this entry should go in a new section for v25.2.0.

sigmavirus24: Done & done.

jakirkham:
Do you think there is any value to having a `.txt` extension for the license? There seem to be many authors that do this. Particularly to ensure that Windows is able to open this file correctly.

sigmavirus24:
@jakirkham not really? Almost every project I've ever seen has had the license file sans `.txt`.

jakirkham:
Just to provide some examples, NumPy [does]( https://github.com/numpy/numpy/blob/e1f191c46f2eebd6cb892a4bfe14d9dd43a06c4e/LICENSE.txt ). Similarly so does [SciPy]( https://github.com/scipy/scipy/blob/d2b4e1786bc4737849695c67f32fedd81ec9466b/LICENSE.txt ) and [Cython]( https://github.com/cython/cython/blob/0dc185de9842e1e05b6448ca45ef0e8083f35daa/LICENSE.txt ). If one looks around a bit, one can find many more examples. Though it is true there are examples without the extension.

sigmavirus24:
Yeah, we can both dig up numerous examples either way both inside and outside of the Python ecosystem. I don't think it really matters and this (merged) pull request certainly isn't the place for a discussion of this type (whether or not to put .txt at the end of a LICENSE file name because some operating systems do weird things in an attempt to be *smart*).

PR #732, opened 2016-08-13
fix for extra names containing '-'

mindw:
This is variation of #578 with a more modest goal.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1428,7 +1428,7 @@ def safe_extra(extra):
     Any runs of non-alphanumeric characters are replaced with a single '_',
     and the result is always lowercased.
     """
-    return re.sub('[^A-Za-z0-9.]+', '_', extra).lower()
+    return re.sub('[^A-Za-z0-9.-]+', '_', extra).lower()
 
 
 def to_filename(name):
@@ -2807,8 +2807,8 @@ class DistInfoDistribution(Distribution):
         dm[None].extend(common)
 
         for extra in self._parsed_pkg_info.get_all('Provides-Extra') or []:
-            extra = safe_extra(extra.strip())
-            dm[extra] = list(frozenset(reqs_for_extra(extra)) - common)
+            s_extra = safe_extra(extra.strip())
+            dm[s_extra] = list(frozenset(reqs_for_extra(extra)) - common)
 
         return dm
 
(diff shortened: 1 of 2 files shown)


jaraco:
Thanks for this and especially thanks for the test.

This change feels like it's shifting the spec for what is a valid extra name. A change like this probably should be reviewed by a larger audience. Has this been discussed elsewhere? Is there a reason that this shift is the obvious best choice?

Also, if/when we decide to move forward with this, would you include a CHANGES.rst entry describing for users the effect of the change?

mindw:
I'm a bit at a loss here - this looks to me like a pure bugfix:
1. setuptools fails to record extras with '-' - such as `faster-signatutres`.
2. setuptools fails to resolve `faster-signatures`.

Both items are required by pep508 and pep 426 (even though it's deferred). A good "in the wild" can be found at https://bitbucket.org/pypa/wheel/src/cf4e2d98ecb1f168c50a6de496959b4a10c6b122/setup.py?fileviewer=file-view-default .

jaraco:
Oh, great. That PEP provides an excellent authoritative reference to indicate from whence the change is derived.

I see how from one perspective this feels like a bugfix, but from another perspective, the previous, established implementation seems to think that it will always normalize by replacing certain characters with underscores.

I wonder if accompanying this change if [check_extras](https://github.com/pypa/setuptools/blob/master/setuptools/dist.py#L160-L174) should be updated to restrict extra names to match the PEP requirements.

Still, if you want to leave that for others to address later, that's fine. Just add a CHANGES.txt entry for this change, referencing the PEP, and we're good.

mindw:
As you correctly observed `Foo[baz_lightyear]` will fail lookup with this patch. It is a regression.

To move forward can you please assist me in figuring out what is the correct behavior of looking up extras as pep426 states that :

> The names of extras MUST abide by the same restrictions as those for distribution names

and
> All comparisons of distribution names MUST be case insensitive, and MUST consider hyphens and underscores to be equivalent.

while pep508 defines the extras as:
```
identifier_end = letterOrDigit | (('-' | '_' | '.' )* letterOrDigit)
identifier    = < letterOrDigit identifier_end* >
name          = identifier
extras_list   = identifier:i (wsp* ',' wsp* identifier)*:ids -> [i] + ids
extras        = '[' wsp* extras_list?:e wsp* ']' -> e
```
But makes not statement about how extras are looked up (AFAIK).

Thanks!

jaraco:
> can you please assist me in figuring out what is the correct behavior of looking up extras

I'm afraid I'm just not finding the time to dig into this. I'm inclined to just accept it and let subsequent requests improve on this implementation as necessary.

PR #734, opened 2016-08-14
Update tox.ini envlist to match supported Pythons

timheap:
Pythons 3.1, 3.2 are no longer supported, but Python 3.5, Pypy, and
Pypy3 are.


jaraco:
I think it might be the case that Python 3.2 is still supported (and I just prematurely removed the trove classifier when 3.2 support was considered deprecated). No matter, tox.ini isn't officially supported, so whatever the community agrees on is fine.

PR #737, opened 2016-08-16
Add tests for MANIFEST.in

timheap:
I was looking in to finishing #610, but found that there were no tests for `MANIFEST.in` that I could find. As setuptools customises the FileList a little bit, and #610 customises it some more, some tests for this would be sensible.

I wrote some tests that work much like `setuptools/tests/test_sdist.py`, by creating some files and running with various MANIFEST.in files.

I also copied the FileList tests from distutils, modifying them slightly to fit with py.test and the setuptools environment.

This way, it will be easy to verify that any modifications done to setuptools retain compatibility with distutils.

### DIFF
--- a/setuptools/tests/test_manifest.py
+++ b/setuptools/tests/test_manifest.py
@@ -6,8 +6,10 @@ import os
 import shutil
 import sys
 import tempfile
+from distutils import log
+from distutils.errors import DistutilsTemplateError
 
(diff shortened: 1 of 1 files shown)


timheap:
OK, got the tests passing. Compatibility across so many versions and platforms is apparently tricky!

jaraco:
Awesome. Thanks!

Issue #728, opened 2016-08-08
Windows regression: wrong LIBPATH quoting on command-line (with numpy.distutils)


pitrou:
During the upgrade from setuptools 23.0.0 to 25.1.6, it seems quoting of arguments in Windows command lines got broken (specifically when linking C extensions), which you can witness on those two AppVeyor builds:

* ok with 23.0.0: https://ci.appveyor.com/project/pitrou/numba/build/1.0.1220#L185
* fail with 25.1.6: https://ci.appveyor.com/project/pitrou/numba/build/1.0.1227#L185

This seems to be a setuptools issue since it's the only piece of the environment that changed between the two builds.

anthrotype:
Oh no, again? :(
Supposedtly the behaviour was reverted with https://github.com/pypa/setuptools/pull/725

pitrou:
Well, if you look at the failing build, the quoting is missing, which the reverted patch was supposed to fix (?). But we also use `numpy.distutils`, so perhaps there's an interaction with that?

anthrotype:
> But we also use numpy.distutils, so perhaps there's an interaction with that?

Yes, I think that's it. In the google/brotli project, reverting the patch fixed the issue. Brotli does not use numpy.distutils. 

https://ci.appveyor.com/project/quixdb/brotli/build/1.0.26

pitrou:
Well, we build against the Numpy C API, so it's not possible for us to disable `numpy.distutils` (which seems to monkeypatch itself into distutils as soon as it's imported...).

rgommers:
The only relevant recent change is to add ``msvc9`` to the list of compiler classes that use a patched``gen_lib_options``, in https://github.com/numpy/numpy/commit/e149fac25eefcc93c821ea5cca261aa55abd1f7f (in numpy 1.11.1).

For other compiler classes nothing has changed in ~10 years: https://github.com/numpy/numpy/blame/master/numpy/distutils/ccompiler.py#L602

vallsv:
Here we had the same talk https://github.com/pypa/setuptools/issues/694.
I think the problem is on numpy. They dont need anymore to monkey-patch some function to quote things.

pitrou:
> I think the problem is on numpy. They dont need anymore to monkey-patch some function to quote things.

That doesn't sound like a plausible explanation. If you look at the logs, the builds fail because the arguments are not quoted anymore.

vallsv:
[...]

PR #739, opened 2016-08-16
numpy.distutils and distutils._msvccompiler compatibility

JGoutin:
- Fix compatibility between `numpy.distutils` and `distutils._msvccompiler`. See #728 : Setuptools 24 `msvc.py` improvement import `distutils._msvccompiler` (New Python 3.5 C compiler for MSVC >= 14), but this one is not compatible with `numpy.distutils` (because not patched with `numpy.distutils.ccompiler.gen_lib_options`) and return unquoted libpaths when linking. The problem was patched in Numpy, but need to be patched also in Setuptools for compatibility between older versions of Numpy and `distutils._msvccompiler` (and indirectly Setuptools > 24).

Also some minor changes :
- Replace some residuals `except Exception`.
- PEP8.
- Better history of the problem in Change.rst.

### DIFF
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -6,6 +6,7 @@ import sys
 import platform
 import itertools
 import distutils.errors
+from distutils.version import StrictVersion
 
 from setuptools.extern.six.moves import filterfalse
 
@@ -228,9 +229,9 @@ def msvc14_gen_lib_options(*args, **kwargs):
     """
     if "numpy.distutils" in sys.modules:
         import numpy as np
-        return np.distutils.ccompiler.gen_lib_options(*args, **kwargs)
-    else:
-        return unpatched['msvc14_gen_lib_options'](*args, **kwargs)
+        if StrictVersion(np.__version__) < StrictVersion('1.11.2'):
+            return np.distutils.ccompiler.gen_lib_options(*args, **kwargs)
+    return unpatched['msvc14_gen_lib_options'](*args, **kwargs)
 
 
 def _augment_exception(exc, version, arch=''):
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -79,19 +79,17 @@ def patch_for_specialized_compiler():
         pass
 
     try:
-        # Patch distutils._msvccompiler._get_vc_env
+        # Patch distutils._msvccompiler._get_vc_env for numpy compatibility
         unpatched['msvc14_get_vc_env'] = msvc14compiler._get_vc_env
         msvc14compiler._get_vc_env = msvc14_get_vc_env
     except NameError:
         pass
 
     try:
-        # Apply "gen_lib_options" patch from Numpy to "distutils._msvccompiler"
-        # to fix compatibility between "numpy.distutils" and
-        # "distutils._msvccompiler" (for Numpy < 1.11.2)
-        import numpy.distutils as np_distutils
-        msvc14compiler.gen_lib_options = np_distutils.ccompiler.gen_lib_options
-    except (ImportError, NameError):
+        # Patch distutils._msvccompiler.gen_lib_options
+        unpatched['msvc14_gen_lib_options'] = msvc14compiler.gen_lib_options
+        msvc14compiler.gen_lib_options = msvc14_gen_lib_options
+    except NameError:
         pass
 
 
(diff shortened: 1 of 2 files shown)


vallsv:
Hi. I think there is a problem with `import numpy.distutils as np_distutils`. You dont check if the project is using numpy or not. Then it will force to use `numpy.distutils` anyway the project dotn want it. Maybe there is no problem with that?

Another way would be to check if `CCompile.spawn`

Here is an IPython check
```
> import distutils.ccompiler
> distutils.ccompiler.CCompiler.spawn.__module__
Out: 'distutils.ccompiler'
> import numpy.distutils 
> distutils.ccompiler.CCompiler.spawn.__module__
Out: 'numpy.distutils.ccompiler'
```

Then something like that:
```
if "numpy" in distutils.ccompiler.CCompiler.spawn.__module__:
        # apply the patch
        # you also can check msvc14compiler.gen_lib_options.__module__
```


But it means numpy must be loaded before distutils. Or you have to check it that at execution time.

JGoutin:
True. But setuptools is often imported first.

Fix on execution time is a good idea, I changed the code to try this.

Please review it with your project.

setuptools/msvc.py
```diff
@@ -212,6 +219,19 @@ def msvc14_get_vc_env(plat_spec):
         raise
 
 
+def msvc14_gen_lib_options(*args, **kwargs):
+    """
+    Patched "distutils._msvccompiler.gen_lib_options" for fix 
+    compatibility between "numpy.distutils" and "distutils._msvccompiler"
+    (for Numpy < 1.11.2)
+    """
+    if "numpy" in distutils.ccompiler.CCompiler.spawn.__module__:
```
pitrou: That's very specific. Testing for `"numpy.distutils" in sys.modules` sounds less hackish, no?

vallsv:
Hi, i will try that now.

BTW i think what you write on the comment about the numpy version is good.
Maybe it would be nice to check it programatically. It can avoid further problems, if numpy.distutils API is changed.

```
if distutils.version.StrictVersion(numpy.__version__) < distutils.version.StrictVersion("1.11.2"):
    # your patch
```

vallsv:
Hi again.

Nice job. Here it is working in all cases.

I test with:
- `from numpy.distutils.core import Extension`
- `from setuptools import Extension`
- import first `numpy.distutils`
- import first `setuptools`

Everything was fine on Windows x64, Python 3.5. I dont try other platforms.

JGoutin:
Hi,

Thanks for the review.

:+1:  for the version check, I'll add this.

setuptools/msvc.py
```diff
@@ -212,6 +221,19 @@ def msvc14_get_vc_env(plat_spec):
...
 
+def msvc14_gen_lib_options(*args, **kwargs):
+    """
+    Patched "distutils._msvccompiler.gen_lib_options" for fix 
+    compatibility between "numpy.distutils" and "distutils._msvccompiler"
+    (for Numpy < 1.11.2)
+    """
+    if "numpy.distutils" in sys.modules:
+        import numpy as np
+        if StrictVersion(np.__version__) < StrictVersion('1.11.2'):
```
rgommers: You cannot use ``StrictVersion`` here, that's a broken braindead thing. You will get something like

    ValueError: invalid version number '1.11.1rc1+fa248ad'

for any development version. This is true for many projects, not just numpy. From numpy 1.9.0 on ``numpy.lib.NumpyVersion`` is available, which does proper version comparisons. IIRC something similar that does the right thing for any PEP 440 compliant version string is present in ``distlib`` or another PyPA project.

rgommers: (sorry I'm late with testing/reviewing - just moved and without internet at home)

rgommers: I'm happy to send a fix for this if a ``setuptools`` dev can say which is the preferred version comparison function here.

jaraco: I think you want `pkg_resources.extern.packaging.version.LegacyVersion`.

vallsv: Hi. Sorry for that, i ask for `StrictVersion`. But you can use `LooseVersion` without problem i guess.
```
In [6]: distutils.version.LooseVersion("1.11.1rc1+fa248ad") < distutils.version.LooseVersion('1.11.2')
Out[6]: True
```
I can send a patch. But im not sure you want to use this solution.
Regards.

vallsv: I dont find `pkg_resources.extern.packaging.version`. But `pkg_resources.parse_version`, if you prefer

```
In [20]: pkg_resources.parse_version('1.11.1rc1+fa248ad') < pkg_resources.parse_version('1.11.2')
Out[20]: True
In [21]: pkg_resources.parse_version('1.11.3rc1+fa248ad') < pkg_resources.parse_version('1.11.2')
Out[21]: False
```

jaraco: The problem with parse version is it will return a LegacyVersion when a proper Version can't be constructed and a LegacyVersion always compares less than a Version.

The reason for extern.packaging is because Setuptools vendors the packaging package, and extern exposes the package whether it is vendored or not.

vallsv: One more patch https://github.com/pypa/setuptools/pull/775

Issue #757, opened 2016-08-22
entry_point scripts doesn't pass command line arguments to module:main


sria91:
`entry_point` console (and gui) scripts created using setuptools doesn't pass command line arguments to `module:main`.

Below is the relevant code snippet from `setup.py`

```
...
	entry_points={
		'console_scripts': [
			'module = module:main',
		],
	},
...
```

anthrotype:
you need to use `sys.argv` inside `main` to access command line arguments.

```python
import sys

def main(args=None):
    if args is None:
        args = sys.argv[1:]
    print(args)

if __name__ == "__main__":
    main()
```

jaraco:
Or use [autocommand](https://pypi.org/project/autocommand):

```
from autocommand import autocommand

@autocommand(__name__)
def main(arg):
    print(arg)
```

PR #758, opened 2016-08-22
Fix issue #757

sria91:
allows console (and gui) scripts to receive command line parameters as an argument to the `entry_point` main function.

#757


jaraco:
Thanks for the contrib. Unfortunately, this commit would break every console script out there. Plus, it's better to have a simple interface which expects a simple callable. See the recommendations in the ticket for the recommended approach.

Issue #415, opened 2015-08-17
Following symlinks should be optional


bb-migration:
Originally reported by: **untitaker (Bitbucket: [untitaker](http://bitbucket.org/untitaker), GitHub: [untitaker](http://github.com/untitaker))**

---

In #105 a change was made to follow symlinks while discovering package content. I propose that this behavior should be made optional or reverted. In my case my project directory accidentally included a symlink to another, very large project, and I ended up including a complete installation of ownCloud into a PyPI upload.

I'm willing to provide a patch for this.

This is a continuation of https://bitbucket.org/pypa/pypi/issues/293/re-uploading-of-releases-is-completely

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/415

abadger:
‎We have an additional use case in our project.  We have our own plugin system there.  Sometimes we want to rename a plugin but leave the old name as an alias for a few releases as a deprecation period.  To do that, we rename the file with the actual code and then we symlink to that file with the old filename.  Whenever our plugin system runs across a symlink, it knows that it's dealing with an alias and can do things like show the alias in the docs, display a deprecation warning when the plugin is used via the old name, etc.

If we had a way to mark these symlinks as not to be expanded or a way to toggle that behaviour for all files in the run we'd be able to use that to enable our use-case.

distutils currently hardcodes non-expansion of symlinks in sdists for tarfiles.  zip files expand the symlinks but it looks like a bug: https://bugs.python.org/issue12585  it looks like it's something people desire to make the same but currently there isn't an obvious fix for the bug so the behaviour is asymmetric between tars and zips.

It looks like even distutils expands the symlinks when installing the package.  So I'd probably still have to do more work to make it do what we need at install-time.

Our software requires a *nix like system so symlinks are supported all the places we care about.  This might not be the case for other projects.  being togglable in some manner would allow projects to choose but of course, then setuptools has to support both code paths.

abadger:
Found a second use-case for including symlinks without dereferencing.  We have some code that copies files and has to handle symlinks appropriately.  So we have some integration tests that use a circular symlink structure to test that our code behaves properly.

distutils/setuptools is expanding the symlinks in our integration tests during sdist build time which wastes a bit of space in the tarball and also renders the test useless for consumers of the tarball.

gtristan:
I am curious about this as it presents a bug in my project:
    https://gitlab.com/BuildStream/buildstream/issues/284

[...]

PR #764, opened 2016-08-29
Do not scan whole file tree when making MANIFEST

timheap:
When building a MANIFEST from a MANIFEST.in, setuptools previously
scanned the whole directory tree in to a list, and then picked matching
files based on MANIFEST.in commands from this list.

Now, files are found using the `glob` library from Python 3.5. This only
explores directories that need to be scanned, resulting in a large speed
improvement for projects with large file trees. A modified `glob`
module has been included. It has been changed to support back to
Python 2.6, and to include `.hidden` files in its matches. The previous
functionality included `.hidden` files in its glob matches. It is
unclear if this behaviour is desired and required, or accidental and not
required, but for strict backwards-compatibility, this behaviour is
kept.

Each command in the MANIFEST.in is now represented by its own function
on the FileList (`include`, `exclude`, `graft`, etc.) to allow for an
efficient implementation. The previous commands
`FileList.include_pattern` and `FileList.exclude_pattern` still
exist for backwards compatibility, but these use the slow 'scan all
files' method, so are discouraged. `global_include` by its nature must
scan all directories in the project to work, so this does not receive
any speed improvements.

The changes will speed up creating packages for the vast majority of
users. There are a few unusual corner cases, such as multiple `graft`
commands operating on the same set of directories, that will be slower.
These can be solved by consolidating the overlapping `graft` commands in
to one command.

Note that this is a competing PR to #610. The approach in #610 made it hard to correctly implement `graft`.

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -3,6 +3,7 @@
 Create a distribution's .egg-info directory and contents"""
 
 from distutils.filelist import FileList as _FileList
+from distutils.errors import DistutilsInternalError
 from distutils.util import convert_path
 from distutils import log
 import distutils.errors
@@ -27,6 +28,7 @@ from pkg_resources import (
     parse_requirements, safe_name, parse_version,
     safe_version, yield_lines, EntryPoint, iter_entry_points, to_filename)
 import setuptools.unicode_utils as unicode_utils
+from setuptools.glob import glob
 
 from pkg_resources.extern import packaging
 
@@ -34,6 +36,88 @@ try:
     from setuptools_svn import svn_utils
 except ImportError:
     pass
+
+
+def translate_pattern(glob):
+    """
+    Translate a file path glob like '*.txt' in to a regular expression.
+    This differs from fnmatch.translate which allows wildcards to match
+    directory separators. It also knows about '**/' which matches any number of
+    directories.
+    """
+    pat = ''
+
+    # This will split on '/' within [character classes]. This is deliberate.
+    chunks = glob.split(os.path.sep)
+
+    sep = re.escape(os.sep)
+    valid_char = '[^%s]' % (sep,)
+
+    for c, chunk in enumerate(chunks):
+        last_chunk = c == len(chunks) - 1
+
+        # Chunks that are a literal ** are globstars. They match anything.
+        if chunk == '**':
+            if last_chunk:
+                # Match anything if this is the last component
+                pat += '.*'
+            else:
+                # Match '(name/)*'
+                pat += '(?:%s+%s)*' % (valid_char, sep)
+            continue  # Break here as the whole path component has been handled
+
+        # Find any special characters in the remainder
+        i = 0
+        chunk_len = len(chunk)
+        while i < chunk_len:
+            char = chunk[i]
+            if char == '*':
+                # Match any number of name characters
+                pat += valid_char + '*'
+            elif char == '?':
+                # Match a name character
+                pat += valid_char
+            elif char == '[':
+                # Character class
+                inner_i = i + 1
+                # Skip initial !/] chars
+                if inner_i < chunk_len and chunk[inner_i] == '!':
+                    inner_i = inner_i + 1
+                if inner_i < chunk_len and chunk[inner_i] == ']':
+                    inner_i = inner_i + 1
+
+                # Loop till the closing ] is found
+                while inner_i < chunk_len and chunk[inner_i] != ']':
+                    inner_i = inner_i + 1
+
+                if inner_i >= chunk_len:
+                    # Got to the end of the string without finding a closing ]
+                    # Do not treat this as a matching group, but as a literal [
+                    pat += re.escape(char)
+                else:
+                    # Grab the insides of the [brackets]
+                    inner = chunk[i + 1:inner_i]
+                    char_class = ''
+
+                    # Class negation
+                    if inner[0] == '!':
+                        char_class = '^'
+                        inner = inner[1:]
+
+                    char_class += re.escape(inner)
+                    pat += '[%s]' % (char_class,)
+
+                    # Skip to the end ]
+                    i = inner_i
+            else:
+                pat += re.escape(char)
+            i += 1
+
+        # Join each chunk with the dir separator
+        if not last_chunk:
+            pat += sep
+
+    return re.compile(pat + r'\Z(?ms)')
 
 
 class egg_info(Command):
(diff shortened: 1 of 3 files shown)


timheap:
These tests seem to be failing for unrelated reasons. The tests fail even on `master` currently. I do not know why, and the stack traces are rather confusing.

timheap:
Oh, `py.test` v3 is still broken. Running the tests with `py.test>=2.8,<3` works just fine.

timheap:
OK, test pass again. I've rebased this on top of #765, which excludes `py.test>=3.0` completely.

benoit-pierre:
This patch is great, and fix my issue with [following symlinks](https://github.com/pypa/setuptools/issues/415#issuecomment-202920478) when building the manifest. I however get a test failure on my system:
```python
self = <setuptools.tests.test_manifest.TestFileListTest object at 0x2b1c7b96d850>

    def test_include_pattern(self):
        # return False if no match
        file_list = FileList()
        self.make_files([])
        assert not file_list.include_pattern('*.py')

        # return True if files match
        file_list = FileList()
        self.make_files(['a.py', 'b.txt'])
        assert file_list.include_pattern('*.py')

        # test * matches all files
        file_list = FileList()
        self.make_files(['a.py', 'b.txt'])
        file_list.include_pattern('*')
>       assert file_list.files == ['a.py', 'b.txt']
E       assert ['b.txt', 'a.py'] == ['a.py', 'b.txt']
E         At index 0 diff: 'b.txt' != 'a.py'
E         Use -v to get the full diff
```
This is both with Python 2.7.12 and Python 3.5.2 on Arch Linux.

timheap:
Thanks for checking this.

The order of files in the file list appears to depend on the underlying file system. My system reported the files in the order they were created, but apparently yours does not. I've wrapped the file list in a `set()` for the comparison, as the order is not part of the test.

There is already a separate test that checks that `FileList.sort()` does the sensible thing.

timheap:
The code for this patch is complete, barring any feedback. Is there anything more I can do to help this patch along?

setuptools/command/egg_info.py
```diff
@@ -239,7 +322,151 @@ def check_broken_egg_info(self):
...
+        found = False
+        for i in range(len(self.files) - 1, -1, -1):
+            if predicate(self.files[i]):
+                self.debug_print(" removing " + self.files[i])
+                del self.files[i]
+                found = True
+        return found
+
+    def include(self, pattern):
+        """Include files that match 'pattera'n."""
```
jaraco: This comment looks like it has a typo.

setuptools/command/egg_info.py
```diff
@@ -239,7 +322,151 @@ def check_broken_egg_info(self):
 
 
 class FileList(_FileList):
-    """File list that accepts only existing, platform-independent paths"""
+    # Implementations of the various MANIFEST.in commands
```
jaraco: Previously, this class was a subclass of the implementation in distutils. It's not clear from this diff if there's still value in subclassing `_FileList`. Is this a complete re-implementation or does it also re-use some of the upstream implementation? Do you expect that these changes could be merged upstream at some point?

setuptools/command/egg_info.py
```diff
@@ -341,38 +567,13 @@ def add_defaults(self):
         elif os.path.exists(self.manifest):
             self.read_manifest()
         ei_cmd = self.get_finalized_command('egg_info')
-        self._add_egg_info(cmd=ei_cmd)
-        self.filelist.include_pattern("*", prefix=ei_cmd.egg_info)
-
-    def _add_egg_info(self, cmd):
```
jaraco: Previously, `cmd.egg_base` was processed, but now it appears not to be. What new behavior obviates the need for this behavior?

setuptools/glob.py
```diff
@@ -0,0 +1,169 @@
+"""
+Filename globbing utility. Mostly a copy of `glob` from Python 3.5.
+
+Changes include:
```
jaraco: This file adds a big amount of technical debt to the project. Although I believe it will work, it's a fork of the functionality in Python 3.5 without a merge strategy. Moreover, it has already diverged from the upstream implementation. Thanks, by the way for describing the divergence (I worry other contributors might not be so diligent).

But I wonder - if you could rely on Python 3.5, would the implementation have been different, or would you have had to override all of this behavior not to ignore hidden files? For example, if you look at the recent `setuptools.sdist.py36compat.py`, you'll see functionality that's a backport of code in Python 3.7, which can be removed when Python 3.7 is the minimum supported version. Could you do something similar here, where the forward-compatibility implementation and the divergent behavior are maintained separately (even in separate files)?

setuptools/command/egg_info.py
timheap: I kept it for backwards compatibility. Packages or plugins may depend on calling `FileList.include_pattern`, the old way of adding packages, and I did not want to break these.

setuptools/command/egg_info.py
timheap: :+1:

setuptools/command/egg_info.py
timheap: The behaviour is still required, but is now as simple as `self.filelist.graft(ei_cmd.egg_info)`, assuming I understand what is going in `_add_egg_info()`. I've replaced that whole function with that one-liner.

setuptools/glob.py
timheap: It would not be possible to ship the glob library from Py3.5 without any modifications, as `yield from` is not valid syntax in older Pythons, so at least that needs to change regardless.

To make the library compatible with Py2, I also had to use `six` for dealing with strings/bytes. These changes require modifications to internal function code, so are not easily patched from "outside". I notice that I did not mention this in the list of changes, I will add it now.

The same problem exists with including hidden files - `is_hidden` is called from a number of functions, so these functions need to be changed to not call it. Alternatively, `is_hidden` could be changed to always return `False`, which would have the same effect.

Finally, all the functions in the glob library call each other directly. There is no good way of doing dependency injection or something similar to override the specific functions that need patching.

I completely agree with you that this adds some undesirable technical debt. If you can think of any good solutions to the problems above I am very happy to implement them.

timheap: It is also worth noting that a `diff -U5 /usr/lib/python3.5/glob.py setuptools/glob.py` shows a very nice summary of the changes.

setuptools/command/egg_info.py
jaraco: Okay. Great!

timheap:
Thanks for the merge!

spookylukey:
@timheap thanks for taking this on and succeeding where I ran out of steam! and thanks @jaraco for the merge. This has made a world of difference to my use case - https://github.com/tox-dev/tox/issues/293#issuecomment-284625591

PR #765, opened 2016-08-29
Install py.test>=2.8,<3

timheap:
The 3.x series is currently broken. Until py.test works again, 3.x should not be used by anything.


jaraco:
Yes, sounds reasonable to me.  I would like to see a bug report referencing the defect in pytest, but for now, getting tests passing (and releases working again) is more important.

timheap:
I was about to write up my findings, but spotted that you've already logged pytest-dev/pytest#1888, which contains basically everything I've found out.

PR #770, opened 2016-09-02
Use pytest>=3.0.2

nicoddemus:
The bug that lead to pin it (pytest-dev/pytest#1888) was fixed in 3.0.2


jaraco:
Thank you and thank you.

nicoddemus:
:+1: :smile:

PR #772, opened 2016-09-04
pytest: ignore all .* directories

leorochael:
In Cloud9 a .c9 directory is created. Other IDEs like to add .directories as well.

This PR has pytest ignoring all `.*` instead of just `.tox`

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,6 +1,6 @@
 [pytest]
 addopts=--doctest-modules --ignore release.py --ignore setuptools/lib2to3_ex.py --ignore tests/manual_test.py --ignore tests/shlib_test --doctest-glob=pkg_resources/api_tests.txt --ignore scripts/upload-old-releases-as-zip.py --ignore pavement.py
-norecursedirs=dist build *.egg setuptools/extern pkg_resources/extern .tox
+norecursedirs=dist build *.egg setuptools/extern pkg_resources/extern .*
 flake8-ignore =
     setuptools/site-patch.py F821
     setuptools/py*compat.py F811


.gitignore
```diff
@@ -6,10 +6,8 @@ include
...
 distribute.egg-info
 setuptools.egg-info
-.coverage
-.tox
 *.egg
 *.py[cod]
 *.swp
 *~
-.hg*
+/.*
```
timheap: This will also ignore the .gitignore, .hgignore, .hgtags, and .travis.yml files, which need to be included.

leorochael: Duh, of course... I'll drop this and leave only the pytest exclusion.

leorochael:
There, anything else?

jaraco:
Hi @leorochael: Are you saying that pytest was actually running tests on files in .c9?

timheap:
It runs all the tests it can find in my `setuptools/.venv` directory when I try to run the tests.

leorochael:
@jaraco, yes, like @timheap said. Easily reproduceable: add a `.c9` directory and a `test_fail.py` inside it with these contents:

```
def test_fail():
    assert False, "failed"
```

Then run `python setup.py test`.

I can easily ignore things locally in git by adding them to `$HOME/.config/git/ignore` or `.git/info/exclude`, but it's hard to tell `pytest` to ignore things without changing something else that can't easily be ignored by `git` since `pytest` doesn't overlay configuration files. It only uses the first configuration file it finds.

The only workaround I found was to do run either:

 * `python setup.py test --addopt=--ignore=.c9`

Or

  * `PYTEST_ADDOPTS=--ignore=.c9 python setup.py test`

But those are cumbersome and hard to remember. And in the heat of running all tests it takes a good five minutes to understand why `pytest` is collecting more than it should. It happened me to both this project and to `pep8/pycodestyle`.

jaraco:
We can merge this - but it feels like the wrong place. Perhaps pytest should ignore all .dirs or known .dirs (like .c9 and .tox) or maybe developers using C9 should set PYTEST_ADDOPTS=--ignore=.c9 in their environments. It just feels like a bad precedent to set to have to update each project in Github to accommodate each possible developer environment.

leorochael:
I understand your feeling. I proposed this because I think ignoring all `.directories` is a good non-development-environment-specific middle ground.

Thanks for the merge!

timheap:
I also understand where you are coming from. Would it instead be possible to explicitly tell py.test to search only the `setuptools/` directory (and any others that contain tests)? This would solve these kinds of problems for all environments.

jaraco:
I've considered that, but I've been bitten more than once before where explicitly specifying the files or directories where tests exist might cause tests to be missed. I want to avoid maintaining a manifest of tests (or places tests might be found) and rely on the discovery mechanisms of pytest to find not just these tests, but new tests that get added (or moved) in the future. I think exclusions are preferable in this case.

PR #775, opened 2016-09-06
Use LegacyVersion instead of LooseVersion

vallsv:
- Also apply the patch for numpy 1.11.2 pre release


vallsv:
Tested on Windows, Python 3.5, numpy 1.11.1

rgommers:
Thanks @vallsv, works for me now and should work robustly.

Issue #85, opened 2013-09-25
TypeError: dist must be a Distribution instance


ghost:
Originally reported by: **michaeltripoli (Bitbucket: [michaeltripoli](http://bitbucket.org/michaeltripoli), GitHub: Unknown)**

---

Please forgive if this is misdirected but I'd really like for a setuptools developer to comment on [github.com/Netflix/aminator/issues/147](https://github.com/Netflix/aminator/issues/147#issuecomment-25062382).

quick summary:
aminator depends on stevedore plus other libraries
library stevedore==0.12 converted setup to use pbr
aminator setup.py fails with TypeError following successful installation of stevedore

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/85

Issue #73, opened 2013-08-29
TypeError: dist must be a Distribution instance


ghost:
Originally reported by: **Anonymous**

---

This works on a fresh clean precise vm:

```
wget https://pypi.python.org/packages/source/s/setuptools/setuptools-0.9.8.tar.gz
tar zxvf setuptools-0.9.8.tar.gz
cd setuptools-0.9.8/
sudo python setup.py install
cd
sudo easy_install pip
sudo aptitude install libffi-dev
git clone git@github.com:openstack/swift.git
cd swift/
sudo python setup.py install
```

If I s/setuptools-0.9.8/setuptools-1.1/g those commands, the swift install fails with the Distribution TypeError

```
Processing dependencies for swift==1.9.2.39.g261142d
Searching for python-swiftclient
Reading https://pypi.python.org/simple/python-swiftclient/
Best match: python-swiftclient 1.5.0
Downloading https://pypi.python.org/packages/source/p/python-swiftclient/python-swiftclient-1.5.0.tar.gz#md5=717ade4e91c0f5372f6fa6d545e7bab7
Processing python-swiftclient-1.5.0.tar.gz
Writing /tmp/easy_install-5ywEpr/python-swiftclient-1.5.0/setup.cfg
Running python-swiftclient-1.5.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-5ywEpr/python-swiftclient-1.5.0/egg-dist-tmp-wk8goo
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'd2to1'
  warnings.warn(msg)
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'setup_requires'
  warnings.warn(msg)
Traceback (most recent call last):
  File "setup.py", line 22, in <module>
    pbr=True)
  File "/usr/lib/python2.7/distutils/core.py", line 152, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/clayg/setuptools-1.1/setuptools/command/develop.py", line 27, in run
    self.install_for_development()
  File "/home/clayg/setuptools-1.1/setuptools/command/develop.py", line 129, in install_for_development
    self.process_distribution(None, self.dist, not self.no_deps)
[...]

PR #784, opened 2016-09-12
Don't check if dist is an instance of distutils.dist.Distribution, its not

JensTimmerman:
based on this patch for Distutils (but we can't patch distutils in old python versions can we?)
http://bugs.python.org/issue23102

Whilst we wait for this to get merged into python, it could be fixed this way for setuptools?

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -133,8 +133,50 @@ class Command(_Command):
         """
         Construct the command for dist, updating
         vars(self) with any keyword parameters.
+        Backported from distutils
         """
-        _Command.__init__(self, dist)
+        # late import because of mutual dependence between these classes
+        from distutils.dist import Distribution
+
+        if not hasattr(dist, 'get_requires'):
+            raise TypeError, "dist must be a Distribution instance"
+        if self.__class__ is Command:
+            raise RuntimeError, "Command is an abstract class"
+
+        self.distribution = dist
+        self.initialize_options()
+
+        # Per-command versions of the global flags, so that the user can
+        # customize Distutils' behaviour command-by-command and let some
+        # commands fall back on the Distribution's behaviour.  None means
+        # "not defined, check self.distribution's copy", while 0 or 1 mean
+        # false and true (duh).  Note that this means figuring out the real
+        # value of each flag is a touch complicated -- hence "self._dry_run"
+        # will be handled by __getattr__, below.
+        # XXX This needs to be fixed.
+        self._dry_run = None
+
+        # verbose is largely ignored, but needs to be set for
+        # backwards compatibility (I think)?
+        self.verbose = dist.verbose
+
+        # Some commands define a 'self.force' option to ignore file
+        # timestamps, but methods defined *here* assume that
+        # 'self.force' exists for all commands.  So define it here
+        # just to be safe.
+        self.force = None
+
+        # The 'help' flag is just used for command-line parsing, so
+        # none of that complicated bureaucracy is needed.
+        self.help = 0
+
+        # 'finalized' records whether or not 'finalize_options()' has been
+        # called.  'finalize_options()' itself should not pay attention to
+        # this flag: it is the business of 'ensure_finalized()', which
+        # always calls 'finalize_options()', to respect/update it.
+        self.finalized = 0
+
+
         vars(self).update(kw)
 
     def reinitialize_command(self, command, reinit_subcommands=0, **kw):
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -139,9 +139,9 @@ class Command(_Command):
         from distutils.dist import Distribution
 
         if not hasattr(dist, 'get_requires'):
-            raise TypeError, "dist must be a Distribution instance"
-        if self.__class__ is Command:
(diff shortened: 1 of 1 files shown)


JensTimmerman:
relates to #73

JensTimmerman:
```
#!/usr/bin/env python
''' Installation script that breaks '''

from distutils.command import install
from distutils.core import setup

class installer(install.install):
    def run(self):
        import setuptools
        install.install.run(self)


setup(name='test',
      cmdclass = dict(install=installer)
     )

```
run with: 
python setup.py install --user
result:
```
running install
running build
running install_egg_info
  File "mysetup.py", line 14, in <module>
    cmdclass = dict(install=installer)
  File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib64/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "mysetup.py", line 10, in run
    install.install.run(self)
  File "/usr/lib64/python2.7/distutils/command/install.py", line 575, in run
    self.run_command(cmd_name)
  File "/usr/lib64/python2.7/distutils/cmd.py", line 326, in run_command
  File "/usr/lib64/python2.7/distutils/dist.py", line 970, in run_command
    cmd_obj = self.get_command_obj(command)
  File "/usr/lib64/python2.7/distutils/dist.py", line 846, in get_command_obj
    cmd_obj = self.command_obj[command] = klass(self)
  File "/usr/lib64/python2.7/distutils/cmd.py", line 59, in __init__
    raise TypeError, "dist must be a Distribution instance"
TypeError: dist must be a Distribution instance
```
expected result:
```
running install
running build
running install_egg_info
Writing /home/jens/.local/lib/python2.7/site-packages/test-0.0.0-py2.7.egg-info
```

it seems like this might get fixed in newer distutil releases:
http://bugs.python.org/issue23102

jaraco:
My instinct is to say that this attempted usage of setuptools is not supported.

Can you explain more why the code goes through those contortions to avoid importing setuptools until late?

JensTimmerman:
The code goes trought these contortions to be a small example that fits in a single file.

It just happens every time a setuptools is imported at the wrong time, or when a distutils module is 
reloaded, or when a distutils Command object is patched before setuptools patches it, or if setuptools does patch it but then get's reloaded.

A lot of projects have made this error in the past, and it is pretty hard to debug, usually you don't hit the error in your application, but when one of your dependencies imports setuptools to early.

a list of projects who have run into this:
https://github.com/pypa/setuptools/issues/85
https://github.com/pypa/setuptools/issues/73
http://stackoverflow.com/questions/21136266/typeerror-dist-must-be-a-distribution-instance-when-resolving-beautifulsoup-d
https://github.com/numpy/numpy/issues/7235
https://bugs.launchpad.net/pbr/+bug/1412875
https://bugs.launchpad.net/pbr/+bug/1324784
https://groups.google.com/forum/#!topic/cython-users/VP44IsQrJ60
https://ask.openstack.org/en/question/5478/how-do-i-work-around-the-typeerror-dist-must-be-a-distribution-instance-problem-when-following-the-swift-all-in-one-instructions/
https://github.com/pypa/setuptools/issues/73#issuecomment-247829987
http://lists.science.uu.nl/pipermail/nix-dev/2014-July/013591.html
https://www.mediawiki.org/wiki/Talk:Continuous_integration/Jenkins_job_builder
http://www.gossamer-threads.com/lists/openstack/dev/36874
https://mail.scipy.org/pipermail/scipy-dev/2008-September/009736.html
https://jira.opnfv.org/browse/STORPERF-35
https://twitter.com/simonmcc/status/473490398145298432

the list goes on and on, just google for the error message, people usually get frustrated in python, and barely have any idea how to resolve this issue.
Usually it's either: make sure setuptools is imported earlier, or make sure setuptools is imported later

Read trought a few of the above links to get an idea of how frustrated people can get about this, 
it also took me 3 months of trying all sorts of different things before I found out I just had to import setuptools later in my case, not earlier.

jaraco:
Oh, I think I missed that this was patched in distutils. Yes, setuptools add forward compatibility for what distutils offers.

Can you implement it slightly differently -- copy the code from distutils byte-for-byte into a separate method and commit that change so that we can track any changes that are made to that code. Add a comment indicating from where the code was taken (e.g. Python 3.6.0b1 distutils/cmd.py), and then call that from the original method (keeping separate the divergent behavior), and finally, call the version in distutils directly if it's known to have been patched. This way, when our future selves return to this function, we'll be able to ascertain what was done, why it was done, and when the compatibility layer can be dropped.

jaraco:
Reviewing the upstream bug, it seems like it hasn't been fixed yet, but will be soon. Let's focus on getting the issue patched in distutils, and then prepare a backport in Setuptools based on that implementation and using the technique I describe above.

dabliss:
what is the status with this?  i see the issue's been closed, but i just encountered this error, and i have no idea how to fix it.  @JensTimmerman did a great job above reviewing ways in which people have encountered this error and some hack solutions for it . . . but what is the final, correct, general way to avoid this error?

Issue #406, opened 2015-07-06
Rely on importlib2 and PEP 420 for namespace packages


ghost:
Originally reported by: **jaraco (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

In [this comment](https://bitbucket.org/pypa/setuptools/issue/250/develop-and-install-single-version#comment-19818087), @ncoghlan indicates that it may be possible to get PEP-420 namespace packages on Python 2.7. @rbtcollins indicates that he's worked on this as well with some issues.

I'd like to see setuptools support (and prefer) the PEP 420 model for namespaces. I don't yet have an idea quite what the solution looks like, so I'm opening this discussion here.

Is it possible for setuptools to detect whether the importlib2 hook has been installed on older Pythons? Would that be a sufficient condition in practice to infer that packages could be installed with that model (and eliminate the .pth and other **init**.py techniques)?

As a solution for this issue would likely address issues with namespace packages between different installation models, I'm marking it as higher priority.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/406

asmodehn:
In case someone ends up here, looking for PEP 420 support in python 2, and importing namespace packages, here is a gist that might help you to do just that : https://gist.github.com/asmodehn/d4d69374213be20002ec64ed7ab77dad.
A Finder and a Loader, extracted from importlib2.

Issue #250, opened 2014-08-29
`develop` and `install --single-version-externally-managed` are not compatible with each other for namespace packages


bb-migration:
Originally reported by: **hongqn (Bitbucket: [hongqn](http://bitbucket.org/hongqn), GitHub: [hongqn](http://github.com/hongqn))**

---

To reproduce this issue, do the following in a virtualenv:
1. prepare source code of two packages with same namespace package `ns`:
   
   ```
   mkdir -p pkg1/ns/pkg1 pkg2/ns/pkg2
   echo "from setuptools import setup, find_packages; setup(name='pkg1', packages=find_packages(), namespace_packages=['ns'])" > pkg1/setup.py
   echo "__import__('pkg_resources').declare_namespace(__name__)" > pkg1/ns/__init__.py
   touch pkg1/ns/pkg1/__init__.py
   echo "from setuptools import setup, find_packages; setup(name='pkg2', packages=find_packages(), namespace_packages=['ns'])" > pkg2/setup.py
   echo "__import__('pkg_resources').declare_namespace(__name__)" > pkg2/ns/__init__.py
   touch pkg2/ns/pkg2/__init__.py
   ```
2. install pkg1 using --single-version-externally-managed
   
   ```
   cd pkg1
   python setup.py install --single-version-externally-managed --record record.txt
   cd ..
   ```
3. install pkg2 using develop
   
   ```
   cd pkg2
   python setup.py develop
   cd ..
   ```
4. pkg1 is accessible
   
   ```
   python -c 'import ns.pkg1'
   ```
5. pkg2 is unexpectedly inaccessible
   
   ```
   python -c 'import ns.pkg2'
   Traceback (most recent call last):
     File "<string>", line 1, in <module>
   ImportError: No module named pkg2
   ```

This was discussed in pip's issue tracker https://github.com/pypa/pip/issues/3 years ago (because pip uses `install --single-version-externally-managed` for `pip install` and use `develop` for `pip install --editable`), but they still have not gotten a resolution.

I think this issue should be resolved in setuptools instead of pip.  `install --single-version-externally-managed` creates a `{pkg}-nspkg.pth` in site-packages directory.  It will create a ns module and assign its `__path__` to the ns directory in site-packages.  So `pkg_resources.declare_namespace()` in pkg2's ns/__init__.py has no chance to run.

 I have two proposals to fix it:
[...]

Issue #97, opened 2013-11-05
find_packages() doesn't find PEP 420 packages


bb-migration:
Originally reported by: **gwideman (Bitbucket: [gwideman](http://bitbucket.org/gwideman), GitHub: [gwideman](http://github.com/gwideman))**

---

```
#!python

setup(...
    packages=find_packages(..)
    )
```

on the developer machine will fail to find packages that lack a `__init__.py` file, as is allowed in Python 3.3. However, such packages listed explicitly: packages=['mypkg'] do appear to get included and later installed.

Note: When testing this, before each test be sure to delete all generated metadata, including that which setup may previously have placed in the original source directory, as it seems that setup may use metadata created on a previous run in order to include files.

This is part of a general problem reported in issue #83, but I've logged it separately as it's specifically about setuptools.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/97

jaraco:
> Another issue that's come up recently on distutils-sig is that nspkg.pth support is broken for PEP 420 currently -- the magic really shouldn't be running if PEP 420 support is available.

This issue has since been addressed in #805.

jaraco:
> That's why it would be simplest to change setuptools.dist:check_nsp to automatically add the namespace packages to the list of packages, instead of giving an error -- at that point, nobody should even notice that find_packages() isn't finding the packages in question. (Though the docs should still be updated to clarify that any __init__-less packages must still be manually listed in namespace_packages.)

This sounds like a fine recommendation, which I will attempt to implement.

jaraco:
In the [issue-97 branch](/pypa/setuptools/tree/issue-97), I've attempted to implement the recommendation, but I'm pretty sure this doesn't solve the problem. In particular, it still doesn't address the issue of a very basic use case like `zope.interface`. If `zope/__init__.py` does not exist, `setuptools.find_packages()` will still not find `zope/interface/__init__.py`.

estan:
Anyone know if there has been any more work on this? I'm currently in the situation of wanting to switch from using an "old school" namespace package to a PEP-420 one. In my case I have only a single (company-wide) namespace package, and various distributions that install packages into it, so for me, the solution @jaraco was working on in the [issue-97 branch](https://github.com/pypa/setuptools/tree/issue-97) would be great, if only `find_packages` was updated as well, such that it also looks at `namespace_packages` when taking the decision on whether it should enter a directory in its search.

What I mean is, if `find_packages` encounters:

[...]

PR #789, opened 2016-09-15
Add namespace support for develop

jonparrott:
See https://github.com/pypa/packaging-problems/issues/12 for detailed context, but essentially:

> python setup.py develop does not work with namespace packages

This adds support for namespace packages by adding functionality to the develop command to create `{distribution}-develop-nspkg.pth` files in `site-packages` instead of `.egg-link`s.

This work is mostly based on `install_egg_info` and [this gist](https://gist.github.com/cbrand/a1624ac3e9c81ce45fcb).

There are some outstanding questions I have about this, and of course I need to write tests before merging. I just want to sanity check before I go further.

Questions:
1. `install_egg_info` and `develop` share a lot of common code around generating the `nspkg.pth` files. Should I refactor that into a common module? If so, what should the module be? (There doesn't seem to be any 'private' modules in setuptools).
2. Right now this code path is only activated with you specify the flag _and_ when there are namespace packages around. However, it's very possible that I can generate `pth` files that work the same as the `.egg-info` files today for non-namespace packages. Is that desired?

/cc @ncoghlan

### DIFF
--- a/setuptools/command/develop.py
+++ b/setuptools/command/develop.py
@@ -151,8 +151,6 @@ class develop(easy_install):
             self._gen_nspkg_line(package) for package in namespace_packages]
 
         if self.dry_run:
-            # always generate the lines, even in dry run
-            list(lines)
             return
 
         with open(filename, 'wt') as f:
--- a/setuptools/command/develop.py
+++ b/setuptools/command/develop.py
@@ -20,9 +20,12 @@ class develop(easy_install):
     user_options = easy_install.user_options + [
         ("uninstall", "u", "Uninstall this source package"),
         ("egg-path=", None, "Set the path to be used in the .egg-link file"),
+        ("use-pth-file", None,
+            "Create a pip-compatible .pth file instead of an .egg-link file")
     ]
 
-    boolean_options = easy_install.boolean_options + ['uninstall']
+    boolean_options = easy_install.boolean_options + [
+        'uninstall', 'use-pth-file']
 
     command_consumes_arguments = False  # override base
 
@@ -37,6 +40,7 @@ class develop(easy_install):
     def initialize_options(self):
         self.uninstall = None
         self.egg_path = None
+        self.use_pth_file = False
         easy_install.initialize_options(self)
         self.setup_path = None
         self.always_copy_from = '.'  # always copy eggs installed in curdir
@@ -123,14 +127,43 @@ class develop(easy_install):
             self.easy_install(setuptools.bootstrap_install_from)
             setuptools.bootstrap_install_from = None
 
+        if self.distribution.namespace_packages and self.use_pth_file:
+            self._create_nspth()
+        else:
+            self._create_egg_link()
+
+        # postprocess the installed distro, fixing up .pth, installing scripts,
+        # and handling requirements
+        self.process_distribution(None, self.dist, not self.no_deps)
+
+    def _create_nspth(self):
+        namespace_packages = self._get_all_ns_packages()
+
+        # TODO Give the nspth it's own variable
+        filename, ext = os.path.splitext(self.egg_link)
+        filename += '-nspkg-develop.pth'
+        self.outputs.append(filename)
+
+        log.info("Installing %s (link to %s)", filename, self.egg_path)
+
+        # TODO: support package_dir.
+        lines = [
+            self._gen_nspkg_line(package) for package in namespace_packages]
+
+        if self.dry_run:
+            # always generate the lines, even in dry run
+            list(lines)
+            return
+
+        with open(filename, 'wt') as f:
+            f.writelines(lines)
+
+    def _create_egg_link(self):
         # create an .egg-link in the installation dir, pointing to our egg
         log.info("Creating %s (link to %s)", self.egg_link, self.egg_base)
         if not self.dry_run:
             with open(self.egg_link, "w") as f:
                 f.write(self.egg_path + "\n" + self.setup_path)
-        # postprocess the installed distro, fixing up .pth, installing scripts,
-        # and handling requirements
-        self.process_distribution(None, self.dist, not self.no_deps)
 
     def uninstall_link(self):
         if os.path.exists(self.egg_link):
@@ -171,6 +204,43 @@ class develop(easy_install):
     def install_wrapper_scripts(self, dist):
         dist = VersionlessRequirement(dist)
         return easy_install.install_wrapper_scripts(self, dist)
+
+    def _get_all_ns_packages(self):
+        """Return sorted list of all package namespaces"""
+        nsp = set()
+        for pkg in self.distribution.namespace_packages or []:
+            pkg = pkg.split('.')
+            while pkg:
+                nsp.add('.'.join(pkg))
+                pkg.pop()
+        return sorted(nsp)
+
+    _nspkg_tmpl = (
+        "import sys, types, os",
+        "p = os.path.join(%(develop_path)r, *%(pth)r)",
+        "ie = os.path.exists(os.path.join(p,'__init__.py'))",
+        "m = ie and "
+            "sys.modules.setdefault(%(pkg)r, types.ModuleType(%(pkg)r))",
+        "mp = (m or []) and m.__dict__.setdefault('__path__',[])",
+        "(p not in mp) and mp.append(p)",
+    )
+    "lines for the namespace installer"
+
+    _nspkg_tmpl_multi = (
+        'm and setattr(sys.modules[%(parent)r], %(child)r, m)',
+    )
+    "additional line(s) when a parent package is indicated"
+
+    def _gen_nspkg_line(self, pkg):
+        # ensure pkg is not a unicode string under Python 2.7
+        pkg = str(pkg)
+        develop_path = str(self.egg_path)
+        pth = tuple(pkg.split('.'))
+        tmpl_lines = self._nspkg_tmpl
+        parent, sep, child = pkg.rpartition('.')
+        if parent:
+            tmpl_lines += self._nspkg_tmpl_multi
+        return ';'.join(tmpl_lines) % locals() + '\n'
 
 
 class VersionlessRequirement(object):


setuptools/command/develop.py
```diff
@@ -123,14 +127,43 @@ def install_for_development(self):
...
+
+        log.info("Installing %s (link to %s)", filename, self.egg_path)
+
+        # TODO: support package_dir.
+        lines = [
+            self._gen_nspkg_line(package) for package in namespace_packages]
+
+        if self.dry_run:
+            # always generate the lines, even in dry run
+            list(lines)
```
ekohl: Since lines is already a list comprehension, what does this achieve?

jonparrott: Nothing. This was just left over from where I copied this code from `install_egg_info`. In that function it uses `map` to generate the lines instead of a list comprehension.

jonparrott: (Fixed)

jaraco:
How does this feature play with `setup.py develop --uninstall`?

jonparrott:
> How does this feature play with setup.py develop --uninstall

It'll work as it does today, it'll just remove the `-develop.nspkg.pth` file instead of the `.egg-link` (I haven't coded this part yet because of outstanding questions on the approach).

jaraco:
> Should I refactor that into a common module? If so, what should the module be? (There doesn't seem to be any 'private' modules in setuptools).

Yes please. Feel free to create a new private module for this purpose.

> I can generate pth files that work the same as the .egg-info files today.

That does seem less complex and thus preferable.


My main reservation with this implementation is it adds additional dependence on -nspkg.pth files, where these files are presumably unnecessary on Python 3.3+ (pypa/pip#1924) and possibly could  be obviated on older Pythons (pypa/setuptools#406). I'd be tempted to restrict this functionality to environments where PEP-420 namespace support isn't available (i.e. python 2.6-3.2 where importlib2.hook.__installed__ is not True).

I know that complicates the implementation (and deployment) but it has the added benefit of not masking the problem where a better solution (presumably) exists. What do you think?

jonparrott:
> Yes please. Feel free to create a new private module for this purpose.

Will do.

> I know that complicates the implementation (and deployment) but it has the added benefit of not masking the problem where a better solution (presumably) exists. What do you think?

What better solution for editable install exists in those versions? As far as I can tell, you can omit `-nspkg.pth` files in those interpreters for installed packages, but this is explicitly for editable packages which need to somehow map a package to a folder outside of the normal site-packages tree.

jaraco:
> What better solution for editable install exists in those versions?

Well, in Python, when you start an interpreter, it adds '.' to sys.path (either via `''` or `os.path.abspath('')` in sys.path). Therefore, any packages (namespace or otherwise) being edited in `.` will be importable (and preferred). If there are other packages that one wants to be editing, simply add those to PYTHONPATH:

```
/dev/project_a$ PYTHONPATH=/dev/project_b python
>>> import stuff_from_a
>>> import stuff_from_b
```

Neither project_a nor project_b need to be installed (and if they are, the editable versions in /dev will be imported first). That's of course the simplest use case, but I imagine you have a more complicated use case in mind. Can you expand on how the above wouldn't work?

jonparrott:
Hmm, the `PYTHONPATH` way would be easier, I suppose. However, afaik `PYTHONPATH` isn't treated as a sitedir and I don't think that namespace packages will work properly. There's also the case where `package_dir` is specified in `setup.py` - mapping a package to a different direction e.g. `src/google` -> `google`.

Thoughts?

jaraco:
With PEP 420, any directory in PYTHONPATH can have a namespace package:

```
$ mkdir -p /tmp/foo/pkg
$ touch /tmp/foo/pkg/a.py 
$ mkdir -p /tmp/bar/pkg
$ touch /tmp/bar/pkg/b.py
$ PYTHONPATH=/tmp/bar:/tmp/foo python
Python 3.6.0b1 (v3.6.0b1:5b0ca4ed5e2f, Sep 12 2016, 09:24:46) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import pkg.a
>>> import pkg.b
>>> pkg.__path__
_NamespacePath(['/tmp/bar/pkg', '/tmp/foo/pkg'])
```

As for 'src', why not simply `PYTHONPATH=src`?

Actually, now that I think about it, I'm not opposed to the use of `.pth` files, but the use of `.pth` files to mess with the import mechanism, breaking PEP 420 packages. Here's a `.pth` file linking one of the two namespace packages:

```
$ cat > /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pkg.a.pth
/tmp/foo
$ PYTHONPATH=/tmp/bar python
Python 3.6.0b1 (v3.6.0b1:5b0ca4ed5e2f, Sep 12 2016, 09:24:46) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import pkg.a
>>> import pkg.b
>>> pkg.__path__
_NamespacePath(['/tmp/bar/pkg', '/tmp/foo/pkg'])
```

`setup.py develop` could do something similar, creating `.pth` files (even for non-namespace packages) with only the link to the target for Python 3.3+.

For Python < 3.3, it might still be necessary to fold in the extra behavior.

jonparrott:
> Actually, now that I think about it, I'm not opposed to the use of .pth files, but the use of .pth files to mess with the import mechanism, breaking PEP 420 packages. 

Cool, so I think we might have a way forward here. Please let me know if any of this is incorrect:

develop will *always* create a `{package}-develop.pth` file for every declared top-level package in `setup.py`.

For Python 2 and < 3.3 (or more correctly when `importlib2.hook.installed == False`) the contents of the `.pth` file depend on the presence of namespace packages:
- For non-namespace packages the `.pth` file will just contain the full filesystem path to the package directory.
- For namespace packages the `.pth` file will resemble pip's `-nspkg.pth` files with import machinery to inject the full filesystem path to the package directory into the namespace package's `__path__`.

For > 3.3 (or more correctly when `importlib2.hook.installed == True`) the contents of the `.pth` file will always be the full filesystem path to the package directory.

ncoghlan:
What @jonparrott suggests above is roughly what I had in mind when I suggested using `*.pth` files in https://github.com/pypa/packaging-problems/issues/12, I just completely forgot about the complicating factor of also needing to keep pre-PEP-420 namespace packages working.

jonparrott:
I'm going to be out of office for the rest of this week, but will pick this back up next week. Thanks for helping me work out exactly what needs to be done here @jaraco and @ncoghlan.

jaraco:
@jonparrott: Interestingly, I have an example where I expected to run into this issue but I'm not. In [this build on Travis](https://travis-ci.org/jaraco/jaraco.mongodb/builds/164286068), under `tox`, I'm running tests on the `jaraco.mongodb` package that depends on other `jaraco.*` packages. Tox creates a virtualenv for the target Python version and installs the package under test using the [usedevelop feature](https://testrun.org/tox/latest/config.html#confval-usedevelop=BOOL). I've found it has to do this to run the doctests without running up against tox-dev/tox#373.

What's interesting is that the tests pass - the package under develop seems to be able to import the installed packages from the same namespace... and the tests pass on many Python versions including 2.7.

Right now, I can't explain why it is that I'm not hitting this issue, but perhaps this finding reveals a technique that's already viable. Perhaps I'll find to look into this more later, but I'd welcome your review of this as well to see if you can determine what I'm not seeing.

jonparrott:
@jaraco that is... interesting. Hmm.

jaraco:
And although I use the same technique in another project, I find [its tests are failing](https://travis-ci.org/yougov/vr.cli/jobs/166248061) apparently due to namespace package issues, and here only under Python 3. I wonder if there's some sensitivity to the order in which the package under `develop` is imported.

ncoghlan:
A point potentially worth noting: `*.pth` files are processed by the `site` module, so python's `-S` option will always break them.

The other common source of conflicts is that the current directory is added to the front of `sys.path` when using the `-m` switch, but the directory containing the script itself is added when using direct execution.

jaraco:
I think there's a more insidious problem here - with pip-installed packages (or any other installer that omits `__init__.py` modules for namespace packages), a project that shares a namespace with its dependencies will run into conflicts under Python 3.3+ because the package under development will contain a `__init__.py` file in the namespace package, but the dependencies will not. For example, consider something like `zope.component`, which depends on `zope.interface`. Since the introduction of #805, if one were to install zope.component using `setup.py develop`, `zope` would be loaded by the NamespaceLoader for `zope.interface` (assuming it was pip installed), but by SourceFileLoader for `zope.component` (because there's no opportunity for `zope/__init__.py` to be omitted). For the project under development to be installed in a way that's consistent with its dependencies, the `zope/__init__.py` would have to be removed from the filesystem. Perhaps the project could be reconfigured not to include that file at all, but that would eliminate the ability to use `find_packages` (at least until #97 can be solved). I'm also unsure if this causes conflicts with dependencies installed as eggs.

jaraco:
I should also mention that this PR is designed to address pypa/pip#3 and pypa/setuptools#250.

jonparrott:
Man, this is complex. :| 

I haven't had time to really dig into this as I've been busy re-writing an auth library, but I do want to come back to this eventually. However, If someone wants to take the reigns in the meantime, please feel free to work from what I've done and cc me on the new PR.

ncoghlan:
@jaraco While I agree that situation isn't ideal, I believe it should still *work* as long as the `__init__.py` does the necessary dance to set its own `__path__` correctly using `pkg_resources` or `pkgutil`.

Folks using something like tox to do full install-and-test in a venv would be able to test the post-install configuration that way, even if `setup.py develop` & `pip install -e` gave them something slightly different.

jaraco:
I noticed there was a lot of code in common with `install_egg_info` and `develop`, so I endeavored in the [namespace-module branch](/pypa/setuptools/tree/namespace-module) to create a new module and mix-in to provide the namespace functionality in one place.

jaraco:
> I believe it should still work as long as the __init__.py does the necessary dance to set its own __path__ correctly using pkg_resources or pkgutil.

In my testing, I was unable to get the system to work because the `nspkg.pth` files disable their behavior when a __init__.py exists. Although now it occurs to me that perhaps we could disable that behavior (the `ie` in the `nspkg.pth` line).

I'm going to continue this discussion in #250.

jonparrott:
@jaraco sg, thanks for staying on top of this.

PR #792, opened 2016-09-15
pkg_resources: fix str(Requirement())

benoit-pierre:
Ensure the result of str(Requirement()) is itself a valid requirement line that can be parsed back. Otherwise:
```python
In [1]: import pkg_resources, setuptools

In [2]: setuptools.__version__
Out[2]: '27.2.0'

In [3]: str(pkg_resources.Requirement("python-xlib>=0.16; 'linux' in sys_platform"))
Out[3]: 'python-xlib>=0.16; linux in "sys_platform"'
```
This caused some [regressions in pip](https://github.com/pypa/pip/issues/3781) when environment marker parsing was made consistent by only using the code copied from `setuptools`.

### DIFF
--- a/pkg_resources/_vendor/packaging/markers.py
+++ b/pkg_resources/_vendor/packaging/markers.py
@@ -54,11 +54,15 @@ class Node(object):
 
 
 class Variable(Node):
-    pass
+
+    def __str__(self):
+        return self.value
 
 
 class Value(Node):
-    pass
+
+    def __str__(self):
+        return repr(str(self.value))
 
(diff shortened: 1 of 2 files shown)


benoit-pierre:
Turn out the issue is already [fixed in packaging](https://github.com/pypa/packaging/pull/76).

jaraco:
Great. So once that's released, ping this project here or elsewhere and we can cut a new release. Or bump the vendored requirement and revendor the package and submit a new PR.

PR #796, opened 2016-09-17
Add missing whitespace.

stepshal:


jaraco:
Is that whitespace really missing? I thought usually a item getter should follow immediately after the container, i.e. `thing[item]`.

jaraco:
Nevermind - it's insignificant enough to have any more words about it.

PR #798, opened 2016-09-26
Note find_packages backwards incompatible change

timheap:
Also add a test for the new behaviour.


timheap:
Thanks for the merge!

PR #802, opened 2016-09-28
certifi-2016.9.26

mindw:
so soon ...


jaraco:
Looks like this PR accidentally landed on the `bootstrap` branch. Can you resubmit against master? Also, if you can include a changelog entry (0.1 for something like this), that'll help me cut a release quickly.

mindw:
whoa! sorry about that ...
Will do.

PR #809, opened 2016-10-05
Find nested packages with excluded parent

### LINKED ISSUE #808: find_packages() include does not allow inclusion of subpackages without super packages
Hello! I'm trying to solve [an issue](https://github.com/fedora-infra/bodhi/issues/994) I've got in my project with [the change in setuptools 28.0 that made exclude also exclude all subpackages](https://github.com/pypa/setuptools/pull/733). I had been relying on that behavior to exclude my package's top level package so that I could distribute one of the subpackages as its own add-on package.

I decided to try to include just the subpackage and its subpackages instead, but this also does not seem to work on 28.0 unless I include the top level package (which I am trying to avoid).

To be a little more concrete, here are examples of different behavior. With setuptools-18.0.1, I get this with my project:

```
$ python -c "from setuptools import find_packages; print find_packages(include=['bodhi.server', 'bodhi.server.*'])"
['bodhi.server', 'bodhi.server.static', 'bodhi.server.services', 'bodhi.server.models', 'bodhi.server.views', 'bodhi.server.consumers', 'bodhi.server.scripts']
```

That's the results I want (note that 'bodhi' is not included). With setuptools-28.2.0, I get this with my project:

```
$ python -c "from setuptools import find_packages; print find_packages(include=['bodhi.server', 'bodhi.server.*'])"
[]
```

If I add ```'bodhi'``` (which I do not want), the ```bodhi.server``` packages get included.

timheap:
`find_packages(exclude=['pkg'])` should still find and include the `pkg.subpkg` package.

Fixes #808

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -75,13 +75,17 @@ class PackageFinder(object):
                 rel_path = os.path.relpath(full_path, where)
                 package = rel_path.replace(os.path.sep, '.')
 
-                # Check if the directory is a package and passes the filters
-                if ('.' not in dir
-                        and include(package)
-                        and not exclude(package)
-                        and cls._looks_like_package(full_path)):
+                # Skip directory trees that are not valid packages
+                if ('.' in dir or not cls._looks_like_package(full_path)):
+                    continue
+
+                # Should this package be included?
+                if include(package) and not exclude(package):
                     yield package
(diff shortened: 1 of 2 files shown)


bowlofeggs:
This looks good to me, thanks!

jaraco:
Awesome. Thanks.

PR #813, opened 2016-10-14
Follow PEP8 for keyword arguments syntax in setup

bittner:
I think it makes sense to offer PEP8 compliant code samples in the docs.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -106,9 +106,9 @@ the distutils.  Here's a minimal setup script using setuptools::
 
     from setuptools import setup, find_packages
     setup(
-        name = "HelloWorld",
-        version = "0.1",
-        packages = find_packages(),
+        name="HelloWorld",
+        version="0.1",
+        packages=find_packages(),
     )
 
 As you can see, it doesn't take much to use setuptools in a project.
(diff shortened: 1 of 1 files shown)


docs/setuptools.txt
```diff
@@ -130,29 +130,29 @@ dependencies, and perhaps some data files and scripts::
...
-        author_email = "me@example.com",
-        description = "This is an Example Package",
-        license = "PSF",
-        keywords = "hello world example examples",
-        url = "http://example.com/HelloWorld/",   # project home page, if any
+        author="Me",
+        author_email="me@example.com",
+        description="This is an Example Package",
+        license="PSF",
+        keywords="hello world example examples",
```
bittner: **Side Note:** Must this really be a blank-separated list of word? Or can commas also be used? I've also seen some developers making a list out if this -- Which is the correct way?

bittner:
@jaraco Jason, thanks for merging! Do you know an answer on my question above, too? Is a space-separated string really *the* proposed format? What about commas or even using a list?

jaraco:
In my opinion, the docs should represent what's recommended, the tests should guarantee what's supported, and in lieu of any tests, whatever the implementation supports is what's supported (kinda circular, I know), and we as a community are free to loosen or tighten the spec as we see fit.

bittner:
I agree with you.

- So, the space-separated keywords are the recommended way?
- There doesn't seem to be a test covering the keywords argument.
- I failed to find the implementation of `setup()` where `keywords` is used. Can you help?

jaraco:
> space-separated keywords are the recommended way?

Yes, but that doesn't mean we can't change that if appropriate.

> find the implementation

Look at distutils/dist.py:DistributionMetadata. It looks like the implementation is primarily there. there might be distutils tests that rely on it. It doesn't look like setuptools does much to supplement the behavior in distutils.

PR #814, opened 2016-10-15
Fix tox invocation in .travis.yml

timheap:
To run tox from Travis, you have to specify the tox environment to run with `tox -e ENV`. You can construct the correct env from the TRAVIS_PYTHON_VERSION variable.

I also fixed the Appveyor config a little, but something odd is going on and it is still broken.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,36 @@
 language: python
-python:
-  - 2.6
-  - 2.7
-  - 3.3
-  - 3.4
-  - 3.5
-  - pypy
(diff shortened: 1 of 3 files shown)


timheap:
Py3.5 on Windows is failing, something to do with stevedore. I've got no idea what that is. Everything else is passing, and running in tox.

jaraco:
Thanks for this, but I was seeking to avoid having to matrix out all of the Python Envs. The tests are passing now and more simply invoked using the current Python version.

PR #815, opened 2016-10-15
Add 3.6-dev and nightly to travis configuration

xavfernandez:


xavfernandez:
Hmmm, I was hopinh the tests to crash on 3.6 and 3.7 since pip test suite on these versions triggers an error on `from pkg_resources import Distribution, PathMetadata, ensure_directory` (cf https://travis-ci.org/pypa/pip/jobs/167948406).

Anyway, even if tests pass, it does not hurt to test setuptools on coming versions.

jaraco:
I'll put pypy3 back as well, that's a supported configuration.

xavfernandez:
oops, I must have deleted pypy3 without noticing.
Any clue on https://travis-ci.org/pypa/pip/jobs/167948406 crash ?

jaraco:
I don't really have a clue on either of those failures. I do see the second failure occurs in setuptools, but it's setuptools 0.9.8. Since tests pass on later versions of setuptools, my best guess is that the issue was addressed in later versions of setuptools... or that there's something that pip is doing to the environment that's breaking setuptools install.

yan12125:
> Any clue on https://travis-ci.org/pypa/pip/jobs/167948406 crash ?

This seems the cause: https://github.com/pypa/virtualenv/issues/962

PR #818, opened 2016-10-17
Sort manifest file list in tests

### LINKED ISSUE #816: FileList.files order discrepancy on Arch Linux since 28.5.0
Reproduced in both Python 2.7.12 and 3.5.2. I have tried both ptr and tox and still getting the same test failure:

```
=================================== FAILURES ===================================
____________________ TestFileListTest.test_include_pattern _____________________

self = <setuptools.tests.test_manifest.TestFileListTest object at 0x7fe946df4860>

    def test_include_pattern(self):
        # return False if no match
        file_list = FileList()
        self.make_files([])
        assert not file_list.include_pattern('*.py')

        # return True if files match
        file_list = FileList()
        self.make_files(['a.py', 'b.txt'])
        assert file_list.include_pattern('*.py')

        # test * matches all files
        file_list = FileList()
        self.make_files(['a.py', 'b.txt'])
        file_list.include_pattern('*')
>       assert file_list.files == ['a.py', 'b.txt']
E       assert ['b.txt', 'a.py'] == ['a.py', 'b.txt']
E         At index 0 diff: 'b.txt' != 'a.py'
E         Use -v to get the full diff
```

timheap:
Different OS's and file systems return lists of files in different  orders, not always creation order. This caused intermittent test failures.

The file list is now sorted prior to being checked to ensure a consistent order across all systems.

Fixes #816

### DIFF
--- a/setuptools/tests/test_manifest.py
+++ b/setuptools/tests/test_manifest.py
@@ -351,8 +351,8 @@ class TestFileListTest(TempDirTestCase):
             l('global/one.txt'),
             l('global/two.txt'),
         ]
-        file_list.sort()
 
+        file_list.sort()
         assert file_list.files == wanted
 
     def test_exclude_pattern(self):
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks Tim.

PR #824, opened 2016-10-20
[DO NOT MERGE] Fix write_pkg_file for python 2.6

xavfernandez:
This could be something like that.
Tests still missing.


jaraco:
Based on the comments in the ticket, it sounds as if this issue only affected much older versions of Python 2.6 and newer versions are not subject to the issue, so we'll decline to complicate the code. Thanks anyhow for the effort on this matter!

PR #826, opened 2016-10-23
Same warning message for all standard files

### LINKED ISSUE #825: Warning not filtered as supposed to
The warning emitted for missing standard script name in [py36compat.py](https://github.com/pypa/setuptools/blob/master/setuptools/command/py36compat.py#L76) is not suppressed by [`manifest_maker.warn`](https://github.com/pypa/setuptools/blob/master/setuptools/command/egg_info.py#L558) but seems to be supposed to.

I found this while exploring a setup.py as a directory containing a `__main__.py`.  I happened to run

```
python setup.py/__main__.py
```

and I had a warning about the file `__main__.py` missing.

You can suppress the warning by using the `script_name` keyword argument of `setuptools.setup` like so:

```
from setuptools import setup

setup(
    name="mypackage",
    # ... a plethora of other kwargs
    script_name="setup.py"
)
```

abstrus:
Fixes #825

### DIFF
--- a/setuptools/command/py36compat.py
+++ b/setuptools/command/py36compat.py
@@ -73,7 +73,7 @@ class sdist_add_defaults:
                 if self._cs_path_exists(fn):
                     self.filelist.append(fn)
                 else:
-                    self.warn("standard file '%s' not found" % fn)
+                    self.warn("standard file not found: '%s'" % fn)
 
     def _add_defaults_optional(self):
         optional = ['test/test*.py', 'setup.cfg']


jaraco:
I don't think this solution works because it's editing the backport of the code from newer Pythons. On newer versions of Python, the distutils version of this code will take precedence, and the message from there will be used. This pull request would make sense if the Python 3.6 code was also updated to be the same. Why not edit the filtering code to match either form of the message?

abstrus:
Thank you for your comment. I'll try to rebase this and add a test this
week-end

Le 4 nov. 2016 2:23 PM, "Jason R. Coombs" <notifications@github.com> a
écrit :

PR #830, opened 2016-10-26
Remove _add_defaults_data_files override

### LINKED ISSUE #274: Discrepancy between setuptools and distutils in sdist without MANIFEST.in
Originally reported by: **ikelos (Bitbucket: [ikelos](http://bitbucket.org/ikelos), GitHub: [ikelos](http://github.com/ikelos))**

[This upstream bug](http://bugs.python.org/issue2279) (fixed in to 2010) resolved the need to create a MANIFEST.in file as well as list data_files in the setup.py.  This does not appear to have been fixed in setuptools, and as such sdist and similar functions fail to include the appropriate files on machines that have setuptools, but not those that don't, when a fallback import system is used...

Would it be possible to align the two again so that setuptools is again a drop-in replacement for distutils?

- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/274

### LINKED ISSUE #521: Setuptools' sdist doesn't use data_files while distutils does

tweksteen:
Fixes #274 and fixes #521.

### DIFF
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -141,11 +141,6 @@ class sdist(sdist_add_defaults, orig.sdist):
                     self.filelist.extend([os.path.join(src_dir, filename)
                                           for filename in filenames])
 
-    def _add_defaults_data_files(self):
-        """
-        Don't add any data files, but why?
-        """
-
     def check_readme(self):
         for f in self.READMES:
             if os.path.exists(f):


jaraco:
Okay. Let's give this a try.

Issue #831, opened 2016-10-27
expose structured package data


majuscule:
Package metadata is standardized and structured data that is often reused across projects, (such as in `setup.py` and `--help` output). `<pkg_resources.EggInfoDistribution>` supports retrieving this information via `get_metadata()`, but returns only a text blob. This has lead to projects re-implementing parsing logic in their codebases (i.e. [mozilla/gecko-dev/.../utils.py#L20](https://github.com/mozilla/gecko-dev/blob/master/testing/mozbase/mozrunner/mozrunner/utils.py#L20)). Package data should instead be exposed as a dictionary or first-class object.

jaraco:
Sounds like a good idea to me. Would you be willing to draft a PR for consideration?

PR #834, opened 2016-10-28
expose PKG-INFO metadata from Distribution objs

majuscule:
I've taken an attempt to implement the feature request I proposed (https://github.com/pypa/setuptools/issues/831).

I chose the `Distribution` class to expose the metadata attribute as I believe it is the first in its hierarchy to mandate a `PKG-INFO` resource. The alternative `IMetadataProvider` interface supports arbitrary documents and formats.

I chose to expose the data as an attribute as opposed to method call as I believe it better complements the existing attribute interface (i.e. `.version`). If updated data was desired a method would be preferable but I do not see a use case for that at this time.

### DIFF
--- a/docs/pkg_resources.txt
+++ b/docs/pkg_resources.txt
@@ -959,6 +959,13 @@ version
     metadata file.  If ``PKG-INFO`` is unavailable or can't be parsed,
     ``ValueError`` is raised.
 
+pkg_info
+    A dict containing the mapped fields read from its ``PKG-INFO`` metadata
+    file, as defined by `PEP-0314`_.  If ``PKG-INFO`` is unavailable or can't be
+    parsed, ``ValueError`` is raised.
+
+.. _PEP-0314: https://www.python.org/dev/peps/pep-0314/
+
 parsed_version
     The ``parsed_version`` is an object representing a "parsed" form of the
     distribution's ``version``.  ``dist.parsed_version`` is a shortcut for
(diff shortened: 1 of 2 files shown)


majuscule:
Looking closer, this code is far too optimistic and also lacks support for multi-line fields such as `Description`. I've also found the `pkginfo` module which offers a solution to my original complaint, although I still feel that the functionality should be included here in setuptools. I'm curious what your thoughts would be regarding adding `pkginfo` as a dependency here. It feels a bit heavy-handed to do so, however to implement this feature safely would require a reimplementation of [pkginfo/distribution.py](http://bazaar.launchpad.net/~tseaver/pkginfo/trunk/view/head:/pkginfo/distribution.py).

majuscule:
Perhaps including that code in `setuptools` would be most appropriate however, as users of `pkginfo` should probably be using `pkg_resources` instead to begin with.

jaraco:
I'm not strictly opposed to including pkginfo as a vendored dependency. I do think it would be worthwhile asking @tseaver if he has any opinions about pkg_resources adopting the functionality of pkginfo.

tseaver:
@jaraco Vendoring it would be fine with me.  I just released 1.4.0, re-licensing it under the MIT license (previously it was under the PSF license, which isn't appropriate for non-stdlib projects).

jaraco:
@majuscule Would you put together a pull request with the pkginfo project vendored? The technique isn't documented as far as I recall, but it's pretty simple - update pkg_resources/_vendor/vendored.txt and pkg_resources/extern/__init__ and pavement.py to include the new package, use [paver](https://pypi.org/project/paver) to invoke the vendoring, then commit everything. If all's working well, you should be able to `import pkg_resources.extern.pkginfo`.

PR #837, opened 2016-11-03
Support Python 3.3.0 for extension suffix

reaperhulk:
In setuptools 27.0+ extension building fails on Python 3.3.0 because the extension suffix is obtained using `SO` and not `EXT_SUFFIX` (which is used in 3.3.1 and above). See: https://hg.python.org/cpython/file/v3.3.0/Lib/distutils/command/build_ext.py#l673

This patch tries `EXT_SUFFIX` and falls back to `SO` if `EXT_SUFFIX` is falsey.

### DIFF
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -109,7 +109,7 @@ class build_ext(_build_ext):
                 and get_abi3_suffix()
             )
             if use_abi3:
-                so_ext = get_config_var("EXT_SUFFIX") or get_config_var("SO")
+                so_ext = get_config_var('EXT_SUFFIX') or get_config_var('SO')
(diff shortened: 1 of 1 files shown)


jaraco:
This seems all fine and good.

But it also feels like some undocumented cruft that will linger long after it's needed. Imagine your future self looking at this code and wondering, "why does it grab a value from one of two different places? Under what circumstances will EXT_SUFFIX return Falsey?"

This change is the kind that I like to wrap in some sort of wrapper to explain its purpose and when its purpose will no longer be needed. I'll go ahead and add that here.

reaperhulk:
@jaraco Looks good, thanks for taking this and making it better!

Issue #833, opened 2016-10-28
Installation of packages fails after 28.7.0 upgrade: Documentation object is not iterable


nicpottier:
Getting a weird error: ` TypeError: 'Documentation' object is not iterable` trying to install rjsmin in a brand new virtualenv with the latest setup tools.  This is on Ubuntu 16.0.4 and my OS X machine does the same thing.

```
ubuntu@staging:~$ virtualenv env
New python executable in /home/ubuntu/env/bin/python
Installing setuptools, pip, wheel...done.
ubuntu@staging:~$ source env/bin/activate
(env) ubuntu@staging:~$ pip install rjsmin
Collecting rjsmin
  Using cached rjsmin-1.0.12.tar.gz
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/rjsmin.egg-info
    writing pip-egg-info/rjsmin.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/rjsmin.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/rjsmin.egg-info/dependency_links.txt
    writing manifest file 'pip-egg-info/rjsmin.egg-info/SOURCES.txt'
    warning: manifest_maker: standard file '-c' not found
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-EeZDjY/rjsmin/setup.py", line 42, in <module>
        setup()
      File "/tmp/pip-build-EeZDjY/rjsmin/setup.py", line 33, in setup
        return run(script_args=args, ext=ext, manifest_only=_manifest)
      File "_setup/py2/setup.py", line 421, in run
        return _core.setup(**kwargs)
      File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/home/ubuntu/env/local/lib/python2.7/site-packages/setuptools/command/egg_info.py", line 279, in run
        self.find_sources()
      File "/home/ubuntu/env/local/lib/python2.7/site-packages/setuptools/command/egg_info.py", line 306, in find_sources
        mm.run()
      File "/home/ubuntu/env/local/lib/python2.7/site-packages/setuptools/command/egg_info.py", line 533, in run
        self.add_defaults()
[...]

PR #839, opened 2016-11-03
Add test to verify the install of Pypi top packages

tweksteen:
Following the discussion in #833, I've created a test to verify that we are not breaking the install of the top packages from Pypi.

This test is added at the root test directory as it relies on virtualenv to isolate each package installation. It is disabled by default as it may take a certain amount of time to complete. It is recommended to run it for large scale modification or before each release.

To run on the top 200: `$ python -m pytest tests/test_pypi.py`
To run an arbitrary list of packages: `$ python -m pytest tests/test_pypi.py  --package_name=simplejson --package_name=rcssmin`

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,1 +1,8 @@
+import pytest
+
 pytest_plugins = 'setuptools.tests.fixtures'
+
+def pytest_addoption(parser):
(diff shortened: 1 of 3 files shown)


jaraco:
sweet. Nicely done.

Issue #563, opened 2016-04-26
Meaningful message if corrupted setuptools


idgserpro:
The primary mode of installation of setuptools is to [use ez_setup.py](https://pypi.python.org/pypi/setuptools#unix-wget): but if a beginner downloads a corrupted zip for some reason (check https://github.com/pypa/setuptools/issues/557 for example) and don't delete the corrupted setuptools zip before running ez_setup.py again, the installation can't continue.

Since this is confusing to beginners the idea is to have a similar behavior like bootstrap-buildout.py: always download setuptools, but if you need for some reason to use a local file, use a parameter to have it ([--setuptools-to-dir](https://github.com/buildout/buildout/blob/9ba1e3170a474e6139c9a7eadd32a498c22cb3bb/bootstrap/bootstrap.py#L70) comes to mind).

ez_setup.py has now [a specific branch](https://github.com/pypa/setuptools/tree/bootstrap) and a test suite and this improvement should be done there.

idgserpro:
@jaraco We think adding this option will add more confusion since ez_setup already has a "to_dir" parameter, which sets where setuptools sould be downloaded as well.

In fact, if we change the behavior of `download_setuptools` in ez_setup, removing `if not os.path.exists(saveto)`, it will break bootstrap-buildout.py since --setuptools-to-dir starts "to_dir" parameter, *exactly* for reusing an already downloaded setuptools.

We're not confident into changing this code. It has the potential of breaking other's people infrastructure.

SInce the motivation of this issue was

> Shouldn't we have the same behavior in ez_setup.py? Always download but give the option to use a local copy, if needed? We're asking this because this "bad download" error is really confusing for beginners, since this is the official primary method of installation.

We started to think if we shouldn't just catch the `zipfile.BadZipfile: File is not a zip file` exception and add a message "Maybe you have a corrupt setuptools file. Remove it and run ez_setup again."

jaraco:
Sounds like a simple and elegant improvement to me.

idgserpro:
@jaraco Here we go: https://github.com/pypa/setuptools/pull/842

idgserpro:
Fixed #842

PR #842, opened 2016-11-08
Meaningful message if corrupted setuptools.

idgserpro:
Fixes https://github.com/pypa/setuptools/issues/563.

Is not clear in ez_setup.py that it always uses a setuptools archive if
already downloaded: although useful (and even used as a feature in
buildout-bootstrap.py), it's hard for beginners that may have a
corrupted ez_setup.py.

So we just improve the error message, even showing the full path of the
probably "corrupted" zip.

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -40,6 +40,8 @@ DEFAULT_VERSION = LATEST
 DEFAULT_URL = "https://pypi.io/packages/source/s/setuptools/"
 DEFAULT_SAVE_DIR = os.curdir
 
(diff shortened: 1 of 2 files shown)


idgserpro:
@jaraco can you help us with the AppVeyor error? We're not used to Windows CI.

jaraco:
The AppVeyor error is just saying that there's no appveyor.yml file in that branch of the repo. I've updated the AppVeyor settings for Setuptools to ignore branches without the file.

PR #843, opened 2016-11-08
Remove _add_defaults_data_files override and add unittest

tweksteen:
Reopening a pull request similar to PR https://github.com/pypa/setuptools/pull/830. I've included unit test for that specific feature.

The only noticeable package that will fail with this patch is [django_compressor](https://github.com/django-compressor/django-compressor) because of dependencies on [rcssmin](https://github.com/ndparker/rcssmin) and [rjsmin](https://github.com/ndparker/rjsmin). The main author has been contacted and is aware of the [issue](https://github.com/ndparker/rcssmin/issues/6). It is best to organise a meaningful timeline with him to fix his issues before merging this PR.

### DIFF
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -142,9 +142,10 @@ class sdist(sdist_add_defaults, orig.sdist):
                                           for filename in filenames])
 
     def _add_defaults_data_files(self):
-        """
-        Don't add any data files, but why?
-        """
+        try:
+            sdist_add_defaults._add_defaults_data_files(self)
+        except TypeError:
+            log.warn("data_files contains unexpected objects")
 
     def check_readme(self):
         for f in self.READMES:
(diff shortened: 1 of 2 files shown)


jaraco:
@tweksteen: Would you consider augmenting this pull request to specifically detect the case where non-standard values are being passed for data_files, log an error, and omit the files? That way, for this rare case, setuptools will continue to work as it did before, but for everything else it will work fine. The warning message will also provide incentive to the offending packages to correct their behavior.

tweksteen:
Agreed and done. Let me know if the warning message is clear enough. (Also, should `self.warn`, `log.warn` and `warning.warn` be merged somehow?)

jaraco:
Looks good. And after fixing the test failures due to other flux in the ecosystem, the PR tests are passing. I'm prepared to accept.

jaraco:
I spoke too soon. The tests for Python 3.7 are failing.

PR #848, opened 2016-11-14
expose PKG-INFO metadata from Distribution objs

majuscule:
A second attempt at implementation of feature request https://github.com/pypa/setuptools/issues/831 (see previous PR: https://github.com/pypa/setuptools/pull/834), using a vendored `pkginfo` package.

@jaraco


majuscule:
Not sure what I am doing wrong with vendoring here :-/. Will revisit soon but would appreciate any pointers offhand.

jaraco:
This looks mostly correct to me. I think the issue is with the imports in pkginfo itself. Since they're absolute imports, they won't be found. Here's a case where implicit is better than explicit. You'll either need to write a routine to patch all of the imports to make them relative or (preferably) submit a request upstream to get pkginfo to simple use relative imports natively, cut a new release, then vendor that release.

jaraco:
I [filed a request with pkginfo](https://bugs.launchpad.net/pkginfo/+bug/1653126) to see if pkginfo will help support this approach.

majuscule:
Thanks @jaraco, sorry that I did not get to it before you.

PR #853, opened 2016-11-20
Fix #849 global-exclude globbing

### LINKED ISSUE #849: Since setuptools 28.5.0, a "global-exclude .pyc" no longer matches .pyc files
Since commit bb45468d27615c2ce9f9c9757a367c44d6ee80d2, if a project's MANIFEST.in contains ``global-exclude .py[co]``, the archive generated with ``python setup.py sdist`` will still contain all ``.pyc`` / ``.pyo`` files.

The test below shows that this line worked in setuptools 28.4.0, and broke in setuptools 28.5.0.

I'm not sure whether this can be considered a bug: the previous implementation looked up patterns without any anchoring, whereas the new one forces users to explicitly include wildcards in their ``MANIFEST.in``.
However, I think this change should have been mentioned in the version Changelog.

```patch
diff --cc setuptools/tests/test_manifest.py
index 6360270,ef4beda..0000000
--- a/setuptools/tests/test_manifest.py
+++ b/setuptools/tests/test_manifest.py
@@@ -425,12 -464,22 +425,19 @@@ class TestFileListTest(TempDirTestCase)
          assert file_list.files == ['b.txt']
          self.assertWarnings()
  
+         file_list = FileList()
+         file_list.files = ['a.py', 'b.txt', l('d/c.pyc'), 'e.pyo']
+         file_list.process_template_line('global-exclude .py[co]')
+         file_list.sort()
+         assert file_list.files == ['a.py', 'b.txt']
+         self.assertNoWarnings()
+ 
          # recursive-include
  ```

timheap:
After #764, `global-exclude .pyc` no longer excluded `.pyc` files. This fixes that regression, and adds a test for this behaviour.

Fixes #849.

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -457,7 +457,7 @@ class FileList(_FileList):
         """
         if self.allfiles is None:
             self.findall()
-        match = translate_pattern(os.path.join('**', pattern))
+        match = translate_pattern(os.path.join('**', '*' + pattern))
         found = [f for f in self.allfiles if match.match(f)]
         self.extend(found)
         return bool(found)
@@ -466,7 +466,7 @@ class FileList(_FileList):
         """
         Exclude all files anywhere that match the pattern.
         """
-        match = translate_pattern(os.path.join('**', pattern))
+        match = translate_pattern(os.path.join('**', '*' + pattern))
         return self._remove_files(match.match)
 
     def append(self, item):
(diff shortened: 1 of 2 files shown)


timheap:
Tests seem to be failing for an unrelated reason, which is also affecting `master`

jaraco:
I re-ran the Travis tests and [they're now passing](https://travis-ci.org/pypa/setuptools/builds/177527486).

PR #854, opened 2016-11-21
Upgrade packaging to 16.8

stepshal:


jaraco:
The tests are failing with `ImportError: The 'packaging.requirements' package is required; normally this is bundled with this package so if you get this warning, consult the packager of your distribution.` I suspect that there are new files in this updated packaging library that need to be added to the repository after refreshing the vendoring.

stepshal:
@jaraco fixed.

PR #859, opened 2016-11-25
Make Entrypoint methods docs up to date.

liaoishere:
Entrypoint.load() is deprecated and it's recommend to use `require` and
`resolve` separately.


jaraco:
Thanks for pointing out this discrepancy. I'd like to make some tweaks to wording, but I'll use your contribution as a basis.

PR #865, opened 2016-12-02
Add python_requires to setup.py

xavfernandez:


jaraco:
Ugh. Now we have to advertise and maintain the supported versions in two places.

PR #873, opened 2016-12-07
Tell unittest.main not to exit, fixes #850.

### LINKED ISSUE #850: test command prevents other commands from running
The test command seems to prevent other commands from running. Looks like this issue was introduced in 18.4; in 18.3.2 it doesn't occur. Using https://github.com/pypa/sampleproject:

Setup:

```
$ pyvenv-3.5 venv
$ source venv/bin/activate
```

With setuptools 18.3.2, both "setup.by build test" and "setup.py test build" run the specified commands in the expected order:

```
$ pip install setuptools==18.3.2
Requirement already satisfied (use --upgrade to upgrade): setuptools==18.3.2 in ./venv/lib/python3.5/site-packages
$ python setup.py build test
running build
[...]
running test
[...]
$ python setup.py test build
running test
[...]
running build
[...]
```

...but with 18.4 (and up to 28.8.0), "setup.py build test" runs both build and test, but "setup.py test build" does **not** run build at all.

```
$ pip install setuptools==18.4
[...]
Successfully installed setuptools-18.4
$ python setup.py build test
running build
[...]
running test
[...]
$ python setup.py test build
running test
[...]
# note, "running build" does not appear
```

While fiddling with the build and test commands this way may be a bit silly, the issue exists with other commands as well. Initially I thought this was a flaw in the flake8 command and reported this there, see https://gitlab.com/pycqa/flake8/issues/260. With flake8, the test command behaves differently/worse than with build (with which "build test" works); "flake8 test" runs flake8 but not test, and "test flake8" runs test, not flake8.

scop:


setuptools/command/test.py
```diff
@@ -229,6 +229,7 @@ def run_tests(self):
             None, None, self._argv,
             testLoader=self._resolve_as_ep(self.test_loader),
             testRunner=self._resolve_as_ep(self.test_runner),
+            exit=False,
```
sigmavirus24: This parameter doesn't exist on Python 2.6. You'll need to handle the case where it doesn't exist or just check before passing it.

scop:
Thanks, PR updated.

sigmavirus24:
Thanks for fixing this @scop :)

PR #883, opened 2016-12-12
Fix usage of extras when installing via Wheels; resolves #882

### LINKED ISSUE #882: entry_point with un-satisfied extras_require doesn't throw a DistributionNotFound if you installed your package as a wheel
This is somewhat-complex, but I think I have a handle on what's going on. I have a [project](https://github.com/jimporter/bfg9000) with a `setup.py` that looks like this:

```python
setup(
    extras_require={
        'msbuild': ['lxml'],
    },

    entry_points={
        'bfg9000.backends': [
            'msbuild=bfg9000.backends.msbuild.writer [msbuild]',
        ],
    },
)
```

The goal of this code is to say "the MSBuild backend requires the `lxml` package". Later, I do the following:

```python
from pkg_resources import iter_entry_points, DistributionNotFound
# ...
for i in iter_entry_points('bfg9000.backends'):
    try:
        backend = i.load()
        backends.append((i.name, backend))
    except DistributionNotFound:
        pass
```

The above should load all the backends whose requirements are met (i.e. don't load the MSBuild backend if you don't have lxml). This works fine when the package was **not** installed as a Wheel, but if it's a Wheel, setuptools tries to actually import the MSBuild backend and raises an `ImportError` instead of `DistributionNotFound`.

It appears to be caused by [this code](https://github.com/pypa/setuptools/blob/f783aa36971d78b293fb5d627d1aacd6dbd49afc/pkg_resources/__init__.py#L828-L829), which from my reading says, "If the extras for this requirement are not satisfied, skip this requirement." However, the requirement in this case is `lxml; extra == "msbuild"` (in a non-Wheel installation, it's just `lxml`), so pkg_resources is basically saying, "It's ok if you don't have lxml; just try to load the MSBuild backend." Unfortunately, that's the opposite of what I want, and isn't consistent with the docs:

> [An entry_point] can also include a bracketed list of “extras” that are required for the entry point to be used. When the invoking application or framework requests loading of an entry point, any requirements implied by the associated extras will be passed to pkg_resources.require(), so that an appropriate error message can be displayed if the needed package(s) are missing.

I assume "appropriate error message" in the above is supposed to mean a `DistributionNotFound` exception.

jimporter:
When resolving requirements, we now pass the list of extras we're using along to Marker.evaluate, since we want to include the extra's requirements in our list of required packages.

@jaraco, @dstufft: This should fix the issue I was having in #882. I'm not sure what you want for tests though, or if you think I should make further changes. I don't 100% understand the `_ReqExtras` class (specifically the part where things get added to it via the `dict` interface).

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -786,7 +786,7 @@ class WorkingSet(object):
         self._added_new(dist)
 
     def resolve(self, requirements, env=None, installer=None,
-            replace_conflicting=False):
+                replace_conflicting=False, extras=None):
         """List all distributions needed to (recursively) meet `requirements`
 
         `requirements` must be a sequence of ``Requirement`` objects.  `env`,
@@ -802,6 +802,12 @@ class WorkingSet(object):
         the wrong version.  Otherwise, if an `installer` is supplied it will be
         invoked to obtain the correct version of the requirement and activate
         it.
+
+        `extras` is a list of the extras to be used with these requirements.
+        This is important because extra requirements may look like `my_req;
+        extra = "my_extra"`, which would otherwise be interpreted as a purely
+        optional requirement.  Instead, we want to be able to assert that these
+        requirements are truly required.
         """
 
         # set up the stack
@@ -825,7 +831,7 @@ class WorkingSet(object):
                 # Ignore cyclic or redundant dependencies
                 continue
 
-            if not req_extras.markers_pass(req):
+            if not req_extras.markers_pass(req, extras):
                 continue
 
             dist = best.get(req.key)
@@ -1004,7 +1010,7 @@ class _ReqExtras(dict):
     Map each requirement to the extras that demanded it.
     """
 
-    def markers_pass(self, req):
+    def markers_pass(self, req, extras=None):
         """
         Evaluate markers for req against each extra that
         demanded it.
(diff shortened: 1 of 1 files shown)


jaraco:
After fixing the failing tests in master, this PR now also passes. I haven't yet had a chance to review in depth.

jimporter:
Sounds good. If this is the right strategy, I'll try to add a few more comments explaining it before this lands.

jaraco:
Every time I've tried to review this PR, I've gotten lost. I really don't like how the `extras` parameter is just a pass-through for some calls, but that may be inevitable given the current design.

Would you consider adding more comments explaining the purpose? If the comments can help me understand it, that's a good sign they'll help a future maintainer to understand it as well.

jimporter:
Sure, I can add some comments. I think the method I'm using now is sort of papering over some deeper issues, but those are hard to fix. The biggest problem I see is that in a `DistInfoDistribution` object, you have a dependency mapping that looks like this:

```python
{None: ['common_dep'],
 'my_extra': ['extra_dep; extra = "my_extra"']
}
```

The problem is that `my_extra` is doubled up. It's really saying "when using `my_extra`, `extra_dep` is required, but only when `my_extra` is being used." Unfortunately, by the time we get to `WorkingSet.resolve`, all we have is "`extra_dep` is required, but only when `my_extra` is being used", only we've forgotten about how we want to use `my_extra`, so it just skips `extra_dep`. Probably the best way to fix this would be to remove the redundancy in the dep map, but then I'd need to have ways to clone and mutate `Requirement` objects.

The easier way it just to let `WorkingSet.resolve` know that we want to use `my_extra` so that it can properly check for all its requirements.

jimporter:
Ok, I've added comments and a slightly more-detailed commit message in case anyone from the future is curious as to what's going on here.

PR #892, opened 2016-12-16
Exit on test failure

### LINKED ISSUE #891: test command does not return failure code
Since version 30.2.1, when running the command "python setup.py test" if the tests fails, the exit code is always 0.
I tracked the change of behavior to 53b47e1dfa9dfb1e8b94172a4650409fc03d4048
We found the issue because our CI is no more reporting failure because we run  the test with this command.
About #850, I think it is an expected behavior that if a command in a list fails that the following are not executed. So I think the revision should be reverted.

cedk:
When test fails, it should not continue to run other commands.
Fixes #891

### DIFF
--- a/setuptools/command/test.py
+++ b/setuptools/command/test.py
@@ -226,12 +226,14 @@ class test(Command):
                 list(map(sys.modules.__delitem__, del_modules))
 
         exit_kwarg = {} if sys.version_info < (2, 7) else {"exit": False}
-        unittest_main(
+        test = unittest_main(
             None, None, self._argv,
             testLoader=self._resolve_as_ep(self.test_loader),
             testRunner=self._resolve_as_ep(self.test_runner),
             **exit_kwarg
         )
+        if not test.result.wasSuccessful():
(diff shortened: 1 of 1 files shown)


jaraco:
Does this work on Python 2.6 where exit kwarg is empty?

cedk:
> Does this work on Python 2.6 where exit kwarg is empty?

Yes and no, because on Python2.6 there is no exit option so it always exits no matter the result.

jaraco:
> Yes and no, because on Python2.6 there is no exit option so it always exits no matter the result.

That's acceptable.

jaraco:
Rolling out as v32.1.0.

cedk:
@jaraco can you tell me what are the time frames of the releases? Our CI always fetch the latest version and so for now it always generates false positive which is pointless :-)

jaraco:
Tagged versions are automatically released via Travis. Here's where [v32.1.0 was released](https://travis-ci.org/pypa/setuptools/builds/184562767).

Issue #510, opened 2016-03-01
Avoid full path enumeration on import of setuptools or pkg_resources?


ghost:
Originally reported by: **konstantint (Bitbucket: [konstantint](http://bitbucket.org/konstantint), GitHub: [konstantint](http://github.com/konstantint))**

---

At the moment on my machine, it takes about 1.28 seconds to do a bare `import pkg_resources`, 1.47 seconds to do a bare `import setuptools`, 1.36 seconds to do a bare `from pkg_resources import load_entry_point` and 1.25 seconds to do a bare `from pkg_resources import load_entry_point`.

This obviously affects all of the python scripts that are installed as console entry points, because each and every one of them starts with a line like that. In code which does not rely on entry points this may be a problem whenever I want to use `resource_filename` to consistently access static data.

I believe this problem is [decently](https://github.com/catkin/catkin_tools/issues/297)  [common](https://github.com/Russell91/pythonpy/issues/64), yet I did not find any issue or discussion, hence I'm creating one, hoping I'm not repeating what has been said already elsewhere unnecessarily.

I am using Anaconda Python, which comes along with a fairly large package, alongside several of my own packages, which I commonly add my path via `setup.py develop`, however I do not believe this setup is anything out of the ordinary. There are 37 items on my `sys.path` at the moment. Profiling `import pkg_resources` shows that this leads to 76 calls to `workingset.add_entry` (timing at about a second), of which most of the time is spent in 466 calls to `Distribution.from_location`.

Obviously, the reason for the problem lies in the two `_call_aside` methods at the end of `pkg_resources` which lead to a full scan of the python path at the moment when the package is imported, and the only way to alleviate it would be to somehow avoid or delay the need for this scan as much as possible.

I see two straightforward remedies:
 a) Make the scanning lazy. After all, if all one needs is to find a particular package, the scan could stop as soon as the corresponding package is located. At the very least this would allow me to "fix" my ipython loading problem by moving it up in the path. This might break some import rules which do not respect the precedence of the path, which I'm not aware.
 b) Cache a precomputed index and update it lazily. Yes, this might requre some ad-hoc rules for resolving inconsistencies, and this may lead to ugly conflicts with external tools that attempt to install multiple versions of a package, but this will basically avoid the current startup delay in 99.99% of cases and solve so much of my problems, that I'd be willing to pay the price.

[...]

PR #901, opened 2016-12-28
Speed up console_ and gui_scripts

untitaker:
See #510

- [ ] Add testcase for `my_mod:MyClass.my_func` entrypoint (where to put it?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ setuptools.egg-info
 *.swp
 *~
 .hg*
+.cache
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -2018,12 +2018,11 @@ class ScriptWriter(object):
         __requires__ = %(spec)r
         import re
         import sys
-        from pkg_resources import load_entry_point
 
         if __name__ == '__main__':
             sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
             sys.exit(
-                load_entry_point(%(spec)r, %(group)r, %(name)r)()
+                __import__(%(ep_module)r, fromlist=['__name__'], level=0).%(ep_attrs)s()
             )
     """).lstrip()
 
@@ -2060,6 +2059,10 @@ class ScriptWriter(object):
             group = type_ + '_scripts'
             for name, ep in dist.get_entry_map(group).items():
                 cls._ensure_safe_name(name)
+                if not ep.attrs:
+                    raise ValueError('Invalid script', ep)
+                ep_module = ep.module_name
+                ep_attrs = '.'.join(ep.attrs)
                 script_text = cls.template % locals()
                 args = cls._get_script_args(type_, name, header, script_text)
                 for res in args:
--- a/setuptools/tests/test_easy_install.py
+++ b/setuptools/tests/test_easy_install.py
@@ -44,7 +44,9 @@ class FakeDist(object):
     def get_entry_map(self, group):
         if group != 'console_scripts':
             return {}
-        return {'name': 'ep'}
+        return {
+            'name': pkg_resources.EntryPoint.parse('name = my_module:my_func')
+        }
 
     def as_requirement(self):
         return 'spec'
@@ -73,12 +75,11 @@ class TestEasyInstallTest:
             __requires__ = 'spec'
             import re
             import sys
-            from pkg_resources import load_entry_point
 
             if __name__ == '__main__':
                 sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
                 sys.exit(
-                    load_entry_point('spec', 'console_scripts', 'name')()
+                    __import__('my_module', fromlist=['__name__'], level=0).my_func()
                 )
             """)
         dist = FakeDist()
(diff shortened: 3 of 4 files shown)


untitaker:
Is there a possibility that some sort of changes entry points at runtime? Because then we might not be able to resolve and import them at this point.

jaraco:
I know one case where this will definitely break - `multi-version installs`. If you use `easy_install --multi-version` to install your package, the package won't be visible on `__import__` without first being `required`, which is part of what `load_entry_point` does.

I also worry that all of this effort is for nothing since `easy_install` is largely deprecated in favor of installing with pip, which uses distlib, and [simply relying on distlib in easy_install scripts](https://github.com/pypa/setuptools/pull/595) as well.

Considering the backward-incompatibility concerns here and the fact that it conflicts with the preferred approach, I'd rather suggest that we focus on advancing the work started in #595.

untitaker:
>If you use easy_install --multi-version to install your package, the package won't be visible on __import__ without first being required, which is part of what load_entry_point does.

I see. Could we use the old logic for `easy_install` and this one for `pip`? Such that only scripts that are installed through pip get this speedup.

>I also worry that all of this effort is for nothing since easy_install is largely deprecated in favor of installing with pip

This PR has a direct effect on the scripts `pip install -e` produces.

I don't see how #595 is related.

jaraco:
If one were to use distlib to install scripts, those scripts would behave the same as those installed by pip.

untitaker:
That's a lot of ifs there. I don't understand whether it is possible to install
scripts using distlib at the moment.

jaraco:
> Could we use the old logic for easy_install and this one for pip?

Isn't it the case that only scripts installed by easy_install are affected by this issue?

untitaker:
No, this PR is about "pip install". I didn't test easy_install manually.

ninjaaron:
pip install will use the easy_install module internally unless the the package was built with wheel.

as far as I know, there is no good way to do it with `pip install -e` (though I'm sure someone will correct me if I'm wrong). you have to do something like `pip wheel .; pip install .` to get the benefits of distlib with pip on a local project. The alternative would be to use your own loader script in the project directory. (or use [fastentrypoints](https://github.com/ninjaaron/fast-entry_points), of course!)

jaraco:
> No, this PR is about "pip install". I didn't test easy_install manually.

Well, when I use `pip install`, it installs the packages without the `pkg_resources` hooks.

```
$ pip install --user --no-binary :all: pytest 
Collecting pytest
  Downloading pytest-3.0.5.tar.gz (740kB)
    100% |████████████████████████████████| 747kB 1.6MB/s 
Requirement already satisfied: py>=1.4.29 in /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages (from pytest)
Installing collected packages: pytest
  Running setup.py install for pytest ... done
Successfully installed pytest-3.0.5
$ cat Library/Python/3.6/bin/py.test               
#!/Library/Frameworks/Python.framework/Versions/3.6/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'pytest==3.0.5','console_scripts','py.test'
__requires__ = 'pytest==3.0.5'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('pytest==3.0.5', 'console_scripts', 'py.test')()
    )
```

I think that's because `pip install` will build from the sdist a wheel and install it if no wheel is available (or if binaries are excluded).

It sounds like what you're talking about is `pip install -e`, which under the hood just invokes `setup.py develop` which itself invokes `easy_install`.

Since this issue only affects `develop` installs and not normal installs, it seems largely mitigated by the recommendation to primarily use `pip install`.

I'd like to focus the remaining energy we have onto some of the larger goals, such as removing easy_install. One step that will have to take place to remove easy_install is to solve the `setup.py develop`/`pip install -e` use case, preferably in a library that pip or setuptools can invoke (probably packaging+distlib, but maybe just distlib). Solving that would reduce pip's dependency on setuptools but will also solve this issue.

@vsajip, @dstufft: Do you have any opinion on where the logic for "in-place installs" (as currently implemented in setup.py develop) should go?

dstufft:
@jaraco Honestly, I think the best thing to do is to continue to have setuptools handle this, and wait until the PEPs for abstract build systems come to completion and let the standardization process hammer out who should be responsible for what in terms of in place builds.

hobarrera:
@dstufft TBH, that's too long-term. We're currently dealing with an issue of very slow start-ups (for a 200ms app, the start-up is an additional ~900ms for me).
I'm also constantly seeing reports of slowness at start-up for my projects and others I follow.

While I think it's great that at some point in the future the standardization process gives an elegant solution for this, we really need some workaround or at least a not-so-elegant fix for this in the meantime.

Maybe an extra `setup.py` flag to *not* use `pkg_resources` and do a plain ol' import would suffice for now (at least for those who don't need the multi-version support).

dstufft:
@hobarrera If your users are installing using pip, the only time they should get the slow down is they are not installing from a wheel.

hobarrera:
> If your users are installing using pip, the only time they should get the slow down is they are not installing from a wheel.

I can't really control how users install. TBH, most (including myself) use the distribution packages, which use `python setup.py install`.

I can try and convince packagers of my packages to start using wheels, but that will only affect a few packages for my own projects, and not others (like I said: some apps I use/follow have the same issue, though I'm not a dev there).

PR #913, opened 2017-01-05
strip trailing slash from package_dir before counting slashes

davidszotten:


jaraco:
What's the purpose of this change? What issue does it solve? Should it have a test (what would prevent someone from optimizing this change away)? Are there conditions where this change will cause problems?

davidszotten:
Hi! Apologies, wasn't sure of your process so figured i'd just throw this over the wall and see what response, if any it provoked.

I was getting `Can't get a consistent path to setup script from installation directory` when trying to `pip install -e` a package the other day, and eventually tracked it down to a trailing slash in `package_dir`. The docs i've seen don't use trailing slashes in the examples, but also don't mention that they aren't allowed.

It looks like the edited code is using slashes as a heuristic for how far up the tree to walk, and i guessed that it might be getting confused by the trailing slash

unfortunately i'm new to this codebase, so was hoping for some guidance from yourself or some other core dev as to other (potentially bad) implications of this change, as well as how to best test it

Thanks!

jaraco:
I figured it was something like that. Thanks for providing the background. It sure would be nice to have a test for this - I'm not sure where it would go, if it should be a test for the develop command, or for the package_dir values. The fix also should consider non-Unix paths. Perhaps the whole routine should be written to use `os.path.relpath` or similar in order to avoid re-inventing the wheel.

PR #915, opened 2017-01-08
Add namespace_packages parameter to find_packages

### LINKED ISSUE #97: find_packages() doesn't find PEP 420 packages

estan:
The user must pass any implicit (PEP 420) namespace packages in this parameter. Packages in this list will be recognized as such, despite not having any `__init__.py`.

Closes #97

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ v33.0.0
   still desired, consider adding the functionality to
   setuptools_svn in setuptools_svn #2.
 
+* #97: ``find_packages`` now has a ``namespace_packages``
+  parameter for explicitly specifying implicit (PEP 420)
+  namespace packages. These will be considered packages
+  during the search, despite not having an ``__init__.py``.
+
 v32.3.1
 -------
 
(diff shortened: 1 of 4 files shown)


jaraco:
While still discussing this in #97, I'm going to close this here. Since a change like this will have substantial effects on end-users, I want to be sure to get it right. If we end up settling on this approach, I'll be sure to pull the implementation from here.

PR #930, opened 2017-01-20
Add timestamp-based dependency system to build_clib

GandaG:
`build_clib` is sorely missing one thing in distutils - checking outdated dependencies when compiling. This makes it so that, when building a static c library, **all** objects are always **all** compiled.

The function to create the static library itself *does* implement something of this sort, but since all objects have been recompiled it has no practical effect (everything still runs **every** time).

This is a small patch designed to fix that, with a couple of qol improvements along the way. Two new keys were added to the `build_info` dictionary:

-  obj_deps - Allows specifying extra dependencies for each source so they can be recognized as being newer than the object being compiled. Example:

```
setup(
...
libraries=[(
    'examplelib', 
    {'sources': 'example.c', 
     'obj_deps': {'': 'global.h',  # use an empty string to define global (for all sources) dependencies
                  'example.c': 'example.h'}
    }
)]
...
)
```

- cflags - This just allows the user to provide an extra set of flags to the compiler.

That is pretty much it, a simple (but important) quality of life improvement for those of us who need to compile static libs with python.

If the implementation is fine I can get started on documentation, just not sure where to put it.

### DIFF
--- a/setuptools/command/__init__.py
+++ b/setuptools/command/__init__.py
@@ -2,7 +2,7 @@ __all__ = [
     'alias', 'bdist_egg', 'bdist_rpm', 'build_ext', 'build_py', 'develop',
     'easy_install', 'egg_info', 'install', 'install_lib', 'rotate', 'saveopts',
     'sdist', 'setopt', 'test', 'install_egg_info', 'install_scripts',
-    'register', 'bdist_wininst', 'upload_docs', 'upload',
+    'register', 'bdist_wininst', 'upload_docs', 'upload', 'build_clib',
 ]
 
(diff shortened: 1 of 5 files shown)


jaraco:
This seems reasonable, and I appreciate the effort put into making this first request a complete and thorough one.

I welcome subsequent improvements to the documentation. I also encourage you to see this implementation made the default in distutils (stdlib) with setuptools providing forward compatibility but eventually delegating that role back to distutils. I won't stipulate that as a requirement, however.

GandaG:
I'll definitely look into that, time to start learning mercurial :) Thanks!

PR #943, opened 2017-01-24
WIP: For issue #97 find_packages finds namespace packages.

illume:


jaraco:
I don't think this approach is suitable. It will have too many false positives, creating packages for `docs` and `scripts` and `examples` in many projects. Furthermore, #915 provides another proposal under consideration.

illume:
I wrote a note in #97 about checking this assumption about false positives(I guess there's already a tool for checking setuptools changes on existing packages?). I think the false positives can be avoided in a similar manner to how I avoided site-packages folders in this patch. I also think it's worth weighing up any potential breakage over the breakage the bug is already causing to valid python packages that fail to be packaged.

PR #949, opened 2017-01-29
Fix mock import for build_clib tests on python-2.7

radhermit:


felixonmars:
It's not needed when backports.unittest_mock is present.

radhermit:
Ah thanks, I guess I haven't recently looked at the test requirements. It might be worth dropping the similar try/except mock import workarounds in the other tests as well then.

PR #955, opened 2017-02-02
Add support for overriding .pypirc location

### LINKED ISSUE #954: Allow environment variable override of .pypirc
When `setuptools` is run in a CI environment such as Jenkins, it is desirable to place the `.pypirc` file in the local directory of each job rather than the global location `~/.pypirc`. Please allow users to override the default location via an environment variable such as `PYPI_CONFIG_FILE`. Happy to provide PR if we agree that this is a feature worth having.

srikanth-viswanathan:
Add support for overriding .pypirc location via an environment variable `PYPI_CONFIG_FILE`. I've added tests as well. Happy to address any review feedback. This fixes #954


jaraco:
Closing per discussion in #954.

PR #956, opened 2017-02-02
Fix hard link to URLs that should have been internal (method 1: via manual reference labels)

smheidrich:
There is another suggestion for how to solve this at #957.

In this case, I'm doing what [the Sphinx documentation](http://www.sphinx-doc.org/en/stable/markup/inline.html#cross-referencing-arbitrary-locations) suggests by manually adding a reference label to the linked section.

There are some more "outdated hard links" like this, but I'm waiting for you guys' input on what the preferred way of fixing this is before changing them all.

### DIFF
--- a/docs/pkg_resources.txt
+++ b/docs/pkg_resources.txt
@@ -1090,6 +1090,8 @@ simply by creating an appropriate `IResourceProvider`_ implementation; see the
 section below on `Supporting Custom Importers`_ for more details.
 
 
+.. _pkg-resources-resource-manager-api:
+
 ``ResourceManager`` API
 =======================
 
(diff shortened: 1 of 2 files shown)


jaraco:
Can you describe what the issue is?

smheidrich:
@jaraco In the docs there are links to pages on ``telecommunity.com`` that are online (and possibly obsolete?) versions of documents that exist in the docs. Changing them to "internal links" should be easy, but the problem is that the links are to specific sections, and it seems to me like Sphinx doesn't have very good support for links to sections *in other documents* (linking to a section within the same document is easy, as is linking to another document as a whole).

It would be great if anyone has a better idea for how to achieve this. These are the only two options I've found.

PR #957, opened 2017-02-02
 Fix hard link to URLs that should have been internal (method 2: via autosectionlabel Sphinx extension)

smheidrich:
There is another suggestion for how to solve this at #956.

Here I'm using the [Sphinx autosectionlabel extension](http://www.sphinx-doc.org/en/stable/ext/autosectionlabel.html) as suggested in [this answer on Stack Overflow](http://stackoverflow.com/a/40225340/2748899).
I like this approach better than the one in #956, but it does lead to some warnings during doc build about duplicate headings, because every section heading has to be unique if you want to link to it. What would be best is if there was a version of ``sphinx.ext.autosectionlabel`` that concatenates the document and section heading to form a unique heading, but that doesn't seem to exist... Any other ideas?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,7 +29,7 @@ import setup as setup_script
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['rst.linker']
+extensions = ['rst.linker', 'sphinx.ext.autosectionlabel']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/pkg_resources.txt
+++ b/docs/pkg_resources.txt
@@ -143,10 +143,10 @@ namespace package for Zope Corporation packages, and the ``peak`` namespace
 package for the Python Enterprise Application Kit.
 
 To create a namespace package, you list it in the ``namespace_packages``
-argument to ``setup()``, in your project's ``setup.py``.  (See the `setuptools
-documentation on namespace packages`_ for more information on this.)  Also,
-you must add a ``declare_namespace()`` call in the package's ``__init__.py``
-file(s):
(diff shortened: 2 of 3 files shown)


jaraco:
This approach seems fine to me. I hope it doesn't break the History generation. We'll try it and see.

jaraco:
Hmm. This change seems to have [broken the docs builds](https://readthedocs.org/projects/setuptools/builds/4975495/) with this error:

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/setuptools/envs/latest/lib/python3.4/site-packages/sphinx/application.py", line 431, in setup_extension
    mod = __import__(extension, None, None, ['setup'])
ImportError: No module named 'sphinx.ext.autosectionlabel'
```

That plugin probably needs to be added to docs/requirements.txt.

jaraco:
@smheidrich Can you send a PR with the requisite dependency declared?

smheidrich:
@jaraco It comes with Sphinx itself since version 1.4, but doesn't seem to be available as a standalone project. However, it's just one short file and I could just copy it from Sphinx source to setuptools docs source, then add it according to [those instructions](http://www.sphinx-doc.org/en/1.5.2/ext/thirdparty.html). Or does that create license problems? Sphinx is GPL but setuptools is MIT, isn't it?

jaraco:
Ugh. Let's not bundle it. Thanks for tracing the issue. If it comes with Sphinx 1.4, let's just force that version. I'll give that a go.

jaraco:
Good news - that fixed it. Thanks again.

smheidrich:
Thank you as well! :+1:

smheidrich:
Good news: https://github.com/sphinx-doc/sphinx/pull/3388
Once this makes its way into a Sphinx release, we can even get rid of the duplicate label warnings that result from the changes made in this PR.

jaraco:
Great!

PR #958, opened 2017-02-02
Add --skip-dep-install to boostrap.py script

### LINKED ISSUE #950: bootstrap script cannot be executed without setuptools already installed
The `bootstrap.py` script cannot be used to bootstrap from an environment without setuptools.  The script imports `pip` which requires setuptools to be be installed.  This behavior was introduced in #933.

jjhelmus:
The --skip-dep-install flag skips installation of setuptools dependencies
using pip, delegating this responsibility to other tools.  This can be used to
install setuptools without the need for pip.

closes #950

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -7,6 +7,7 @@ egg-info command to flesh out the egg-info directory.
 
 from __future__ import unicode_literals
 
+import argparse
 import os
 import io
 import re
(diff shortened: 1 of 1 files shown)


jaraco:
This implementation probably won't work due to its dependency on argparse, which is Python 2.7 and we still support Python 2.6. But your implementation gives me some ideas so I'll work from it.

jaraco:
@jjhelmus Will this implementation work for your case?

jjhelmus:
@jaraco Yes, the new changes work great.  Thanks for the quick update!

PR #959, opened 2017-02-02
Refs for self upgrade/installation deprecation

idgserpro:
This can save some time for people that still use ez_setup.py but don't follow setuptools development or mailing lists.


jaraco:
Thanks.

PR #965, opened 2017-02-08
Deprecation message if not doing 'python ez_setup.py'

idgserpro:
Useful for users of https://bootstrap.pypa.io/bootstrap-buildout.py

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -20,7 +20,6 @@ import subprocess
 import platform
 import textwrap
 import contextlib
-import warnings
 
 from distutils import log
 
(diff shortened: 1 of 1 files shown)


ez_setup.py
```diff
@@ -409,5 +410,12 @@ def main():
...
-    warnings.warn("ez_setup.py is deprecated, check https://github.com/pypa/setuptools/issues/581 for more info; use pip to install setuptools")
+    warnings.warn(DEFAULT_DEPRECATION_MSG)
     sys.exit(main())
+
+# There are situations where you don't use 'python ez_setup.py' but ez_setup.py
+# is still executed (specially when using bootstrap.py from zc.buildout, check
+# https://github.com/buildout/buildout/blob/9ba1e3170a474e6139c9a7eadd32a498c22cb3bb/bootstrap/bootstrap.py#L88
+# for more details). We still need to show the deprecation message in that
+# context.
+log.warn(DEFAULT_DEPRECATION_MSG)
```
jaraco: Won't this cause the warning to be logged twice, once at exec and once when `__name__ == '__main__'`? Perhaps the warning should happen exactly once and closer to the top of the file?

idgserpro: When using `exec`, `__name__` is `__builtin__` instead of `__main__`, so it isn't logged twice (actually the way it's implemented now is to exactly avoid that).

But your question gave us na idea to improve ez_setup.py without using the warnings module. We'll update this PR.

idgserpro:
@jaraco done.

PR #966, opened 2017-02-10
support dist-info distributions inside zip files

daa:
I had a problem when ``pkg_resources.get_distribution()`` didn't work inside executable zip, and I've found that it doesn't work for packages inside zip archives mentioned in ``sys.path``. So this pull request adds support for such cases.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1956,6 +1956,12 @@ def find_eggs_in_zip(importer, path_item, only=False):
             subpath = os.path.join(path_item, subitem)
             for dist in find_eggs_in_zip(zipimport.zipimporter(subpath), subpath):
                 yield dist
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for the contrib.

While I can see how this behavior is desirable, I feel a bit uncomfortable with the placement. The function it's in is called `find_eggs_in_zip`, and this function isn't finding eggs. I also see other apparently inappropriately-named functions like `_is_unpacked_egg` which doesn't seem to check anything about packing. Looking at it more, I think I understand why this implementation is correct, and there's nothing about what you've written that's improper, but the expanded scope is making the function name less appropriate.

Overall, I think the benefits outweigh the issues, and indeed it's probably best if structural issues like these are addressed by the package maintainers, so let's go with it.

daa:
Thank you very much

PR #968, opened 2017-02-11
A local version label starts with a '+' sign

moriyoshi:
As per https://www.python.org/dev/peps/pep-0440/#local-version-identifiers:

```
Local version identifiers MUST comply with the following scheme:

<public version identifier>[+<local version label>]
```

But EGG_FRAGMENT's character set does not contain the plus sign, which seems to be a defect.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ v34.2.0
 * #966: Add support for reading dist-info metadata and
   thus locating Distributions from zip files.
 
+* #968: Allow '+' and '!' in egg fragments
+  so that it can take package names that contain
+  PEP 440 conforming version specifiers.
+
 v34.1.1
 -------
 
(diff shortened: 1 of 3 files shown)


jaraco:
This looks good to me. If you add an entry to the CHanges.rst file, that'll save me time getting this out. Probably us an 0.1 bump. Thanks.

moriyoshi:
I found '!' also needs to be accepted on a further investigation:

https://www.python.org/dev/peps/pep-0440/#public-version-identifiers
```
The canonical public version identifiers MUST comply with the following scheme:

[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

https://github.com/pypa/packaging/blob/master/packaging/specifiers.py#L301

jaraco:
Brilliant. I love it.

PR #982, opened 2017-02-23
Feature/bundle build deps

jaraco:
In an attempt to relieve the pressure for setuptools to once again vendor its dependencies, I'm attempting another technique in which setuptools bundles its dependencies for the purposes of building, but continues to rely on its declared dependencies for installation.

As I noted in #980, this change won't affect the most common use-case, that of pip installing setuptools from source when dependencies aren't met. It fails because the setuptools shim is invoked before setuptools has a chance to inject its build dependencies.

I expect this change to be stabilizing, even though it adds a 200k zip file to the sdist. I don't expect it to alter the wheel-based install technique in any way.

I'm interested in feedback from @dstufft in particular.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ setuptools.egg-info
 *.swp
 *~
 .hg*
-requirements.txt
+build-deps.zip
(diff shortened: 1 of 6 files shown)


jaraco:
The more I consider this, and in light of the conversation in #980, I'm thinking this provides little value.

PR #990, opened 2017-03-06
Fixes "AttributeError: module 'packaging' has no attribute 'specifiers'"

andvgal:
Related pypa/setuptools#967

The issue is clearly reproducible on Debian stretch with latest setuptools & virtualenv on both Python 2.7.13 and 3.5.3.
__init__.py of packaging does not explicitly export specifiers sub-module.

I'm not sure how and why it worked before. Perhaps, specifiers got loaded some other way.
However, it is not reproducible on Ubuntu Xenial or Yakkety

P.S. it's really a blind fix reproducing change I made in local virtualenv.

UPD:
I believe it's related to this change:
bpo-27419: Standard __import__() no longer look up “__import__” in
globals or builtins for importing submodules or “from import”. Fixed
handling an error of non-string package name.
https://bugs.python.org/issue27419

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -15,6 +15,7 @@ from distutils.util import rfc822_escape
 import six
 from six.moves import map
 import packaging
+import packaging.specifiers
 
 from setuptools.depends import Require
 from setuptools import windows_support


xavfernandez:
This seems to be needed indeed:

```
(tmp-d0f9b05efca91b9) [xfernandez@xaferiba tmp-d0f9b05efca91b9]$ python
Python 3.6.0 (default, Jan 16 2017, 12:12:55)
[GCC 6.3.1 20170109] on linux
>>> import packaging
>>> packaging.__version__
'16.8'
>>> packaging.specifiers.InvalidSpecifier
AttributeError: module 'packaging' has no attribute 'specifiers'
```

andvgal:
@jaraco any reason to ignore the problem and skipping it for recent 34.3.2 release?
Soon, the issue may be hitting many users as they upgrade to latest Python 2.x/3.x.

Debian Stretch is going to be released with broken setuptools. Then, most likely Ubuntu 17.04 will get broken setuptools as well.

jaraco:
I ended up reviewing #998 first, so I'll decline this change, but thanks. I'll get a release out promptly.

PR #993, opened 2017-03-11
Fix documentation upload by fixing content_type in _build_multipart on Python 3.x.

fschulze:


jaraco:
Thanks for tracking this down and making the patch.

PR #995, opened 2017-03-20
Update MSVC support for Microsoft Visual Studio Build Tools 2017

JGoutin:


AraHaan:
As much as I like this what if the person still has Visual Studio 2015 installed? Will it have some sort of fallback to VS2015 if VS2017 is not installed or no?

JGoutin:
If "Visual Studio 2015" is installed, distutils will work as intended, so this patch will not apply. Maybe this need to be changed for give priority to VS2017 ?

If "Visual C++ BuildTools 2015" and VS2017 are installed, the patch will detect Visual C++ compiler (Theoretically 2017, but this need to be checked, I'll test this).

I there is VS2017 is not installed, it will detect "Visual C++ BuildTools 2015".

setuptools/msvc.py
```diff
@@ -525,9 +528,24 @@ def VCInstallDir(self):
...
+        self.VSInstallDir
+
+        # Default path starting VS2017
+        guess_vc = ''
+        if self.vc_ver > 14.0:
+            default = r'VC\Tools\MSVC'
+            guess_vc = os.path.join(self.VSInstallDir, default)
+            # Subdir with VC exact version as name
+            try:
+                vc_exact_ver = os.listdir(guess_vc)[-1]
```
jaraco: This literal `-1` makes me a little uneasy. What's happening here? Is it assuming that listdir is returning the directories sorted alphanumerically and that the latest version will also be the largest alphanumeric? What sorts of values does one expect to find in os.listdir(guess_vc)?

setuptools/msvc.py
```diff
@@ -525,9 +528,24 @@ def VCInstallDir(self):
         """
         Microsoft Visual C++ directory.
         """
-        # Default path
-        default = r'Microsoft Visual Studio %0.1f\VC' % self.vc_ver
-        guess_vc = os.path.join(self.ProgramFilesx86, default)
+        self.VSInstallDir
```
jaraco: I appreciate your forward thinking in this block to give precedence to the future and indicate the past as legacy. Nice work.

One change I would make here is to implement these two techniques as a single assignment, like:

```
guess_vc = self._guess_vc() or self._guess_vc_legacy()
```

setuptools/msvc.py
```diff
@@ -933,8 +974,11 @@ def SdkTools(self):
         """
         Microsoft Windows SDK Tools
         """
-        bin_dir = 'Bin' if self.vc_ver <= 11.0 else r'Bin\x86'
-        tools = [os.path.join(self.si.WindowsSdkDir, bin_dir)]
+        if self.vc_ver < 15.0:
+            bin_dir = 'Bin' if self.vc_ver <= 11.0 else r'Bin\x86'
+            tools = [os.path.join(self.si.WindowsSdkDir, bin_dir)]
+        else:
+            tools = []
```
jaraco: This probably deserves a comment - why are there no tools for vc_ver < 15?

jaraco: Oh, I see now, you initialize it empty so you can populate it later. This is getting complicated.

setuptools/msvc.py
```diff
@@ -1170,14 +1231,16 @@ def _unique_everseen(self, iterable, key=None):
                     seen_add(k)
                     yield element
 
-    def _get_content_dirname(self, path):
+    def _get_content_dirname(self, path, slash=True):
```
jaraco: One should not use boolean parameters to twiddle inner behavior on functions. Instead of passing `slash=False`, just use `_get_content_dirname().rstrip('\\')`.

jaraco: Even better would be to unconditionally return the value without the slash and let the caller add a slash when appropriate.

jaraco:
My comments above are largely stylistic, so I'd like to get these changes incorporated soon. I'll pull in the branch and adapt to my tastes and get this out.

jaraco:
I've rolled out these changes in 34.4.0. Please test it out and report any issues.

Issue #997, opened 2017-03-23
setuptools.dist imports packaging, but check_specifier accesses packaging.specifiers


jimfulton:
This causes an attribute error in check_specifier.

jimfulton:
FWIW, this is breaking a deployment I'm dealing with.

jaraco:
This issue probably emerged in Setuptools 34 when the vendoring was removed (ff371f1), as dist no longer imports pkg_resources to get the externs (which would import these submodules).

jimfulton:
The bug became effective then. :)

jaraco:
I believe this report is a duplicate of #967.

jimfulton:
Yup

PR #998, opened 2017-03-24
fixed incomplete import of packaging.specifiers and packaging.version

jimfulton:
This fixes: #997 

I didn't write a test because it's likely a test would pass because of imports elsewhere.


jaraco:
Thanks for putting this together, and sorry for the trouble.

Issue #436, opened 2015-09-14
Updating MANIFEST.in does not correctly update the package sdist creates


bb-migration:
Originally reported by: **spookylukey (Bitbucket: [spookylukey](http://bitbucket.org/spookylukey), GitHub: [spookylukey](http://github.com/spookylukey))**

---

The behaviour of sdist depends on previous contents of MANIFEST.in, not just the current. This is not fixed even by running `setup.py clean` or `setup.py clean --all` (although this should not be necessary).

This is very surprising behaviour, and potentially dangerous too - if someone accidentally adds a MANIFEST.in rule that includes a file that must not be distributed and notice the problem, they would expect that removing the rule will remove the file, but it does not.

I've attached a bash script that demonstrates the problem.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/436

miccoli:
I have been hit by this same bug. My current remedy is to delete the `*.egg-info` directory before running `setup.py sdist`.

miccoli:
The problema appears to be in the logic with which the `SOURCES.txt` file is updated by `python setup.py egg_info`:

```
$ python setup.py egg_info
running egg_info
writing sample.egg-info/PKG-INFO
writing dependency_links to sample.egg-info/dependency_links.txt
writing entry points to sample.egg-info/entry_points.txt
writing requirements to sample.egg-info/requires.txt
writing top-level names to sample.egg-info/top_level.txt
reading manifest file 'sample.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'sample.egg-info/SOURCES.txt'
```

Files already in `SOURCES.txt` are preserved across subsequent runs of `python setup.py egg_info`. Could please one of the developers clarify if this is intended behaviour and why?

miccoli:
For setuptools v34.3.0 the behaviour is different in Python 3.6.0 (this bug not present) and Python 2.7.13 (this bug present.)

In fact:

```
$ python2.7 setup.py egg_info
running egg_info
writing requirements to sample.egg-info/requires.txt
writing sample.egg-info/PKG-INFO
writing top-level names to sample.egg-info/top_level.txt
writing dependency_links to sample.egg-info/dependency_links.txt
writing entry points to sample.egg-info/entry_points.txt
reading manifest file 'sample.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'sample.egg-info/SOURCES.txt'
```
[...]

PR #1014, opened 2017-04-13
addresses #436: Updating MANIFEST.in does not correctly update the package sdist creates

miccoli:
The `egg_info` command reads an existing `.egg-info/SOURCES.txt` before writing the new one. Therefore state is preserved, and files once added to `SOURCES.txt` (e.g. by an include line in `MANIFEST.in`) will remain forever in `SOURCES.txt` even if there is no more reason for inclusion.

The problem described in #436  is due to the fact that `.egg-info/SOURCES.txt` is used for building the list of files in a source distribution (`sdist` command).

This PR is surgical, in the sense that solves the issue by removing 2 lines of code. However after this modification the `setuptools.command.sdist.sdist.read_manifest` method is not called anymore within `setuptools` and its removal could be considered, along with the relevant unit tests.

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -564,8 +564,6 @@ class manifest_maker(sdist):
         rcfiles = list(walk_revctrl())
         if rcfiles:
             self.filelist.extend(rcfiles)
-        elif os.path.exists(self.manifest):
-            self.read_manifest()
         ei_cmd = self.get_finalized_command('egg_info')
         self.filelist.graft(ei_cmd.egg_info)
 


jaraco:
I can see you mentioned in this ticket that you don't understand why this logic was authored in the first place. I don't know the reason either.

Perhaps it was a premature optimization. Perhaps it was a mistake.

Since there are no tests capturing the use-case that would have protected this functionality and because the functionality appears to only have detrimental effects, I'm inclined to accept the change and see if any adverse effects are reported, at which point we can distill a use-case and write a unit test to capture that expectation.

jaraco:
On further consideration, I think I've imagined a scenario where this functionality might have been expected - if a project were for whatever reason manually populating that file or using a third-party tool or other command to populate that file, then this change will cause those entries to be dropped. For that reason, I think this change is backward-incompatible and I'll release it as such.

miccoli:
After reviewing the legacy [distutils](https://docs.python.org/3.6/library/distutils.html), I would guess that it was an attempt to maintain compatibility with the `distutils`'s [`MANIFEST` logic](https://docs.python.org/3.6/distutils/sourcedist.html#manifest-related-options):

> - if the manifest file (`MANIFEST` by default) exists and the first line does not have a comment indicating it is generated from `MANIFEST.in`, then it is used as is, unaltered

However this makes sense in `distutils`, where the `MANIFEST` is either automatically generated or authored, and not in `setuptools` where `.egg-info/SOURCES.txt` is always generated.

Thanks for accepting the change, hoping not to break to much existing projects.

voidlily:
I found a project that broke with this change: https://bitbucket.org/vangheem/pyzipcode/src

This package installs properly before 35.0.0 and no longer does. Test script:

```python
from pyzipcode import ZipCodeDatabase
z = ZipCodeDatabase()
z[94103]
```

```
  File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/pyzipcode/__init__.py", line 101, in __getitem__
    zip = self.get(str(zip))
  File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/pyzipcode/__init__.py", line 98, in get
    return format_result(self.conn_manager.query(ZIP_QUERY % zip))
  File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/pyzipcode/__init__.py", line 32, in query
    cursor.execute(sql)
pysqlite2.dbapi2.OperationalError: no such table: ZipCodes
```

The file `zipcode.db` is missing in the installed package in 35.0.0 due to the MANIFEST changes (though the package could also be packaged improperly and lacking a MANIFEST as well, not sure what the exact intention with this change is).

miccoli:
@voidlily From my tests there is no regression from `setuptools==34.4.1`: starting from a fresh clone of pyzipcode, the data file `pyzipcode/zipcodes.db` is not installed:

```
$ hg clone https://bitbucket.org/vangheem/pyzipcode
$ pip install setuptools==34.4.1
$ pip install ./pyzipcode
```

To have `pyzipcode/zipcodes.db` properly included you can try installing `setuptools_scm` which will automagically include all files under version control in the manifest file (which is `.egg-info/SOURCES.txt`) and therefore in the installed package.

pjdelport:
I think this change may have broken sdist installs with package data: #1016

haydensun:
I agree with @miccoli about the behavior of MAINFEST‘s logic. Whenever updating MAINFEST.in, the SOURCES.txt in .egg-info needs to be updated consistently. At present, delete the .egg-info directory before sdist may be a feasible solution.

Issue #1016, opened 2017-04-18
Regression: sdists no longer install package data correctly with 35.0.0


pjdelport:
# Problem description

I have a package that uses `setuptools_scm` and `include_package_data=True` to include some Django templates and static files.

With setuptools 34.4.1, installing an sdist will install these package data files without any problems. However, with setuptools 35.0.0, installing the same sdist file will no longer install these package data files.

Note:

* Only installing sdists seem to be affected by this problem. In particular:
  * _Generating_ sdists seems unaffected: both setuptools 34 and 35 generate sdists with the same (correct) distribution content and `SOURCES.txt`.
  * Installing sdists previously generated with either setuptools version works correctly with 34.4.1, and fails with 35.0.0.
* Wheels seem to be unaffected by this, and build and install fine with both setuptools 34 and 35.

# Cause?

As far as I can tell, this regression is due to #436 / #1014, which no longer reads the sdist package's existing manifest. When the manifest gets regenerated (?), the files that were previously included by `setuptools_scm` while building the sdist are no longer included anymore.

pjdelport:
Here's a reproducing test script using [cartridge-downloads](https://pypi.python.org/pypi/cartridge-downloads), which uses `setuptools_scm` and `include_package_data=True` to include some templates:

```
PACKAGE=cartridge-downloads  # Change this to test other packages

pip install 'setuptools==34.4.1'
pip install --no-binary "$PACKAGE" "$PACKAGE"
pip show -f "$PACKAGE" >good-info

pip uninstall "$PACKAGE"

pip install 'setuptools==35.0.0'
pip install --no-binary "$PACKAGE" "$PACKAGE"
pip show -f "$PACKAGE" >bad-info

diff -u good-info bad-info
```

(Note that this uses `pip install --no-binary` to install the sdist rather than the wheel.)

Diff output:

```diff
--- good-info	2017-04-18 16:25:18.364959212 +0200
+++ bad-info	2017-04-18 16:25:32.172929357 +0200
@@ -37,12 +37,6 @@
   cartridge_downloads/models.py
   cartridge_downloads/page_processors.py
   cartridge_downloads/signals.py
-  cartridge_downloads/templates/email/form_response.html
-  cartridge_downloads/templates/email/form_response.txt
-  cartridge_downloads/templates/email/order_receipt.html
-  cartridge_downloads/templates/email/order_receipt.txt
-  cartridge_downloads/templates/shop/complete.html
[...]

PR #1017, opened 2017-04-18
Revert "addresses #436: Updating MANIFEST.in does not correctly update the package sdist creates"

miccoli:
Reverts pypa/setuptools#1014 see #1016


miccoli:
I'm not very proficient with the PR and revert PR etiquette: please let me know if i've done something wrong by asking to revert my previous PS.

jaraco:
Thanks for this, but I'm going to perform a different git revert command that includes the hash of the commit being reverted.

jaraco:
I used a similar revert here - 1163c86b3b6.

PR #1018, opened 2017-04-19
Tox maintenance

pjdelport:
This adds instructions for running `rwt -- bootstrap.py` to `tox.ini`, to help new contributors get started, and extends the default Tox `envlist` to follow the Python versions being tested on Travis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ setuptools.egg-info
 *~
 .hg*
 requirements.txt
+.cache
--- a/tox.ini
+++ b/tox.ini
@@ -1,3 +1,6 @@
+# Note: Before running Tox, you need to install rwt and run
+# "rwt -- bootstrap.py" (outside of Tox) to generate the necessary files.
+
 [testenv]
 deps=
     -rtests/requirements.txt
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,15 @@
 # Note: Before running Tox, you need to install rwt and run
 # "rwt -- bootstrap.py" (outside of Tox) to generate the necessary files.
 
+[tox]
+envlist =
+    py2{6,7}
+    py3{3,4,5,6}
+    pypy
+# Most people probably don't have all the above installed,
+# so skip unavailable ones without failing, by default.
+skip_missing_interpreters = True
+
 [testenv]
 deps=
     -rtests/requirements.txt


jaraco:
Thanks for this. I'm disinclined to accept 3e364c9 because travis uses tox to run the tests, so this change makes the tests run on multiple interpreters and not the activated one. Plus, my preference is for the default invocation of tox to use the default Python interpreter and to let developers if they're so inclined elect to test on more environments. Plus, as you can see, it seems to break tests on Windows.

The last commit of the .cache directory is fine (though I use a global .gitignore that ignores that dir for all projects, so I find it unnecessary personally).

The first commit, with the rwt bootstrap instructions, may not be correct. Where did you get the indication that rwt is necessary? It's not used by travis. I think the message in setup.py about rwt is outdated (also the requirement in .travis.yml).

I wonder if instead of invoking bootstrap.py in .travis.yml and appveyor.yml, it would be better to invoke it using tox. I think I explored that once and found it wasn't possible... and thus we have to rely on implicit (or documentation only) behavior.

jaraco:
I just revisited and I don't see any way to ask tox to invoke bootstrap.py prior to build/install/test, so that's probably why it can only be in documentation. Feel free to re-submit with instructions to invoke `python bootstrap.py` prior to running tox. Of course, that's what tox will report if you fail to do so, so perhaps that's sufficient (and less likely to grow stale).

pjdelport:
Thanks for the feedback!

> Where did you get the indication that rwt is necessary?

I just got it from the mention in `setup.py`. Just running `bootstrap.py` directly seems to work fine here too, so that's fine. :)

> [instructions to invoke python bootstrap.py] is what tox will report if you fail to do so, so perhaps that's sufficient (and less likely to grow stale).

Hmm, I don't see that. When I run Tox on a clean checkout, the failure is a bit inscrutable:

```
ERROR: invocation failed (exit code 1), logfile: …/setuptools/.tox/python/log/python-1.log
ERROR: actionid: python
…
Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
…
______________________________________________________________________ summary _______________________________________________________________________
ERROR:   python: could not install deps [-rtests/requirements.txt, -rrequirements.txt]; v = InvocationError('…', 1)
```

I'll address the feedback and resubmit.

Issue #1024, opened 2017-04-26
setuptools + pbr broken on Jython 2.7.0


LordGaav:
I'm running into an issue using a combination of setuptools and pbr on Jython, that produces this error:

```
Traceback (most recent call last):
  File "setup.py", line 24, in <module>
    setup(
  File "/home/chaos/jython2.7.0/Lib/distutils/core.py", line 112, in setup
    _setup_distribution = dist = klass(attrs)
  File "/home/chaos/jython2.7.0/Lib/distutils/core.py", line 112, in setup
    _setup_distribution = dist = klass(attrs)
  File "/home/chaos/workspace/omniconf/vendor/Lib/site-packages/setuptools/dist.py", line 320, in __init__
    _Distribution.__init__(self, attrs)
  File "/home/chaos/jython2.7.0/Lib/distutils/dist.py", line 287, in __init__
    self.finalize_options()
  File "/home/chaos/workspace/omniconf/vendor/Lib/site-packages/setuptools/dist.py", line 387, in finalize_options
    ep.load()(self, ep.name, value)
  File "/home/chaos/workspace/omniconf/.eggs/pbr-3.0.0-py2.7.egg/pbr/core.py", line 98, in pbr
    _monkeypatch_distribution()
  File "/home/chaos/workspace/omniconf/.eggs/pbr-3.0.0-py2.7.egg/pbr/core.py", line 62, in _monkeypatch_distribution
    core.Distribution = dist._get_unpatched(core.Distribution)
  File "/home/chaos/workspace/omniconf/vendor/Lib/site-packages/setuptools/dist.py", line 29, in _get_unpatched
    return get_unpatched(cls)
  File "/home/chaos/workspace/omniconf/vendor/Lib/site-packages/setuptools/monkey.py", line 30, in get_unpatched
    return lookup(item)
  File "/home/chaos/workspace/omniconf/vendor/Lib/site-packages/setuptools/monkey.py", line 44, in get_unpatched_class
    base = next(external_bases)
StopIteration
```

All versions `>= 32.3.0` are affected, which leads me to believe that [this change](https://github.com/pypa/setuptools/commit/07a7b06dd8c6dc08c789505f263986e5f084f802) introduced the bug.

Steps to reproduce:

1. Install Jython 2.7.0 (see https://github.com/LordGaav/omniconf/blob/develop/.travis.yml for steps)
[...]

PR #1025, opened 2017-04-27
Use a different method to lookup base classes on Jython

LordGaav:
Jython seems to implement inspect.getmro differently, which causes
any classes with the same name as a class lower in the MRO not to
be returned.

This patch offloads the MRO lookup to a separate function, which
implements different logic for Jython only.

Ref #1024

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v35.0.2
+-------
+
+* #1024: Add workaround for Jython #2581 in monkey module.
+
 v35.0.1
 -------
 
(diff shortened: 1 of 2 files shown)


LordGaav:
I have run the test suite on Python 3.4, and all tests are green. I haven't been able to run it on Jython however, I receive long lists of these kinds of errors:

```
___________ ERROR collecting setuptools/tests/test_unicode_utils.py ____________
import file mismatch:
imported module 'setuptools.tests.test_unicode_utils' has this __file__ attribute:
  __pyclasspath__/setuptools/tests/test_unicode_utils.py
which is not the same as the test file we want to collect:
  /app/setuptools/tests/test_unicode_utils.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

Issue #917, opened 2017-01-09
Remove easy_install


jaraco:
There are many open issues with Setuptools that stem from the differences between easy_install and pip install. Now that #250 is complete and there exists a mechanism by which pip install can largely satisfy the use cases currently filled by easy_install. Perhaps at some point I will link to or enumerate the issues that would be addressed by such a replacement but not right now.

In the replacement, I imagine the following:

Invocations of `easy_install` whether from a command-line entry point or an implicit invocation as part of a `setup.py install` will instead shell out to pip to perform the equivalent operation. Some options may become unavailable while others will be translated to their pip equivalent.

One big open question is whether installs should still be done as eggs (using the deprecated `pip install --egg`) or if it should rely on pip's use of wheel. I'm leaning toward the latter.

There are probably other open issues that I haven't yet considered.

graingert:
probably worth just deprecating easy_install and adding a header from the docs that points to pip

jaraco:
It's not so simple. Easy-install is used by tests requires and setup requires, which has not replacement. Perhaps it makes sense to only replace those functions.

xoviat:
If I'm not mistaken, running `pip install` while pip is already running can cause problems. The only way that this is going to work is if pip understands `setup_requires`.

xoviat:
See pypa/pip#2745.

graingert:
Tests requires should be deprecated in favour of tox

jaraco:
> If I'm not mistaken, running pip install while pip is already running can cause problems. The only way that this is going to work is if pip understands setup_requires or setuptools does some sort of runtime hack/monkeypatch on pip.

I don't think these issues are relevant. In the case where pip is invoking setup.py to build a package from source and that package has setup_requires dependencies, those dependencies would still be "installed", but to a temporary location (perhaps as eggs to ./.eggs as they are now), meaning they would still be isolated from the target environment via subprocesses and temporary directories, so I would expect not to encounter those issues.

dstufft:
FWIW longer term PEP 518 is the solution to ``setup_requires``.

graingert:
maybe for now, just a docs header saying, "hey unless you know what you're doing: don't use easy_install, you probably want pip instead"

marscher:
https://github.com/pypa/pip/issues/1820#issuecomment-151329572 gives a reference on why recursive pip calls are not a good solution. I think this should be handled by setuptools, because setup_requires is a setuptools feature, which pip is currently unaware of.

[...]

PR #1030, opened 2017-05-15
easy_install: wrap pip

xoviat:


xoviat:
one thing to keep in mind is that module mode I don't think works on py26, meaning that pip will not be used for that version.

xoviat:
@jaraco Before I proceed, is this a valid approach?

jaraco:
This PR relates to #917.

This is interesting. The first thing I notice is there are several stylistic changes here. While I'm not opposed to stylistic changes, they do distract from the meaningful changes. May I suggest resisting the temptation to make stylistic changes, or if they're important to the implementation, make them as a separate pull request first so they're not visible during the review of the functional changes?

I think you're on the right track here, and I'm eager to see where you take this. I'm a bit concerned about the inclusion of a dependency on pip, as it starts to create a weird circular dependency where pip requires setuptools to build packages during install, but setuptools uses pip to install packages during builds.

Other challenges that need to be considered:

- Should pip install affect all easy_install or just setup_requires/tests_require? (the former will affect more 
- What are the use cases expected by the current implementation? Do you intend to replace them in a compatible way or will it be backward-incompatible? If the latter, what are the implications (and actions necessary for a system upgrading)?
- Although the test suite is meant to reflect the required use cases, it's not 100%, so wherever possible, consider ensuring the presence of tests to validate the behavior as it is now (or how we want it to be).

I'll be honest, while I'm hopeful that pip might replace easy_install, it may prove a difficult transition to make. I welcome the effort, but with the understanding that it may prove infeasible to roll out. Anything you can do to help with that effort (soliciting feedback on distutils-sig) will help a lot in getting this ultimately accepted.

xoviat:
> I'm a bit concerned about the inclusion of a dependency on pip, as it starts to create a weird circular dependency where pip requires setuptools to build packages during install, but setuptools uses pip to install packages during builds.

I was concerned about this but I remember that other people assured me it would not be a problem. 

> I think you're on the right track here, and I'm eager to see where you take this.

Actually given that egg and whl are both zip formats, it would be possible to use the wheel command and then add those to the pythonpath.

PR #1037, opened 2017-05-18
egg_info: allow specification of setup requirements

xoviat:
So I though about the setuptools dependency on pip and really did not like that idea because is more closely couples the two. This is a simple change that will allow pip to ask for build dependencies from setuptools if possible. 

This doesn't work all of the time and doesn't cover every corner case, but it is an improvement from the current situation.


xoviat:
@jaraco Can you help with the sandbox issues?

jaraco:
I've opened #1038 to track down the issue with the tests failing.

PR #1039, opened 2017-05-22
Implement PEP 517 Build Backend

xoviat:
This complies with the pep that will probably be adopted in any case, and doesn't break any existing functionality.

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -26,7 +26,7 @@ from .py36compat import Distribution_parse_config_files
 
 
 skip_install_eggs = False
-
+ 
 
 class SetupRequirementsError(BaseException):
     def __init__(self, specifiers):
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -25,6 +25,14 @@ import pkg_resources
 from .py36compat import Distribution_parse_config_files
 
 
+skip_install_eggs = False
+
+
+class SetupRequirementsError(BaseException):
+    def __init__(self, specifiers):
+        self.specifiers = specifiers
+
+
 def _get_unpatched(cls):
     warnings.warn("Do not call this function", DeprecationWarning)
     return get_unpatched(cls)
@@ -315,7 +323,10 @@ class Distribution(Distribution_parse_config_files, _Distribution):
             self.dependency_links = attrs.pop('dependency_links', [])
             assert_string_list(self, 'dependency_links', self.dependency_links)
         if attrs and 'setup_requires' in attrs:
-            self.fetch_build_eggs(attrs['setup_requires'])
+            if skip_install_eggs:
+                raise SetupRequirementsError(attrs['setup_requires'])
+            else:
+                self.fetch_build_eggs(attrs['setup_requires'])
         for ep in pkg_resources.iter_entry_points('distutils.setup_keywords'):
             vars(self).setdefault(ep.name, None)
         _Distribution.__init__(self, attrs)
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -25,7 +25,7 @@ import pkg_resources
 from .py36compat import Distribution_parse_config_files
 
 
-skip_install_eggs = False
+_skip_install_eggs = False
 
 
 class SetupRequirementsError(BaseException):
@@ -323,10 +323,9 @@ class Distribution(Distribution_parse_config_files, _Distribution):
             self.dependency_links = attrs.pop('dependency_links', [])
             assert_string_list(self, 'dependency_links', self.dependency_links)
         if attrs and 'setup_requires' in attrs:
-            if skip_install_eggs:
+            if _skip_install_eggs:
                 raise SetupRequirementsError(attrs['setup_requires'])
-            else:
-                self.fetch_build_eggs(attrs['setup_requires'])
+            self.fetch_build_eggs(attrs['setup_requires'])
         for ep in pkg_resources.iter_entry_points('distutils.setup_keywords'):
             vars(self).setdefault(ep.name, None)
         _Distribution.__init__(self, attrs)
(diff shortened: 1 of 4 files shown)


xoviat:
@jaraco This should be okay to go if the tests pass. It's not bulletproof but pip can always fall back.

jaraco:
Can you provide a link to the pep? Why use `py517` instead of `pep517`?

xoviat:
Wow, it looks like it's been accepted.

https://www.python.org/dev/peps/pep-0517/


@jaraco I'll rename the module. I think pip was originally going to have it's own setuptools backend but if we put the backend in the setuptools repository, then pip can just require the latest setuptools by default and then execute the backend according to the standards. In other words, it move all of the monkeypatching and hackery out of pip and into setuptools, where it belongs.

xoviat:
@ncoghlan Can you specify the additonal concerns for out-of-tree builds?

xoviat:
@pfmoore 

> Hello all:
> 
> I have been attempting to resolve the horrible easy_install behavior where compiled wheels published on pypi are not used to fulfill setup requirements. To address this issue, I have submitted pull requests to both setuptools and pip that would allow a proprietary communication method that can later be easily adjusted to be fully PEP 517 compliant, saving work while resolving a specific issue.
> 
> However, everyone appears to be ignoring these pull requests, in my estimation, because PEP 517 is not fully completed. That’s fine, but partial PEP 517 compliance in one area is better than nothing at all, which is currently the default scenario.
>

setuptools/pep517.py
```diff
@@ -0,0 +1,96 @@
...
+def build_wheel(wheel_directory, config_settings=None,
+                metadata_directory=None):
+    wheel_directory = os.abspath(wheel_directory)
+    sys.argv = sys.argv[:1] + ['bdist_wheel'] + \
+        config_settings["--global-option"]
+    _run_setup()
+    if wheel_directory != 'dist':
+        shutil.rmtree(wheel_directory)
+        shutil.copytree('dist', wheel_directory)
+
```
xoviat: Note that we need to copy and extract the sdist first to comply with the point about equivalent wheels.

xoviat:
So just as a reference we need to use `wheel.egg2wheel` to convert egg_info to dist_info.

setuptools/dist.py
```diff
@@ -315,7 +323,10 @@ def __init__(self, attrs=None):
             self.dependency_links = attrs.pop('dependency_links', [])
             assert_string_list(self, 'dependency_links', self.dependency_links)
         if attrs and 'setup_requires' in attrs:
-            self.fetch_build_eggs(attrs['setup_requires'])
+            if skip_install_eggs:
+                raise SetupRequirementsError(attrs['setup_requires'])
+            else:
```
jaraco: You don't need the else block here.

setuptools/dist.py
```diff
@@ -25,6 +25,14 @@
 from .py36compat import Distribution_parse_config_files
 
 
+skip_install_eggs = False
```
jaraco: This variable appears to be undocumented and untested. It could be optimized away.

I wonder if instead of setting a global flag that tweaks the inner behavior of the Distribution object for all users if it wouldn't be better for the caller that would use this functionality to instead create a:

```
class Distribution(setuptools.dist.Distribution):
    def fetch_build_eggs(self, eggs):
        raise SetupRequirementsError(...)
```

It's not yet clear to me where a user or application would hook into this behavior.

Also, the name 'skip_install_eggs' doesn't match the behavior. It's 'fail on install eggs'.

jaraco:
What's the status of this PR? It's still tagged WIP, yet one comment says it's good to go. Other comments still say "we need", suggesting there's more work to be done.

I've added some comments about the implementation. I'm generally fine with new pep517 module, but I'm reluctant to add implicitly-defined hooks that complicate the (already complicated) behavior. I've suggested another possible approach which may prove simpler, but first, I think we need to understand how it is that this behavior is intended to be used.

xoviat:
> What's the status of this PR? It's still tagged WIP, yet one comment says it's good to go. Other comments still say "we need", suggesting there's more work to be done.

The reason for this is that the PEP was modified after it was initially accepted. It's not currently acceptable to merge. I'm working on the pip side of the chain first and will then update this.

setuptools/dist.py
xoviat: This variable is for internal use only. The public api is `get_requires_for_build_wheel`

xoviat:
@jaraco Let's go ahead and get this in here as the pip people seem to be taking their time.

setuptools/dist.py
```diff
@@ -25,6 +25,14 @@
 from .py36compat import Distribution_parse_config_files
 
 
+_skip_install_eggs = False
+
```
xoviat: So again I want to emphasize that this is an internal setuptools implementation detail that should not be used outside of setuptools.

setuptools/pep517.py
```diff
@@ -0,0 +1,91 @@
...
+def get_requires_for_build_wheel(config_settings=None):
+    config_settings = fix_config(config_settings)
+    return get_build_requires(config_settings)
+
+
+def get_requires_for_build_sdist(config_settings=None):
+    config_settings = fix_config(config_settings)
+    return get_build_requires(config_settings)
+
```
xoviat: These two `get_requires` functions are the public API. They will be used by pip to determine the setup requirements.

xoviat:
>  I've suggested another possible approach which may prove simpler, but first, I think we need to understand how it is that this behavior is intended to be used.

I added some tests that should help clarify this matter. Let me know if you have any other questions.

setuptools/pep517.py
```diff
@@ -0,0 +1,91 @@
...
+
+def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
+    sys.argv = sys.argv[:1] + ['dist_info', '--egg-base', metadata_directory]
+    _run_setup()
+
```
xoviat: This will start working once pypa/wheel#190 is merged but there is no need to wait for that.

Issue #940, opened 2017-01-24
Undocumented dependency on new-ish pip


jcsp:
As of now, anyone running "pip install --upgrade setuptools" on a CentOS 6 machine will find it fails horribly:
```
pip install --upgrade setuptools
Downloading/unpacking setuptools from https://pypi.python.org/packages/b1/d1/3df86b914d72b18d37266f964a11576f135eadbb7d2e3c5165377e4263bc/setuptools-34.0.1.zip#md5=9e907950a94d38c0295df0da2507c280
  Downloading setuptools-34.0.1.zip (617kB): 617kB downloaded
  Running setup.py egg_info for package setuptools
    Traceback (most recent call last):
      File "<string>", line 3, in <module>
      File "setuptools/__init__.py", line 10, in <module>
        from six.moves import filter, map
    ImportError: No module named six.moves
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 3, in <module>

  File "setuptools/__init__.py", line 10, in <module>

    from six.moves import filter, map

ImportError: No module named six.moves
```

I discovered this when a long-functioning build script started failing all of a sudden.

I gather from #937 and #938 that a more recent pip is required, and indeed installing it with `pip install --upgrade pip` does allow installation of setuptools to proceed.

May I suggest that the README should specify what version of pip is required, as it is a newer version than is present on many systems.

jaraco:
I don't think there's a specific version of pip that's required. It probably varies by platform and maybe by Python version. At the very least, whatever pip version is used needs to rely on the wheel format and not the sdist. If you would like to try different versions, I could update the README or the CHANGES to indicate which version to use.

jcsp:
I can tell you that pip 1.4.1 does not work, and pip 9.0.1 does work.

If the required version was not known/is not knowable at the point the new dependency was introduced and pushed into the wild, I guess my best suggestion is a big health warning that people should install the latest pip if they're using the latest setuptools.

Alternatively, you could indicate which version of setuptools people should pin to in order to have something that will run on stable distributions that ship an older pip.

shadowmint:
We're seeing this too with pip 1.5.4

adambkaplan:
I'm also seeing this issue with our builds on CircleCI, running Ubuntu 12.04 (Precise).

zroadhouse-rmn:
We have a build pipeline that packs up all the sdists for our application and its dependencies and installs then using only the local index.

[...]

PR #1043, opened 2017-06-01
fixed #1042 -- corrected an import

### LINKED ISSUE #1042: No module named 'six'
Flask's Tox/Travis builds all started failing. Looks like a new version of setuptools was just released that has a problem with six.

~~~pytb
Obtaining file:///home/david/Projects/flask
Installing collected packages: Flask
  Running setup.py develop for Flask
    Complete output from command /home/david/Projects/flask/.tox/py/bin/python3 -c "import setuptools, tokenize;__file__='/home/david/Projects/flask/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps:
    /home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/dist.py:336: UserWarning: Normalizing '0.13-dev' to '0.13.dev0'
      normalized_version,
      File "/home/david/Projects/flask/setup.py", line 109, in <module>
        '''
      File "/usr/lib64/python3.6/distutils/core.py", line 134, in setup
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/dist.py", line 363, in parse_command_line
        result = _Distribution.parse_command_line(self)
      File "/usr/lib64/python3.6/distutils/dist.py", line 472, in parse_command_line
        args = self._parse_command_opts(parser, args)
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/dist.py", line 674, in _parse_command_opts
        nargs = _Distribution._parse_command_opts(self, parser, args)
      File "/usr/lib64/python3.6/distutils/dist.py", line 528, in _parse_command_opts
        cmd_class = self.get_command_class(command)
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/dist.py", line 495, in get_command_class
        self.cmdclass[command] = cmdclass = ep.load()
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2303, in load
        return self.resolve()
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2309, in resolve
        module = __import__(self.module_name, fromlist=['__name__'], level=0)
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/command/develop.py", line 11, in <module>
        from setuptools.command.easy_install import easy_install
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 49, in <module>
        from setuptools.py27compat import rmtree_safe
      File "/home/david/Projects/flask/.tox/py/lib/python3.6/site-packages/setuptools/py27compat.py", line 7, in <module>
        import six
    ModuleNotFoundError: No module named 'six'
~~~

Example failed build log: https://travis-ci.org/pallets/flask/jobs/238166427#L242

alex:


dstufft:
👍

glyph:
Presumably this should be force-merged since the former released version's breakage is what's breaking the build?

alex:
That does appear to be the case.

dmick:
Please merge and release asap

jakepruitt:
^ Seconded, we really would appreciate this getting merged as soon as possible.

davidism:
I'm sure they're already aware that everyone would like this merged, so can we please not all chime in with the equivalent of "+1"? It just spams everyone's notifications. Use the reaction buttons.

bertjwregeer:
The reaction buttons don't get an email sent to any maintainers/watchers of the project. Adding comments does.

It really is not acceptable to break the world, then when notified leave the breakage for hours (at this point it has been 4+ hours)

alex:
You're almost certainly sending notifications to someone who is asleep, which isn't productive.

fake-name:
I kinda feel that for packages as core to basically everything python as pip is, maybe schedule the release for the start of the day, rather then right before leaving to go to sleep?

Having someone who can do an emergency rollback/fix available for a few hours after a new version is released doesn't seem like a huge overhead. Just put that late-afternoon release off to the next morning.

kornicameister:
Seems like this bug affects majority of OpenStack [bug](https://bugs.launchpad.net/devstack/+bug/1694904).

Guess this is how it sounds like ;/

mosquito:
Guys it's super critical for me. All deployment process was broken.

mosquito:
@jaraco all python ecosystem was broken. Please merge it ASAP!!!

2mf:
please merge

punkr0ck:
When installing `Let's Encrypt`, get this error:

```
Installing Python packages...
  Failed building wheel for pip
  Failed building wheel for wheel
Command "/root/.local/share/letsencrypt/bin/python2.7 -u -c "import setuptools, tokenize;__file__='/tmp/pip-qv6is_-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-S16j4X-record/install-record.txt --single-version-externally-managed --compile --install-headers /root/.local/share/letsencrypt/include/site/python2.7/pip" failed with error code 1 in /tmp/pip-qv6is_-build/
  File "/tmp/tmp.rKuEI000oW/pipstrap.py", line 146, in <module>
    exit(main())
  File "/tmp/tmp.rKuEI000oW/pipstrap.py", line 133, in main
    shell=True)
  File "/usr/lib/python2.7/subprocess.py", line 574, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command 'pip install --no-index --no-deps -U /tmp/pipstrap-Ln63ah/pip-8.0.3.tar.gz /tmp/pipstrap-Ln63ah/setuptools-20.2.2.tar.gz /tmp/pipstrap-Ln63ah/wheel-0.29.0.tar.gz' returned non-zero exit status 1
```

setuptools/py27compat.py
```diff
@@ -4,7 +4,7 @@
 
 import platform
 
-import six
+from setuptools.extern import six
```
fwdevmobile: Why using six???

In package six the Python version is determined by calling sys.

PY2 = sys.version_info[0] == 2
See: https://bitbucket.org/gutworth/six/src/92e1c746e1abfbdb94705b821bc93766083efc54/six.py?at=default&fileviewer=file-view-default

Best solution is to use "sys" to check if it is python 2 version.

If it must be version 2.7 check for major and minor version!

sbidoul: Revert https://github.com/pypa/setuptools/commit/1d928cbc7b2cfcf1ffd2ec27f83ee33f0af39dfe ?

mosquito: actually reverting should looks better.

fwdevmobile: yes, reverting is a much cleaner solution.

westsouthnight: all people still wait, if you can please revert this shortly

The-Compiler: `six` is already used elsewhere, so it makes sense to use it here as well (after the import is corrected).

@westsouthnight This was worked around by pulling the affected version from PyPI, so things should be fine again for now.

fwdevmobile: I have seen that it was used in other parts to check if an object is of the type "isinstance(value, six.string_types)". Does that mean that this part of the code must have six? I just can't see that "import sys" must be replaced by six. It doesn't solve a problem in my opinion.

anthrotype:
Thanks @jaraco!

jaraco:
Thanks for providing this change. I apologize that I wasn't available to review and cut a release. I have made the release process thoroughly documented and readily available to anybody on PyPA. I typically also leave time after cutting a release to respond to any regressions, but due to the two botched attempts, I ran out of time and was relying on the passing tests to assure the release. Perhaps setuptools test suite needs a test that it can be installed in a clean environment.

mosquito:
Please provide a new release.

setuptools/py27compat.py
The-Compiler: @fwdevmobile It doesn't mean it *must* have six. But it makes sense for it to use it when available, to - even if only for a tiny bit - increase readability and reduce technical debt. It's unfortunate that there was a mistake in it, but generally, it - IMHO - is a slight improvement.

fwdevmobile: I do agree with the new import will solve all deployment issues. Please merge.

```
```

thanatos:
> I apologize

I for one don't believe I'm owed an apology here. We've all broken the build/production from time-to-time, but I don't think I'm capable of imaging the enormous pressure that must come with breaking *everyone's* builds. 😉 Python, and the tooling around it like PyPI, `setuptools`, and `pip`, are all **free** — but the value we as developers derive from it is enormous. Thanks for taking the time to be a maintainer of such a critical piece of infrastructure.

A patch for the problem was available within minutes, and I got an instant response on IRC channel. I think if anything, we owe you an apology for the incessant clamoring for an instant fix/release.

> I have made the [release process](http://setuptools.readthedocs.io/en/latest/releases.html) thoroughly documented and readily available to anybody on PyPA.

:+1: Thanks for this. There were also several good suggestions for us as *users* on the issue.

hynek:
@jaraco unless you have a paid SLA with anyone you don’t owe us any apologies.  It’s incredible how robust Python’s packaging has become due to PyPA’s tireless efforts.  So thank you once more!  The complaining wouldn’t be half as loud, if the baseline reliability weren’t so good.

***

Those complaining loudly about broken deployments should be thankful for this reminder, that *everything* can break and it’s part of your jobs to deal with it.  Google goes internally as far as introducing outages if a system has been too reliable for too long (see “Don’t overachieve” in the [SLO chapter of the SRE book](https://landing.google.com/sre/book/chapters/service-level-objectives.html)).

This one was a minor incident that got fixed within a short period.  There are totally scenarios where infrastructure you take for granted and have not influence on goes away for a week or longer.  How will you cope?

willingc:
@jaraco Thank you! I agree with @hynek that you and PyPA's maintainers and contributors have done a fantastic job the past few years. Thank you for being professional, responsive, helpful, and enhancing the user experience tremendously. 🌷 🌹 :tulip:

@hynek Thank you for adding thoughtful perspective on this too.

mosquito:
I apologize if I offended someone. However, I wrote only the facts. The ecosystem was actually broken. The worst thing is that there was no way to get around this problem for me. @jaraco did everything right after all. Everyone could make such a mistake. Setuptools is a very reliable package. Thank you.

bertjwregeer:
@jaraco and other setuptools maintainers I sincerely apologise for my comment, and the strife it has caused.

----

I've been getting a lot of flack for my comment above, especially on the Twitter sphere, and in private via email. Lot of people saying people should have known to pin the package or anything along those lines.

Let me state up front that when I posted that comment I was incredibly frustrated, mainly because it wasn't until this started happening that I found out that `virtualenv` by default -- upon creation of a new virtualenv -- will happily go fetch and download the latest version from `pypi` and install it. Everything else in production at $WORK is pinned to a private PyPI server that contains the packages we want with the versions we want, however unlike `python3 -mvenv` there was no realisation (from anyone on the team) that `virtualenv` would go download stuff from the internet.

Ansible with a pip statement will run `virtualenv` if you provide a virtualenv path, there is no flag to disable the default behaviour (to for example add `--no-download`), tox uses virtualenv internally, there is no way to pass `--no-download`. There is no way to specify with the `virtualenv` command to go fetch from an alternative PyPI location (`-i` flag is missing, I've opened an issue for this) other than exporting an environment flag that is not well documented (and happens to be consumed by `pip` during a run of `virtualenv`).

The `virtualenv` usage docs: https://virtualenv.pypa.io/en/stable/userguide/#usage don't specify it is going to download code from the internet. I was fully expecting it to deploy with the `setuptools` it bundles.

Sure there are work-arounds, and I am sure there is a place in the `virtualenv` docs where it describes that by default it will go fetch things from PyPI, but the amount of long-time Python developers that were not aware of the default `virtualenv` behaviour within not just my team but people that I've spoken to in the community makes this a much bigger problem.

----

I maintain a variety of open source projects, I know how thankless of a job it is, and had I put on my maintainer hat when I posted the comment above I wouldn't have hit the comment button, I don't envy the position that `setuptools` has by being such an important part of the Python eco-system and thus being in a place where breakage is felt so far and wide.

mmerickel:
Expanding on @bertjwregeer's excellent point about `--no-download`, the issue here is that even if we pinned setuptools in our `requirements.txt` we would still be broken because the `virtualenv` package installed the latest setuptools already which broke the virtualenv. We would not be able to downgrade to our pinned version after that. Secondly, `pip freeze` **by default** excludes setuptools making it that much harder and non-obvious to pin it - **even when** a package in your tree depends on setuptools explicitly for runtime pkg_resources features. In this new age of setuptools bumping the major version number frequently it's possible these defaults should be re-evaluated to make things easier to pin and to create more stable environments without requiring things like `VIRTUALENV_NO_DOWNLOAD=1` and `pip freeze --all`.

dstufft:
We're likely going to be able to stop excluding setuptools in ``pip freeze`` soon (or installing it by default inside of a virtual environment at all). The work done in PEP 518 opens up a path to being able to do that. We'll likely want some time with the feature actually being released before we get to that point, but that's what I expect to happen in the near future.

shadowmint:
This isn't really the place for this discussion, but if you recall https://github.com/pypa/setuptools/issues/940 from a couple of months ago, this isn't the first time virtualenv behaviour has been broken with setuptools in recent history.

Some kind of mitigation for this repeated type of failure should be put in place.

Where's the best place to take this discussion to move it forward?

leorochael:
As someone who, using `zc.buildout`, has been exposed to breakages arriving from eventual mismatch between `setuptools` and `buildout`, here is what I use for repeatable setups:

    vitualenv --no-setuptools --no-pip sandbox
    sandbox/bin/python <(curl https://bootstrap.pypa.io/get-pip.py) --upgrade setuptools==33.1.1 pip==9.0.1

The trick is that `get-pip.py`, when passed additional arguments, behaves like `pip install`, so you can now make it install exactly the packages you need, in whatever version you need from whatever location you need, and virtualenv won't have gone over the internet to fetch some arbitrary version of setuptools or pip.

PR #1046, opened 2017-06-01
fix: No module named 'six' #1042

kulapard:


The-Compiler:
As mentioned in https://github.com/pypa/setuptools/pull/1043/files I don't think this makes sense - since `six` is already used elsewhere in `setuptools`, why not use it for this as well? It's just the import which needs to be fixed up.

kulapard:
@The-Compiler you are totally right. Thanks!

davidism:
Duplicate of #1043.

PR #1050, opened 2017-06-01
[WIP] Add a test that setuptools can be installed in a clean environment

alexwlchan:
A two-minute attempt to pick up @jaraco’s comment from #1043:

> Perhaps setuptools test suite needs a test that it can be installed in a clean environment.

This script spins up a clean virtualenv without any of the usual pip/setuptools gubbins, then tries to install setuptools. What could go wrong.

I’ve confirmed this script fails on v36.0.0 and passes on v36.0.1 (so would likely have caught #1042). Not entirely sure it’s in the right place for Travis though (and might need to run on Appveyor too?).

Probably needs more tidying/cleanup, I just wanted to put up as a proof-of-concept.

### DIFF
--- a/clean_install.sh
+++ b/clean_install.sh
@@ -5,7 +5,9 @@ set -o xtrace
 
 # Create a temporary directory to install the virtualenv in
 VENV_DIR="$(mktemp -d)"
-function cleanup { rm -rf "$VENV_DIR" }
+function cleanup() {
+  rm -rf "$VENV_DIR"
+}
 trap cleanup EXIT
 
 # Create a virtualenv that doesn't have pip or setuptools installed


jaraco:
Good start, and thanks for the proof-of-concept. I think ultimately, I'd like something that's integrated with the test suite, and maybe something that doesn't rely on virtualenv (even if it only runs on Python 3.3+). Perhaps porting this test to the test_integration.py module would achieve those goals.

But for now, this test serves the purpose, so let's start with it.

jaraco:
Hmm... except that the tests aren't passing as written. Strange error about unexpected end of file. Is there a missing carriage return in the last line?

alexwlchan:
I think it was an error in the way I’d laid out the cleanup function – more sensitive to newlines than I thought. I think the script works now.

PR #1058, opened 2017-06-05
Set license to MIT in setup.py

all3fox:
Currently, `pip show setuptools` shows license field as "UNKNOWN", this PR fixes that.


jaraco:
The issue is a bug in pip that it doesn't honor the [documented preferred declaration in classifiers](https://docs.python.org/2/distutils/setupscript.html#additional-meta-data), which setuptools does present.

Issue #439, opened 2015-09-17
develop command should not pin the distribution version precisely


ghost:
Originally reported by: **RonnyPfannschmidt (Bitbucket: [RonnyPfannschmidt](http://bitbucket.org/RonnyPfannschmidt), GitHub: [RonnyPfannschmidt](http://github.com/RonnyPfannschmidt))**

---

When easy_install or pip installs console entry points (scripts), it does so by referencing the installed version of the package. The `develop` command uses the same logic to create executable scripts for a package under development.

However, it is common when developing a package that the version number will change, thereafter making the script references invalid (and causing errors when attempting to invoke those scripts).

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/439

PR #1061, opened 2017-06-12
remove version from develop scripts

dsludwig:
This extends https://github.com/pypa/setuptools/issues/439 to remove the specific version for scripts generated by `python setup.py develop` from projects using `scripts` instead of `entrypoints.console_scripts`.


jaraco:
Thanks for the contrib. Unfortunately, the tests are failing. Please revisit and resubmit once the tests are once-again passing.

PR #1065, opened 2017-06-20
travis: upgrade and cache pip

xoviat:


jaraco:
What's the value in upgrading pip if all it installs is tox?

xoviat:
I was thinking that we may want integration tests for PEP 517.

.travis.yml
```diff
@@ -8,13 +8,19 @@ python:
   - 3.6
   - nightly
   - pypy
+cache:
```
webknjaz: Just `cache: pip` would work. And also there's this syntax:
```yaml
cache:
  pip: true
```

I don't really like the idea of hardcoding paths.

webknjaz: (ref: https://github.com/pypa/setuptools/pull/1053/files?diff=split#diff-354f30a63fb0907d4ad57269548329e3R34)

webknjaz:
Is this needed for all envs?

jaraco:
This PR needs a better explanation of its motivations and value. I can infer that enabling pip cache is a good thing, but if it's a best practice, then we should be able to reference documentation or a conversation that indicates that this setting is a best practice (and if it is, I'll want to incorporate it with my other projects).

As for upgrading pip, that feels like a separate change with separate motivations that aren't even borne out. I don't know PEP 517 by name, but even if I did, it's not clear to me why upgrading pip in this commit helps. There should be a separate pull request that summarizes what PEP 517 does and how Setuptools can better support that and including the integration tests that go along with the upgrading pip stage.

PR #1068, opened 2017-06-21
Sort file lists

bmwiedemann:
to generate reproducible zip files
that do not differ depending on (random) filesystem order

See https://reproducible-builds.org/ for why this matters.

### DIFF
--- a/setuptools/command/bdist_egg.py
+++ b/setuptools/command/bdist_egg.py
@@ -38,6 +38,14 @@ def strip_module(filename):
         filename = filename[:-6]
     return filename
 
+def sorted_walk(dir):
+    """Do os.walk in a reproducible way,
+    independent of indeterministic filesystem readdir order
+    """
+    for base, dirs, files in os.walk(dir):
+        dirs.sort()
+        files.sort()
+        yield base, dirs, files
 
 def write_stub(resource, pyfile):
     _stub_template = textwrap.dedent("""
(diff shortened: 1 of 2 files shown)


setuptools/command/bdist_egg.py
```diff
@@ -303,7 +303,8 @@ def get_ext_outputs(self):
 
         paths = {self.bdist_dir: ''}
         for base, dirs, files in os.walk(self.bdist_dir):
-            for filename in files:
+            dirs.sort()
+            for filename in sorted(files):
```
jaraco: Why use `sorted` for files but `.sort` for dirs?

setuptools/command/bdist_egg.py
```diff
@@ -331,6 +332,8 @@ def walk_egg(egg_dir):
     """Walk an unpacked egg's contents, skipping the metadata directory"""
     walker = os.walk(egg_dir)
     base, dirs, files = next(walker)
+    dirs.sort()
```
jaraco: It's starting to feel like this change is getting sprinkled around a few places. Maybe it would make more sense to have a "sorted walk" function or similar.

bmwiedemann:
Thanks for your nice feedback.

I added a sorted_walk function with docstring, but I have not fully tested it
and I could use some help with adding a test, because my python-foo is not the best.
Something like:
 

1. create numbered file entries 1..9 (in random order)
2. create an .egg from it
3. check that entries in the zipfile are ordered 1..9

jaraco:
This new PR is much cleaner. And given that eggs are largely deprecated in favor of wheels, I'm not going to be a stickler about the test.

PR #1063, opened 2017-06-14
make `minimal_egg_info` string in `bootstrap.py` unicode

mattgiles:
I believe this is likely to be safe in general, but have only tested with 2.7 on OS X.

Without this change, I am seeing error:

```
$ git clone git@github.com:pypa/setuptools
Cloning into 'setuptools'...
remote: Counting objects: 24321, done.
remote: Compressing objects: 100% (23/23), done.
remote: Total 24321 (delta 11), reused 20 (delta 7), pack-reused 24291
Receiving objects: 100% (24321/24321), 28.77 MiB | 7.38 MiB/s, done.
Resolving deltas: 100% (8669/8669), done.
$ cd setuptools
$ git rev-parse HEAD
995d309317c6895a123c03df28bc8f51f6ead5f5
$ python bootstrap.py
adding minimal entry_points
Traceback (most recent call last):
  File "bootstrap.py", line 63, in <module>
    __name__ == '__main__' and main()
  File "bootstrap.py", line 59, in main
    ensure_egg_info()
  File "bootstrap.py", line 37, in ensure_egg_info
    build_egg_info()
  File "bootstrap.py", line 47, in build_egg_info
    ep.write(minimal_egg_info)
TypeError: must be unicode, not str
```

PR #1070, opened 2017-06-26
Tests releated fixes and tweaks

benoit-pierre:
* move pytest report arguments to `pytest.ini`, so running tox with arguments does not loose those
* add xpassed tests to the pytest report
* fix `fail_on_ascii` fixture: correct detection
* mark `test_unicode_filename_in_sdist` with `fail_on_ascii`
* fix Travis ASCII builds: as evidenced by the log, all tests marked with `fail_on_ascii` were not xfailed

~~Note: first commit is a fix for the same issue as #1063, so I'll rebase once this is merged.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ python:
 matrix:
   include:
   - python: 3.6
-    env: LC_ALL=C LC_CTYPE=C
+    env: LANG=C
   - python: 2.7
-    env: LC_ALL=C LC_CTYPE=C
+    env: LANG=C
 script:
  # need tox to get started
  - pip install tox
--- a/setuptools/tests/test_easy_install.py
+++ b/setuptools/tests/test_easy_install.py
@@ -30,6 +30,7 @@ from setuptools.dist import Distribution
 from pkg_resources import normalize_path, working_set
 from pkg_resources import Distribution as PRDistribution
 import setuptools.tests.server
+from setuptools.tests import fail_on_ascii
 import pkg_resources
 
 from .py26compat import tarfile_open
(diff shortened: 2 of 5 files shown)


benoit-pierre:
Rebased on master.

.travis.yml
```diff
@@ -11,9 +11,9 @@ python:
 matrix:
   include:
   - python: 3.6
-    env: LC_ALL=C LC_CTYPE=C
+    env: LANG=C
```
jaraco: I seem to recall having once started with this flag and later found I had to use LC_ALL and LC_CTYPE. In 4660f43, I made the change. I don't recall why I made that change though. Do you have any reference that indicates that this is the proper mechanism?

tox.ini
```diff
@@ -7,5 +7,5 @@
 [testenv]
 deps=-rtests/requirements.txt
 passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR
-commands=py.test {posargs:-rsx}
```
jaraco: The reason I had the arguments here was I wanted them to be readily overwritten. Otherwise, I don't think there's a way to omit or change the reporting. I do see how it's somewhat counter-intuitive when one specifies an argument for another purpose, though. Yes, perhaps it's better the way you've done it.

setuptools/tests/test_easy_install.py
```diff
@@ -166,6 +167,7 @@ def sdist_unicode(self, tmpdir):
         sdist_zip.close()
         return str(sdist)
 
+    @fail_on_ascii
```
jaraco: I'm guessing this change is the reason you're having to pass LANG=C.

jaraco:
Yes, Travis indicates that the tests are xfailing where expected. Thanks!

PR #1075, opened 2017-07-03
Add new long_description_content_type kwarg

di:
This is a re-opening of #708 which was closed.

Since then, https://github.com/pypa/python-packaging-user-guide/pull/258 has been merged and is now documented here: https://packaging.python.org/specifications/#description-content-type

I've rebased it and amended the commit messages to reflect that https://github.com/pypa/python-packaging-user-guide/pull/258 was accepted.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+v36.4.0
+-------
+
+* #1075: Add new ``Description-Content-Type`` metadata field. `See here for
+  documentation on how to use this field.
+  <https://packaging.python.org/specifications/#description-content-type>`_
+
 v36.3.0
 -------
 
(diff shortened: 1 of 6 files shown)


lynncyrin:
ping! markdown readme's would be great to have

jaraco:
Looks good. Can you just add a CHANGES.txt entry notifying users of what's changed and how they can use it? Also, are there relevant docs changes that need to address this functionality?

di:
@jaraco I added an entry to `CHANGES.rst` and updated a table in the docs which lists the available fields.

I didn't really see anywhere where the documentation goes in-depth about `long_description` where it felt appropriate to add more explanation about `long_description_content_type`, but if you feel there is a place, I am happy to add it.

It might be worth mentioning here (but not in the docs) that until this is implemented in  [pypa/readme_renderer](https://github.com/pypa/readme_renderer), this field will not actually be very useful to the user.

PR #1081, opened 2017-07-10
Fix environment markers handling

benoit-pierre:
While the [documentation](https://github.com/pypa/setuptools/blob/master/docs/setuptools.txt#L767) now points to using environment markers with `install_requires` for declaring platform dependencies, this result in invalid wheels because those are stripped from the final requirements in the embedded metadata.

E.g, using something like this:
```python
from setuptools import setup
setup(
    name='foo',
    install_requires='barbazquux; "foobar" in sys_platform',
)
```
result in `foo.egg-info/requires.txt` containing:
```
barbazquux; "foobar" in sys_platform
```
instead of:
```python
import sys

from setuptools import setup

setup(
    name='foo',
    extras_require={':"foobar" in sys_platform': 'barbazquux'},
)
```
which yields:
```
[:"foobar" in sys_platform]
barbazquux
```

This PR automatically move requirements in `install_requires` that use environment markers to `extras_require` so the egg info is valid and the generated wheels work as expected.

Additionally, I added a check to error out if an `extras_require` requirement is itself using an environment marker, as it's not supported.

~~Note: first commit is a fix for the same issue as #1063, so I'll rebase once this is merged.~~

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -8,6 +8,7 @@ import distutils.log
 import distutils.core
 import distutils.cmd
 import distutils.dist
+from collections import defaultdict
 from distutils.errors import (DistutilsOptionError, DistutilsPlatformError,
     DistutilsSetupError)
 from distutils.util import rfc822_escape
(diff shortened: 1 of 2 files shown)


benoit-pierre:
Rebased on master.

PR #1083, opened 2017-07-13
when trying to create a folder gracefully fall back when it already exists

dirk-thomas:
When installing multiple packages into the same location in parallel each packages installation step tries to create the same directories (e.g. the `bin` folder). Currently the `isdir` and `makedirs` call are performed sequentially which makes this logic vulnerable to a race condition. The `isdir` call will return `False` but the following `makedirs` call will fail since the directory has been created in the meantime by another package being installed in parallel.

This patch changes the logic to use the Pythonic approach to try to create the directory and in case it already exists ignore the exception.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -12,7 +12,7 @@ import subprocess
 import io
 
 
-minimal_egg_info = textwrap.dedent("""
+minimal_egg_info = textwrap.dedent(u"""
     [distutils.commands]
     egg_info = setuptools.command.egg_info:egg_info
 
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -17,6 +17,7 @@ method.
 
 from __future__ import absolute_import
 
+import errno
 import sys
 import os
 import io
@@ -2958,8 +2959,11 @@ def _find_adapter(registry, ob):
 def ensure_directory(path):
     """Ensure that the parent directory of `path` exists"""
     dirname = os.path.dirname(path)
-    if not os.path.isdir(dirname):
+    try:
         os.makedirs(dirname)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            raise
 
 
 def _bypass_ensure_directory(path):


benoit-pierre:
See #1063 for the CI failures.

dirk-thomas:
@benoit-pierre Thank you for the quick pointer. With that change CI passes.

bootstrap.py
```diff
@@ -12,7 +12,7 @@
 import io
 
 
-minimal_egg_info = textwrap.dedent("""
+minimal_egg_info = textwrap.dedent(u"""
```
jaraco: Use `from __future__ import unicode_literals` if unicode literals are required... which I don't think they are here. Scrap this change.

pkg_resources/__init__.py
```diff
@@ -2958,8 +2959,11 @@ def _find_adapter(registry, ob):
 def ensure_directory(path):
     """Ensure that the parent directory of `path` exists"""
     dirname = os.path.dirname(path)
-    if not os.path.isdir(dirname):
+    try:
```
jaraco: This is good, and I was going to accept it straight-up, but given that this technique is superseded in Python 3.2 with the [exists_ok parameter](https://docs.python.org/3.2/library/os.html#os.makedirs). Therefore, I'd prefer to use the py27compat or py31compat module to house this legacy approach. Can you put it in py31compat (as a makedirs function that accepts exists_ok kwarg)?

jaraco:
Ugh. I now I note that since this is in pkg_resources, and pkg_resources can't import setuptools, we need a new py31compat module in pkg_resources. I'll just make the changes.

bootstrap.py
dirk-thomas: The change from ##1063 is only being pulled in here to make CI pass.

pkg_resources/__init__.py
dirk-thomas: That is why the branch is called "makedirs_exist_ok" :wink: But since the code needs to work in Python < 3.2 I didn't use the new parameter.

dirk-thomas:
> I'll just make the changes.

Great, thank you for getting this fixed.

Can you please comment here with a reference to the applied change once that is done. That will allow future users to trace as of which version the fix from this PR will be available.

jaraco:
Should  be fixed in v36.1.0, going out now (assuming tests pass).

dirk-thomas:
I took a look at the patch you committed but it seems to be incorrect: see https://github.com/pypa/setuptools/commit/925dd35ec77d4cfd79c899f34255e3a753f33ca5#commitcomment-23081646

jaraco:
Should be fixed in 2cd7f22eb62b413d65fe4208d66f7e2738f71a56. Release imminent.

jaraco:
I'm trying to figure out now what's going on with https://travis-ci.org/pypa/setuptools/jobs/253323051.

jaraco:
Ugh. So exists_ok had some extra complication with it before Python 3.3.6. From [the docs](https://docs.python.org/3.3/library/os.html#os.makedirs):

> Changed in version 3.3.6: Before Python 3.3.6, if exist_ok was True and the directory existed, makedirs() would still raise an error if mode did not match the mode of the existing directory. Since this behavior was impossible to implement safely, it was removed in Python 3.3.6. See issue 21082.

Seems it [also affects Python 3.2 prior to 3.2.5](https://hg.python.org/cpython/rev/9186f4a18584) and [Python 3.4 prior to 3.4.1](https://docs.python.org/3.4/library/os.html#os.makedirs).

jaraco:
v36.1.1 going out now.

dirk-thomas:
Great, thank you.

PR #1085, opened 2017-07-14
fix possible error when finalizing `install_requires`

benoit-pierre:
I assumed `install_requires` would always be defined, I guess I was wrong... One of my AppVeyor builds [failed](https://ci.appveyor.com/project/benoit-pierre/setuptools/build/1.0.32/job/hpwal39jn8f5e23q#L65), and this PR builds will fail too, now that 36.2.0 has been released.

PR #1088, opened 2017-07-14
Fix extras require handling

### LINKED ISSUE #1087: 'extras_require' requirements cannot include environment markers
I just upgraded to the latest setuptools by accident, and now get following error:

```
 'extras_require' requirements cannot include environment markers, in 'notebook': 'enum34==1.1.6; python_version < "3.4"'
```

Portion of my `extras_notebook_requirement` file:

```
enum34==1.1.6; python_version < '3.4'
```

This is probably related to the #1081.

benoit-pierre:
Rework handling to not error out but automatically convert internally:
```python
extras_require={ 
   'notebook': ["enum34==1.1.6; python_version < '3.4'"]
}
```
to the equivalent of
```python
extras_require={ 
   'notebook: python_version < "3.4"': ['enum34==1.1.6']
}
```
(a syntax I did not know was possible as it's not documented). Fix #1087.

Note:
* include the fix from #1085 
* tested with a patched tox env so the tests can run:
```diff
diff --git a/tests/requirements.txt b/tests/requirements.txt
index 6e2e78e2..2a8e6767 100644
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -2,3 +2,4 @@ pytest-flake8
 pytest>=3.0.2
 # pinned to 1.2 as temporary workaround for #1038
 backports.unittest_mock>=1.2,<1.3
+setuptools==35.0.0
```

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -369,11 +369,12 @@ class Distribution(Distribution_parse_config_files, _Distribution):
         Move requirements in `install_requires` that
         are using environment markers to `extras_require`.
         """
-        if not self.install_requires:
+        if not getattr(self, 'install_requires', None):
             return
+        extras_require = getattr(self, 'extras_require', None)
         extras_require = defaultdict(list, (
             (k, list(pkg_resources.parse_requirements(v)))
-            for k, v in (self.extras_require or {}).items()
+            for k, v in (extras_require or {}).items()
         ))
         install_requires = []
         for r in pkg_resources.parse_requirements(self.install_requires):
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -9,7 +9,6 @@ import distutils.core
 import distutils.cmd
 import distutils.dist
 import itertools
-import operator
 from collections import defaultdict
 from distutils.errors import (
     DistutilsOptionError, DistutilsPlatformError, DistutilsSetupError,
@@ -17,7 +16,7 @@ from distutils.errors import (
 from distutils.util import rfc822_escape
 
 from setuptools.extern import six
-from setuptools.extern.six.moves import map, filter
+from setuptools.extern.six.moves import map
 from pkg_resources.extern import packaging
 
 from setuptools.depends import Require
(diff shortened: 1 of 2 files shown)


benoit-pierre:
As a follow up, it seems like requirements in `install_requires` that are using extras should be moved to `extras_require` too:
```bash
> cat setup.py
import sys
from setuptools import setup
setup(
    name='foo',
    install_requires=[
        'bar',
        'testing; "invalid" in sys_platform',
        'barbazquux [test]',
    ],
    extras_require={
        'notebook': ["enum34==1.1.6; python_version < '3.4'"],
    },
)
>  python setup.py -q egg_info && cat foo.egg-info/requires.txt
bar
[:"invalid" in sys_platform]
testing
[notebook:python_version < "3.4"]
enum34==1.1.6
[test]
barbazquux
```

```diff
 setuptools/dist.py | 12 +++++++++---
 1 file changed, 9 insertions(+), 3 deletions(-)

diff --git i/setuptools/dist.py w/setuptools/dist.py
index da08d613..a5b94601 100644
--- i/setuptools/dist.py
+++ w/setuptools/dist.py
@@ -356,7 +356,7 @@ class Distribution(Distribution_parse_config_files, _Distribution):
         Fix environment markers in `install_requires` and `extras_require`.
 
         - move requirements in `install_requires` that are using environment
-          markers to `extras_require`.
+          markers or extras to `extras_require`.
         - convert requirements in `extras_require` of the form
           `"extra": ["barbazquux; {marker}"]` to
           `"extra:{marker}": ["barbazquux"]`.
@@ -377,11 +377,17 @@ class Distribution(Distribution_parse_config_files, _Distribution):
             getattr(self, 'install_requires', None) or ()
         ):
             marker = r.marker
-            if not marker:
+            extras = r.extras
+            if not marker and not extras:
                 install_requires.append(r)
                 continue
+            r.extras = ()
             r.marker = None
-            extras_require[':' + str(marker)].append(r)
+            for e in extras or ('',):
+                section = e
+                if marker:
+                    section += ':' + str(marker)
+                extras_require[section].append(r)
         self.extras_require = dict(
             (k, [str(r) for r in v])
             for k, v in extras_require.items()
```

setuptools/dist.py
```diff
@@ -133,7 +132,12 @@ def check_nsp(dist, attr, value):
 def check_extras(dist, attr, value):
     """Verify that extras_require mapping is valid"""
     try:
-        list(itertools.starmap(_check_extra, value.items()))
+        for k, v in value.items():
```
jaraco: I would keep the call to _check_extra, but just omit the latter portion.

benoit-pierre:
The tests fail because setuptools 36.2.0 is being installed by tox, and then using `setup.py develop` fail due to #1085. That's why I pinned setuptools to an earlier version locally to pass the tests.

alex:
Thanks for implementing this fix.

benoit-pierre:
Closing in favor of #1089.

PR #1091, opened 2017-07-17
update Travis builds

benoit-pierre:
Update to the [future defaults](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming):
* `dist: trusty`
* `sudo: required`

Note:
* include #1090 
* the version of PyPy now being tested is `pypy-5.4`: `PyPy 5.4.0 [Python 2.7.10]` (previously: `PyPy 2.5.0 [Python 2.7.8]`)
* `pypy3-5.4` is also available, should it be added to the matrix?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: false
 language: python
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jobs:
   - python: &latest_py3 3.6
   - python: nightly
   - python: pypy
+  - python: pypy3
   - python: *latest_py3
     env: LANG=C
   - python: *latest_py2


benoit-pierre:
Dropped #1090, now that the CI builds are working again.

.travis.yml
```diff
@@ -1,4 +1,5 @@
-sudo: false
+dist: trusty
+sudo: required
```
jaraco: I think this is wrong. Sudo wasn't required before and won't be the default with Trusty. I don't think we need to change this setting.

.travis.yml
```diff
@@ -1,4 +1,5 @@
-sudo: false
+dist: trusty
```
jaraco: Maybe it makes more sense to move to Xenial?

jaraco:
I'm definitely in favor of testing against late releases of everything, so including PyPy 5.4 and PyPy3 5.4 sounds good to me.

.travis.yml
benoit-pierre: OK, after reading the blog post again, `sudo: required` will become the default in September, but only for `precise`, so I'll revert this part.

.travis.yml
benoit-pierre: Unfortunately, there's no support for Xenial. Trusty is the most recent supported version.

benoit-pierre:
mmm, I'm pretty sure `pypy3-5.4` was working when I tested, but now the job is failing... Unfortunately, Travis' documentation does not detail what versions of PyPy are supported.

After a few tests, those appear to work:
* `pypy-5.4`: `PyPy 5.4.0 [Python 2.7.10]`
* `pypy-5.7.1`: `PyPy 5.7.1 [Python 2.7.13]`
* `pypy2.7-5.8.0`: `PyPy 5.8.0 [Python 2.7.13]`
* `pypy3.5-5.8.0`: `PyPy 5.8.0-beta0 [Python 3.5.3]`

So which combination shall be used?

benoit-pierre:
Using `pypy` and `pypy3` is [supported again](https://github.com/travis-ci/travis-build/pull/1088)! So I switched to that.

jaraco:
Any idea if the failure in Py26 is due to the `dist: trusty` or if something else has changed in the ecosystem?

I'm really ready to drop support for Python 2.6, but I hate to conflate that change with this one.

benoit-pierre:
It's because of a change related to pytest-flake, looking into it.

benoit-pierre:
It's the switch from `flake8` version 3.3.0 to 3.4.1.

webknjaz:
@benoit-pierre please also consider moving normal `python: py_version` instructions back to root-level matrix. I think it would look cleaner this way. Unless you're going to add some linters stage, which would run before the main test suite.

PR #1107, opened 2017-07-26
Add an easy way to build simple manylinux1 compatible extensions.

mbachry:
Current recommended way of building manylinux1 wheels is using Docker. I find it a bit cumbersome for simple C extensions, so I'd like to propose using this nice solution for glibc version pinning from AppImage project: https://github.com/AppImage/AppImageKit/tree/master/LibcWrapGenerator.

I tested it with my small, obscure php2json package (https://github.com/mbachry/php2json). I built a wheel under Fedora 26 and successfully installed it on Centos 6. It also passed `auditwheel show`.

I think a simpler and faster way of building manylinux1 packages may help pep-513 adoption a lot. The pull request is not necessarily immediately mergeable (maybe setuptools repo is a wrong place for that?), but I hope you like the idea and the code can be developed into something more production ready.

@rmcgibbo @njsmith

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -1640,6 +1640,17 @@ installed, and that your source releases will be similarly usable with or
 without Pyrex.
 
 
+Distributing simple manylinux1 extensions (Linux only)
+------------------------------------------------------
+
+If your C extension module doesn't link to any external libraries
+beyond basic set of glibc libraries (like ``libc.so`` or
+``libpthread.so``), you can pass ``glibc_compat=True`` to
+``setuptools.Extension`` class to compile the extension in Centos 5
+compatibility module.  This makes your wheel file compatible with
+manylinux1 standard without resorting to cumbersome Docker build.
+
+
 -----------------
 Command Reference
 -----------------
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ with io.open(readme_path, encoding='utf-8') as readme_file:
     long_description = readme_file.read()
 
 package_data = dict(
-    setuptools=['script (dev).tmpl', 'script.tmpl', 'site-patch.py'],
+    setuptools=['script (dev).tmpl', 'script.tmpl', 'site-patch.py', 'glibc-compat.h'],
 )
 
 force_windows_specific_files = (
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -2,6 +2,8 @@ import os
 import sys
 import itertools
 import imp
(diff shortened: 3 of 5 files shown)


njsmith:
Huh, it's neat that worked! I did not know about this trick.

I'm surprised – I think you must have gotten lucky and your GCC runtime library (libgcc_s) must have stayed the same between centos5 and the version of GCC you used? That does suggest that libgcc changes very rarely and is unlikely to change soon, but... this is a pretty big risk. The next time libgcc does change then AFAICT this approach suddenly becomes permanently unviable, so I'd recommend that setuptools *not* commit to shipping this by default – setuptools shouldn't put its fate in the hands of the GCC developers like that.

But, you could still probably refactor it into a setuptools plugin and make it a separate package. Personally, I prefer the docker approach: with this you still have to install the python versions you want to build with, install auditwheel, run auditwheel ... so even if you have a rare library that's pure C with no external dependencies, it's still not *that* much simpler than using docker, and the limitations mean that you'll likely end up needing to switch to docker eventually anyway.

You're absolutely right that building manylinux wheels is too annoying, though, and I applaud your trying to do something about that!

If you want to continue with this, then a few comments:

* There will be a manylinux2 in the future, and probably more after it. So instead of `glibc_compat=True` you should probably use something like `glibc_compat="manylinux1"

* You wrap the c++ compiler, but this trick can't work for c++ – libstdc++ has definitely changed since centos5, just like glibc has. It would be better to error out if someone tries to compile c++.

* If I were you I'd include more detailed docs on how that header was generated, in case it needs to be done again later...

Alternatively, if you want to consider other ways to make manylinux wheels more approachable, you might check with the wheel-builders mailing list if they have any ideas. One that occurred to me was a script like `manylinux-builder` that you could pip install, pass a source tree, and it would automatically pull the docker image, build against specified python versions, run auditwheel, etc.

And in any case, thanks for working on this!

anthrotype:
> a script like manylinux-builder that you could pip install, pass a source tree, and it would automatically pull the docker image, build...

Sounds to me like https://github.com/joerick/cibuildwheel

jaraco:
I appreciate the effort on this, and thanks to njsmith for the detailed review. He has raised what look like legitimate concerns that have not yet been addressed. Also, it seems like what anthrotype has linked might obviate this effort. I'm inclined to keep the code more simple and let the modularity of plugins or the like fill this need.

If someone can make a compelling reason why having this functionality in setuptools would provide substantial value over cibuildwheel or a third-party and mitigating the risks raised by njsmith, then we can reconsider.

PR #1108, opened 2017-07-27
fix requires handling when using setup.cfg

### LINKED ISSUE #1105: requirements declared in `setup.cfg` and using environment makers are not correctly handled
```bash
> >setup.py <<\EOF
from setuptools import setup
setup(name='foo', install_requires='barbazquux; python_version < "3.4"')
EOF
> >setup.cfg <<\EOF
EOF
> python setup.py -q egg_info && cat foo.egg-info/requires.txt

[:python_version < "3.4"]
barbazquux
> >setup.py <<\EOF
from setuptools import setup
setup(name='foo')
EOF
> >setup.cfg <<\EOF
[options]
install_requires =
        barbazquux; python_version < "3.4"
EOF
> python setup.py -q egg_info && cat foo.egg-info/requires.txt
barbazquux; python_version < "3.4"
```
This is because parsing config files (like `setup.cfg`) is done after a `Distribution` initialization. Other issue stemming from this: #1054.

benoit-pierre:
Fix #1105.

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -349,14 +349,15 @@ class Distribution(Distribution_parse_config_files, _Distribution):
                     "setuptools, pip, and PyPI. Please see PEP 440 for more "
                     "details." % self.metadata.version
                 )
-        if getattr(self, 'python_requires', None):
-            self.metadata.python_requires = self.python_requires
         self._finalize_requires()
 
     def _finalize_requires(self):
         """
-        Fix environment markers in `install_requires` and `extras_require`.
+        Set `metadata.python_requires` and fix environment markers
+        in `install_requires` and `extras_require`.
         """
+        if getattr(self, 'python_requires', None):
+            self.metadata.python_requires = self.python_requires
         self._convert_extras_requirements()
         self._move_install_requirements_markers()
 
@@ -424,8 +425,7 @@ class Distribution(Distribution_parse_config_files, _Distribution):
         _Distribution.parse_config_files(self, filenames=filenames)
 
         parse_configuration(self, self.command_options)
-        if getattr(self, 'python_requires', None):
-            self.metadata.python_requires = self.python_requires
+        self._finalize_requires()
 
     def parse_command_line(self):
         """Process features after parsing command line options"""
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -333,7 +333,7 @@ class TestOptions:
             ])
             assert dist.install_requires == ([
                 'docutils>=0.3',
-                'pack ==1.1, ==1.3',
+                'pack==1.1,==1.3',
                 'hey'
             ])
             assert dist.setup_requires == ([
@@ -403,7 +403,7 @@ class TestOptions:
             ])
             assert dist.install_requires == ([
                 'docutils>=0.3',
-                'pack ==1.1, ==1.3',
+                'pack==1.1,==1.3',
                 'hey'
             ])
             assert dist.setup_requires == ([
@@ -508,7 +508,7 @@ class TestOptions:
         with get_dist(tmpdir) as dist:
             assert dist.extras_require == {
                 'pdf': ['ReportLab>=1.2', 'RXP'],
-                'rest': ['docutils>=0.3', 'pack ==1.1, ==1.3']
+                'rest': ['docutils>=0.3', 'pack==1.1,==1.3']
             }
 
     def test_entry_points(self, tmpdir):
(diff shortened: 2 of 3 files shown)


suutari-ai:
Any comments on this? I'd love to get this merged.

jaraco:
Wow, quite a bit of work went into the tests refactor. I can see why you went there - wanting to extract the commonality of all the tests.

I'm pretty sure pytest_generate_tests has been superseded by parametrized fixtures, but that can be transformed later if needed.

The main concern I have about the refactoring is what does it do for the output? Is it still possible, for example, to select just one test to run? Is it still clear which test is being performed?

I'm also slightly concerned about the `xfail` syntax, which I presume was an unexpected introduction late. It's concerning in part because it's part of its own language, so it's not possible, for example to provide more detail (i.e. `@pytest.mark.xfail(reason="...")`)... without expanding the DSL.

I do like that the code seems to be more consolidated with this change. It does still feel a little uneasy that `_finalize_requires` gets called twice, but if that's what needs to happen, that's probably better than what was being done before.

If you're comfortable with the change being stable and addressed any concerns above to your satisfaction, would you add a CHANGELOG entry and then cut a release (release docs are in the repo)?

benoit-pierre:
> I'm pretty sure pytest_generate_tests has been superseded by parametrized fixtures, but that can be transformed later if needed.

This seemed to be the simplest way to parametrize those tests while still keeping the code in-class. But I agree it could become unwieldily if multiple methods are parametrized using the same mechanism. Would [this](https://github.com/benoit-pierre/setuptools/blob/fix_requires_handling%2C_again_fixup/setuptools/tests/test_egg_info.py#L190) be better?

> The main concern I have about the refactoring is what does it do for the output? Is it still possible, for example, to select just one test to run? Is it still clear which test is being performed?

Yes, that where the `ids` parameter come into play:
```bash
> python -m pytest --collect-only -qk test_requires
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_marker]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_marker_in_setup_cfg]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_in_setup_cfg]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_and_marker]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_and_marker_in_setup_cfg]

=========================== short test summary info ============================
SKIP [2] /home/bpierre/progs/src/setuptools/setuptools/tests/test_msvc.py:17: could not import 'distutils.msvc9compiler'
============================= 293 tests deselected =============================
2 skipped, 293 deselected in 0.65 seconds

> python -m pytest --collect-only -qk install_requires
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_marker]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_marker_in_setup_cfg]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_in_setup_cfg]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_and_marker]
setuptools/tests/test_egg_info.py::TestEggInfo::test_requires[install_requires_with_extra_and_marker_in_setup_cfg]

=========================== short test summary info ============================
SKIP [2] /home/bpierre/progs/src/setuptools/setuptools/tests/test_msvc.py:17: could not import 'distutils.msvc9compiler'
============================= 293 tests deselected =============================
2 skipped, 293 deselected in 0.64 seconds
```

> I'm also slightly concerned about the `xfail` syntax, which I presume was an unexpected introduction late. It's concerning in part because it's part of its own language, so it's not possible, for example to provide more detail (i.e. `@pytest.mark.xfail(reason="...")`)... without expanding the DSL.

Yeah, I could not find a way to mark a test by name.

benoit-pierre:
OK, I went ahead with the switch to `pytest.mark.parametrize` as it does make the code a little bit simpler.

suutari-ai:
Just tested this fix in setuptools 36.2.7 and it seems to work: I can now generate a wheel for Prequ with correct environment markers (with out-of-the-box setuptools and wheel packages). Thanks a lot! :smile:

PR #1112, opened 2017-07-29
tests: disable pytest-flake8 on Python 2.6

benoit-pierre:
Python 2.6 support has been official dropped from flake8 for some time, and broken since 3.4.0.


benoit-pierre:
By the way, why is pytest-flake8 even installed? It's not used by the CI builds.

jaraco:
> why is pytest-flake8 even installed

I don't know.

PR #1116, opened 2017-07-31
Better detect unpacked eggs

### LINKED ISSUE #462: pkg_resources.find_on_path() thinks the zc.recipe.egg wheel is an egg
Originally reported by: **mgedmin (Bitbucket: [mgedmin](http://bitbucket.org/mgedmin), GitHub: [mgedmin](http://github.com/mgedmin))**

---

Steps to reproduce: try to `pip install zc.recipe.egg`.  You'll get an error:

```
zc.recipe.egg is in an unsupported or invalid wheel
```

If you hack pip's sources to expose the real traceback, you'll see [this](https://github.com/pypa/pip/issues/3028#issuecomment-148614984):

```
  File "/home/qwcode/p/pypa/pip/pip/basecommand.py", line 200, in main                                                                                                                                                                       
  File "/home/qwcode/p/pypa/pip/pip/commands/install.py", line 302, in run                                                                                                                                                                   
    root=options.root_path,                                                                                                                                                                                                                  
  File "/home/qwcode/p/pypa/pip/pip/req/req_set.py", line 657, in install                                                                                                                                                                    
    **kwargs
  File "/home/qwcode/p/pypa/pip/pip/req/req_install.py", line 805, in install
    version = pip.wheel.wheel_version(self.source_dir)
  File "/home/qwcode/p/pypa/pip/pip/wheel.py", line 559, in wheel_version
    wheel_data = dist.get_metadata('WHEEL')
  File "/home/qwcode/p/pypa/pip/pip/_vendor/pkg_resources/__init__.py", line 1611, in get_metadata
    return self._get(self._fn(self.egg_info, name))
  File "/home/qwcode/p/pypa/pip/pip/_vendor/pkg_resources/__init__.py", line 1722, in _get
    with open(path, 'rb') as stream:
IOError: [Errno 2] No such file or directory: '/tmp/pip-build-ucnzjI/zc.recipe.egg/EGG-INFO/WHEEL'
```

AFAIU the problem is the following check in pkg_resources.find_on_path():

``` python
    if os.path.isdir(path_item) and os.access(path_item, os.R_OK):
        if path_item.lower().endswith('.egg'):
            # unpacked egg
```

This is wrong: the unpacked zc.recipe.egg _wheel_ has a top-level directory called `zc.recipe.egg`.

See https://github.com/pypa/pip/issues/3028 for more background.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/462

leorochael:
Do not assume a directory named in `.egg` is an egg, unless it has an
actual egg metadata directory.

Closes #462

Should also close pypa/pip#3028

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1553,7 +1553,7 @@ class EggProvider(NullProvider):
         path = self.module_path
         old = None
         while path != old:
-            if _is_unpacked_egg(path):
+            if _is_egg_path(path):
                 self.egg_name = os.path.basename(path)
                 self.egg_info = os.path.join(path, 'EGG-INFO')
                 self.egg_root = path
@@ -1956,7 +1956,7 @@ def find_eggs_in_zip(importer, path_item, only=False):
         # don't yield nested distros
         return
     for subitem in metadata.resource_listdir('/'):
-        if _is_unpacked_egg(subitem):
+        if _is_egg_path(subitem):
             subpath = os.path.join(path_item, subitem)
             for dist in find_eggs_in_zip(zipimport.zipimporter(subpath), subpath):
                 yield dist
@@ -2033,7 +2033,7 @@ def find_on_path(importer, path_item, only=False):
                     yield Distribution.from_location(
                         path_item, entry, metadata, precedence=DEVELOP_DIST
                     )
-                elif not only and _is_unpacked_egg(entry):
+                elif not only and _is_egg_path(entry):
                     dists = find_distributions(os.path.join(path_item, entry))
                     for dist in dists:
                         yield dist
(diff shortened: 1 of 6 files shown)


jaraco:
Looks good to me. It's unclear from the changeset which calls are affected by the change (as the lines that were tweaked will be unaffected by the change).

More importantly, should this change have a test? It seems unlikely that it would be optimized away, but having a test would ensure that the use-case is preserved.

Also, could you add an entry to the changelog? 0.0.1 bump is good, as this seems like a bugfix.

leorochael:
> Looks good to me. It's unclear from the changeset which calls are affected by the change (as the lines that were tweaked will be unaffected by the change).

The only call that is actually affected is the first branch of `find_on_path()`, which is the only place that still calls the old function name `_is_unpacked_egg()` (which now has a more correct implementation).

I'll try to come up with some tests later and a changelog.

pkg_resources/__init__.py
```diff
@@ -2221,12 +2221,21 @@ def _normalize_cached(filename, _cache={}):
         return result
 
 
+def _is_egg_path(path):
```
gst: Could eventually be a one liner ?

leorochael: > Could eventually be a one liner ?

It's just a renaming of the `_is_unpacked_egg()'. Yes it's a one-liner, but I think it saves enough typing (specially in an editor with completion), that I think it's worth it.

leorochael:
> Few test cases should be straightforward?

@gst, I spent **way** more time on the test cases than the fix. Sure writing the test case of the primary motivation was easy enough, but there weren't any other existing test cases for the current behavior of the function, so had to minimally cover the other uses as well to be sure I wasn't causing a regression.

@jaraco, added changelog and tests, and bumped version. Should be ready for merging.

jaraco:
Big thanks @leorochael. Often it is the case that writing tests is more onerous than the fix. The sacrifice will be appreciated by our future selves.

PR #1119, opened 2017-08-02
Fix exception on mingw built Python 2

### LINKED ISSUE #1118: DistutilsPlatformError in mingw built Python 2 (msys2)
Trying to do many things using setuptools in a mingw built Python 2 (For example `msys2`'s `mingw-w64-x86_64-python2` fails with this exception:
```pytb
$ python setup.py build_ext
  File "setup.py", line 1, in <module>
    from setuptools import setup, Extension
  File "mingw64/lib/python2.7/site-packages/setuptools/__init__.py", line 160, in <module>
    monkey.patch_all()
  File "mingw64/lib/python2.7/site-packages/setuptools/monkey.py", line 104, in patch_all
    patch_for_msvc_specialized_compiler()
  File "mingw64/lib/python2.7/site-packages/setuptools/monkey.py", line 156, in patch_for_msvc_specialized_compiler
    msvc = import_module('setuptools.msvc')
  File "mingw64/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "mingw64/lib/python2.7/site-packages/setuptools/msvc.py", line 49, in <module>
    from distutils.msvc9compiler import Reg
  File "mingw64/lib/python2.7/distutils/msvc9compiler.py", line 306, in <module>
    raise DistutilsPlatformError("VC %0.1f is not supported by this module" % VERSION)
distutils.errors.DistutilsPlatformError: VC 6.0 is not supported by this module
```

`distutils.msvc9compiler` doesn't like being imported on mingw built Python, and this hack isn't necessary there.

P.S. I'm going to send a PR.

segevfiner:
The exception is caused by the VCForPython27 monkey patch. It's not needed on mingw built Python. Disable it and avoid the import by using the same function that `distutils` uses to decide it's running on a mingw built Python.

Hopefully this doesn't break in some random old Python version. 😩

Fixes #1118

### DIFF
--- a/setuptools/monkey.py
+++ b/setuptools/monkey.py
@@ -4,6 +4,7 @@ Monkey patching of distutils.
 
 import sys
 import distutils.filelist
+from distutils.util import get_platform
 import platform
 import types
 import functools
@@ -152,12 +153,12 @@ def patch_for_msvc_specialized_compiler():
     Patch functions in distutils to use standalone Microsoft Visual C++
     compilers.
     """
-    # import late to avoid circular imports on Python < 3.5
-    msvc = import_module('setuptools.msvc')
-
-    if platform.system() != 'Windows':
+    if not get_platform().startswith('win'):
         # Compilers only availables on Microsoft Windows
         return
+
+    # import late to avoid circular imports on Python < 3.5
+    msvc = import_module('setuptools.msvc')
 
     def patch_params(mod_name, func_name):
         """
--- a/setuptools/monkey.py
+++ b/setuptools/monkey.py
@@ -4,7 +4,6 @@ Monkey patching of distutils.
 
 import sys
 import distutils.filelist
-from distutils.util import get_platform
 import platform
 import types
 import functools
@@ -153,12 +152,12 @@ def patch_for_msvc_specialized_compiler():
     Patch functions in distutils to use standalone Microsoft Visual C++
     compilers.
     """
-    if not get_platform().startswith('win'):
-        # Compilers only availables on Microsoft Windows
-        return
-
     # import late to avoid circular imports on Python < 3.5
     msvc = import_module('setuptools.msvc')
+
+    if platform.system() != 'Windows':
+        # Compilers only availables on Microsoft Windows
+        return
 
     def patch_params(mod_name, func_name):
         """
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -47,7 +47,7 @@ else:
 
 try:
     from distutils.msvc9compiler import Reg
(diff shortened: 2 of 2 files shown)


setuptools/monkey.py
```diff
@@ -152,13 +153,13 @@ def patch_for_msvc_specialized_compiler():
     Patch functions in distutils to use standalone Microsoft Visual C++
     compilers.
     """
-    # import late to avoid circular imports on Python < 3.5
-    msvc = import_module('setuptools.msvc')
-
-    if platform.system() != 'Windows':
+    if not get_platform().startswith('win'):
```
jaraco: This fix doesn't feel right to me. It's using a non-standard method to determine the platform. Why is `platform.system()` not suitable? If using a `startswith` comparator, I think a separate function should be supplied with a docstring explaining what's being checked here.

setuptools/monkey.py
```diff
@@ -152,13 +153,13 @@ def patch_for_msvc_specialized_compiler():
...
         # Compilers only availables on Microsoft Windows
         return
 
+    # import late to avoid circular imports on Python < 3.5
```
jaraco: Having this import appear before the Windows check means that when running tests on non-Windows, this module is still importable. Ideally, this import would appear at the top of the file except for the circular import issue (and will move to the top when Python 3.5 is the minimum version supported).

There is already a protection in `setuptools.msvc:48` to suppress errors importing msvc9compiler, but it's currently only catching an ImportError. Perhaps it should also catch the reported DistutilsPlatformError.

Presumably, this same issue would occur if someone were to install MSVC 6.0 or 7.0 (if that's even still possible).

setuptools/monkey.py
segevfiner: `platform.system()` just returns Windows on mingw built Python. It's this hack: [Lib/distutils/util.py:42-54](https://github.com/python/cpython/blob/master/Lib/distutils/util.py#L42-L54) which detects it. It seems that running the monkey patch doesn't raise though, so I removed this code.

setuptools/monkey.py
segevfiner: The issue will only occur if `msvc9compiler` thinks that Python was built with VC 6.0 or similar, which it defaults to when Python is built with mingw since this function [Lib/distutils/msvc9compiler.py:169-192](https://github.com/python/cpython/blob/778928b0c7aa438c282727535814d73df850693a/Lib/distutils/msvc9compiler.py#L169-L192) won't find the MSVC version in the `sys.version`. (Yes it's *"parsing"* `sys.version`... 🙄)

PR #1127, opened 2017-08-05
workaround easy_install bug

### LINKED ISSUE #196: tests_require=['pytest', 'pytest-cov'] breaks setuptools.setup
Originally reported by: **slashfoo (Bitbucket: [slashfoo](http://bitbucket.org/slashfoo), GitHub: [slashfoo](http://github.com/slashfoo))**

---

```
(test)[20:24:14] [jamiel@moeka:masterskel/] $ python setup.py coverage
This command (coverage) requires the installation of pytest-cov
running coverage
Searching for pytest-cov
Reading https://pypi.python.org/simple/pytest-cov/
Best match: pytest-cov 1.6
Downloading https://pypi.python.org/packages/source/p/pytest-cov/pytest-cov-1.6.tar.gz#md5=6da54d74bde9d200de45068ba2ea637a
Processing pytest-cov-1.6.tar.gz
Writing /tmp/easy_install-xvMxzZ/pytest-cov-1.6/setup.cfg
Running pytest-cov-1.6/setup.py -q bdist_egg --dist-dir /tmp/easy_install-xvMxzZ/pytest-cov-1.6/egg-dist-tmp-BGJIES

Installed /home/jamiel/dev/masterskel/pytest_cov-1.6-py2.7.egg
Searching for pytest
Best match: pytest cov-1.6
Downloading https://pypi.python.org/packages/source/p/pytest-cov/pytest-cov-1.6.tar.gz#md5=6da54d74bde9d200de45068ba2ea637a
Processing pytest-cov-1.6.tar.gz
Writing /tmp/easy_install-EweVbo/pytest-cov-1.6/setup.cfg
Running pytest-cov-1.6/setup.py -q bdist_egg --dist-dir /tmp/easy_install-EweVbo/pytest-cov-1.6/egg-dist-tmp-zxCrHa

Installed /home/jamiel/dev/masterskel/pytest_cov-1.6-py2.7.egg
  File "setup.py", line 203, in <module>
    zip_safe=False,
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/jamiel/.virtualenvs/test/local/lib/python2.7/site-packages/setuptools/command/test.py", line 130, in run
    self.distribution.fetch_build_eggs(self.distribution.tests_require)
  File "/home/jamiel/.virtualenvs/test/local/lib/python2.7/site-packages/setuptools/dist.py", line 264, in fetch_build_eggs
    replace_conflicting=True
  File "/home/jamiel/.virtualenvs/test/local/lib/python2.7/site-packages/pkg_resources.py", line 588, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: pytest
```

since 'pytest-cov' depends on 'pytest', making `tests_require=['pytest-cov']` fixes the issue, same if it is `tests_require=['pytest-cov', 'pytest']`.

Setting a breakpoint on `${HOME}/.virtualenvs/slashpyenv/lib/python3.4/site-packages/pkg_resources.py:620` (replace for your path to pkg_sources.py) shows:

```
dist = best[req.key] = env.best_match(req, ws, installer)
```

after running on for `req = Requirement.parse('pytest-cov')`, when I run `!env.best_match(Requirement.parse('pytest'), ws, installer)` it returns the best match to be 'pytest-cov 1.6' and exits.

_note_: In the report you see 3.4 in the outputs ect... but this is also present in 2.7.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/196

benoit-pierre:
When run in a fresh virtualenv, the following command:
```bash
easy_install pytest-runner pytest
```
will return with no error but only install `pytest-runner`.

This is because after parsing the PyPi index page for `pytest-runner`, easy_install will build a list of inferred distribution (in [interpret_distro_name](https://github.com/pypa/setuptools/blob/master/setuptools/package_index.py#L141)) that looks like this:

Project name | Version 
--- | ---
pytest | runner-2.2.1
pytest-runner | 2.2.1
pytest-runner-2.2.1 |
[...]
pytest | runner-2.11.1
pytest-runner | 2.11.1
pytest-runner-2.11.1 |

Which works fine for finding the best matching distribution for `pytest-runner`, but gets also reused for the next requirement: `pytest`....

Switching the requirements around works fine: `easy_install pytest pytest-runner`.

An easy workaround for setuptools' setup internal use is to avoid reusing the same `easy_install` command instance for multiple calls to `Distribution.fetch_build_egg`.

Fix #196.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ v36.4.0
 * #1068: Sort files and directories when building eggs for
   deterministic order.
 
+* #196: Remove caching of easy_install command in fetch_build_egg.
+  Fixes issue where ``pytest-runner-N.N`` would satisfy the installation
+  of ``pytest``.
+
 v36.3.0
 -------
 
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -485,36 +485,30 @@ class Distribution(Distribution_parse_config_files, _Distribution):
 
     def fetch_build_egg(self, req):
         """Fetch an egg needed for building"""
-
-        try:
-            cmd = self._egg_fetcher
-            cmd.package_index.to_scan = []
-        except AttributeError:
-            from setuptools.command.easy_install import easy_install
-            dist = self.__class__({'script_args': ['easy_install']})
-            dist.parse_config_files()
-            opts = dist.get_option_dict('easy_install')
-            keep = (
-                'find_links', 'site_dirs', 'index_url', 'optimize',
-                'site_dirs', 'allow_hosts'
-            )
-            for key in list(opts):
-                if key not in keep:
-                    del opts[key]  # don't use any other settings
-            if self.dependency_links:
-                links = self.dependency_links[:]
-                if 'find_links' in opts:
-                    links = opts['find_links'][1].split() + links
-                opts['find_links'] = ('setup', links)
-            install_dir = self.get_egg_cache_dir()
-            cmd = easy_install(
-                dist, args=["x"], install_dir=install_dir,
-                exclude_scripts=True,
-                always_copy=False, build_directory=None, editable=False,
-                upgrade=False, multi_version=True, no_report=True, user=False
-            )
-            cmd.ensure_finalized()
-            self._egg_fetcher = cmd
+        from setuptools.command.easy_install import easy_install
+        dist = self.__class__({'script_args': ['easy_install']})
+        dist.parse_config_files()
+        opts = dist.get_option_dict('easy_install')
+        keep = (
+            'find_links', 'site_dirs', 'index_url', 'optimize',
+            'site_dirs', 'allow_hosts'
+        )
+        for key in list(opts):
+            if key not in keep:
+                del opts[key]  # don't use any other settings
+        if self.dependency_links:
+            links = self.dependency_links[:]
+            if 'find_links' in opts:
+                links = opts['find_links'][1].split() + links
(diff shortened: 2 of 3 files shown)


jaraco:
I love it. This code fixes a long-standing bug _and_ makes the code shorter. The caching of the _egg_fetcher was written before my time, and I'm unsure if it's a premature optimization or if there were cases where it legitimately needed to be cached for efficiency. I think it's probably worth dropping that behavior simply for this fix unless it turns out to be grossly more expensive.

More importantly, I have a concern that some code (mine included) depends on the _egg_fetcher attribute being present in order to override the fetch behavior. Indeed [ptr itself relies on that attribute to supply a fetcher](https://github.com/pytest-dev/pytest-runner/blob/228cbcb5672579f21601707ba0b23169fb4f423f/ptr.py#L188). Yes, I know... it was bad practice to rely on a private attribute to inject behavior. I held my nose at the time and now it's come to bite me.

I don't think we can safely accept this change without first addressing the need to more formally support the needs of pytest runner for customizing the fetcher, or perhaps pytest-runner could drop support for whatever drove that need.

PR #1133, opened 2017-08-21
Improve README file list handling and add Markdown to the current list

sgaist:
Markdown is a widely used format to write README files and documentation. 

This patch aims to simplify adding new formats and at the same time adds that one to the list.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v36.3.1
+
+* Improved handling of README files extensions and added 
+  Markdown to the list of searched READMES. 
+
 v36.3.0
 -------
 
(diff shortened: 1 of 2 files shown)


jaraco:
I think this project has rejected this change before, probably because adding this change gives the impression that .md is supported, but there are several aspects of the ecosystem in which .md is poorly supported. Searching for it, though, I can find no references. That presumption is changing, however (as you can see in other PRs that are pending/WIP with setuptools and others).

I'd prefer to get away from special-casing the README, but as long as that's what distutils does, we should probably extend the behavior to match users' expectations.

Can you add an entry to CHANGES.txt to inform users of how their expectations should change with this change?

sgaist:
Sure, I'll modify it.

Do you think that it would be possible to update distutils to improve the situation ?

PR #1134, opened 2017-08-24
pkg_resources: filter entries before parsing them

jd:
`_by_version_descending` is quite heavy as it tries to parse versioning and
sort ALL files listed. This is one of the heaviest method.

Actually, there might not be any need to parse those file names it there will
be ignored in the loop just after, so let filter them before sorting.

On my system, this brings down the loading of time of pkg_resources by 20%:

$ python -m timeit -n 5 -r 3 -c 'import pkg_resources; reload(pkg_resources)'
5 loops, best of 3: 1.11 sec per loop

$ python -m timeit -n 5 -r 3 -c 'import pkg_resources; reload(pkg_resources)'
5 loops, best of 3: 889 msec per loop

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2041,8 +2041,17 @@ def find_on_path(importer, path_item, only=False):
                or hasattr(e, "winerror") and e.winerror == 267):
                 return
             raise
+        # first filter entries that interest us before sorting
+        def of_interest(entry):
+            """Return whetever `entry` is interesting to parse as an egg"""
+            return (entry.lower().endswith('.egg-info')
+                    or entry.lower().endswith('.dist-info')
+                    or (not only and _is_egg_path(entry))
+                    or (not only and entry.lower().endswith('.egg-link')))
+
         # scan for .egg and .egg-info in directory
-        path_item_entries = _by_version_descending(entries)
+        path_item_entries = _by_version_descending(
+            filter(of_interest, entries))
         for entry in path_item_entries:
             lower = entry.lower()
             if lower.endswith('.egg-info') or lower.endswith('.dist-info'):


pkg_resources/__init__.py
```diff
@@ -2016,8 +2016,17 @@ def find_on_path(importer, path_item, only=False):
                 )
             )
         else:
+            # first filter entries that interest us before sorting
+            entries = (
+                entry
+                for entry in os.listdir(path_item)
+                if (entry.lower().endswith('.egg-info')
```
jaraco: Would it be possible to apply this test as an argument to `filter`, such that `entries = filter(of_interest, os.listdir(path_item))`? That function `of_interest` could then have a docstring describing its purpose. My only worry is that the additional n function calls might obviate the performance advantages.

jd: Considering how long is the parsing of version (50% cost of exec time IIRC), I don't think it can be worse. I'll change that.

jd:
Pull-request updated, HEAD is now cb989a2e666b5a68f00faf8905f71407792d27ad

jaraco:
Thanks for this. I was going to accept this change, but then I went to extract the of_interest function and realized that the 'only' parameter has to be included. This fact makes all of this code quite complicated. I'm not even quite sure what 'only' is for. Blaming the code, I find it's very old (c721485).

I think I want to mull over this one a bit more before rolling it in because it does add complication to the already complicated code, adds redundant operations, and it's not clear to me if the benefit outweighs the cost.

jd:
I feel you @jaraco. Run `cProfile` and you'll see how heavy the parsing is. I've 400 packages installed on my system, so the time `pkg_resources` lose trying to parse things it does not want in the end has a real cost. That being said, if you have a better way to refactor this, I'm all for it.

The branch is in conflict for now, I'll update it.

jd:
Pull-request updated, HEAD is now 4481608015fa6d8a6f7bf4c3008409f91958570a

jaraco:
@jd Please have a look at the latest commit (referenced above). I believe it accomplishes the same thing you were after.

jd:
@jaraco looks great, thanks for taking care!

PR #1135, opened 2017-08-25
pkg_resources: do not call stat() and access()

jd:
The current code in find_on_path is doing a lot of stat() calls which are
actually useless and prone to race conditions.

As described in Python documentation
(https://docs.python.org/3/library/os.html#os.access), os.access must not be
used before opening a file. Same goes for a directory.

This patch removes those checks by handling exceptions correctly when using
os.listdir() instead, which improves pkg_resources import time.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -34,6 +34,7 @@ import platform
 import collections
 import plistlib
 import email.parser
+import errno
 import tempfile
 import textwrap
 import itertools
@@ -79,6 +80,11 @@ __import__('pkg_resources.extern.packaging.markers')
 
 if (3, 0) < sys.version_info < (3, 3):
     raise RuntimeError("Python 3.3 or later is required")
+
+if six.PY2:
+    # Those builtin exceptions are only defined in Python 3
+    PermissionError = None
+    NotADirectoryError = None
 
 # declare some globals that will be defined later to
 # satisfy the linters.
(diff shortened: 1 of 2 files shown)


pkg_resources/__init__.py
```diff
@@ -2008,46 +2014,55 @@ def find_on_path(importer, path_item, only=False):
     """Yield distributions accessible on a sys.path directory"""
     path_item = _normalize_cached(path_item)
 
-    if os.path.isdir(path_item) and os.access(path_item, os.R_OK):
-        if _is_unpacked_egg(path_item):
-            yield Distribution.from_filename(
-                path_item, metadata=PathMetadata(
-                    path_item, os.path.join(path_item, 'EGG-INFO')
-                )
+    if _is_unpacked_egg(path_item):
```
jaraco: If the code heads down this path now, but path_item is not a dir or is not readable, you'll get a different result than before. That seems to be a backward-incompatible change. Is this code path not expected to be impacted by this change?

jaraco:
I like your instinct here. I'm concerned about the test failure on Windows, which does seem implicated in the change. What's your thought?

jd:
Interesting… The Windows failure is exactly why Python 3 has a `NotADirectoryError`. There seems to be no smart way to catch that in Python 2. I've updated the patch to catch that too, it's ugly, but if you know any better way let me know.

pkg_resources/__init__.py
jd: You're right, it probably needs some more work to, I'll dig into it.

jd: Actually… the code check that `path_item/EGG-INFO/PKG-INFO` is a file, so if that's the case, `path_item` has to be a directory. So the check seems useless.

The permission is not that useful either: the fact that the directory is readable is only useful when calling `os.listdir`, but no code in that path uses `os.listdir` – and if it does, it'd needs to learn to handle errors. What is interesting is to be sure the directory has `+x` permission, and that is done by using `os.path.isfile` in `_is_unpacked_egg`: it'll return `False` if it can't traverse the directory.

So… I can't see of a case where things would go wrong. If there's any we should definitely add a test!

jd:
Pull-request updated, HEAD is now df23966ac5bae1028a37af53c644b9e0839ac93b

Issue #878, opened 2016-12-09
Drop support for Python 2.6


jaraco:
As pip has dropped support for Python 2.6, setuptools should do so as well.

xavfernandez:
Erm, pip will drop 2.6 support in pip 10, i.e. in >=6 months.
pip will likely have a 9.1 release in between, with support for 2.6 ⭕️ 

cf https://github.com/pypa/pip/issues/3955

jaraco:
Oh my. Okay. I was confused. I'll shelve this change for later then.

jaraco:
I've moved the commits to the [drop-py26 branch](https://github.com/pypa/setuptools/tree/drop-py26), to be merged at an appropriate time.

ScoreUnder:
What was the last version with Python 2.6 support?

PR #1140, opened 2017-09-02
Added the namespace_packages argument to find_packages()

agronholm:
This allows find_packages() to work with PEP 420 namespace packages without considering every directory a namespace package.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -427,8 +427,11 @@ such projects also need something like ``package_dir={'':'src'}`` in their
 ``setup()`` arguments, but that's just a normal distutils thing.)
 
 Anyway, ``find_packages()`` walks the target directory, filtering by inclusion
-patterns, and finds Python packages (any directory). On Python 3.2 and
-earlier, packages are only recognized if they include an ``__init__.py`` file.
+patterns, and finds Python packages. By default, it only recognizes directories
+containing an ``__init__.py`` file as valid packages. To enable it to find
+new-style (:pep:`420`) namespace packages, a comprehensive list of such
+packages must be explicitly passed to it, like
+``find_packages(namespace_packages=['foo', 'foo.bar'])``.
 Finally, exclusion patterns are applied to remove matching packages.
 
 Inclusion and exclusion patterns are package names, optionally including
@@ -2281,7 +2284,7 @@ Configuring setup() using setup.cfg files
 
 .. note:: New in 30.3.0 (8 Dec 2016).
 
-.. important:: ``setup.py`` with ``setup()`` function call is still required even 
+.. important:: ``setup.py`` with ``setup()`` function call is still required even
                 if your configuration resides in ``setup.cfg``.
 
 ``Setuptools`` allows using configuration files (usually `setup.cfg`)
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -31,7 +31,8 @@ run_2to3_on_doctests = True
 lib2to3_fixer_packages = ['lib2to3.fixes']
 
 
-def find_packages(where='.', exclude=(), include=('*',)):
+def find_packages(where='.', exclude=(), include=('*',),
+                  namespace_packages=()):
     """Return a list all Python packages found within directory 'where'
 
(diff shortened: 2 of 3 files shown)


agronholm:
Only after submitting this, I learned about `find:`. I would really like to make this work with the `namespace_packages` argument in setup(), but it requires it to be parsed before `packages`. For this to happen, the parsers in `ConfigOptionsHandler` need to be processed in a deterministic order, for which `OrderedDict` is required. This would not be a problem if setuptools wasn't still trying to support Python 2.6. Would it be possible to drop this support, as it has been EOL'd quite some time ago?

estan:
Looks similar to the #915 I submitted back in February, but maybe a little more elegant.

My PR was rejected by @jaraco back then, since he felt discussion was active on #97. That issue has been silent since then though, so I guess perhaps it's time to just accept the state of things and go ahead with a `namespace_packages=` solution. What do you think @jaraco?

estan:
Personally, I'd say that's a better solution than having a growing number of people adding their own ad-hoc `find_packages` alongside their `setup.py`.

estan:
@agronholm What does `find:` in your comment mean (I've been out of the loop for a while).

agronholm:
It means that you can now do `packages='find:` which is equivalent to `packages=find_packages()`, but you can use that when specifying packages under `[metadata]` in your setup.cfg. The obvious advantage is that you could reuse the (otherwise useless on py3) `namespace_packages` option with that one.

estan:
@agronholm Ah, I see.

jaraco:
> Would it be possible to drop this support, as it has been EOL'd quite some time ago?

See #878 for rationale for Python 2.6 support.

> I would really like to make this work with the namespace_packages argument in setup()

Me too.

> OrderedDict is required

Although Setuptools can't accept dependencies naturally, it does have a well-defined way to vendor dependencies, so that is an option.

But it sounds as if the optimal approach here might be to use what you've envisioned, re-using the namespace_packages directive. I much prefer that because it's less duplicative.

Because this change does affect the way packagers' interface, it will have long-lasting effects, so I want to be sure to select the best approach before proceeding.

Let's come to an enthusiastic consensus in #97 and then proceed with an implementation.

agronholm:
To me, this seems to ultimately hinge on pip's continued support of Python 2.6. When that's gone, #878 can be resolved and with that, we won't have to vendor OrderedDict and can proceed with this PR. Right?

jaraco:
This PR as presented doesn't depend on OrderedDict unless the suggested adaptation is made, in which case, dropping Python 2.6 does help with that. I'm also willing to accept a vendored library for ordered dict to add that support with Python 2.6 support.

But I did realize later there's another issue with relying on the `namespace_packages` declaration in the call to `setup`, and that's that the use of `namespace_packages` is [discouraged](https://python-packaging-user-guide.readthedocs.io/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages).

This all leads me to believe that this explicit though sometimes duplicative declaration indicated in this PR as is may be the the best option.

agronholm:
> This all leads me to believe that this explicit though sometimes duplicative declaration indicated in this PR as is may be the the best option.

I disagree. These days I use setup.cfg exclusively for specifying metadata and most setup options too. Now what happens if only find_packages() gets the namespace package finding feature? I'm screwed again.

agronholm:
What is the problem with repurposing `namespace_packages` for use with `find:`? It's even backwards compatible when used this way, yes? We don't **have** to create any placeholder `__init__.py` files on Python 3 when this option is given, do we?

jaraco:
IIRC, if you currently specify `setup(namespace_packages=[...])`, those directories must contain `pkg_resources.declare_namespace` or setuptools will reject it. Perhaps based on the newer recommendations, we should drop that constraint.

agronholm:
Right, at least I don't see any other good way to make namespace packages work with `find:`. If you have a better idea, I'm all ears.

agronholm:
I think that might warrant an entirely new PR though, rather than trying to shoehorn it into this one.

jaraco:
Okay. Let's continue the discussion in #97.

PR #1141, opened 2017-09-03
Fix Python 2.6 support

benoit-pierre:
Fix `python2.6 bootstrap.py` failing on Travis.


jaraco:
I really should have noticed that the tests were failing in #1133. Yes, thanks for this.

jaraco:
I've force pushed v36.4.0, so it should re-attempt the release.

PR #1142, opened 2017-09-06
dist_info: implement command

xoviat:
@jaraco I had a test for this command [here](https://github.com/xoviat/wheel/commit/3c0126e35068a2ccd34a8e59307f2c833b47f420) but it requires a temp_pkg fixture. Is there such a fixture already in the setuptools tests?

### DIFF
--- a/setuptools/command/dist_info.py
+++ b/setuptools/command/dist_info.py
@@ -1,0 +1,37 @@
+"""
+Create a dist_info directory
+As defined in the wheel specification
+"""
+
+import os
+import shutil
+
+from distutils.core import Command
+
+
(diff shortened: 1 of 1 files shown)


jaraco:
Sorry for the delayed response. I'm slammed these days.

> Is there such a fixture already in the setuptools tests?

Not that I recall. I'd have to search. There are probably a few, so any effort you put into re-using what's there (even if that means refactoring) would be most appreciated.

xoviat:
Actually I just need #1143 to be merged; it covers the dist_info command with the pep517 tests.

> Sorry for the delayed response. I'm slammed these days.

You're fine.

Issue #1132, opened 2017-08-11
iter_entry_point often spending a lot of time on parsing a dummy requirements string


tmmorin:
When investigating a performance issue on a tool that makes use of setuptools entry points [1], profiling pointed at iter_entry_points as a hot point (see flamegraph attached to [1]), with iter_entry_points surprisingly spending a lot of time in pyparsing.

Digging further it appears that when an EntryPoint is created, even without anything in "extras", Requirements.parse is still called (see [2]) and ultimately results in parsing a dummy/empty requirements string ("x[]").

[1] https://bugs.launchpad.net/python-openstackclient/+bug/1702483
[2] https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L2289

tmmorin:
Avoiding the call to Requirements.parse when extras is empty (patch below), leads to a significant improvement of iter_entry_points performance.

```
--- /usr/lib/python2.7/dist-packages/pkg_resources/__init__.py	2017-08-02 00:40:42.000000000 +0200
+++ /var/tmp/alt__init__.py	2017-08-11 15:05:16.319781794 +0200
@@ -2286,7 +2286,10 @@
         self.name = name
         self.module_name = module_name
         self.attrs = tuple(attrs)
-        self.extras = Requirement.parse(("x[%s]" % ','.join(extras))).extras
+        if extras:
+            self.extras = Requirement.parse(("x[%s]" % ','.join(extras))).extras
+        else:
+            self.extras = ()
         self.dist = dist
 
     def __str__(self):
```

test:
```
timeit.timeit(
    setup='import pkg_resources',
    stmt='for ep in pkg_resources.iter_entry_points("openstack.cli.extension"): pass',
    number=100)
```
(with 17 entry points in the "openstack.cli.extension" group)

time with original code: **0.969233036041**
time with the patch below: **0.0956060886383**

tmmorin:
(a possibly better fix may possibly be, in `pkg_resources._vendor.packaging.requirements`, to avoid calling pyparsing in the case of a trivial requirements string)

jaraco:
I've been unable to replicate your findings. When I try running timeit on iter_entry_points, I'm seeing results in the ~0.09 ballpark:

```
$ python -m timeit -s "import pkg_resources" -n 100 "for ep in pkg_resources.iter_entry_points('distutils.commands'): pass"
100 loops, best of 3: 95 usec per loop
[...]

PR #1144, opened 2017-09-08
EntryPoint: avoid costly pyparsing for dummy requirement string

tmmorin:
In the case where an EntryPoint is created with no extra, a Requirement
object is instanciated with a dummy requirement string 'x[]', which incurs
a useless significant cost due to the use of pyparsing to parse this string.

This change skips the instanciation of a Requirements objects in the
case where extra is empty.

See github issue #1132 for more details, including information on the
performance improvement brought by this change.


pkg_resources/__init__.py
```diff
@@ -2286,7 +2286,10 @@ def __init__(self, name, module_name, attrs=(), extras=(), dist=None):
         self.name = name
         self.module_name = module_name
         self.attrs = tuple(attrs)
-        self.extras = Requirement.parse(("x[%s]" % ','.join(extras))).extras
+        if extras:
+            self.extras = Requirement.parse(("x[%s]" % ','.join(extras))).extras
```
webknjaz: It looks like `Requirement.parse` and underlying `parse_requirements` aren't so expensive. But anyway I'd add this check there.

jaraco:
Thanks very much for the ticket with a detailed description of the problem and a clear indication of a performance impact. This PR is an excellent example showing how to communicate a problem and a proposed solution.

I do have some critiques that I'd like to see if we can address.

My main concern lies with the lack of comment or test protecting this new behavior. I could easily see myself optimizing this change away, not realizing the affect on performance. If it would be possible to write a test, that would be best... but a separate function documenting the intention would also be good.

My preference, though, would be for packaging to expose a 'parse_extras' function or method that could be readily invoked without the expense of parsing a dummy requirement, such that the setuptools implementation was simple and direct. Let's explore that first.

tmmorin:
Thanks @jaraco and I agree that documenting this properly is required to avoid regressions.
I also agree that having this done right in `packaging` would be even better.

I'm I interpreting your suggestion correctly as meaning to add a `parse_extras` method to `packaging.requirements.Requirement` [1].

But then I fail to see how to write such a function: in the case where extras is non empty, it would have to call the "regular" parse(<generated requirement string for the extras>).   But `packaging.requirements.Requirement` does *not* have a `parse` method, the `parse` method used at [2] is in fact added in the wrapper at [3].

Given that, I would prefer adding a `parse_extras` method in this wrapper, with proper documentation on why we shortcut the case where `extras` is empty.

[1] https://github.com/pypa/packaging/blob/master/packaging/requirements.py#L78
[2] https://github.com/tmmorin/setuptools/blob/94fa402a207e27139499d3347a69a7d8f7036939/pkg_resources/__init__.py#L2290
[3] https://github.com/tmmorin/setuptools/blob/94fa402a207e27139499d3347a69a7d8f7036939/pkg_resources/__init__.py#L2950-L2953

jaraco:
In d1f16fdd, I added one small optimization, bypassing much of the unnecessary processing in `pkg_resources.Requirement.__init__`, though I suspect from what I see in the original ticket that profiling revealed `pyparsing` as the primary culprit, so this structural optimization won't provide the desired performance optimization.

jaraco:
In 4ee5c651, I found what I think is an even better solution. If there was a purpose behind the serialization/parsing step, it's not captured by comments nor tests, so it's subject to change.

Issue #844, opened 2016-11-09
bdist_egg --exclude-source-files does not work with Python3


dmiyakawa:
(Coming from here: https://bugs.python.org/issue28630)

I'm trying to prepare an egg file without source code.
I had used `python setup.py bdist_egg --exclude-source-files`, which does not work with Python3.

After digging into the issue more, I found PEP-3147 is related to the problem.
https://www.python.org/dev/peps/pep-3147/

```
$ git clone https://github.com/dmiyakawa/python_greeting.git
..
$ cd python_greeting
$ python --version
Python 3.5.2
$ python setup.py bdist_egg --exclude-source-files
..
$ unzip -l dist/greeting-1.0.0-py3.5.egg
Archive:  dist/greeting-1.0.0-py3.5.egg
  Length      Date    Time    Name
---------  ---------- -----   ----
        1  11-09-2016 16:31   EGG-INFO/dependency_links.txt
       57  11-09-2016 16:31   EGG-INFO/entry_points.txt
      319  11-09-2016 16:31   EGG-INFO/PKG-INFO
      213  11-09-2016 16:31   EGG-INFO/SOURCES.txt
        9  11-09-2016 16:31   EGG-INFO/top_level.txt
        1  11-09-2016 16:31   EGG-INFO/zip-safe
      139  11-09-2016 16:31   greeting/__pycache__/__init__.cpython-35.pyc
      392  11-09-2016 16:31   greeting/__pycache__/greeting.cpython-35.pyc
---------                     -------
     1131                     8 files
```

Here, `greeting/__pycache__/__init__.cpython-35.pyc` and `greeting/__pycache__/greeting.cpython-35.pyc` are not where they should be.
See the following results (with Python 2.7.12)

```
$ python --version
Python 2.7.12
$ python setup.py bdist_egg --exclude-source-files
..
$ unzip -l dist/greeting-1.0.0-py2.7.egg
Archive:  dist/greeting-1.0.0-py2.7.egg
  Length      Date    Time    Name
---------  ---------- -----   ----
        1  11-09-2016 16:32   EGG-INFO/dependency_links.txt
       57  11-09-2016 16:32   EGG-INFO/entry_points.txt
      319  11-09-2016 16:32   EGG-INFO/PKG-INFO
      213  11-09-2016 16:32   EGG-INFO/SOURCES.txt
        9  11-09-2016 16:32   EGG-INFO/top_level.txt
[...]

PR #1145, opened 2017-09-13
python 3 bdist_egg --exclude-source-files __pycache__ issue

haobibo:
This PR is to mend the `__pycache__` subfolder issue in Python 3 for bdist_egg.
There is an earlier issue (https://github.com/pypa/setuptools/issues/844) discussed the undesired `__pycache__/xxx.cpython-3x` in `pyc` file path for Python 3 when excluding source files.


haobibo:
This PR fails one Python 2.6 based test because of one line code of `wheel` does not work (SyntaxError) under Python 2.6:
https://github.com/pypa/wheel/blob/3c9cc212a76b334307985cebfa7ebf269a419223/wheel/bdist_wheel.py#L407

Please consider accepting the PR (https://github.com/pypa/wheel/pull/194) under `wheel` to fix the SyntaxError under Python 2.6 or remove Python 2.6 tests from this repo.

haobibo:
Hi @benoit-pierre , as @dholth suggested in https://github.com/pypa/wheel/pull/194 ,  would you prefer to remove Python 2.6 CI job?

Thank you @bRsatUoh ! Is there anything I can do to help this PR merged?

benoit-pierre:
The Python 2.6 CI job has been fixed on master, so you can rebase your branch to fix the issue.

haobibo:
Thank you @benoit-pierre !
I rebased the PR and now all tests passed.

setuptools/command/bdist_egg.py
```diff
@@ -238,13 +239,26 @@ def run(self):
 
     def zap_pyfiles(self):
         log.info("Removing .py files from temporary directory")
+        py3 = sys.version_info >= (3, 0)
```
jaraco: Perhaps it's possible to invoke this behavior unconditionally (on Python 2 and Python 3). Presumably, it will be a null operation when no matching files are found.

setuptools/command/bdist_egg.py
```diff
@@ -238,13 +239,26 @@ def run(self):
...
+        re_pycache_file = re.compile('(.+)\.cpython-3\d(.pyc)$')
```
jaraco: Probably best not to hard-code some of the "magic tag" (i.e. `cpython-3\d`). Also, there's a bug in the use of unescaped `.` in `.pyc`. Also, I think there's an issue with this string not being a raw string (Python 3.6 has warnings about these). Maybe instead:

```
pattern = r'(?P<name>.+)\.(?P<magic>[^.]+)\.pyc'
```

Then use `m.groupdict()` or `m.group('name')` to extract the name.

setuptools/command/bdist_egg.py
```diff
@@ -238,13 +239,26 @@ def run(self):
...
                     log.debug("Deleting %s", path)
                     os.unlink(path)
 
+                if py3 and base.endswith('__pycache__'):
+                    path_old = os.path.join(base, name)
+                    
+                    m = re_pycache_file.match(name)
+                    path_new = os.path.join(base, os.pardir, m.group(1) + m.group(2))
+                    log.info("Renaming Python 3 .pyc file from [%s] to [%s]" % (path_old, path_new))
+                    if os.path.exists(path_new):
```
jaraco: There's a race condition here. Better to simply attempt to remove it and suppress any failure Exception.

PR #1146, opened 2017-09-13
Drop Python 2.6 support

benoit-pierre:
With the new wheel 0.30.0 dropping support for Python 2.6, and the next version of pip doing the same, I think it's time to do it too.


benoit-pierre:
@jaraco: I see that you already have a branch in progress for doing that, I'll check the differences.

jaraco:
Also, there's more discussion in #878.

benoit-pierre:
OK, in the mean time, how about pinning wheel to 0.29.0 for the 2.6 job?

PR #1148, opened 2017-09-13
travis: make deploy stage conditional

benoit-pierre:
Skip the deploy stage if no tag is present.


webknjaz:
+1, I've already played with this one in a few repos. @jaraco you may merge this w/o any fear

jaraco:
Thanks both!

PR #1152, opened 2017-09-16
Clean up resource warnings during tests

jdufresne:
When tests are invoked with the Python flag -Wall, warnings appear in the form of:

```
  ResourceWarning: unclosed file ...
```

Close all files and resources deterministically to avoid such warnings. Most often, easiest to do using a context manager.


webknjaz:
@jaraco this PR is 100% ready to be merged

PR #1155, opened 2017-09-20
Tests: Run `python -m pip` instead of plain `pip`

hroncok:
This allows to run the testsuite on systems where `/usr/bin/pip` is a specific Python version, but the tests are run on a different one.

For example in Fedora RPM, when the Python 3 tests are invoked, they should use Python 3 pip, not Python 2 `/usr/bin/pip`.

Unlike other approaches, like using `pip2`/`pip3` which is currently done in Fedora RPM (downstream patch), this way it Works Everywhere™ and the downstream patch can be dropped.

See https://src.fedoraproject.org/rpms/python-setuptools/blob/54eaa03a4dc97f93a5e4c92c55e580a4ab55a058/f/0001-Run-test-on-a-version-specific-pip.patch

cc @torsava (author of that Fedora downstream patch)

### DIFF
--- a/setuptools/tests/test_develop.py
+++ b/setuptools/tests/test_develop.py
@@ -169,7 +169,7 @@ class TestNamespaces:
         install_cmd = [
             sys.executable,
             '-m',
-            'pip',
+            'pip.__main__',
             'install',
             str(pkg_A),
             '-t', str(target),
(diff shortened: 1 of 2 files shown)


hroncok:
Oh, this breaks Python 2.6 :(

My Works Everywhere™ assumption was wrong.

Let me see what can i do about that.

hroncok:
Fixed.

torsava:
Great solution @hroncok !

This would greatly help downstream distributors of Python and setuptools, especially now that Python 2 is finally on its death bed.

Please consider merging!

PR #1166, opened 2017-10-04
MSVC: Add API to create EnvironmentInfo from a compiler

xoviat:
Why this is needed *in principle*: for any project building C++ on windows, it may be necessary to include additional DLLs from the "redist" folder. We can provide a function to do that, or we can create a generic function that will allow us to easily create an `EnvironmentInfo` that we can use to query about various MSVC information.

This *is not necessarily* the correct implementation of this function. If you have a better idea, I am open to it. But the requirement is that the function return an `EnvironmentInfo` with the same version number and architecture as the compiler (not necessarily the same installation).


xoviat:
Also, we need test coverage, but would like feedback first.

xoviat:
Actually it would probably be better to make EnvironmentInfo a property of the compiler through monkeypatching.

xoviat:
Closing for now.

PR #1171, opened 2017-10-13
Allow to use `file:` directive for requirements.

### LINKED ISSUE #1074: Allow 'file:' for tests_require and install_require in setup.cfg
Hi !

Most of the people use a `requirements.txt` file for managing their dependencies. Unfortunately, `setuptools` does not support reading from those files for the requirements tags, which means that devs have to either duplicate their requirements file in the `install_requires` parameter, or not use a requirement file at all. Since `setuptools` already supports reading some metadata from a file, it should probably be allowed to read those parameters as well.

tohin:
Fix #1074.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2430,8 +2430,8 @@ Options
 Key                      Type
 =======================  =====
 zip_safe                 bool
-setup_requires           list-semi
-install_requires         list-semi
+setup_requires           file:, list-semi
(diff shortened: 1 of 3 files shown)


jaraco:
I think there are some important design considerations that first need to be addressed. I've raised those in the parent ticket. I'm going to close this PR for now, but hang onto it as it may form the basis for an acceptable solution once the design concerns are addressed.

Issue #1174, opened 2017-10-15
Help debugging entry_points failure


xoviat:
While trying to get pip to use the new `build_meta`, I ran into a problem: setuptools loaded the entry points from the global environment rather than the temporary build environment: [see here](https://travis-ci.org/xoviat/pip/jobs/288084422#L1036). How does `pkg_resources` discover `entry_points` and how can I override it?

xoviat:
This was a stupid issue. Closing.

benoit-pierre:
Entry points are loaded from active distributions. On import, `pkg_resources` will active all distributions found in `sys.path`. The problem with using `concurrent.futures` is that, like `multiprocessing`, the main module will be imported, resulting in a lot of modules being imported before you get a chance to fix the path.

I took a stab at monkey-patching the backend to use `pip._internal.utils.misc.call_subprocess` instead, and fix a few things, see here: https://travis-ci.org/benoit-pierre/pip/jobs/288266999

There are some issues with the method used by setuptools' backend to run `setup.py`:
- the environment handling of locals/globals is wrong (that's why `test_install_with_hacked_egg_info` is failing)
- when a package use `if __name__ == '__main__':` to make its `setup.py` importable (like setuptools itself does), it does not work (so pip installing setuptools itself from source does not work)

Both those issues are resolved with the following patch:
```diff
 setuptools/build_meta.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git i/setuptools/build_meta.py w/setuptools/build_meta.py
index 54f2987b..f89f82d6 100644
--- i/setuptools/build_meta.py
+++ w/setuptools/build_meta.py
@@ -65,10 +65,11 @@ def _run_setup(setup_script='setup.py'):
     # Note that we can reuse our build directory between calls
     # Correctness comes first, then optimization later
     __file__ = setup_script
+    __name__ = '__main__'
     f = getattr(tokenize, 'open', open)(__file__)
     code = f.read().replace('\\r\\n', '\\n')
     f.close()
-    exec(compile(code, __file__, 'exec'))
+    exec(compile(code, __file__, 'exec'), locals())
 
 
 def _fix_config(config_settings):
```

benoit-pierre:
I don't think it was stupid :)

[...]

PR #1175, opened 2017-10-15
Build meta: fixes and cleanups

xoviat:
Attached are some minor cleanups and fixes to the build_meta module that I added in #1143.

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -93,6 +93,11 @@ def _get_build_requires(config_settings):
     return requirements
 
 
+def _get_immediate_subdirectories(a_dir):
+    return [name for name in os.listdir(a_dir)
+            if os.path.isdir(os.path.join(a_dir, name))]
+
+
 def get_requires_for_build_wheel(config_settings=None):
     config_settings = _fix_config(config_settings)
     return _get_build_requires(config_settings)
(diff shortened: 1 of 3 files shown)


xoviat:
Okay, this should be ready to go.

setuptools/build_meta.py
```diff
@@ -64,11 +66,13 @@ def patch(cls):
 def _run_setup(setup_script='setup.py'):
     # Note that we can reuse our build directory between calls
     # Correctness comes first, then optimization later
+    _initialize_master_working_set()
```
benoit-pierre: This is not necessary anymore, no?

jaraco:
This PR needs more detail. In particular, it needs a bug report or comment describing the undesirable behavior this is correcting. I see some description in the commit message of the first commit (ffb2e69)... and I'd like to know more. In particular, what are the two setuptools versions under which different entry points are loaded?  I feel like I've encountered some of these issues before and want to be sure we're coordinating approaches.

benoit-pierre:
Related: https://github.com/pypa/setuptools/issues/1174.

Adding 1 or 2 tests for the change to `_run_setup` would be great, check if those examples can be run:
```python
variable = True
def function():
  return variable
```
```python
assert __name__ == '__main__'
```

xoviat:
@benoit-pierre Can you explain the local() and __main__ thing here?

xoviat:
>  In particular, it needs a bug report or comment describing the undesirable behavior this is correcting. 

In particular, some setup scripts ([particularly here](https://github.com/pypa/pip/blob/66c53519c2629ec784659fba85c5a84c88144663/tests/data/packages/HackedEggInfo/setup.py)) failed because the `__name__` attribute wasn't set to `__main__`. In addition, passing `locals()` executes the script in the context of the function, which doesn't pass in the classes defined in the module as part of the `setup.py` namespace.

The other change with respect to `dist_info` is a minor cleanup that reduces the number of times that files are copied and, in the case of a read-only directory with a `setup.py`, doesn't fail trying to create `.egg-info`. 

The `pkg_resources` change has been dropped because it actually isn't necessary and I am afraid it would bog down this PR.

xoviat:
Tests have been added. 

@jaraco Hopefully I've done the right thing.

xoviat:
Also many thanks to @benoit-pierre for the excellent PR reviews.

xoviat:
@jaraco ping.

jaraco:
Released as v36.7.3

xoviat:
Thanks. It looks like the pip side has gone off the rails though.

Issue #701, opened 2016-07-28
tests run twice with setup.py test in python3.5


wtayyeb:
Hi.
The `setup.py test` in python3.5 result to twice test run.

the very simple project to reproduce the problem is located at https://github.com/wtayyeb/setuptools-twice-test which have three tests

by running it via `tox` you could see that for py27 no test found (legacy behaviour, need `discover` to find them) and for py34 the behaviour is fixed and three tests were run. but in py35 it run 6 tests, as they are doublicated. 

if we use `setup.py test -s dummy` all three environment will report 3 test run. but it is not the general way. also when run `python -m unittest discover` it also result correctly 3 test for each python version.

In case of using legacy woraround that create test_command that add `discover` to unittest args, the problem still exists for py35.

the result of tox is like below

<pre>
**py27 installed:**
py27 runtests: PYTHONHASHSEED='545'
py27 runtests: commands[0] | python setup.py test
running test
running egg_info
writing top-level names to dummy.egg-info\top_level.txt
writing dependency_links to dummy.egg-info\dependency_links.txt
writing dummy.egg-info\PKG-INFO
reading manifest file 'dummy.egg-info\SOURCES.txt'
writing manifest file 'dummy.egg-info\SOURCES.txt'
running build_ext

----------------------------------------------------------------------
**Ran 0 tests in 0.000s**

OK
**py34 installed:**
py34 runtests: PYTHONHASHSEED='545'
py34 runtests: commands[0] | python setup.py test
running test
running egg_info
writing dummy.egg-info\PKG-INFO
writing top-level names to dummy.egg-info\top_level.txt
writing dependency_links to dummy.egg-info\dependency_links.txt
reading manifest file 'dummy.egg-info\SOURCES.txt'
writing manifest file 'dummy.egg-info\SOURCES.txt'
running build_ext
test_one (dummy.test_dummy.Test) ... ok
test_three (dummy.test_dummy.Test) ... ok
test_two (dummy.test_dummy.Test) ... ok

----------------------------------------------------------------------
**Ran 3 tests in 0.000s**

OK
**py35 installed:**
py35 runtests: PYTHONHASHSEED='545'
py35 runtests: commands[0] | python setup.py test
running test
running egg_info
writing dummy.egg-info\PKG-INFO
writing top-level names to dummy.egg-info\top_level.txt
writing dependency_links to dummy.egg-info\dependency_links.txt
reading manifest file 'dummy.egg-info\SOURCES.txt'
[...]

PR #1183, opened 2017-10-31
Keep track of processed test modules to avoid duplicate run.

yoichi:
Fixes https://github.com/pypa/setuptools/issues/701
which is reproduced with python3.6.3 and setuptools 403bfce

### DIFF
--- a/setuptools/command/test.py
+++ b/setuptools/command/test.py
@@ -18,6 +18,10 @@ from setuptools.py31compat import unittest_main
 
 
 class ScanningLoader(TestLoader):
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


benoit-pierre:
This would be great if you could add a test. I actually have a fix for this I forgot to PR, see here: https://github.com/pypa/setuptools/compare/master...benoit-pierre:fix_test_command

yoichi:
@benoit-pierre It seems better than this. Would you submit a PR? I'll decline this PR.

benoit-pierre:
@yoichi: OK, PR here: #1187.

PR #1190, opened 2017-11-08
use ssl.create_default_context and SNI if available

gweis:
I ran into a ssl certificate verification problem using buildout and a private package index using SNI.

The usual workaround to make sure that the ssl cert of the private registry server defaults to the package index hostname was not possible this time, so I thought I'll fix it at the core and use SNI when connecting to a ssl server.

Happy to ammend this pull request in any way, but I do believe that something like this would be good to have as it also uses good practice ssl settings maintained in python core ssl.py module, and sticks to officially document methods in ssl module.


3bsilversmiths:
Ok

PR #1204, opened 2017-11-20
[distutils] fix test failures

xoviat:
On Python 3.6 and above, setuptools replaces the builtin distutils with it's own version. Python 3.5 and below is unaffected. The idea is to start with Python 3.6 and then roll in future versions as they come, until setuptools always uses it's own copy of distutils. Eventually, python can remove distutils and add an ensuresetuptools module for building.

/cc @jaraco @zooba

### DIFF
--- a/distutils/__init__.py
+++ b/distutils/__init__.py
@@ -9,5 +9,13 @@ used from a setup script as
 """
 
 import sys
+import os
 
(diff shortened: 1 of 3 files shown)


xoviat:
Windows/2.7 seems to be failing. Not sure why.

xoviat:
Okay, that looks good.

PR #1205, opened 2017-11-21
Support PEP 345 Project-URL metadata list

fungi:
A Project-URL list is supported in v1.2 PKG-INFO metadata with
pkginfo 0.6 or later, and is used primarily by PyPI to display
helpful hyperlinks in a generic manner. Add support here to be able
to pass it through setup.py. See the corresponding section of the
Core Metadata Specifications from the Python Packaging User Guide
for details:

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2406,6 +2406,7 @@ name                                               str
 version                                            attr:, str
 url                             home-page          str
 download_url                    download-url       str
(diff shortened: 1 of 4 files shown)


fungi:
After some reflection on the core metadata spec and PEP 345, I believe I've interpreted the multi-use option format incorrectly and will need to re-spin this.

Issue #1206, opened 2017-11-21
long_description_content_type supported in setup.cfg but does not work


cryvate:
I tried setting long_description_content_type in my setup.cfg metadata section, however the tarball PKG-INFO does not have it set. Setting it in setup.py does work.

PR #1207, opened 2017-11-21
Add setup.cfg support for long_description_content_type

cryvate:
[This was already claimed in the docs.](https://github.com/pypa/setuptools/blob/d45be2cc4f7a1e4ddc70b363baaa613c6b068b98/docs/setuptools.txt#L2417) That this did not work before can be seen by cherry picking the changed test in this PR without the other changes: it will fail. Fixes issue #1206

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -598,7 +598,7 @@ def write_pkg_info(cmd, basename, filename):
         metadata.version, oldver = cmd.egg_version, metadata.version
         metadata.name, oldname = cmd.egg_name, metadata.name
         metadata.long_description_content_type = getattr(
-            cmd.distribution,
+            cmd.distribution.metadata,
             'long_description_content_type'
         )
         try:
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -597,10 +597,7 @@ def write_pkg_info(cmd, basename, filename):
         metadata = cmd.distribution.metadata
         metadata.version, oldver = cmd.egg_version, metadata.version
         metadata.name, oldname = cmd.egg_name, metadata.name
-        metadata.long_description_content_type = getattr(
-            cmd.distribution.metadata,
-            'long_description_content_type'
-        )
+
         try:
             # write unescaped data to PKG-INFO, so older pkg_resources
             # can still parse it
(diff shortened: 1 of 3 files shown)


fungi:
I was just noticing this today while working on Project-URL metadata support, and your try/except there is exactly what I was missing to get my added test passing. Thanks!

setuptools/dist.py
```diff
@@ -332,6 +332,13 @@ def __init__(self, attrs=None):
         for ep in pkg_resources.iter_entry_points('distutils.setup_keywords'):
             vars(self).setdefault(ep.name, None)
         _Distribution.__init__(self, attrs)
+
+        try:
```
jaraco: This code block bothers me for several reasons.

First, it seems overly defensive. Is it really true that sometimes metadata.long_description_content_type will be set, but other times not? Even if that's really true, it should use `vars(self.metadata).setdefault('long_description_content_type', self.long_description_content_type`.

Second, why is `long_description_content_type` being singled-out here? Why do other metadata fields not need this special handling?

Third, it seems bad form to be copying an attribute by value from one object to another. Ideally, and especially for newer attributes, there should be one place where that value is defined.

I think perhaps the bug that this PR is seeking to address is revealing a design flaw in the implementation of the attribute.

Can we make long_description_content_type work like other attributes so it doesn't require this special treatment?

setuptools/command/egg_info.py
```diff
@@ -598,7 +598,7 @@ def write_pkg_info(cmd, basename, filename):
         metadata.version, oldver = cmd.egg_version, metadata.version
         metadata.name, oldname = cmd.egg_name, metadata.name
         metadata.long_description_content_type = getattr(
-            cmd.distribution,
+            cmd.distribution.metadata,
```
jaraco: This probably should just be metadata.

setuptools/dist.py
cryvate: This code block does not make me happy. I know that distutils' DistributionMetadata will not set this attribute, and setup.cfg is not honoured because of [this check](https://github.com/pypa/setuptools/blob/dce4750aed2d5b0a4ba677b0e308cd10dca2f6ee/setuptools/config.py#L162). I think it is different because it does not appear in any PEP, and hence not in distutils.

I'll have a look at making things more sleek :) I think there will still need to be special casing, but I will at least try to reduce it down to one place.

cryvate: I think fundamentally, the only way not to special case it, is to patch distutils' DistributionMetadata. As you can see in the new commit I just pushed, there was some special casing of this attribute that is slightly simplified (in my view) by setting it in the constructor.

Issue #458, opened 2015-11-06
Please make the generated requires.txt files reproducible


bb-migration:
Originally reported by: **lambyuk (Bitbucket: [lambyuk](http://bitbucket.org/lambyuk), GitHub: Unknown)**

---

```
Whilst working on the "reproducible builds" effort [0], we noticed that
python-setuptools generates non-deterministic output when creating
requires.txt files:

│   │   │   │  [all]
│   │   │   │ -fake-factory==0.5.2
│   │   │   │ -pytz
│   │   │   │  numpy>=1.9.0
│   │   │   │  pytz
│   │   │   │ +pytz
│   │   │   │  django>=1.7
│   │   │   │ +fake-factory==0.5.2


Patch attached.

 [0] https://wiki.debian.org/ReproducibleBuilds
```

Filed in Debian as https://bugs.debian.org/804249

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/458

anthraxx:
@jaraco any updates on this? debian is already applying a custom patch to this issue. currently this is blocking reproducible python builds for our distro. I hope we can get some movement here after 2 years

jaraco:
Apologies for the delay. It’s the unfortunate situation that passing time affords little benefit for open Setuptools tickets (with the exception of regressions). Pull Requests get the lion’s share of attention.

But I’ll see if I can knock this one out.

jaraco:
I'm trying to understand the issue. By using the patch as a starting point, I'm looking at how `dist.install_requires` is generated, and it seems to be taking the sequence of requirements directly from the package declaration, and if that declaration was deterministically ordered, so also would be the requires.txt file.

I'll see if I can replicate the finding as reported, but my suspicion is it happens only if the package being built uses a non-deterministically ordered sequence (set or dict) for their `install_requires` declaration.

jaraco:
Although i pushed that commit with the 'fix', I've stripped that commit and I'm pushing the tests separately. Using c461d513, I've determined that the issue exists only when the requirements were declared by the user in a non-deterministically-ordered sequence:

```
$ tox --hashseed 1983024029 -- -v -k 'install_requires and deterministic and not in_setup_cfg'
python develop-inst-nodeps: /Users/jaraco/Dropbox/code/main/setuptools
[...]

PR #1210, opened 2017-11-22
Support PEP 345 Project-URL metadata

fungi:
By including one or more Project-URL entries in PKG-INFO metadata,
PyPI can display helpful hyperlinks in a generic manner. Add support
here to be able to pass it through setup.cfg and setup.py with a
project_urls dict. See the corresponding section of the Core
Metadata Specifications from the Python Packaging User Guide for
details:


fungi:
An example built with this can be seen in the "Project Links" box in the top-right area of https://testpypi.python.org/pypi/mudpy/0.0.1.dev232 where the custom Project-URL entries for Bug Tracker, Documentation and Source Code are displayed.

setuptools/dist.py
```diff
@@ -57,6 +57,8 @@ def write_pkg_file(self, file):
     file.write('License: %s\n' % self.get_license())
     if self.download_url:
         file.write('Download-URL: %s\n' % self.download_url)
+    for project_url in self.project_urls.items():
```
fungi: Should this be changed to operate on a sorted list of keys instead so that output order remains consistent between Python versions?

cryvate: Presumably the order here is going to be the same as it is displayed in on PyPi?

The user will probably want to present a certain order that is not alphabetical. If the config is in code (i.e. setup.py) the user will have to make sure this preserves ordering (i.e. use a collections.OrderedDict, or dict on Python3.6+).

However, in the setup.cfg the dict creation is done by setuptools, and it isn't ordered. As python2.6 is not supported anymore, we can replace [the dictionary creation](https://github.com/pypa/setuptools/pull/1210/files#diff-360050608fe825a457c8093298edf207R219) with a collections.OrderedDict to preserve the order.

setup.py
```diff
@@ -98,6 +98,11 @@ def pypi_link(pkg_filename):
     long_description_content_type='text/x-rst; charset=UTF-8',
     keywords="CPAN PyPI distutils eggs package management",
     url="https://github.com/pypa/setuptools",
+    project_urls={
```
jaraco: I can't help but feel like this setting is adding a lot of repetition. Now `https://github.com/pypa/setuptools` appears three times in the metadata. Perhaps if this only included the `Documentation` link, I'd feel better than a usage that demands so much repetition. I don't feel strongly about it, though.

setup.py
```diff
@@ -98,6 +98,11 @@ def pypi_link(pkg_filename):
...
+        "Bug Tracker": "https://github.com/pypa/setuptools/issues",
+        "Documentation": "http://setuptools.readthedocs.io/",
```
jaraco: Should be https.

setuptools/dist.py
jaraco: In light of #458, I suspect it's going to be a hard requirement that any package metadata be captured or stored in a deterministic order. Even in Python 3.6, the ordering "[should not be relied upon](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation)." Therefore, I advise that the user be required to provide an ordering or that the metadata always sort the keys before storing the values.

setup.py
fungi: Happy to adjust this to whatever you like honestly. This was mostly just a placeholder while seeking feedback from the setuptools maintainers as to what they'd want to see added for this set (if anything). I'll scale this back to just the Documentation link here so it's still dogfooded, and then you can add/update any URLs you like in the future obviously.

setup.py
fungi: Will fix, thanks!

setuptools/dist.py
fungi: Okay, so dict type structures from setup.cfg will need switching to collections.ordereddict under the covers anyway, and I shouldn't need to make any corresponding adjustments to this PR to accomodate?

As for "same as it is displayed in on PyPI" that's a bit uncertain at the moment. I agree that PyPI should make some attempt to preserve ordering of this, but my test upload is showing the reverse order from what's in the sdist PKG-INFO and wheel METADATA files on the cheeseshop (I haven't been able to sufficiently untangle the source there but I have a feeling order is being determined by DB query results). By contrast, the implementation in warehouse exhibits the same sort of list-of-tupes-fed-into-unordered-dict design as here (and is also not working at the moment, per discussion in https://github.com/pypa/warehouse/issues/1419 ).

fungi:
jaraco: Where you say "more documentation" where did you have in mind? It's already documented in the Core Metadata Specifications from the Python Packaging User Guide. Should I add a "Project URLs" subsection in docs/formats.txt under "Standard Metadata" summarizing the feature?

setuptools/command/egg_info.py
```diff
@@ -597,6 +597,7 @@ def write_pkg_info(cmd, basename, filename):
         metadata = cmd.distribution.metadata
         metadata.version, oldver = cmd.egg_version, metadata.version
         metadata.name, oldname = cmd.egg_name, metadata.name
+        metadata.project_urls = cmd.distribution.metadata.project_urls
```
fungi: Realizing this line is likely unnecessary.

fungi:
On closer inspection I'm seeing that the metadata sections in formats.txt are about various sorts of metadata files rather than specific entries. Maybe just update examples in the "Basic Usage" section of docs/setuptools.txt instead?

fungi:
Are there any other adjustments you'd want to see made to this?

PR #1220, opened 2017-11-30
fix `data_files` handling when installing from wheel

### LINKED ISSUE #1219: setuptools wheel install will fail if a data dir is a subdirectory in a package
With the new wheel support, I ran into an issue where if `data_files` specifies a subdirectory in a package, `setuptools` will fail on moving the data.  This worked in previous versions of setuptools w/o wheel support, and also works with pip install.

Perhaps this is not a supported use case (should `package_data` be used instead?)

I was able to work around the issue by removing the subdirectory, but in case someone else runs into this, here is a repro test case that can be added to `test_wheel.py`

```
    dict(
        id='data_in_package',
        file_defs={
            'foo': {
                '__init__.py': '',
                'data_dir': {
                    'data.txt': DALS(
                        '''
                        Some data...
                        '''
                    ),
                }
            }
        },
        setup_kwargs=dict(
            packages=['foo'],
            data_files=[('foo/data_dir', ['foo/data_dir/data.txt'])],
        ),
        install_tree=DALS(
            '''
            foo-1.0-py{py_version}.egg/
            |-- EGG-INFO/
            |  |-- DESCRIPTION.rst
            |  |-- PKG-INFO
            |  |-- RECORD
            |  |-- WHEEL
            |  |-- metadata.json
            |  |-- top_level.txt
            |-- foo/
            |  |-- __init__.py
            |  |-- data_dir/
            |     |-- data.txt
            '''
        ),
    ),
```

In the test case, `foo` is a package and `/foo/data_dir/` is a data dir. Since `foo` already exists as a package, the `os.rename` fails, even though `/foo/data_dir/` does not exist. Should it check if the directory already exists?

The error looks something like the following (on OSX):

```
                for entry in os.listdir(subdir):
                    os.rename(os.path.join(subdir, entry),
>                             os.path.join(destination_eggdir, entry))
E                   OSError: [Errno 66] Directory not empty: '/var/folders/07/c66dj8613n37ssnxsvyz8dcm0000gn/T/tmpybp6f4ft/foo-1.0-py3.5.egg/foo-1.0.data/data/foo' -> '/var/folders/07/c66dj8613n37ssnxsvyz8dcm0000gn/T/tmpybp6f4ft/foo-1.0-py3.5.egg/foo'
```

benoit-pierre:
Fix #1219.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v38.2.4
+-------
+
+* #1220: Fix `data_files` handling when installing from wheel.
+
 v38.2.3
 -------
 
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -444,6 +444,42 @@ WHEEL_INSTALL_TESTS = (
             '''),
     ),
 
+    dict(
+        id='data_in_package',
+        file_defs={
+            'foo': {
+                '__init__.py': '',
+                'data_dir': {
+                    'data.txt': DALS(
+                        '''
+                        Some data...
+                        '''
+                    ),
+                }
+            }
+        },
+        setup_kwargs=dict(
+            packages=['foo'],
+            data_files=[('foo/data_dir', ['foo/data_dir/data.txt'])],
+        ),
+        install_tree=DALS(
+            '''
+            foo-1.0-py{py_version}.egg/
+            |-- EGG-INFO/
+            |  |-- DESCRIPTION.rst
+            |  |-- PKG-INFO
+            |  |-- RECORD
+            |  |-- WHEEL
+            |  |-- metadata.json
+            |  |-- top_level.txt
+            |-- foo/
+            |  |-- __init__.py
+            |  |-- data_dir/
+            |  |  |-- data.txt
+            '''
+        ),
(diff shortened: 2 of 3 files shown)


reinout:
I tried your fix locally and it works like a charm.

I saw the problem when installing a django-celery wheel. Here's the first part of unzipping the wheel:

  inflating: django_celery-3.1.10.data/data/djcelery/static/djcelery/style.css
  inflating: django_celery-3.1.10.data/data/djcelery/templates/admin/djcelery/change_list.html
  inflating: django_celery-3.1.10.data/data/djcelery/templates/djcelery/confirm_rate_limit.html
  inflating: djcelery/__init__.py
  inflating: djcelery/admin.py
  inflating: djcelery/admin_utils.py
  inflating: djcelery/app.py
  inflating: djcelery/common.py
  inflating: djcelery/db.py
  ...

So: exactly the same problem. And your fix works. So it looks good to merge!

benoit-pierre:
@reinout, @ikreymer: thanks for testing! I'll merge and release 38.2.4.

reinout:
I've tested 38.2.4 and it works like a charm.
Thanks!

Issue #1221, opened 2017-12-01
Add more tests for PEP 517


ghost:
One issue with [this code](https://github.com/pypa/setuptools/blob/b729553a6c6b92dfc7cfd9c577f2399bfde3000e/setuptools/build_meta.py#L120-L124) is that it's completely untested. In some cases, setup.py egg_info can create a directory not in the `egg_base`, but in a subdirectory. We need to figure out when this can happen, and then add a test for it.

xoviat:
Here's the test failure that occured at the time:
```

tests/functional/test_uninstall.py Collecting pd.requires==0.0.3
  Downloading pd.requires-0.0.3.tar.gz
  Installing build dependencies
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
    Complete output from command /tmp/pytest-of-travis/pytest-0/test_uninstall_namespace_packa0/workspace/venv/bin/python3.6 -I -c "
    import importlib
    import os
    import sys
    
    py_path = '/tmp/pytest-of-travis/pytest-0/test_uninstall_namespace_packa0/workspace/tmp/pip-build-env-7b1av7_j/lib/python3.6/site-packages'
    if py_path is not None:
        sys.path[0:0] = py_path.split(os.pathsep)
    mod = importlib.import_module('setuptools.build_meta')
    res = getattr(mod, 'prepare_metadata_for_build_wheel')(*('/tmp/pytest-of-travis/pytest-0/test_uninstall_namespace_packa0/workspace/tmp/pip-install-4ayg6_0p/pd.requires/pip-dist-info',), **{})
    with open('/tmp/pytest-of-travis/pytest-0/test_uninstall_namespace_packa0/workspace/tmp/tmpmmbnxxmx', 'w') as fp:
        fp.write(repr(res))
    ":
    running dist_info
    writing src/pd.requires.egg-info/PKG-INFO
    writing dependency_links to src/pd.requires.egg-info/dependency_links.txt
    writing entry points to src/pd.requires.egg-info/entry_points.txt
    writing namespace_packages to src/pd.requires.egg-info/namespace_packages.txt
    writing requirements to src/pd.requires.egg-info/requires.txt
    writing top-level names to src/pd.requires.egg-info/top_level.txt
    reading manifest file 'src/pd.requires.egg-info/SOURCES.txt'
    writing manifest file 'src/pd.requires.egg-info/SOURCES.txt'
    Traceback (most recent call last):
      File "<string>", line 10, in <module>
[...]

PR #1223, opened 2017-12-01
Build meta: read dist_info directly from command

xoviat:
This is an effort to improve the reliability of build_meta in situations where egg_base contains more than one directory. It should not be merged until #1221 is resolved.


xoviat:
Actually, it's not clear what the advantage is of `sys.argv[1:]`, given that a frontend could theoretically inject problematic arguments and still comply with the standard.

jaraco:
Given that this depends on an abandoned ticket and that the contributor is gone and it was a draft issue, I'm inclined to abandon it. Happy for someone to revive it at any point.

PR #1225, opened 2017-12-04
All extra names must be converted to lower case. use pkg_resources.sa…

GabrielC101:
Fixes  pypa/pip#4617. Supposedly.

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -378,11 +378,15 @@ class Distribution(Distribution_parse_config_files, _Distribution):
         spec_ext_reqs = getattr(self, 'extras_require', None) or {}
         self._tmp_extras_require = defaultdict(list)
         for section, v in spec_ext_reqs.items():
+            try:
+                pkg_resources.invalid_marker(section)
+            except SyntaxError as e:
+                section = pkg_resources.safe_extra(section)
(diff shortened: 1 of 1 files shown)


benoit-pierre:
This patch is mangling the environment markers.

GabrielC101:
I'll have to come back to this later.

GabrielC101:
Note: setuptools makes the extras safe when parsing the Requirement object args.  https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L3006

GabrielC101:
setuptools.pkg_resources.evaluate_marker claims to determine if a string is a valid marker. Yet the string ':python_version=="2.7"' is considered invalid. 

Is evaluate_marker reliable? Ultimately the decision is made by the pyparsing grammer. I'm having difficulty deciphering it.

benoit-pierre:
`':python_version=="2.7"'` is not a valid marker, `'python_version=="2.7"'` is.

GabrielC101:
This commit uses the same process used in _check_extra to separate an extra name from the environment markers. It then uses the safe_extras method to convert the extra name before recreating the section.

This ensures setuptools views extra names as case insensitive.

Issue #127, opened 2013-12-29
UnicodeDecodeError when install in Windows 7


ghost:
Originally reported by: **tkdchen (Bitbucket: [tkdchen](http://bitbucket.org/tkdchen), GitHub: [tkdchen](http://github.com/tkdchen))**

---

Install version 2.0.1 in Windows 7, Python 2.7.6 64bit. This error happened again when installing by running ez_setup.py downloaded from this repo.

```
#!python

Extracting in c:\users\chen\appdata\local\temp\tmprpne9d
Now working in c:\users\chen\appdata\local\temp\tmprpne9d\setuptools-2.0.1
Installing Setuptools
Traceback (most recent call last):
  File "setup.py", line 17, in <module>
    exec(init_file.read(), command_ns)
  File "<string>", line 8, in <module>
  File "c:\users\chen\appdata\local\temp\tmprpne9d\setuptools-2.0.1\setuptools\__init__.py", line 11, in <module>
    from setuptools.extension import Extension
  File "c:\users\chen\appdata\local\temp\tmprpne9d\setuptools-2.0.1\setuptools\extension.py", line 5, in <module>
    from setuptools.dist import _get_unpatched
  File "c:\users\chen\appdata\local\temp\tmprpne9d\setuptools-2.0.1\setuptools\dist.py", line 15, in <module>
    from setuptools.compat import numeric_types, basestring
  File "c:\users\chen\appdata\local\temp\tmprpne9d\setuptools-2.0.1\setuptools\compat.py", line 19, in <module>
    from SimpleHTTPServer import SimpleHTTPRequestHandler
  File "c:\Python27\lib\SimpleHTTPServer.py", line 27, in <module>
    class SimpleHTTPRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
  File "c:\Python27\lib\SimpleHTTPServer.py", line 208, in SimpleHTTPRequestHandler
    mimetypes.init() # try to read system mime.types
  File "c:\Python27\lib\mimetypes.py", line 358, in init
    db.read_windows_registry()
  File "c:\Python27\lib\mimetypes.py", line 258, in read_windows_registry
    for subkeyname in enum_types(hkcr):
  File "c:\Python27\lib\mimetypes.py", line 249, in enum_types
    ctype = ctype.encode(default_encoding) # omit in 3.x!
UnicodeDecodeError: 'ascii' codec can't decode byte 0xb0 in position 1: ordinal not in range(128)
[...]

PR #1231, opened 2017-12-10
Removed warning when PYTHONDONTWRITEBYTECODE is enabled

zsimic:
This is an old warning (from 8 years ago) that isn't useful anymore.
If one disables generation of .pyc files, one constantly gets these warnings, polluting the output.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v38.2.5
+-------
+
+* Removed warning when PYTHONDONTWRITEBYTECODE is enabled
+
 v38.2.4
 -------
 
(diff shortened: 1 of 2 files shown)


jaraco:
> old warning isn't useful anymore.

Can you reference the commit where the change was added and provide some justification for what has changed since it was added until now that changes its usefulness? Or can you provide some other reason why it was never useful?

zsimic:
Yes, the warning was introduced [here](https://github.com/zsimic/setuptools/commit/3a9c591f47b69048b513e5654e5d98efe3f2365a), in order to fix [issue 127](https://github.com/pypa/setuptools/issues/127).

I believe the warning was never useful, as that is the [official way](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE) of turning off `.pyc` generation.
This setuptools warning basically nags users that use this feature with noise on stderr.

PR #1232, opened 2017-12-12
Fix trailing slash handling in pkg_resources.ZipProvider

### LINKED ISSUE #1188: Inconsistent handling of trailing slashes in pkg_resources.ZipProvider.resource_listdir()
Given a ZipProvider, if the underlying ZipImporter `prefix` is empty, then `resource_listdir('')` and `resource_listdir('subdir/')` fail, while `resource_listdir('/')` and `resource_listdir('subdir')` succeed.

On the other hand, if the underlying ZipImporter `prefix` is not empty, then `resource_listdir('/')` fails but `resource_listdir('')` succeeds.

Testcase demonstrating bug:
https://github.com/duggelz/setuptools/commit/0d37f59682905bd69a2674c045017a67cdf9822d

Proposed fix:
https://github.com/duggelz/setuptools/commit/ccd83dbd3bae9e35fe8dbdbdddfdf0ac46965b62

If this seems reasonable, let me know and I will open a pull request.  I have tried to match the style, structure, and naming, let me know if you would like anything modified.

Note that resource_names with '.' or '..', e.g. 'subdir/.' or 'subdir/../mod.py' are not handled by either the original code or my proposed fix, as this seems beyond the desired scope of the API.

duggelz:
Given a ZipProvider, if the underlying ZipImporter prefix is empty,
then resource_listdir('') and resource_listdir('subdir/') fail, while
resource_listdir('/') and resource_listdir('subdir') succeed.

On the other hand, if the underlying ZipImport prefix is not empty,
then resource_listdir('/') fails but resource_listdir('') succeeds.

With this change, the cases listed succeed with or without trailing
slashes.

Fixes #1188

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1693,6 +1693,10 @@ class ZipProvider(EggProvider):
     def _zipinfo_name(self, fspath):
         # Convert a virtual filename (full path to file) into a zipfile subpath
         # usable with the zipimport directory cache for our target archive
+        while fspath.endswith(os.sep):
+            fspath = fspath[:-1]
+        if fspath == self.loader.archive:
+            return ''
         if fspath.startswith(self.zip_pre):
             return fspath[len(self.zip_pre):]
         raise AssertionError(
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1693,8 +1693,7 @@ class ZipProvider(EggProvider):
     def _zipinfo_name(self, fspath):
         # Convert a virtual filename (full path to file) into a zipfile subpath
         # usable with the zipimport directory cache for our target archive
-        while fspath.endswith(os.sep):
-            fspath = fspath[:-1]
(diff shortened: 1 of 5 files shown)


pkg_resources/__init__.py
```diff
@@ -1693,6 +1693,10 @@ def __init__(self, module):
     def _zipinfo_name(self, fspath):
         # Convert a virtual filename (full path to file) into a zipfile subpath
         # usable with the zipimport directory cache for our target archive
+        while fspath.endswith(os.sep):
+            fspath = fspath[:-1]
```
jaraco: Could these two lines be `fspath = fspath.rstrip(os.sep)`?

duggelz: Done

duggelz:
Added changelog entry, bumped version number, and rebased to head.  Thanks.

jaraco:
Excellent. With those changes, I can cut the release from my comm.

Issue #1218, opened 2017-11-29
Better setup_requies


ghost:
I realize that it's not the role of the backend, but since pip is taking their time, it would be nice if setuptools could (immediately when imported):

1. Check to determine whether there is a pyproject.toml in the setup.py directory
2. If there is, parse the requirements
3. Execute the following on importing setuptools

```
Distribution({'setup_requires': setup_requires})
```

xoviat:
/cc @benoit-pierre 

If approved, l think we can have PEP 518 rolled out by tomorrow! It's trivial to implement this way and it wouldn't conflict at all with pip's implementation, since the build requirements would already be installed.

benoit-pierre:
I don't think setuptools should be doing this kind of things on import. And this still would not allow for specifying a version of setuptools to be used that is different from the one currently installed.

benoit-pierre:
Setuptools can't upgrade itself: yes, if the version currently being executed is not the right one, then the correct version will be installed to `.eggs` and activated, but that does not change the fact that the setuptools module as already been imported, and so the wrong version is currently in use.

xoviat:
Okay, but setuptools is not the only build dependency and this would still be a measurable improvement.

benoit-pierre:
Seems like a missed opportunity would have been to use something similar to pkg_resources' `__requires__`:
```diff
 setuptools/__init__.py | 7 +++++++
 1 file changed, 7 insertions(+)

diff --git i/setuptools/__init__.py w/setuptools/__init__.py
index 7da47fbe..a2719dcd 100644
--- i/setuptools/__init__.py
+++ w/setuptools/__init__.py
@@ -178,3 +178,10 @@ def findall(dir=os.curdir):
 
 
 monkey.patch_all()
+
+try:
+    from __main__ import __setup_requires__
+except ImportError:
+    pass
+else:
+    Distribution().fetch_build_eggs(__setup_requires__)
```

And then your example become:
```python
__setup_requires__ = '''
numpy
'''

import setuptools  # <-- numpy installed
import numpy

# Do numpy things

setup()
```

xoviat:
LGTM, but can that be a list?

xoviat:
@jaraco Would you be okay with the proposed patch?

benoit-pierre:
Yes, like `__requires__`, a list would work too.

jaraco:
[...]

PR #1235, opened 2017-12-18
Implement a better setup-requires

xoviat:
Install setup requirements on import of setuptools rather than having to later specify them.

Closes gh-1218.

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -178,3 +178,10 @@ def findall(dir=os.curdir):
 
 
 monkey.patch_all()
+
+try:
+    from __main__ import __setup_requires__
+except ImportError:
+    pass
+else:
+    Distribution().fetch_build_eggs(__setup_requires__)


jaraco:
This change feels a little orphaned (and in a top-level module). I'd like to see it at least in a function that describes its purpose. Is there a reason the functionality can't be found alongside the other behavior for setup_requires?

xoviat:
The function is `fetch_build_eggs`. It needs to be in a top-level module so that it's run on importing setuptools.

xoviat:
I guess it could be put in one of the other modules imported in `__init__`. Which one would you prefer?

benoit-pierre:
I believe this won't work when installing with pip from the same reasons `__requires__` does not work: https://github.com/pypa/pip/issues/4648.

This is what I've been resorting to on another project for handling setup requirements: https://github.com/openstenoproject/plover/pull/881

xoviat:
Nice. If the `pip` folks would have just merged pypa/pip#4799, we wouldn't have these issues.

xoviat:
Sorry, but I just find it completely ironic when Paul complained about not having any PEP 518 support when he refused to merge any PRs.

PR #1237, opened 2017-12-23
Fix distutils path in certain circumstances.

xoviat:
Handle the case where site-packages is named dist-packages


xoviat:
Note: test failure on appveyor was a pre-existing issue not related to this.

Issue #233, opened 2014-07-12
Add support for extra_index_url (config and/or option)


ghost:
Originally reported by: **myaskevich_kyriba (Bitbucket: [myaskevich_kyriba](http://bitbucket.org/myaskevich_kyriba), GitHub: Unknown)**

---

We've got both find-links and index-url config parameters available in user pydistutils config file. But none of them fit in my case.

Why not just mimic the pip behavior and try extra one and fall back if nothing's found there?

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/233

sharksforarms:
Hello, is there any updates on this issue?

Here are the dependencies of my local package `foo`:
```
foo (local repo)
     \
       bar (remote repo)
```
Here is my `~/.pydistutils.cfg`
```
[easy_install]
index-url = http://pypi.local
find_links = http://pypi.remote
```
Here is my test case:
`easy-install foo`

I am getting the following output:
```
Searching for foo
Reading http://pypi.remote
Reading http://pypi.local
Downloading http://pypi.local/.../foo.whl
Best match: foo 0.0.0
Processing ....
Installing ...
...
Processing dependencies for foo
Searching for bar
Reading http://pypi.local/.../simple/bar/
Couldn't find index page for 'bar' (maybe misspelled?)
No local packages or working download links found for 'bar'
```

I would like to be able to specify multiple index-urls, as you can with pip.

pganssle:
I think it's not unreasonable to do this, but I think it depends on how much longer we'll have easy_install around. The use case for this is a lot less compelling with PEP 518.

abravalheri:
Hello, I suppose we can close this issue now that `easy_install` has been deprecated and widely replace by other installation means, such as `pypa/pip`.

If anyone would like to reopen this issue, please feel free to write a new comment with more information or other use cases that we might be missing 😄.

PR #1253, opened 2018-01-22
extra-index-urls #233

sharksforarms:
First time looking at code base, let me know if it needs work!

For issue #233

### DIFF
--- a/docs/easy_install.txt
+++ b/docs/easy_install.txt
@@ -837,6 +837,16 @@ Command-Line Options
     package, and those download pages will be searched for links to download
     an egg or source distribution.
 
+``--extra-index-urls=URLS, -I URLS`` (New in x.xxx; )
+    Specifies extra base URLs of the Python Package Index. Indexes specified here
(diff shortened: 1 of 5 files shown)


jaraco:

I have two concerns about this change.

The main concern is the change in the interface on `PackageIndex.__init__`. The insertion of a parameter will break any usage that is passing positional arguments. And the fact that `*args` does signal that positional arguments are relevant. That said, I would be a little surprised if PackageIndex is used outside of the internal usage, so maybe it's safe. I would feel more comfortable if we had a documented, supported public interface is for PackageIndex.

Another concern is about how much of the optional multi-index handling bleeds through the implementation, requiring changes to several aspects of the system. In an ideal world, adding support for multiple indexes would be a simple loop over the behavior supporting a single index.

But overall, I'd be comfortable accepting the change if we can have a reasonable confidence that no other packages are relying on the API that's changing. Can you do a search in Github for packages that use package_index and ensure they won't break if we accept this change?

Issue #1257, opened 2018-01-23
scan_module(): ValueError: bad marshal data (unknown type code)


jwilk:
This happens with CPython from git master (e768c86ef442ef89004089a8a34ce5909ffb90f2):
```pytb
Traceback (most recent call last):
  File "setup.py", line 16, in <module>
    install_requires=["appdirs"],
  File ".../python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File ".../python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File ".../python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File ".../python3.7/site-packages/setuptools/command/install.py", line 67, in run
    self.do_egg_install()
  File ".../python3.7/site-packages/setuptools/command/install.py", line 109, in do_egg_install
    self.run_command('bdist_egg')
  File ".../python3.7/distutils/cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File ".../python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File ".../python3.7/site-packages/setuptools/command/bdist_egg.py", line 209, in run
    os.path.join(archive_root, 'EGG-INFO'), self.zip_safe()
  File ".../python3.7/site-packages/setuptools/command/bdist_egg.py", line 245, in zip_safe
    return analyze_egg(self.bdist_dir, self.stubs)
  File ".../python3.7/site-packages/setuptools/command/bdist_egg.py", line 355, in analyze_egg
    safe = scan_module(egg_dir, base, name, stubs) and safe
  File ".../python3.7/site-packages/setuptools/command/bdist_egg.py", line 392, in scan_module
    code = marshal.load(f)
ValueError: bad marshal data (unknown type code)
```
I believe this is because [PEP 552](https://www.python.org/dev/peps/pep-0552/) has been implemented, which adds extra 4 bytes to the .pyc header; but setuptools is not aware of that yet.

benoit-pierre:
See #1263.

jaraco:
Release on v38.4.1 pending...

jenisys:
Same problem occurred w/ Python 3.7.0a4+ on Travis CI (reported as: travis-ci/travis-ci#9181), probably w/o the fix.

smagafurov:
@jaraco Same problem
https://travis-ci.org/aio-libs/aiopg/jobs/341326946

[...]

PR #1263, opened 2018-01-25
fix Python 3.7 support

benoit-pierre:
- update scanning code to handle pyc header change
- handle change to `Exception.__repr__` output


benoit-pierre:
Fix #1257.

Issue #1274, opened 2018-02-10
pkg_resources use system python version instead of py_version for requires()


ignatenkobrain:
```
⋊> ~ python3 ./t.py /usr/lib/python2.7/site-packages/hypothesis-3.44.24-py2.7.egg-info                                                                                                                     22:10:41
PyVersion: 2.7
[Requirement.parse('attrs>=16.0.0'), Requirement.parse('coverage')]
⋊> ~ python2 ./t.py /usr/lib/python2.7/site-packages/hypothesis-3.44.24-py2.7.egg-info                                                                                                                     22:10:46
PyVersion: 2.7
[Requirement.parse('attrs>=16.0.0'), Requirement.parse('coverage'), Requirement.parse('enum34')]
```

So `pkg_resources.Distribution` found that this egg belongs to Python 2.7, but if script was run with Python 3 it doesn't find that enum34 is dependency.

Testing code:
```python
#!/usr/bin/python3

import os
import sys

import pkg_resources

def main():
    path = sys.argv[1]

    dist_name = os.path.basename(path)
    if os.path.isdir(path):
        path_item = os.path.dirname(path)
        metadata = pkg_resources.PathMetadata(path_item, path)
    else:
        path_item = path
        metadata = pkg_resources.FileMetadata(path)
    dist = pkg_resources.Distribution.from_location(path_item, dist_name, metadata)
    print("PyVersion: {}".format(dist.py_version))
    print(dist.requires())

if __name__ == "__main__":
    main()
```

```
⋊> ~ grep "enum34" -B1 /usr/lib/python2.7/site-packages/hypothesis-3.44.24-py2.7.egg-info/requires.txt                                                                                                     22:13:13
[:python_version == "2.7"]
enum34
```

Let me know if I can provide more details.

ignatenkobrain:
I think I know how to fix this... Give me a bit of time.

PR #1275, opened 2018-02-10
pkg_resources: pass python_version to marker.evaluate()

ignatenkobrain:
We should be passing more, but unfortunately we don't know full
environment just based on filename.

$ python2 t.py hypothesis-3.44.24-py2.7.egg-info
* Requirement.parse('attrs>=16.0.0')
* Requirement.parse('coverage')
* Requirement.parse('enum34')

$ python3 t.py hypothesis-3.44.24-py2.7.egg-info
* Requirement.parse('attrs>=16.0.0')
* Requirement.parse('coverage')

But if pkg_resources.Distribution already knows python version it is
targeting, we could guide evaluation.

Fixes: https://github.com/pypa/setuptools/issues/1274
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2687,23 +2687,26 @@ class Distribution(object):
         return self.__dep_map
 
     @staticmethod
-    def _filter_extras(dm):
+    def _filter_extras(dm, evaluate_markers=True):
         """
         Given a mapping of extras to dependencies, strip off
-        environment markers and filter out any dependencies
-        not matching the markers.
+        environment and filter out any dependencies
+        not matching the markers unless asked not to.
         """
         for extra in list(filter(None, dm)):
             new_extra = extra
             reqs = dm.pop(extra)
-            new_extra, _, marker = extra.partition(':')
-            fails_marker = marker and (
-                invalid_marker(marker)
-                or not evaluate_marker(marker)
-            )
-            if fails_marker:
-                reqs = []
+            new_extra, sep, marker = extra.partition(':')
             new_extra = safe_extra(new_extra) or None
+            if evaluate_markers:
+                fails_marker = marker and (
+                    invalid_marker(marker)
+                    or not evaluate_marker(marker)
+                )
+                if fails_marker:
+                    reqs = []
+            elif marker:
+                new_extra = "{}{}{}".format(new_extra or "", sep, marker)
 
             dm.setdefault(new_extra, []).extend(reqs)
         return dm


ignatenkobrain:
I was thinking to change evaluate_marker(), but then realized that it is kinda public API....

ignatenkobrain:
So how do I get python2 requires if I use python3 to run tool which uses pkg_resources?

benoit-pierre:
I'd say you can't.

jaraco:
What’s the use case?

benoit-pierre:
Looking at the linked issue above it's for finding a python package dependencies in RPM's pythondistdeps.

jaraco:
After reviewing this in depth, I do feel like the reported defect is a legitimate defect.

If pkg_resources is only meant to open/manage/inspect packages that were installed/intended for the current Python, then it should probably raise an exception if one tries to do what's described (i.e. create a PathMetadata for a Python 2.7 (or 3.5) package on Python 3.6). If we went that route, then one would be forced to use the relevant Python environment to inspect its packages.

I'm not aware of any formal support for this cross-platform or cross-Python functionality (do correct me if I'm mistaken), but that doesn't mean it can't gain that functionality. If setuptools wishes to support this behavior, it should do so formally, with tests and documentation (at least in code).

I also feel there are other environment considerations (such as platform, which is also tracked by the metadata and guarded by extras). If it has cross-python support, it should probably have cross-platform support. And what about other environment markers for which there isn't sufficient package metadata to detect the environment?

Would you consider expanding this PR to address these concerns (documenting the intention, testing the expectation, and consider what should happen for other markers)?

In the meantime, having just spent some time in this code, I wish to refactor it in a way that will conflict with this PR, but in a way that would allow for changes like this proposal to be easier to implement at a less complicating scope.

jaraco:
Another interpretation of what this code should do is not to filter based on environment markers at all. In fact, looking at the code, I think the current filtering based on markers should probably happen in `requires` or between `_dep_map` and `requires`. With that kind of refactoring, it would enable a consumer like t.py to retrieve the raw dependency map and use its own logic for evaluating the markers.

ignatenkobrain:
> And what about other environment markers for which there isn't sufficient package metadata to detect the environment?

We just ignore them (aka keep current behavior). Because I don't think there is way how to find those things.

> Would you consider expanding this PR to address these concerns (documenting the intention, testing the expectation, and consider what should happen for other markers)?

Sure!

> With that kind of refactoring, it would enable a consumer like t.py to retrieve the raw dependency map and use its own logic for evaluating the markers.

This definitely makes a lot of sense and this probably would be best thing to do for us in RPM.

ignatenkobrain:
I pushed first commit which added ability to not evaluate markers, will work on rest soon.

arekm:
How this ended in redhat python deps evaluation script since no progress on this pull request. Was this handled differently?

jaraco:
Thanks for the effort on this, but I think ultimately, pkg_resources won't implement this functionality. Instead, consumers who wish to perform specialized dependency management scenarios should rely on [packaging](https://pypi.org/project/packaging) and [importlib.metadata](https://docs.python.org/3.8/library/importlib.metadata.html) to inspect environments and packages.

PR #1279, opened 2018-02-20
enable pip cache in appveyor build

thijstriemstra:


jaraco:
This change seems reasonable, even if a little overkill for just tox and its dependencies. Can you say a little but about what motivated this change and why it’s a good idea?

thijstriemstra:
@jaraco I think it's waste of bandwidth and pypi's resources to re-download **any** packages during CI builds, hence this pull request.

jaraco:
By that reasoning, all projects should implement this change. I wonder if that view then suggests AppVeyor should themselves set this value by default, rather than going to each project to inject this setting. For now, we'll accept the change here.

thijstriemstra:
> By that reasoning, all projects should implement this change.

Agreed. I opened many cache-related pull requests for my most used Python projects:

screen shot 2018-02-20 at 14 46 08

Thanks for merging.

PR #1286, opened 2018-03-06
Updates for Metadata 2.1 (PEP 566)

di:
This PR makes the necessary updates for Metadata 2.1 ([PEP 566](https://www.python.org/dev/peps/pep-0566/)):

* Handles the new `Description-Content-Type` and `Provides-Extra` fields
  * The former was already added in #1075, I made a slight tweak from #1207 to only set this field if it is present
* Correctly sets the `Version` field to `2.1` if either of the new fields are present

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+v38.5.3
+-------
+
+* Support Metadata 2.1 (`PEP 566 <https://www.python.org/dev/peps/pep-0566/>`_)
+
 v38.5.2
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
-v38.5.3
+v38.6.0
 -------
 
-* Support Metadata 2.1 (`PEP 566 <https://www.python.org/dev/peps/pep-0566/>`_)
+* #1286: Add support for Metadata 2.1 (PEP 566).
 
 v38.5.2
 -------
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -46,6 +46,8 @@ def write_pkg_file(self, file):
     # Setuptools specific for PEP 345
     if hasattr(self, 'python_requires') or self.project_urls:
         version = '1.2'
+    if self.long_description_content_type or self.provides_extras:
+        version = '2.1'
 
     file.write('Metadata-Version: %s\n' % version)
     file.write('Name: %s\n' % self.get_name())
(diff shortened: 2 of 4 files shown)


di:
Hi @jaraco, anything I can do to help get this merged? Since [`pkginfo 1.4.2`](https://pypi.org/project/pkginfo/1.4.2/) was released today, and since it only provides these fields if the `Metadata-Version` is `2.1`, this PR is soon going to become a blocker for anyone wanting to use them. Thanks!

jaraco:
These changes look to be backwards-compatible. Can you add an entry to the CHANGES.rst with an 0.1 bump from the last? Then we can pull that and cut a release.

di:
> The warning /usr/local/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'description_content_type' comes from readme_renderer, right? I presume it does because:

It comes from `distutils`: https://github.com/python/cpython/blob/8837dd092fe5ad5184889104e8036811ed839f98/Lib/distutils/dist.py#L273

It can be safely ignored.

> The wheel building step still makes a DESCRIPTION.rst (Forms990_analysis-1.0.5.dev1.dist-info/DESCRIPTION.rst). This feels suboptimal to me given that, when I inspect the DESCRIPTION.rst here, it's my Markdown README.

This is a good catch, but I'm pretty sure this happens in `wheel`, not `setuptools`.

di:
Yeah, here's the relevant lines in `wheel`: https://github.com/pypa/wheel/blob/master/wheel/bdist_wheel.py#L428-L436

I'm not really sure what that's used for since the description also exists as the message body in `/METADATA` as well, and IIRC that's where `twine` gets it from.

di:
@jaraco Added a changelog entry, thanks!

jaraco:
Releasing now as 38.6.0.

di:
Oops, misread your request for a version bump. Thanks!

wimglenn:
This caused a regression for me.  PR incoming.

sayanarijit:
Hi

I notices it loads normal markdown as usual. But I believe lots of markdowns are in GFM (GitHub flavored) format. For e.g. it supports
```
print('Hello')

# ```
# print('Hello')
# ```
```
But not
```python
print('Hello')

# ```python
# print('Hello')
# ```
```
We can think about code highlighting and all the advanced features later but for now I think just adding the support to render it would be nice.

brainwane:
@sayanarijit Thank you for the comment! Several people have asked about Warehouse supporting GitHub-flavored Markdown -- for that let's talk in pypa/packaging-problems#126, as this is a request that spans a few different codebases.

Issue #1292, opened 2018-03-16
Spurious "Provides-Extra:" generated in metadata


wimglenn:
In the recently released version of setuptools, when installing my distribution [copyingmock](https://pypi.python.org/packages/e7/4b/857c745721ed3d5dc33c426d6e31aaff13a5cd5caaf444117139b2d9936f/copyingmock-0.2.tar.gz#md5=9aa6ba13542d25e527fe358d53cdaf3b), it claims that it provides the `extras=['']`.  Of course, I don't provide the empty-string extra, and the correct metadata should be `extras=[]`.  This seems to be related to [using the conditional deps feature](https://github.com/wimglenn/copyingmock/blob/c80ec21e80a733138b99eed6020563ceae8c20ef/setup.py#L26), somehow:

    install_requires=["mock ; python_version<'3.3'"],

To reproduce the issue, you can get the sdist:

    $ pip download copyingmock==0.2 --no-binary copyingmock

Then use `git bisect` with this script:

```
#!/usr/bin/env python
# check.py
import json
import subprocess
import sys
from zipfile import ZipFile

if __name__ == '__main__':
    subprocess.check_call(['pip', 'wheel', '--no-deps', '--no-index', './copyingmock-0.2.tar.gz'])
    whl = ZipFile('copyingmock-0.2-py2.py3-none-any.whl')
    metadata = json.loads(whl.read('copyingmock-0.2.dist-info/metadata.json'))
    print(metadata['extras'])

    if metadata['extras']:
        sys.exit(1)
    else:
        sys.exit(0)
```

I did bisect and found 

    $ git bisect start
    $ git bisect bad tags/v38.6.0
    $ git bisect good tags/v28.8.0
    Bisecting: 375 revisions left to test after this (roughly 9 steps)
    [2cd7f22eb62b413d65fe4208d66f7e2738f71a56] Correctly honor exist_ok. Ref #1083.
    $ git bisect run ./check.py
    ...

https://github.com/pypa/setuptools/commit/b2ea3c4a20d008622caec445f5b6916ddd420d16 is the first bad commit.

PR #1293, opened 2018-03-16
Don't claim to provide the extra ''

wimglenn:
Resolves https://github.com/pypa/setuptools/issues/1292

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -89,7 +89,8 @@ def write_pkg_file(self, file):
         )
     if self.provides_extras:
         for extra in self.provides_extras:
-            file.write('Provides-Extra: %s\n' % extra)
+            if extra:
+                file.write('Provides-Extra: %s\n' % extra)
 
 
 # from Python 3.4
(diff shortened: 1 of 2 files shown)


di:
@wimglenn Sorry #1286 broke things for you. I think you probably want to move the check farther up and just never add the "empty" extra to `self.metadata.provides_extras`:

```diff
                 # Since this gets called multiple times at points where the
                 # keys have become 'converted' extras, ensure that we are only
                 # truly adding extras we haven't seen before here.
-                self.metadata.provides_extras.add(extra.split(':')[0])
+                extra = extra.split(':')[0]
+                if extra:
+                    self.metadata.provides_extras.add(extra)

         self._convert_extras_requirements()
         self._move_install_requirements_markers()
```

wimglenn:
@di  Thanks, I suspected it could have been handled earlier but had trouble to find the source and wasn't sure if this was distutils fault somehow. I'm not too familiar with setuptools and this code is a little hard to follow with the module level "self" function and the monkeypatches. I've just edited the PR in the online github UI, hope that works OK.

PR #1294, opened 2018-03-16
Add support for maintainer in PKG-INFO

### LINKED ISSUE #1288: maintainer= generates AUTHOR: line in PKG-INFO
This seems like a deliberate choice, but it seems that at the moment the `Author:` metadata is pulled preferentially from `maintainer=`, and `Maintainer:` is never actually generated.

The implementation is [here](https://github.com/pypa/setuptools/blob/d10a1bb476db25dc888b5b4aad9d81ffa245ea30/setuptools/dist.py#L55), and the implementation of `get_contact` that prioritizes `maintainer` is [here](https://github.com/python/cpython/blob/8837dd092fe5ad5184889104e8036811ed839f98/Lib/distutils/dist.py#L1119).

I suspect that this is an artifact of the fact that [PEP 314](https://www.python.org/dev/peps/pep-0314/#author-optional) does not provide a `Maintainer` field, and that was a later addition in [PEP 345](https://www.python.org/dev/peps/pep-0345/#maintainer-email-optional)?

I can make a PR for this, but there is an implementation question. The options I see right now are:

1. Maintain the current behavior and simply *add* a "Maintainer" field. This is the most "backwards compatible" way to do things.
2. Change the behavior so that `author=` is given preference over `maintainer=` when generating the `AUTHOR` field, but still use `maintainer` if no `author` is specified. Additionally always add the `Maintainer` field if `maintainer=` is present.
3. Change the behavior so that *only* `author=` generates the `Author` field and *only* `maintainer=` generates the `Maintainer:` field. This is the least backwards-incompatible change.

I suspect that the choice of these probably should come down to how PyPI and Warehouse handle the metadata. If both PyPI and Warehouse are able to handle an either/or Author/Maintainer field, then I suggest that the best option is 3. Basically the fallback logic should be on the metadata consumer side, not on the metadata *producer* side.

pganssle:
Fixes #1288.

I have not really gone through this with a fine-toothed comb to see if it meets all of the specifications of PEP 345, but this is at least a step in the right direction.

### DIFF
--- a/setuptools/tests/test_dist.py
+++ b/setuptools/tests/test_dist.py
@@ -1,10 +1,13 @@
+# -*- coding: utf-8 -*-
 from setuptools import Distribution
 from setuptools.extern.six.moves.urllib.request import pathname2url
 from setuptools.extern.six.moves.urllib_parse import urljoin
+from setuptools.extern.six import StringIO
 
 from .textwrap import DALS
 from .test_easy_install import make_nspkg_sdist
 
+import pytest
 
 def test_dist_fetch_build_egg(tmpdir):
     """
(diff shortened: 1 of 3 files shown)


setuptools/tests/test_dist.py
```diff
@@ -45,3 +48,91 @@ def sdist_with_index(distname, version):
             for r in reqs
         ]
     assert [dist.key for dist in resolved_dists if dist] == reqs
+
+
+
+def roundtrip_tofile(attr_dict):
```
pganssle: Oops! This is an artifact.

fungi:
Looks like you dropped checking for project_urls when moving the sieve from write_pkg_file() to get_metadata_version(), where it was previously setting a minimum metadata version of 1.2 per the Project-URL entry for PEP 345. Was that intentional, or an oversight? I'm just trying to figure out whether I should push a PR to readd the check for project_urls or whether it wasn't really necessary in the first place.

pganssle:
@fungi I believe the reason that I dropped checking for `project_urls` because [PEP 345](https://www.python.org/dev/peps/pep-0345/#requires-python) specifies that both `Requires-Python` and `Project-URL` are required, and so anything using version >= 1.2 *must* have `Requires-Python`.

That said, this is a heuristic based on metadata generated mostly from `setup()`, so probably the original checking for "if you request *any* features of 1.2, we should create a valid version 1.2+ file".

So it was intentional but probably wrong. Feel free to make a PR to add it back.

fungi:
Thanks, will do. For the record, I found this specifically because I noticed after a new setuptools release I was suddenly ending up with PKG-INFO files claiming to be metadata version 1.1 compliant even though they included Project-URL entries.

PR #1295, opened 2018-03-17
config: Allow `file:` directive with `version`

AntyMew:
Reading version number directly from a plaintext file, e.g. VERSION, is a language-independent alternative to importing a module attribute. For example, it is very easily readable in shell scripts, batch scripts, makefiles, powershell, nix packages, you name it.

This change allows the `version` option to use either the `attr:` directive or the `file:` directive. As far as I know this would make it the first option to accept multiple different directives, so I'm unsure if this implementation is the best option. The potential alternative I see is accepting nested directives... but that seems to me like it'd be overcomplicating things.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2421,7 +2421,7 @@ Metadata
 Key                             Aliases            Type
 ==============================  =================  =====
 name                                               str
-version                                            attr:, str
+version                                            file:, attr:, str
 url                             home-page          str
 download_url                    download-url       str
 project_urls                                       dict
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -237,7 +237,7 @@ class ConfigHandler(object):
         return value in ('1', 'true', 'yes')
 
     @classmethod
-    def _parse_file(cls, value):
+    def _parse_file(cls, value, separator='\n'):
         """Represents value as a string, allowing including text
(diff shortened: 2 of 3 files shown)


AntyMew:
...whoops, I couldn't get the tests working on Linux. Well, serves me right for being lazy. Will fix those build failures.

AntyMew:
Oh wow, my bad, I thought I installed my build properly on top of my setuptools install, and somehow all my ad-hoc tests worked. This is embarrassing.

AntyMew:
There. Not taking any chances trying to be clever anymore :U

jaraco:
@antymew: Thanks for the contrib.

Does this feature work if there are newlines in the VERSION file?

@idlesign Do you have any opinion on this change?

AntyMew:
> Does this feature work if there are newlines in the VERSION file?

Ah, no. Fixed, thanks for the catch

I guess that makes the separator unnecessary, too... that change could be removed

idlesign:
Nice feature, I'm ok with it.
Thank you!

jaraco:
Thanks both. I'll revisit merging this until after the codebase has settled around the maintainer metadata.

pganssle:
@jaraco Looks like #1359 is a duplicate of this PR, actually.

pganssle:
Seems like this was fixed in #1359.

PR #1301, opened 2018-03-18
Stop patching write_pkg_info

### LINKED ISSUE #1289: Stop patching write_pkg_info?
According to [this comment](https://github.com/pypa/setuptools/blob/89ef5bb7e4812193d6f593d531bad36191517981/setuptools/monkey.py#L114), `write_pkg_info` is only patched because of a bug in 3.2.2. Now that setuptools has dropped support for Python 3.2, maybe that patch is no longer necessary?

pganssle:
Fixes #1289. Since setuptools no longer supports Python 3.2, this patch is unnecessary.

I am aware that this may be an unnecessary risk considering this patch isn't actively hurting anything, but it's probably worth some cleanup.

I tried to dig around in the git history to figure out which, if any, tests test that this works correctly, but I couldn't find anything. @jaraco Do you know if there are any? If not, should I add them?

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -123,15 +123,6 @@ def write_pkg_file(self, file):
             file.write('Provides-Extra: %s\n' % extra)
 
 
-# from Python 3.4
-def write_pkg_info(self, base_dir):
-    """Write the PKG-INFO file into the release tree.
-    """
-    with open(os.path.join(base_dir, 'PKG-INFO'), 'w',
-              encoding='UTF-8') as pkg_info:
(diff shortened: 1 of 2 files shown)


jaraco:
Good catch. Yes, happy to see it go.

It's quite possible there aren't any tests that capture this functionality. Setuptools has always struggled with test coverage, though it continues to improve bit by bit as new contributions (especially with public interface or behavior expectations) should have tests and tests are often added for regressions. Still, it's a best-effort approach probably with lots of gaps. I'm happy to see them closed, but not at the expense of more pressing improvements. So, feel free to add them but not obliged.

PR #1304, opened 2018-03-21
Set up code coverage

### LINKED ISSUE #1302: setuptools repository does not have code coverage
To help understand how much of the codebase is already covered by the test suite and to, hopefully, help new contributors write tests that cover at least their patches, I think we should add code coverage to the repository, and some sort of service like [CodeCov](https://codecov.io) or [Coveralls](https://coveralls.io/).

Currently my plan is to shamelessly copy the `tox` setup that @gaborbernat set up for [`dateutil`](https://github.com/dateutil/dateutil/blob/master/tox.ini#L23) as a start:

Things to do:
- [ ] Add coverage to tox
- [ ] Set up codecov or coveralls (I'm leaning towards codecov since it's what I know)

pganssle:
Coverage metrics work with tox locally, need to test it out with CI.

Resolves #1302.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -5,3 +5,5 @@ virtualenv>=13.0.0
 pytest-virtualenv>=1.2.7
 pytest>=3.0.2
 wheel
+coverage
+pytest-cov
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -5,5 +5,5 @@ virtualenv>=13.0.0
 pytest-virtualenv>=1.2.7
 pytest>=3.0.2
 wheel
-coverage
-pytest-cov
+coverage>=4.5.1
+pytest-cov>=2.5.1
--- a/tox.ini
+++ b/tox.ini
@@ -33,6 +33,7 @@ commands=codecov --file {toxworkdir}/coverage.xml
 
 [coverage:run]
 source=setuptools
+omit=setuptools/_vendor/*
 
 [coverage:report]
 skip_covered=True
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,28 @@
 
 [testenv]
 deps=-rtests/requirements.txt
+setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR
-commands=py.test {posargs}
+# These are separate so xfail tests don't count towards code coverage
+commands=pytest -m "not xfail" {posargs: "{toxinidir}" --cov-config={toxinidir}/tox.ini --cov=setuptools}
+         pytest -m "xfail"  {posargs: "{toxinidir}"}
 usedevelop=True
+
+
+[testenv:coverage]
+description=Combine coverage data and create reports
+deps=coverage
+skip_install=True
+changedir={toxworkdir}
+setenv=COVERAGE_FILE=.coverage
+commands=coverage erase
+         coverage combine
+         coverage report --rcfile={toxinidir}/tox.ini
+         coverage xml
+
+[coverage:run]
+source=setuptools
+
+[coverage:report]
+skip_covered=True
+show_missing=True
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ commands=codecov --file {toxworkdir}/coverage.xml
 
 [coverage:run]
 source=setuptools
-omit=setuptools/_vendor/*
+omit=*/_vendor/*
 
 [coverage:report]
 skip_covered=True
--- a/tox.ini
+++ b/tox.ini
@@ -4,28 +4,28 @@
 #
 # export TOXENV='py27,py3{3,4,5,6},pypy,pypy3'
 
+[tox]
+envlist=python
+
 [testenv]
 deps=-rtests/requirements.txt
 setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 # TODO: The passed environment variables came from copying other tox.ini files
 # These should probably be individually annotated to explain what needs them.
 passenv=APPDATA HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
-# These are separate so xfail tests don't count towards code coverage
-commands=pytest -m "not xfail" {posargs: "{toxinidir}" --cov-config={toxinidir}/tox.ini --cov=setuptools}
-         pytest -m "xfail"  {posargs: "{toxinidir}"}
+commands=pytest --cov-config={toxinidir}/tox.ini --cov-report= {posargs}
 usedevelop=True
 
 
 [testenv:coverage]
-description=Combine coverage data and create reports
+description=Combine coverage data and create report
 deps=coverage
 skip_install=True
 changedir={toxworkdir}
 setenv=COVERAGE_FILE=.coverage
 commands=coverage erase
          coverage combine
-         coverage report --rcfile={toxinidir}/tox.ini
-         coverage xml
+         coverage {posargs:xml}
 
 [testenv:codecov]
 description=[Only run on CI]: Upload coverage data to codecov
@@ -34,9 +34,8 @@ skip_install=True
 commands=codecov --file {toxworkdir}/coverage.xml
 
 [coverage:run]
-source=setuptools
-omit=*/_vendor/*
-
-[coverage:report]
-skip_covered=True
-show_missing=True
+source=
+	pkg_resources
+	setuptools
+omit=
+	*/_vendor/*
--- a/tox.ini
+++ b/tox.ini
@@ -25,6 +25,12 @@ commands=coverage erase
          coverage report --rcfile={toxinidir}/tox.ini
          coverage xml
 
+[testenv:codecov]
+description=[Only run on CI]: Upload coverage data to codecov
+deps=codecov
+skip_install=True
+commands=codecov --file {toxworkdir}/coverage.xml
+
 [coverage:run]
 source=setuptools
 
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,9 @@
 [testenv]
 deps=-rtests/requirements.txt
 setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
-passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
+# TODO: The passed environment variables came from copying other tox.ini files
+# These should probably be individually annotated to explain what needs them.
+passenv=APPDATA HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
 # These are separate so xfail tests don't count towards code coverage
 commands=pytest -m "not xfail" {posargs: "{toxinidir}" --cov-config={toxinidir}/tox.ini --cov=setuptools}
          pytest -m "xfail"  {posargs: "{toxinidir}"}
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@
 [testenv]
 deps=-rtests/requirements.txt
 setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
-passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR
+passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
 # These are separate so xfail tests don't count towards code coverage
 commands=pytest -m "not xfail" {posargs: "{toxinidir}" --cov-config={toxinidir}/tox.ini --cov=setuptools}
          pytest -m "xfail"  {posargs: "{toxinidir}"}
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,1 @@
+comment: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ python:
 - &latest_py3 3.6
 - nightly
 - pypy
-- pypy3
 env:
     TOXENV=py
 
@@ -23,6 +22,10 @@ jobs:
             - TOXENV=py
   - python: *latest_py2
     env: *cenv
+  - python: pypy3
+    # Running coverage breaks the pypy3 build on Travis only.
+    # Until that's fixed the pypy3 build needs to be run without coverage
+    script: tox -e py -- $TRAVIS_BUILD_DIR
   - stage: deploy (to PyPI for tagged commits)
     if: tag IS present
     python: *latest_py3
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,19 @@ python:
 - nightly
 - pypy
 - pypy3
+env:
+    TOXENV=py
 
 jobs:
   fast_finish: true
   include:
   - python: *latest_py3
-    env: LANG=C
+    env: &cenv
+        global:
+            - LANG=C
+            - TOXENV=py
   - python: *latest_py2
-    env: LANG=C
+    env: *cenv
   - stage: deploy (to PyPI for tagged commits)
     if: tag IS present
     python: *latest_py3
@@ -49,3 +54,6 @@ install:
 - python bootstrap.py
 
 script: tox
+
+after_success:
+    - if [[ $TOXENV == "py" ]]; then tox -e coverage; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,4 +56,5 @@ install:
 script: tox
 
 after_success:
-    - if [[ $TOXENV == "py" ]]; then tox -e coverage; fi
+    - if [[ $TOXENV == "py" ]]; then tox -e coverage,codecov; fi
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,29 +8,20 @@ python:
 - 3.5
 - &latest_py3 3.6
 - nightly
-- pypy
-env:
-    TOXENV=py
 
 jobs:
   fast_finish: true
   include:
   - python: *latest_py3
-    env: &cenv
-        global:
-            - LANG=C
-            - TOXENV=py
+    env: LANG=C
   - python: *latest_py2
-    env: *cenv
-  - python: pypy3
-    # Running coverage breaks the pypy3 build on Travis only.
-    # Until that's fixed the pypy3 build needs to be run without coverage
-    script: tox -e py -- $TRAVIS_BUILD_DIR
+    env: LANG=C
   - stage: deploy (to PyPI for tagged commits)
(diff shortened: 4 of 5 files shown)


pganssle:
@jaraco @benoit-pierre This is good to go. I more or less used the settings I used for `dateutil`.

I think the most controversial part might be that I have excluded `xfail` tests from coverage. The reason I've done this in the past on `dateutil` is that since `xfail`s allow failures, it may mask the fact that we have no *passing* test that hits a given line.

I'm not sure how it deals with conditional xfails, though.

tox.ini
```diff
@@ -1,11 +1,48 @@
 # Note: Run "python bootstrap.py" before running Tox, to generate metadata.
-#
-# To run Tox against all supported Python interpreters, you can set:
-#
-# export TOXENV='py27,py3{3,4,5,6},pypy,pypy3'
+[tox]
+envlist = py27,
```
jaraco: My preference here is to run tox only for the main python version ('python') and leave the rest to be executed by CI runners. Even if we wish to change this expectation, we should do it in a separate pull request.

tox.ini
```diff
@@ -1,11 +1,48 @@
...
+          pypy3,
+          coverage
+skip_missing_interpreters=true
 
 [testenv]
 deps=-rtests/requirements.txt
-passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR
-commands=py.test {posargs}
+setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
+passenv=APPDATA USERPROFILE HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
```
jaraco: I suspect CODECOV_* is for code coverage but the addition of these other variables is more boilerplate. Is there someplace this boilerplate is maintained? I've been thinking of merging /jaraco/skeleton into this project, but have mainly resisted as this project can't use setuptools_scm.

tox.ini
pganssle: OK, I'll revert this part.

In the past I've just had CI manually override the environments and the bare `tox` command just run on all supported environments present on the user's local machine, but that's just because that's how it was set up in the projects I've worked on.

tox.ini
pganssle: Some fraction of these extra variables are necessary for `codecov` to detect that the run was on Travis or Appveyor (I think at *least* `CI` and `APPVEYOR` are needed) - if it doesn't detect that, it requires a user upload token and will fail to upload.

To be honest, I just started copying over the values from [dateutil's `tox.ini`](https://github.com/dateutil/dateutil/blob/master/tox.ini) until things worked. @gaborbernat slaved over the CI configuration getting all that stuff working for pretty much a full day when I ran a sprint in London and he actually knows `tox` in and out, so I was not amazingly confident straying from his example.

I think it might be a good idea to leave these as is for now, possibly with a comment that says, "These have been cargo-culted, we should actually go through and understand why each is necessary". As it is, the code coverage seems to look a bit different when it's *only* in a PR and not on master, so any experimentation might be better done on a PR against a master that already has codecov set up.

benoit-pierre: I don't see `APPDATA`, `HOMEDRIVE`, `HOMEPATH` or `windir` being set in [dateutil's tox.ini](https://github.com/dateutil/dateutil/blob/master/tox.ini#L18).

tox.ini
gaborbernat: I would say it's more a best practice to set the skip missing interpreters but still run the most we can on the user machine, instead of forcing just whatever version the user has as ``python``.

tests/requirements.txt
```diff
@@ -5,3 +5,5 @@ virtualenv>=13.0.0
 pytest-virtualenv>=1.2.7
 pytest>=3.0.2
 wheel
+coverage
```
gaborbernat: shouldn't this be version qualified with a lower bound; I mean the interface of pytest coverage e.g. changed over time; so I don't think this will run with any version

tox.ini
```diff
@@ -6,6 +6,35 @@
...
+# These are separate so xfail tests don't count towards code coverage
+commands=pytest -m "not xfail" -k test_dist {posargs: "{toxinidir}/setuptools/tests" --cov-config={toxinidir}/tox.ini --cov=setuptools}
```
benoit-pierre: Is only testing `test_dist` intended?

benoit-pierre: `pkg_resources` is also not tested anymore.

benoit-pierre:
I'm not sure I agree with filtering out xfailed tests: IMHO the goal of the tests it not to show everything is working but rather to detect when things go wrong... And the point of coverage is to ensure all code paths are tested, an xfailed test fits the bill.

tox.ini
pganssle: @benoit-pierre Those were already in the `setuptools` configuration, so I left them in place.

tox.ini
pganssle: Nope, oops.

pganssle: I did that to make the CI runs faster and meant to rebase delete it. That explains the super low code coverage metrics :P

tox.ini
benoit-pierre: Right! I see `windir` is referenced in `pkg_resources`. Not sure about the other ones, pip's tox.ini does not set any of those.

pganssle:
@benoit-pierre Filtering out xfails was a tough choice for me, but I don't see what the point of covering all code paths if the code paths are only covered down some blind alley that is *allowed to fail*. For example, if you hit lines 30-36, then it fails on line 37 because lines 30 and 36 were producing nonsense for a given input, you haven't *actually* tested that lines 30-36 work as expected, just that in a certain failure mode they seem to be hit.

Ideally there would be a way to just mark those lines as "hit in xfail", so we can have the best of both worlds.

tests/requirements.txt
pganssle: Good idea.

pganssle:
This is an improvement but for some reason it's not omitting `setuptools._vendor`.

pganssle:
Not entirely sure what the problem is. I dug around in coverage's insides a bit and I am starting to suspect that `coverage` ended up picking the version of `setuptools` that was vendored somewhere or another, and that all the filenames are being collapsed down into one, which is why the `omit=` flag was not working. Using a `*` will cover anything under a `_vendor` folder (which we don't want to include anyway), so that solves the problem.

That said, this may indicate that some of the coverage is hitting not something within the test suite but some other version of setuptools that is installed somewhere. I'm not sure how we'd know the difference.

jaraco:
I haven't been following the discussion, but I can give some background on the `_vendor`. It's possible that packages in `_vendor` bypass the code coverage discovery protocols because they're loaded using a custom finder/loader in `pkg_resources.extern`. That is, you `import pkg_resources.extern.six` but get `pkg_resources.vendor.six` or `six` (depending on if the vendored package is present). I would not be surprised if that breaks the code coverage discovery.

pganssle:
@jaraco I don't think that's it - I dumped some logging into `coverage.py` and it seems like the matching all happens at the level of the filename, not the import name.

pganssle:
I'm not entirely sure why the `pypy3` tests are failing, but I suspect it may be related to [slower performance on pypy3](https://bitbucket.org/pypy/pypy/issues/1871/10x-slower-than-cpython-under-coverage) (though that issue seems resolved). Maybe the solution is to drop coverage for `pypy3` until the situation improves?

The weird thing is it works fine on my laptop, even using the same version of pypy3 that Travis is using.

pganssle:
For now I've disabled coverage metrics on pypy3 which fixes the problem. I don't think it's a major problem because the other Python 3.x tests should be hitting all the same lines anyway.

I think coverage is also causing some signfiicant slowdown in the `pypy` build, but not actually *failing*, so I left that in.

@jaraco @benoit-pierre I think this is finally ready to merge if you agree.

pganssle:
Ping @benoit-pierre. I'd like to merge this, but I'm a bit gunshy after you found that glaring error where I was turning off 75% of the tests. Mind giving this a quick once over?

benoit-pierre:
I'm looking into it right now.

benoit-pierre:
OK, so after taking the time to look more into it, I'd like some changes:

```diff
commit d38a935425ce534c0d7ca011bc67e68464a2538d
Author: Benoit Pierre <benoit.pierre@gmail.com>
Date:   Mon Mar 26 18:34:40 2018 +0200

    tmp

diff --git a/.travis.yml b/.travis.yml
index fb659f13..544b8df3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,29 +8,20 @@ python:
 - 3.5
 - &latest_py3 3.6
 - nightly
-- pypy
-env:
-    TOXENV=py
 
 jobs:
   fast_finish: true
   include:
   - python: *latest_py3
-    env: &cenv
-        global:
-            - LANG=C
-            - TOXENV=py
+    env: LANG=C
   - python: *latest_py2
-    env: *cenv
-  - python: pypy3
-    # Running coverage breaks the pypy3 build on Travis only.
-    # Until that's fixed the pypy3 build needs to be run without coverage
-    script: tox -e py -- $TRAVIS_BUILD_DIR
+    env: LANG=C
   - stage: deploy (to PyPI for tagged commits)
     if: tag IS present
     python: *latest_py3
     install: skip
     script: skip
+    after_success: skip
     before_deploy: python bootstrap.py
     deploy:
       provider: pypi
@@ -43,6 +34,13 @@ jobs:
       distributions: release
       skip_cleanup: true
       skip_upload_docs: true
+  # Don't run coverage on pypy.
+  - python: pypy
+    script: tox
+    after_success: skip
+  - python: pypy3
+    script: tox
+    after_success: skip
 
 cache: pip
 
@@ -56,8 +54,6 @@ install:
 # update egg_info based on setup.py in checkout
 - python bootstrap.py
 
-script: tox
-
-after_success:
-    - if [[ $TOXENV == "py" ]]; then tox -e coverage,codecov; fi
+script: tox -- --cov
 
+after_success: env TRAVIS_JOB_NAME="${TRAVIS_PYTHON_VERSION} (LANG=$LANG)" CODECOV_ENV=TRAVIS_JOB_NAME tox -e coverage,codecov
diff --git a/appveyor.yml b/appveyor.yml
index 2b659459..ff7122b4 100644
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,13 @@
 environment:
 
   APPVEYOR: True
+  CODECOV_ENV: APPVEYOR_JOB_NAME
 
   matrix:
-    - PYTHON: "C:\\Python35-x64"
-    - PYTHON: "C:\\Python27-x64"
+    - APPVEYOR_JOB_NAME: "python35-x64"
+      PYTHON: "C:\\Python35-x64"
+    - APPVEYOR_JOB_NAME: "python27-x64"
+      PYTHON: "C:\\Python27-x64"
 
 install:
   # symlink python from a directory with a space
@@ -20,10 +23,9 @@ cache:
 test_script:
   - "python bootstrap.py"
   - "python -m pip install tox"
-  - "tox -e py"
+  - "tox -- --cov"
 
 after_test:
     - tox -e coverage,codecov
 
 version: '{build}'
-
diff --git a/tox.ini b/tox.ini
index 42d01a2b..ee3fa902 100644
--- a/tox.ini
+++ b/tox.ini
@@ -4,28 +4,28 @@
 #
 # export TOXENV='py27,py3{3,4,5,6},pypy,pypy3'
 
+[tox]
+envlist=python
+
 [testenv]
 deps=-rtests/requirements.txt
 setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 # TODO: The passed environment variables came from copying other tox.ini files
 # These should probably be individually annotated to explain what needs them.
 passenv=APPDATA HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_*
-# These are separate so xfail tests don't count towards code coverage
-commands=pytest -m "not xfail" {posargs: "{toxinidir}" --cov-config={toxinidir}/tox.ini --cov=setuptools}
-         pytest -m "xfail"  {posargs: "{toxinidir}"}
+commands=pytest --cov-config={toxinidir}/tox.ini --cov-report= {posargs}
 usedevelop=True
 
 
 [testenv:coverage]
-description=Combine coverage data and create reports
+description=Combine coverage data and create report
 deps=coverage
 skip_install=True
 changedir={toxworkdir}
 setenv=COVERAGE_FILE=.coverage
 commands=coverage erase
          coverage combine
-         coverage report --rcfile={toxinidir}/tox.ini
-         coverage xml
+         coverage {posargs:xml}
 
 [testenv:codecov]
 description=[Only run on CI]: Upload coverage data to codecov
@@ -34,9 +34,9 @@ skip_install=True
 commands=codecov --file {toxworkdir}/coverage.xml
 
 [coverage:run]
-source=setuptools
-omit=*/_vendor/*
-
-[coverage:report]
-skip_covered=True
-show_missing=True
+source=
+	pkg_resources
+	setuptools
+omit=
+	*/_vendor/*
+	*/tests/*
```

Rational:
* using `tox` should work as before (same as `tox -e python`, by adding `envlist=python` to `tox.ini`)
* coverage is not enabled by default (it's slow)
* disable coverage for pypy and pypy3 (way too slow)
* add coverage info for `pkg_resources`, and make sure tests are ignored (fix `coverage:run` section in `tox.ini`)
* make sure all CI jobs are identifiable (using `CODECOV_ENV`/`APPVEYOR_JOB_NAME`/`TRAVIS_JOB_NAME`)
* again, I don't see the point in treating xfailed tests differently, especially when this results in executing the tests suite 2 times: now I have to parse 2 different reports, and what about the interaction with pytest options (like `--last-failed`, or `--last-failed-no-failures`)?
* disable report on terminal: for the CI jobs, that's what codecov is for, and locally, it's much more user-friendly to call `tox -e coverage -- html` and look at the output for example

gaborbernat:
@benoit-pierre I would generally agree, other than the last one. The only reason ``tox -e coverage -- html`` is nicer to view for now is because we have a lot of lines of code untested (so the output gets a bit noisy). And doing that also forces the user to fire up a browser just to see what's happening :thinking:

benoit-pierre:
I guess it's a matter of preference: I have to open a tab, but then I can refresh it between runs to see the changes. Anyway, you can also use `tox -- -k egg_info --cov-report term --cov`, or use another format with the coverage pass or ask for the report `tox -e coverage -- report` . The important point, IMHO, is that, particularly for the CI jobs, showing the coverage data on the "terminal" just add unnecessary noise.

benoit-pierre:
It does not help that the full path to each source file is shown... I've not found a way to change that.

pganssle:
@benoit-pierre I'm basically fine with all that, but I don't really know how to turn that patch into a commit. Can you just push it to my branch?

The big thing I still disagree with is xfail tests, but admittedly it's not a *big* problem (and to some extent it's actually a bit weird in setuptools because many xfail tests actually only fail on a specific platform). I think the best way to articulate how I feel about `xfail` tests is that they *aren't actual tests*. They're more like "TODO" list options because you don't actually test any behavior. The only guarantee you get out of an xfail is that the application *doesn't have the expected behavior*. You're not actually testing that it fails in any specific way, so the nature of the failure could change from version to version. Nothing that is *only* hit by a failing tests is *actually* covered, which is why I exclude them from coverage metrics.

With regards to them being a separate output, I actually like that, because when I run the test suite, I want to see all '......s....' for the covered portion and all 'xxxxxxx' in the un-covered portion.

> I would generally agree, other than the last one. The only reason tox -e coverage -- html is nicer to view for now is because we have a lot of lines of code untested (so the output gets a bit noisy). And doing that also forces the user to fire up a browser just to see what's happening thinking

@gaborbernat This may be true, but in a lot of libraries you have some paths that are only taken on one interpreter or the other, not both, so 100% coverage on each platform is not possible even if you can get 100% combined coverage. As such, you're *always* going to see a mess of lines on any one platform (unless we define some complicated per-platform coverage pragma, which would likely make the *code* messier).

gaborbernat:
@pganssle true, I consider that a limitation of tox. though You should be able to instead define dependencies between your tox tasks, in which case one would make all interpreter version run with coverage but don't output anything and then just specify a post task that puts it together and prints the coverage. Hopefully the next iteration of tox will support such flow.

pganssle:
@benoit-pierre I reformatted your diff a bit to get it into a `git format-patch` format and then applied it, so no need to push to my branch now I guess.

One thing I noticed that I didn't see originally and that I disagree with is that you are now ignoring the `tests` folder. I quite disagree with this choice. The reason I explicitly *include* tests in the test coverage is that every line of every test *should* be hit during the run, and if it's not then part of your test suite *isn't running*.

This has pointed out a bug for me when submitting a PR to CPython - I accidentally shadowed an earlier test in the test suite, making it so the first one never actually ran (oops). It also would have helped me notice the issue with `test_dist.py` if `codecov` were actually showing what lines of code are getting hit and missed (this is either due to a misconfiguration or due to the fact that we have no code coverage on the master branch, I'm not sure). When I went to improve test coverage, I would have seen that most of the tests were not being run at all.

I strongly suggest that we include the test suite in the coverage.

benoit-pierre:
But you would know you overshadowed a test because the test coverage decreased, no? If not then it could be argued said shadowed test is unnecessary. And with tests included the coverage percentage does not answer the question of how much of the pkg_resources/setuptools code is covered since this now includes the testsuite... I did try to [split and group coverage info with flags on codecov](https://github.com/pganssle/setuptools/compare/coverage...benoit-pierre:coverage) but the results were [not as expected](https://codecov.io/gh/benoit-pierre/setuptools/tree/fb805494cc2877874696b80eb49b46012a04e8b6)!

Anyway, note that changing the current working directory can have an impact on whether or not an include/ommit pattern match.

Also, at the very least if tests are included, `setuptools/tests/mod_with_constant.py` should be excluded.

pganssle:
> But you would know you overshadowed a test because the test coverage decreased, no? If not then it could be argued said shadowed test is unnecessary.

In that case no because I either added both tests in the same PR or the tests were hitting the C api which doesn't have any code coverage metrics set up. Still, it's also not true in the general case, consider:

```python
def my_func(x, y):
    return x + y

def test_my_func():
    assert my_func(3, 7) == 10

def test_my_func():
    a = my_func(1, float('nan'))
    assert a != a
```

Say I accidentally copy-pasted or auto-completed the second "def test_my_func", when it should have been `test_my_func_nan`. Both tests are necessary and both induce the same line coverage, but I'd only detect this accidental shadowing if I have coverage on the tests themselves.

> And with tests included the coverage percentage does not answer the question of how much of the pkg_resources/setuptools code is covered since this now includes the testsuite... I did try to split and group coverage info with flags on codecov but the results were not as expected!

You can get those actual numbers from the detailed breakdowns - the bulk number is more or less meaningless anyway, what's important is that the number trends towards 100% over time. If someone adds a test to the test suite, I want to be able to look at the diff, see which lines are covered both in their test and in the library itself, and see that every line of their patch was executed as part of the test suite. Excluding tests hurts that in favor of having the arbitrary metric of "library code coverage" be more "accurate", which I think is not a useful goal.

benoit-pierre:
OK.

pganssle:
Pushed a new version that now covers the tests as well. I think we're in general agreement. @benoit-pierre or @jaraco you can merge when ready (assuming the tests all pass).

Thanks for the help with this all.

PR #1308, opened 2018-03-26
tests: Add regression test for #1081

### LINKED ISSUE #1307: Change in behavior for 'requires.txt' in 36.2.0
It seems `setuptools` v36.2.0 changed the behavior of `{package}.egg-info/requires.txt` generation. I suspect this was introduced by #1081. We saw this in `pbr` as it [broke our unit tests](https://launchpad.net/bugs/1758877). We have since [fixed this](https://review.openstack.org/#/c/556481/), but ideally this behavior shouldn't have changed in a *minor* release.

I don't want to revert that change but I also would like this behavior to remain as consistent as possible (for as long as `requires.txt` files are used, anyway). I can work on a test to ensure this is the case.

stephenfin:
This ensures we don't unknowingly change the behavior of this again.

Signed-off-by: Stephen Finucane <stephen@that.guru>
Fixes #1307

### DIFF
--- a/setuptools/tests/test_egg_info.py
+++ b/setuptools/tests/test_egg_info.py
@@ -245,6 +245,18 @@ class TestEggInfo(object):
         ''',
 
         '''
+        install_requires_ordered
+
+        install_requires=["fake-factory>=1.12.3,!=2.0"]
+
+        [options]
+        install_requires =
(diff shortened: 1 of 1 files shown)


pganssle:
This looks good to me, good to have a test of this behavior.

PR #1319, opened 2018-04-06
Update wheel tests to reflect latest version

### LINKED ISSUE #1318: Test suite failing on wheel tests
[Example here](https://travis-ci.org/pypa/setuptools/jobs/362236950#L767)

I can reproduce this locally with master and jumping back to known good commits, so I think maybe some dependency has changed behavior.

pganssle:
Per @benoit-pierre's detective work on #1318, this should fix the failing test suite.

Fixes #1318.

### DIFF
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -161,11 +161,9 @@ WHEEL_INSTALL_TESTS = (
             '''
             foo-1.0-py{py_version}.egg/
             |-- EGG-INFO/
-            |  |-- DESCRIPTION.rst
             |  |-- PKG-INFO
             |  |-- RECORD
             |  |-- WHEEL
-            |  |-- metadata.json
             |  |-- top_level.txt
             |-- foo/
             |  |-- __init__.py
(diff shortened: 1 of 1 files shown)


pganssle:
I should have known it wouldn't be that easy - this was succeeding on 3.3 before this change so it's failing there now. I'm guessing the version of wheel that dropped these also dropped support for Python 3.3.

benoit-pierre:
As of 2017-09-29, Python 3.3.x reached end-of-life status, maybe it's time to drop support for it?

pganssle:
I'm generally OK with dropping support for 3.3, but I don't think it's necessary just to make this test more convenient to write.

I'm wondering if the best thing to do here is to change these tests such that they only check that the essentials are provided. Per your comment on #1318:

```
entry_points.txt
PKG-INFO
requires.txt
top_level.txt
zip-safe
```

pganssle:
I've switched the tests over to one that checks that *at least* the files listed are provided, not that *exactly* those files are provided. It seems to me that this is more in line with the actual contract we want to test.

pganssle:
@jaraco @benoit-pierre Are you comfortable with this change to the testing philosophy? It does make an important change to the contract that we are testing here.

jaraco:
Sounds right to me.

pganssle:
Cool, merging.

PR #1325, opened 2018-04-12
Bump appveyor Python 3 to 3.6

pganssle:
I'm guessing the test strategy for Appveyor is intended to be "Test on the latest Python 3 and latest Python 2". Python 3.6 is now the latest and probably the most common Python 3 version.


jmbowman:
From this PR's AppVeyor build:

```
setuptools\tests\test_sdist.py ......xxx.
...
XFAIL setuptools/tests/test_sdist.py::TestSdistTest::()::test_read_manifest_skips_non_utf8_filenames
  System does not support latin-1 filenames
XFAIL setuptools/tests/test_sdist.py::TestSdistTest::()::test_sdist_with_utf8_encoded_filename
  System does not support latin-1 filenames
XFAIL setuptools/tests/test_sdist.py::TestSdistTest::()::test_sdist_with_latin1_encoded_filename
  System does not support latin-1 filenames
```

From a recent build using 3.5 on AppVeyor:

```
setuptools\tests\test_sdist.py ..........
```

It looks like this is probably a consequence of [PEP 529](https://www.python.org/dev/peps/pep-0529/), implemented in Python 3.6 and only impacting Windows.

PR #1332, opened 2018-04-18
pep425tags: turn warning into a debug log entry

### LINKED ISSUE #1314: Py_DEBUG and WITH_PYMALLOC warnings [Windows]
Hi,

I'm on Windows 7 x64, running the latest Python 3.6.5 (64-bit), with setuptools 39.0.1. Each time a setuptools internal function is invoked (for example by running zc.buildout's [bootstrap.py](https://github.com/buildout/buildout/blob/master/bootstrap/bootstrap.py)) and somehow calls ``pep425tags.get_abi_tag``, 2 warnings relative to ``PyDEBUG`` and ``WITH_PYMALLOC`` are raised. I don't remember seeing these warnings with previous versions of ``setuptools`` (I had 36.x if I recall correctly)

I've tried to search issues relative to these warnings but the only I could find are from 2016 and do not seem to be relevant.

Here is a minimal command that triggers the warnings:

```
$ python -c "from setuptools.pep425tags import get_abi_tag; get_abi_tag()"
C:\Python\3.6\lib\site-packages\setuptools\pep425tags.py:89: RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
  warn=(impl == 'cp')):
C:\Python\3.6\lib\site-packages\setuptools\pep425tags.py:93: RuntimeWarning: Config variable 'WITH_PYMALLOC' is unset, Python ABI tag may be incorrect
  warn=(impl == 'cp')):
```

I am not sure I fully understand what is going on though, so it may have something to do with my setup. Hopefully this can be reproduced somehow.

tkhyn:
fixes #1314 

See https://github.com/pypa/pip/issues/3383 and https://github.com/pypa/pip/pull/3399/commits/c67e3ba987de53f6cb08fb287d2dbb446e6a5574

### DIFF
--- a/setuptools/pep425tags.py
+++ b/setuptools/pep425tags.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import distutils.util
+from distutils import log
 import platform
 import re
 import sys
@@ -69,8 +70,8 @@ def get_flag(var, fallback, expected=True, warn=True):
     val = get_config_var(var)
     if val is None:
         if warn:
-            warnings.warn("Config variable '{0}' is unset, Python ABI tag may "
-                          "be incorrect".format(var), RuntimeWarning, 2)
+            log.debug("Config variable '%s' is unset, Python ABI tag may "
+                      "be incorrect", var)
         return fallback()
     return val == expected
 


pganssle:
Should we change the `warn` parameter in any way? It doesn't actually raise a warning anymore, nor does it use `log.warn`.

benoit-pierre:
I'd rather keep the changes to a minimum so it's easier to merge-in updates from pip/wheel.

PR #1338, opened 2018-04-25
Prepare submodules for dependency links to Git repositories.

Czxck001:
## Problem

I need to manage dependency between several Python projects hosted in private Git repositories. The orthodox way to do this is writting `dependency_links` in `setup.py` (or something similar purposed in `pip 10.0.0`).

Here is an example, one of my project `rabbit` is dependent on another private project `carrots`.

```Python
from setuptools import setup


DEPENDENCIES = [
    'carrots',
]
DEPENDENCY_LINKS = [
    'git+ssh://git@github.com/Czxck001/carrots@v0.0.1#egg=carrots-0.0.1'
]

setup(
    name='rabbit',
    version='0.0.1',
	...
    install_requires=DEPENDENCIES,
    dependency_links=DEPENDENCY_LINKS,
)

```

However, some of these depended Git repositories (`carrots`) may have submodules. These submodules are necessary to in their build. The problem is, `setuptools` only clone the depended repositories and check it out to user-specified branch. `setuptools` won't prepare the submodules.

In such circumstances, the user of `rabbit` will have to manually install `carrots` (after using `git submodule` to get all submodules prepared). That is to say, `setup.py` in `rabbit` cannot independently install `rabbit`.

I believe this is not a so-personal user case. Using submodules to integrate code is common for Git, especially for the C language. If one want to have a Python wrapper for a C project, using git submodules would be a good choice.

## Solution

I browsed the source code of `setuptools`. I found the git-clone operation is hard-coded in `setuptools/package_index.py`. Here is the method for downloading git repository:

```Python
    def _download_git(self, url, filename):
        filename = filename.split('#', 1)[0]
        url, rev = self._vcs_split_rev_from_url(url, pop_prefix=True)

        self.info("Doing git clone from %s to %s", url, filename)
        os.system("git clone --quiet %s %s" % (url, filename))

        if rev is not None:
            self.info("Checking out %s", rev)
            os.system("(cd %s && git checkout --quiet %s)" % (
                filename,
                rev,
            ))

        return filename
```

To solve the problem, I appended the git command recursively clone submodules after checkout:

```Python
    def _download_git(self, url, filename):
        filename = filename.split('#', 1)[0]
        url, rev = self._vcs_split_rev_from_url(url, pop_prefix=True)

        self.info("Doing git clone from %s to %s", url, filename)
        os.system("git clone --quiet %s %s" % (url, filename))

        if rev is not None:
            self.info("Checking out %s", rev)
            os.system("(cd %s && git checkout --quiet %s)" % (
                filename,
                rev,
            ))

        os.system("(cd %s && git submodule update --init --recursive)" % (
            filename
        ))

        return filename
```

Then it works fine for me.

I tested the code in both MacOS and Linux. I believe it will works for Windows too. 

This fix will not influence the case when user did not have submodules in their Git repository.

### DIFF
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -898,9 +898,10 @@ class PackageIndex(Environment):
                 filename,
                 rev,
             ))
-            os.system("(cd %s && git submodule update --init --recursive)" % (
-                filename
-            ))
+
+        os.system("(cd %s && git submodule update --init --recursive)" % (
+            filename
+        ))
 
         return filename
 
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -898,6 +898,9 @@ class PackageIndex(Environment):
                 filename,
                 rev,
             ))
+            os.system("(cd %s && git submodule update --init --recursive)" % (
+                filename
+            ))
 
         return filename
 
--- a/setuptools/tests/test_packageindex.py
+++ b/setuptools/tests/test_packageindex.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 import sys
 import os
 import distutils.errors
+import mock
 
 from setuptools.extern import six
 from setuptools.extern.six.moves import urllib, http_client
@@ -169,6 +170,28 @@ class TestPackageIndex:
         url, rev = vsrfu('https://example.com/bar@2995')
         assert url == 'https://example.com/bar'
         assert rev == '2995'
+
+    @mock.patch('os.system')
+    def test__download_git(self, mocked_os_system):
+        # start an index server
+        server = IndexServer()
+        server.start()
+        index_url = server.base_url() + 'test__download_git/simple/'
+
+        # download from a git repository (when os.system is mocked)
+        pi = setuptools.package_index.PackageIndex(index_url)
+        pi._download_git('git@github.com:pypa/carrots@favourite', 'tempdir')
+        server.stop()
+
+        from mock import call
+        # assert calls by os.system
+        calls = [
+            call('git clone --quiet git@github.com:pypa/carrots tempdir'),
+            call('(cd tempdir && git checkout --quiet favourite)'),
+            call('(cd tempdir && git submodule update --init --recursive)')
+        ]
+        mocked_os_system.assert_has_calls(calls)
+        assert mocked_os_system.call_count == 3
 
     def test_local_index(self, tmpdir):
         """


ZhaoxiongCuiKolmostar:
Made a small improvement: update submodules even if no custom branch is specified:
```python
    def _download_git(self, url, filename):
        filename = filename.split('#', 1)[0]
        url, rev = self._vcs_split_rev_from_url(url, pop_prefix=True)

        self.info("Doing git clone from %s to %s", url, filename)
        os.system("git clone --quiet %s %s" % (url, filename))

        if rev is not None:
            self.info("Checking out %s", rev)
            os.system("(cd %s && git checkout --quiet %s)" % (
                filename,
                rev,
            ))

        os.system("(cd %s && git submodule update --init --recursive)" % (
            filename
        ))

        return filename
```

pganssle:
Should this really be using `os.system` instead of `subprocess`?

pganssle:
I think this needs tests.

ZhaoxiongCuiKolmostar:
@pganssle I agree with you. Obviously `subprocess` is better to me. I use `os.system` because it is used in the original function. Not for sure if it is to support older Python version (where `subprocess` is not matured enough).
As for the tests, in general I agree this method should be tested. However, before I made the change this method is not yet tested. I guess maybe this simple wrapper of a subprocess call is not that easy and necessary to be tested.

Czxck001:
@pganssle Sorry for confusion. The above @ZhaoxiongCuiKolmostar is another account of myself. I forgot to switch back then.

Czxck001:
@pganssle I agree with you. Obviously `subprocess` is better to me. I use `os.system` because it is used in the old code (before I made the change). Not for sure if it is to be compatible with older Python version (where subprocess is not matured enough).
As for the tests, in general I agree a method should be tested. However, before I made the change this method is not yet tested. I guess maybe this simple wrapper of a subprocess call is not that easy and necessary to be tested.

pganssle:
@Czxck001 Yes, I think for now we should leave it as `os.system`, that was just an initial first comment that I noticed.

With regards to the test, I personally don't feel that comfortable accepting patches for code that has no tests because it means I have no way to "sanity check" that the patch didn't break anything. Obviously that means that some code is easier to change than others, but that's always the case, really.

Czxck001:
@pganssle I added a unittest for `_download_git` method by mocking `os.system` call. The unittest is verified by running
```
python3 -m pytest setuptools/tests/test_packageindex.py
```
and it works.

pganssle:
@Czxck001 There are two major problems with this test:

1. It tests the implementation, not the interface. It's directly calling a private method - it's an implementation detail that that private method even exists, so it's not a good idea to test it - and all it does is test that that implementation actually does exactly what the code says it's going to do. What we really care about (and thus want to test) is that some part of the public interface that deals with `git` repositories does the right thing when confronted with a repo that has submodules.
2. It doesn't actually test that this *works*, just that these git calls are made. If these git commands don't do what we expect them to do, this test will still pass.

I'll have to think this through and look at the code more throughly, but I think the right test is one that creates a dummy repo and a dummy submodule, and tests that `PackageIndex.download` correctly downloads the git repository.

jaraco:
I agree with Paul, the test is just a repeat of the code itself, so validates only the specific implementation and not its intended purpose. It's only a little better than running a hash of the source code and ensuring the hash matches.

On the other hand, it may be too much to ask a contributor to write a dramatically more sophisticated test. @Czxck001 what do you think about Paul's latest suggestion?

Czxck001:
@jaraco I choose to test by mocking `os.system` is because of the following reasons:
1. The original `_download_git` method is totally untested. The exact reason why it is not tested yet remains unknown to me. 
2. My suspect is this lack of test to some extent show that a test of such a method merely call `os.system` is of little meaningness and heavy (consider the lack of decoupling in this module).
3. Another suspect is in canonical software engineering, one shouldn't test private methods. As the method name goes '_download_git`, the prepended single underscore indicates that this method (should) be private. In book of *Pragmatic Unit Testing*, Dave Thomas and Andy Hunt write:
> Most of the time, you should be able to test a class by exercising its public methods. If there is significant functionality that is hidden behind private or protected access, that might be a warning sign that there's another class in there struggling to get out.
4. If you propose that this method really need a test, then my suggestion is by mocking `os.system`. Using mocking to test this function is based on the fact that I trust the correctness of Git commands. So as long as the commands passed to `os.system` is correct, then result would be correct. If we really create a Git repository for the test, this will make the test very heavy. It will also add a dependency of Git to this repo.

Open to discussion.

jaraco:
I think the big questions for me - how do we know that this additional `git` invocation doesn't have any failure consequences? For example, are there git versions in the wild that don't accept these parameters? Are there repo configurations that would behave poorly with these parameters? Are we certain that this command can't crash? Will it produce noisy output (for example, the other commands are invoked with `--quiet`)?

We could develop tests to answer these questions to some extent, but for a real-world integration, we probably have to release it and get feedback.

> Are we certain that this command can't crash?

I just tested `os.system` and while it does return the exit code, it doesn't crash even if the command fails, so this feels like a small risk. Which basically only indicates that noisy output might be the biggest risk, which is something we can address later. I'm inclined to accept this for now and iterate as needed later.

benoit-pierre:
Why not use pip?

jaraco:
That's a really good point, Benoit. @Czxck001 Have you tried using pip for your installation operations? Is there a reason using pip doesn't work around the issue?

Czxck001:
@benoit-pierre I'm really confused about the relationship between `pip install` and `setuptools`. We have been using custom pip repository to workaround this problem for long.

pganssle:
@Czxck001 Generally speaking, `setuptools` is trying to get out of the business of installing things and specialize in becoming a build backend. As such, `python setup.py install` is deprecated.

`pip install` will invoke `setuptools` as a build backend, but it will handle dependency management itself. I think @benoit-pierre may be right that we may want to move this issue to `pip` itself.

Regardless of whether it's here or in `pip`, I think the core of the test should be similar - create a dummy repo and install it from `git`, but you'll probably want to raise an issue on [pip's issue tracker](https://github.com/pypa/pip/issues) to see if this is something they are planning on supporting before putting a lot of effort into it (it's also possible that `pip` already supports this - try coming up with a [minimal reproducing example](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) to make sure it definitely fails with pip).

jaraco:
Thanks Czxck001 for the contribution here, but I think we're going to decline. We're now actively in the process of deprecating `easy_install` and other download/install operations that happen in setuptools, so adding this feature will only encourage the discouraged use and add flux to a codebase that's stable and predictable (if incomplete). For your needs, please install everything using pip and work with the pip team if it's not supporting submodules as your uses require.

PR #1340, opened 2018-04-26
Update all pypi.python.org URLs to pypi.org

jdufresne:
For details on the new PyPI, see the blog post:

https://pythoninsider.blogspot.ca/2018/04/new-pypi-launched-legacy-pypi-shutting.html

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -319,7 +319,7 @@ class easy_install(Command):
                     self.all_site_dirs.append(normalize_path(d))
         if not self.editable:
             self.check_site_dir()
-        self.index_url = self.index_url or "https://pypi.python.org/simple"
+        self.index_url = self.index_url or "https://pypi.org/simple/"
         self.shadow_path = self.all_site_dirs[:]
         for path_item in self.install_dir, normalize_path(self.script_dir):
             if path_item not in self.shadow_path:
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -297,7 +297,7 @@ class PackageIndex(Environment):
     """A distribution index that scans web pages for download URLs"""
 
     def __init__(
-            self, index_url="https://pypi.python.org/simple", hosts=('*',),
+            self, index_url="https://pypi.org/simple/", hosts=('*',),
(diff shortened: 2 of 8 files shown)


setuptools/command/easy_install.py
```diff
@@ -319,7 +319,7 @@ def finalize_options(self):
                     self.all_site_dirs.append(normalize_path(d))
         if not self.editable:
             self.check_site_dir()
-        self.index_url = self.index_url or "https://pypi.python.org/simple"
+        self.index_url = self.index_url or "https://pypi.org/simple"
```
pganssle: @ewdurbin @brainwane CCed for sanity check - is this the right thing to do?

Also, does this mean that the old versions of `setuptools` out there now are going to start breaking at the end of the week?

jdufresne: > Also, does this mean that the old versions of setuptools out there now are going to start breaking at the end of the week?

I don't think so. The linked blog states:

> The address pypi.python.org will continue to redirect to Warehouse.

ewdurbin: I might suggest changing the default to `https://pypi.org/simple/` to save a redirect.

```shell
$ curl -I https://pypi.org/simple
HTTP/2 301 
location: https://pypi.org/simple/
```

But otherwise this is 👍.

The `pypi.python.org` service will continue to operate for the foreseeable future as it is currently:

```shell
$ curl -I https://pypi.python.org/simple
HTTP/2 301 
location: https://pypi.org/simple
```

```shell
$ curl -I https://pypi.python.org/simple/
HTTP/2 301 
location: https://pypi.org/simple/
```

setuptools/package_index.py
```diff
@@ -297,7 +297,7 @@ class PackageIndex(Environment):
     """A distribution index that scans web pages for download URLs"""
 
     def __init__(
-            self, index_url="https://pypi.python.org/simple", hosts=('*',),
+            self, index_url="https://pypi.org/simple", hosts=('*',),
```
ewdurbin: again, might suggest changing to `https://pypi.org/simple/` to save a redirect.

jdufresne:
I have updated all instances of `https://pypi.org/simple` to `https://pypi.org/simple/`.

Issue #1339, opened 2018-04-25
Time for a new release?


jaraco:
I see we have quite a few commits pending in master since the last release.

I'd like to allow the other maintainers to cut the release, and I'm happy to help guide the process. There is a [documented process] which should be up-to-date, though I haven't validated it.

Would @pganssle or @benoit-pierre care to cut the release? Ping me in Gitter or #pypa-dev or here.

pganssle:
I'm kinda busy this week, @benoit-pierre do you want to take this one?

Otherwise I can do it next week if it can wait that long (I'm happy to get the next one).

benoit-pierre:
I'll see if I can do it before next week.

Note:
* it's not possible to run the Python 3.3 tests with tox 3.0 (`ERROR: unknown environment 'py33'`), I have to downgrade my local version to 2.9.1.
* I think we should look into using [towncrier](`https://pypi.org/project/towncrier/`), like the pip project does, for managing `CHANGES.rst` and improve the workflow around it

jaraco:
I’ve not set up town crier before but I’ve used it a couple of times and like it. I’d be delighted if someone implemented it for setuptools.

Also fine to drop support for Py33. Just need to make a backward incompatible release. Best to do that after releasing pending changes, even if they have to go out untested.

pganssle:
I think we can rely on the CI for testing the Python 3.3 version, but I'm more or less fine dropping support as well.

The one thing is that I tend to prefer to have the "we support version X" be descriptive rather than proscriptive - do we have any changes in the wings that are just waiting for us to drop Python 3.3 support? Also, is there anything special we have to worry about if we drop it in `setuptools` since it's a lot harder to pin your `setuptools` version than it is to pin most other package versions?

pganssle:
I set up a separate ticket for tracking the `towncrier` thing - #1341. I think there are a few policy questions about how we would use it to work out, only partially related to the actual configuration.

benoit-pierre:
I released v39.1.0.

PR #1342, opened 2018-04-26
Drop support for EOL Python 3.3

hugovk:
As suggested in https://github.com/pypa/setuptools/issues/1339, as it's EOL and Tox no longer supports it.

Also remove redundant code and update other code.

### DIFF
--- a/setuptools/py31compat.py
+++ b/setuptools/py31compat.py
@@ -1,3 +1,5 @@
+__all__ = []
+
 try:
     # Python >=3.2
     from tempfile import TemporaryDirectory
--- a/setuptools/py31compat.py
+++ b/setuptools/py31compat.py
@@ -1,17 +1,3 @@
-__all__ = ['get_config_vars', 'get_path']
-
-try:
-    # Python 2.7 or >=3.2
-    from sysconfig import get_config_vars, get_path
-except ImportError:
-    from distutils.sysconfig import get_config_vars, get_python_lib
-
-    def get_path(name):
-        if name not in ('platlib', 'purelib'):
-            raise ValueError("Name must be purelib or platlib")
-        return get_python_lib(name == 'platlib')
-
-
 try:
     # Python >=3.2
     from tempfile import TemporaryDirectory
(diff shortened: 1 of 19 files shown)


setuptools/py31compat.py
```diff
@@ -1,17 +1,3 @@
-__all__ = ['get_config_vars', 'get_path']
```
pganssle: Why drop `__all__` here?

hugovk: `get_config_vars` and `get_path` have been removed from py31compat.py because Python 2.7 and 3.4+ can directly `from sysconfig import get_config_vars, get_path`.

pganssle: Shouldn't we change this to `__all__ = []` then?

pganssle:
I think the diff coverage that is failing is not a problem. The changes seem sane in the uncovered portions, and it's just that it's changing things that aren't changed in the first place.

Per #1339, this should be deferred until the next breaking release.

setuptools/py31compat.py
hugovk: I don't think it's necessary, if nothing is exposed.

The others don't. For example: https://github.com/pypa/setuptools/blob/master/setuptools/py36compat.py

But happy to add back if you prefer!

pganssle: I think it's probably better to be explicit, since if `__all__` is not specified, `from py31compat.py import *` will import everything, but if `__all__` is set to `[]` it will import nothing.

I think it also sends a clearer message that this module provides no public members.

benoit-pierre:
It looks like the `wheel` package has [dropped support for it](https://travis-ci.org/pypa/setuptools/jobs/380159997#L650) too.

pganssle:
@jaraco @benoit-pierre Now that `wheel` is starting to break the Travis build, it's probably time to land this.

Do we have any breaking changes already in master? If not, should we cut a minor release now and then land this PR?

pganssle:
It's been a few days since the last release, so I think we can merge this now.

CC: @jmbowman

PR #1346, opened 2018-05-01
Pin pytest-flake8 to 1.0.0

ianw:
pytest-flake8 1.0.1 has a hard dependency on pytest>=3.5 [1], which
does not support python3.3 [2].  Pin it until python3.3 is removed
[3].

[1] https://github.com/tholo/pytest-flake8/commit/25bbd3b42d3aa0962fb736202115dae9e5d2cd7c
[2] https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23
[3] https://github.com/pypa/setuptools/pull/1342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,8 @@ jobs:
 cache: pip
 
 install:
+# ensure we have recent pip/setuptools
+- pip install --upgrade pip setuptools
 # need tox to get started
 - pip install tox 'tox-venv; python_version!="3.3"'
 
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,6 +1,7 @@
 importlib; python_version<"2.7"
 mock
-pytest-flake8; python_version>="2.7"
+pytest-flake8<=1.0.0; python_version>="3.3" and python_version<"3.5"
+pytest-flake8; python_version>="2.7" and python_version!="3.3" and python_version!="3.4"
 virtualenv>=13.0.0
 pytest-virtualenv>=1.2.7
 pytest>=3.0.2


jaraco:
Probably better would be not to pin... allowing latest versions (3.5.0+) to install where viable but on Python 3.3, let the Python-Requires directive screen out 3.5+.

If this project really wishes to take on the burden of additionally restricting versions, then there should be multiple entries, each with environment markers... so that modern Pythons aren’t forced to accept older libs.

jaraco:
I had a typo in my last message (omitted `!` in `!="3.4"`); there may be others.

benoit-pierre:
How about dropping the requirement altogether? Without passing the `--flake8` command line flag to `pytest`, the plugin is not used.

jaraco:
Looks like the dep was added in c7c6522000. It does make sense to drop it if it's not in use.

jaraco:
On the other hand, having the dep installed makes the `--flake8` command-line available. Let's keep it.

benoit-pierre:
IMHO it would be more convenient to have 2 dedicated tox environments, [like pip](https://github.com/pypa/pip/blob/master/tox.ini#L36).

pganssle:
I weakly disagree with keeping this in unless we're actually going to start using `--flake8` in the CI. This is not required for running the tests and anyone who *wants* to use `--flake8` can always run `pip install pytest-flake8`.

I also don't really think we should be complicating the build environment to support old versions of `pip`. This one was easy enough, but soon enough we'll start seeing incompatibilities with transitive dependencies, so we'll have to start actively adding and pinning specific versions of transitive dependencies. `Python-Requires` should be enough.

Issue #1350, opened 2018-05-13
Installing a wheel may fail due to case sensitivity on Windows


wiggin15:
Trying to install a package with wrong letter case (on Windows) causes the installation to fail. For example, "Logbook" is a package that is registered with a capital "L", but on Windows we can try to install its whl file with a lower-case "L". This makes "easy_install" fail, like this:

```
> wget https://files.pythonhosted.org/packages/13/71/3c4903fc0690059b574396e57f3d8bdf2466cd27b0b756bccd56ba7b0110/Logbook-1.3.3-cp27-cp27m-win_amd64.whl
> easy_install logbook-1.3.3-cp27-cp27m-win_amd64.whl
Processing logbook-1.3.3-cp27-cp27m-win_amd64.whl
Installing logbook-1.3.3-cp27-cp27m-win_amd64.whl to c:\cygwin\home\administrator\python\lib\site-packages
Traceback (most recent call last):
  File "c:\cygwin\home\administrator\python\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\cygwin\home\administrator\python\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Cygwin\home\Administrator\python\scripts\easy_install.exe\__main__.py", line 9, in <module>
  File "c:\cygwin\home\administrator\python\lib\site-packages\setuptools\command\easy_install.py", line 2312, in main
    **kw
  File "c:\cygwin\home\administrator\python\lib\site-packages\setuptools\__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "c:\cygwin\home\administrator\python\lib\distutils\core.py", line 151, in setup
    dist.run_commands()
  File "c:\cygwin\home\administrator\python\lib\distutils\dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "c:\cygwin\home\administrator\python\lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "c:\cygwin\home\administrator\python\lib\site-packages\setuptools\command\easy_install.py", line 412, in run
    self.easy_install(spec, not self.no_deps)
  File "c:\cygwin\home\administrator\python\lib\site-packages\setuptools\command\easy_install.py", line 654, in easy_install
    return self.install_item(None, spec, tmpdir, deps, True)
[...]

PR #1351, opened 2018-05-14
 Read wheel metadata from extracted content instead of from zip

wiggin15:
See issue #1350.

package_name is read from the filename, which can be case-insensitive (on Windows). zipfile is case-sensitive, so trying to read the dist-info from it can fail if the filename case does not match the project name case. Extracting the files first allows us to read dist-info from the filesystem, which is also case insensitive on Windows.

When reviewing, please review the two commits separately (there is a lot of indentation change in the second commit).

### DIFF
--- a/setuptools/wheel.py
+++ b/setuptools/wheel.py
@@ -78,84 +78,84 @@ class Wheel(object):
 
     def install_as_egg(self, destination_eggdir):
         '''Install wheel as an egg directory.'''
+        dist_basename = '%s-%s' % (self.project_name, self.version)
+        dist_info = '%s.dist-info' % dist_basename
+        dist_data = '%s.data' % dist_basename
+        # Extract to target directory.
+        os.mkdir(destination_eggdir)
         with zipfile.ZipFile(self.filename) as zf:
-            dist_basename = '%s-%s' % (self.project_name, self.version)
-            dist_info = '%s.dist-info' % dist_basename
-            dist_data = '%s.data' % dist_basename
-            # Extract to target directory.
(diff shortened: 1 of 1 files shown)


benoit-pierre:
What's pip behavior in this case?

wiggin15:
pip manages to install wheel files with wrong letter case successfully (mentioned in the issue, #1350).

wiggin15:
I think that 'pip' goes over all files and looks for the dist-info dir that matches the "canonical name" (PEP 503) https://github.com/pypa/pip/blob/0ae0109901f63b7f133229a9fee41332dd7366b4/src/pip/_internal/wheel.py#L270

pganssle:
@wiggin15 Thanks for taking to time to make the issue and PR. You get an automatic bump in my esteem for mentioning the strategy about reviewing the commits separately, that was very helpful!

Would you mind adding a test so that this does not have a regression in the future?

wiggin15:
Thanks for the reply @pganssle . Actually I'm considering abandoning this PR in favor of an approach more similar to what 'pip' are doing (looking for the ".dist-info" that has the same canonicalized form as the canonicalized name of the package). This also means we won't have to extract the wheel before determining if it is even supported. I uploaded a commit here: https://github.com/Infinidat/setuptools/commit/59e5c8697d1068a257041dffe6dc280e04f85c6a
I also added a test there.
Can you please take the time to review the other branch and tell me what you think? Thanks!

wiggin15:
Abandoned in favor of #1360

PR #1352, opened 2018-05-14
Add tox environment for docs

jmbowman:
Support easy local builds of docs via `tox -e docs`, and update the developer documentation accordingly.


pganssle:
This will be super useful. Thanks @jmbowman!

Issue #1341, opened 2018-04-26
Set up towncrier for changelog management


pganssle:
This is an issue to track discussion of adding `towncrier` to `setuptools`. I wanted to detach this from #1339 since the two are not necessarily related - generally the value of `towncrier` is that the changelog is generated "as you go" so that it's not a bunch of archaeological work for the maintainer to figure out the changes that happened since the last release while preparing a changelog, and that didn't happen for this release anyway, so it shouldn't be a release blocker.

I recently set this up for `python-dateutil` and I'm happy to do it here. Here are the lessons / issues that I took away from my experience:

### Ticket numbering
`towncrier` assumes a specific workflow where there is one or more tracking tickets (issues) for each change that goes in, so every changelog change requires an associated ticket number. It's pretty common for people to just open a PR that both describes their issue *and* fixes it. For `dateutil`, I've always referenced both the github PR and github issue in my changelog entries, so I end up rewriting the history before merging most PRs anyway, and the ticket number can be the PR number or the issue number (the ticket number referenced in the title is not used in my towncrier template).

I think the options are:

- Require that each PR have a corresponding issue which is referenced in the changelog.
- Always use PR number for the changelog - all users would need to push additional commits or rewrite their history after the PR is sent.
- Allow *either* PR number or issue number to be used as the primary ticket number, and users not fixing a reported issue use the PR number.

### Git history policy
I may be more persnickety than most, but I prefer to keep my git history relatively clean, which means that I end up rewriting my history and the history of people submitting PRs pretty frequently. Originally I was splitting the changelog file into its own separate commit, but I have recently decided to always squash the changelog commit into the last commit of the change it describes. That way, if I end up cherry-picking commits into backports, the changelog will come with it "for free". That said, changelog entries tend to be "linear", so you'd probably want the changelog entry to appear in exactly one branch (the current or the backport). Neither strategy really solves this problem.

Options here:

- changelog entry in its own commit: This is easier for most people if they are adding the changelog *after* the PR is made, because it doesn't involve rewriting history and force pushing to the remote repo. On the other hand, it makes `git commit --amend` more of a pain for people who like to keep a clean history, since usually the changelog entry is after the commit you'd want to amend.
- changelog entry in the last commit: Requires rewriting history in the case of PR-only fixes
- No policy - Easiest thing to do, but commit history is messier in some abstract sense.

### Onboarding
[...]

PR #1354, opened 2018-05-14
Add configuration for towncrier

pganssle:
Per #1341, this sets up `towncrier` using this:

```
towncrier --version
```

The current configuration more or less recreates the old changelog format. We can tweak from here if we want to make improvements that take advantage of what `towncrier` can give us.

We now just need documentation on how to use this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,34 @@
+[tool.towncrier]
+    package = "setuptools"
+    package_dir = "setuptools"
+    filename = "CHANGES.rst"
+    directory = "changelog.d"
+    title_format = "v{version}"
+    issue_format = "#{issue}"
+    template = "towncrier_template.rst"
+    underlines = ["-"]
+
+    [[tool.towncrier.type]]
+        directory = "deprecation"
+        name = "Deprecations"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "breaking"
+        name = "Breaking Changes"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "change"
+        name = "Changes"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "doc"
+        name = "Documentation changes"
+        showcontent = true
+
+    [[tool.towncrier.type]]
+        directory = "misc"
+        name = "Misc"
+        showcontent = true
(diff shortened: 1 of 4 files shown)


pganssle:
@jaraco @benoit-pierre Ready for review.

We can add this to the "how to make a release" docs in a subsequent PR.

pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+    package_dir = "setuptools"
+    filename = "CHANGES.rst"
+    directory = "changelog.d"
+    title_format = "v{version}"
+    issue_format = "#{issue}"
+    template = "towncrier_template.rst"
+    underlines = ["-"]
+
+    [[tool.towncrier.type]]
+        directory = "deprecations"
```
jmbowman: Is there a good reason why this is plural while the "directory" entry for Changes is the singular "change"?

towncrier_template.rst
```diff
@@ -0,0 +1,27 @@
+{% for section, _ in sections.items() %}
+{% set underline = underlines[0] %}{% if section %}{{section}}
+{{ underline * section|length }}{% set underline = underlines[1] %}
```
jmbowman: There's only one entry configured for "underlines" in pyproject.toml; should this `set` tag be removed?  `underline` doesn't seem to be referenced after this in the template.

towncrier_template.rst
```diff
@@ -0,0 +1,27 @@
...
+{% endif %}
+{% if sections[section] %}
+{% for category, val in definitions.items() if category in sections[section]%}
+{% if definitions[category]['showcontent'] %}
```
jmbowman: It looks like the category names aren't used in the template; is there any plan to separate the different types of changes in the generated changelog entries?

pyproject.toml
pganssle: No, good catch.

towncrier_template.rst
pganssle: Yeah, I did it this way so that we can eventually start adding section headers if we want.

Maybe I should just pull it out and we'll add it in if and only if we need it.

towncrier_template.rst
pganssle: Possibly, but for now the categories are just used for ordering purposes so that the important stuff is sorted on top.

pganssle:
@jmbowman Made some updates, want to take another look?

kojiromike:
Apparently, there's a "town-crier" and a "towncrier" and I guess they're not the same thing at all. The one linked here appears to be a blockchain interface...

PR #1356, opened 2018-05-14
Small fixes to Docs:Developer Guide

MiriamSexton:
As discussed at the PycCn sprints 2018, day 1. These are small changes to the developer guide section of the docs to make the directions easier to follow for new developers.

Two changes: 
* changed `git checkout` to `git clone` in instructions for initially getting the codebase 
* Added instruction to run `python bootstrap.py` before invoking `tox` for the first time.

I did not update the CHANGES.txt as I was unsure on the protocol there and these are essentially typo fixes. Please let me know if I need to change that.

### DIFF
--- a/docs/developer-guide.txt
+++ b/docs/developer-guide.txt
@@ -86,8 +86,12 @@ from this repository.
 Testing
 -------
 
-The primary tests are run using tox. To run the tests, first make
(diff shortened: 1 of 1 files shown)


jaraco:
Looks great. And thanks for validating the syntax locally before pushing. No need for a changelog update here.

PR #1357, opened 2018-05-14
Fix doc build warnings

jmbowman:
Fixed all the current warnings from documentation builds, and configured tox to fail doc builds if any new warnings are introduced.  The `autosectionlabel` extension had to be disabled because it caused "duplicate label" warnings in `CHANGES.rst` where the same subsections are used in a few different releases.  Only 2-3 of the cross-references seem to have been depending on it anyway, fixed here by adding explicit section labels.

There was one link that no longer seems to correspond to any section in the docs; I just removed that one, let me know if there's something it should actually point to instead.

### DIFF
--- a/changelog.d/1357.doc.rst
+++ b/changelog.d/1357.doc.rst
@@ -1,0 +1,1 @@
+Fixed warnings in documentation builds and started enforcing that the docs build without warnings in tox.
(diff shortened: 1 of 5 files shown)


pganssle:
This is ready to merge when the tests pass.

PR #1360, opened 2018-05-15
Use canonicalize_name to look for .dist-info in wheel files

wiggin15:
wheel files contain a directory that ends with ".dist-info", but its full name may not exactly match the name of the package as written in the wheel filename. e.g. on Windows, the wheel file name may use different letter cases. Using canonicalize_name on both the package name from the wheel filename and the directories in the zip to find a match fixes the problem.
See issue #1350

The general logic was taken from the pip code: https://github.com/pypa/pip/blob/0ae0109901f63b7f133229a9fee41332dd7366b4/src/pip/_internal/wheel.py#L270

### DIFF
--- a/setuptools/wheel.py
+++ b/setuptools/wheel.py
@@ -4,10 +4,12 @@ from distutils.util import get_platform
 import email
 import itertools
 import os
+import posixpath
 import re
 import zipfile
 
 from pkg_resources import Distribution, PathMetadata, parse_version
+from setuptools.extern.packaging.utils import canonicalize_name
 from setuptools.extern.six import PY3
 from setuptools import Distribution as SetuptoolsDistribution
 from setuptools import pep425tags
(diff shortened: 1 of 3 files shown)


setuptools/wheel.py
```diff
@@ -77,14 +79,24 @@ def egg_name(self):
...
 
+    def get_dist_info(self, zf):
+        # find the correct name of the .dist-info dir in the wheel file
+        for member in zf.namelist():
+            dirname = posixpath.dirname(member)
+            if (dirname.endswith('.dist-info') and
+                    canonicalize_name(dirname).startswith(
+                        canonicalize_name(self.project_name))):
+                return dirname
+        raise ValueError("unsupported wheel format. .dist-info not found")
```
pganssle: This line is still not covered. Can you add a test for it?

wiggin15:
I updated the PR with the new test.

pganssle:
@wiggin15 Damn, that's entirely my bad. Do you still have your version of the code? Can you do a force push? I might be able to recover the original test from my git history.

wiggin15:
Updated with force push.

pganssle:
LGTM.

PR #1364, opened 2018-05-15
Feature/issue 1330

### LINKED ISSUE #1330: pkg_resources.Distribution should provide __dir__
pkg_resources.Distribution uses `__getattr__` to proxy attributes to its _provider attribute; it should also implement its own `__dir__` so that (e.g.) tab-completion works correctly.

seawolf42:
## Summary of changes

Add `__dir__()` implementation to `pkg_resources.Distribution()` that includes the attributes in the `_provider` instance variable.

Closes #1330

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2676,6 +2676,10 @@ class Distribution(object):
             )
         )
 
+    if not hasattr(object, '__dir__'):
+        # python 2.7 not supported
+        del __dir__
+
     @classmethod
     def from_filename(cls, filename, metadata=None, **kw):
         return cls.from_location(
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2667,6 +2667,15 @@ class Distribution(object):
             raise AttributeError(attr)
         return getattr(self._provider, attr)
 
+    def __dir__(self):
+        return list(
+            set(super(Distribution, self).__dir__())
+            | set(
+                attr for attr in self._provider.__dir__()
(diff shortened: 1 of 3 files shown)


pkg_resources/__init__.py
```diff
@@ -2667,6 +2667,19 @@ def __getattr__(self, attr):
...
+            set(super(Distribution, self).__dir__())
+            | set(
+                attr for attr in self._provider.__dir__()
+                if not attr.startswith('_')
+            )
+        )
+
+    if not hasattr(object, '__dir__'):
+        # python 2.7 not supported
+        del __dir__
```
jmbowman: Python 2.7 doesn't provide a default implementation of `__dir__`, but it'll use one in `dir()` if present.  I found a pretty thorough [cross-version solution](https://stackoverflow.com/a/32453194) on StackOverflow for making sure a functional base `__dir__` is inherited; you'd probably only need a subset of that implementation to handle this particular case.

jmbowman:
I'm ok with leaving it as is, just looks a little odd to (sometimes) delete a method immediately after it's defined.

pkg_resources/__init__.py
jaraco: I'd be inclined to adopt that implementation for backward compatibility if it were published in six. Then I'd trust that it's hardened and supported. Otherwise, I'd rather not adopt that code for the marginal benefit it would provide. Still, good reference.

jaraco:
Thanks for the feedback.

PR #1365, opened 2018-05-15
Fix attr: with package_dirs

### LINKED ISSUE #1333: package_dir and setup.cfg version attr: <foo> mutually exclusive?
I was messing around with the new ``setup.cfg`` support for metadata, and it looks like there is a poor interaction with ``package_dir`` and the version field support, specifically if you have a ``setup.cfg`` like:

```
[metadata]
version = attr: foobar.__version__

[options]
package_dir =
    = src
```

You'll get an error trying to import foobar. This is because it simply appends the current directory to ``sys.path`` and then attempts to import the above.

jmbowman:
## Summary of changes

Takes the `package_dir` option into account when loading `version` from a module attribute.  This should handle cases where all modules are located under a common parent and where that particular module is in a subdirectory and/or has a different name.  It doesn't attempt to handle cases where `package_dir` contains keys like "foo.bar", but I'm not sure such entries are even valid.

This required adding a dependency on the `options` section parsing in the `metadata` section parsing, in order to get the value of `package_dir`.  This seems pretty harmless since they happen right after each other, but let me know if it poses any problems.

Closes #1333 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -102,13 +102,13 @@ def parse_configuration(
         If False exceptions are propagated as expected.
     :rtype: list
     """
-    meta = ConfigMetadataHandler(
-        distribution.metadata, command_options, ignore_option_errors)
-    meta.parse()
-
     options = ConfigOptionsHandler(
         distribution, command_options, ignore_option_errors)
     options.parse()
+
+    meta = ConfigMetadataHandler(
+        distribution.metadata, command_options, ignore_option_errors, distribution.package_dir)
+    meta.parse()
 
     return meta, options
 
@@ -281,7 +281,7 @@ class ConfigHandler(object):
             return f.read()
 
     @classmethod
-    def _parse_attr(cls, value):
+    def _parse_attr(cls, value, package_dir=None):
         """Represents value as a module attribute.
 
         Examples:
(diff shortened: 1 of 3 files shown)


setuptools/config.py
```diff
@@ -301,7 +301,21 @@ def _parse_attr(cls, value):
         module_name = '.'.join(attrs_path)
         module_name = module_name or '__init__'
 
-        sys.path.insert(0, os.getcwd())
+        parent_path = os.getcwd()
+        if package_dir:
+            if attrs_path[0] in package_dir:
+                # A custom path was specified for the module we want to import
+                custom_path = package_dir[attrs_path[0]]
+                parts = custom_path.rsplit('/', 1)
```
idlesign: There probably should be OS directory separator instead of hardcoded `/`.

jmbowman: The [setuptools documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files) specifies that Unix-style separators should be used for all paths in setup.cfg, in order to facilitate portability (otherwise package builds would fail on platforms using the other separator type).

> Also notice that if you use paths, you must use a forward slash (/) as the path separator, even if you are on Windows. Setuptools automatically converts slashes to appropriate platform-specific separators at build time.

idlesign: Yes, it is so, and the last sentence is what I was concerned with: I've suspected this automatic normalization is done after the parsing stage, but was not sure.

jmbowman: I'm pretty sure it's after; everything up to this point is mostly config file parsing, without much knowledge of what the individual values represent.  Also, AppVeyor builds passed on Windows with the new test cases including forward slashes in `package_dir`.

Issue #1362, opened 2018-05-15
setuptools fails to install wheel files with duplicate entries in .data on Windows


wiggin15:
easy_install fails to install, for example, `selenium`'s wheel file:

```
> wget http://pypi/media/dists/selenium-3.11.0-py2.py3-none-any.whl\#md5\=4644b52a12a4acf352586794ef792bc9
> easy_install selenium-3.11.0-py2.py3-none-any.whl
Processing selenium-3.11.0-py2.py3-none-any.whl
removing 'c:\cygwin\home\administrator\python\lib\site-packages\selenium-3.11.0-py2.7.egg' (and everything under it)
Installing selenium-3.11.0-py2.py3-none-any.whl to c:\cygwin\home\administrator\python\lib\site-packages
error: [Error 183] Cannot create a file when that file already exists
```

This is because selenium contains the following files in the wheel file:

```
selenium/webdriver/firefox/amd64/x_ignore_nofocus.so
selenium-3.11.0.data/data/selenium/webdriver/firefox/amd64/x_ignore_nofocus.so
```
setuptools tries to copy all the files from <package>.data to the base directory (https://github.com/pypa/setuptools/blob/master/setuptools/wheel.py#L146) and `os.renames` (https://github.com/pypa/setuptools/blob/master/setuptools/wheel.py#L38) fails on Windows when the destination already exists.

pganssle:
Thanks @wiggin15. Sounds like this is indeed a bug.

I imagine the hard part of solving this is going to be writing a test for it (as is so often the case).

pganssle:
@benoit-pierre I think I may have misunderstood the issue, I thought that they were files that start in different folders but have the same name, pass through some intermediate stage where they are in the same folder, and then are distributed out to new locations where they have different names.

If this is trying to install two different files in the same location, then I think it rightly *should* be raising an error.

wiggin15:
@pganssle I started discussion about this in https://github.com/pypa/packaging-problems/issues/152 but no action has been taken yet. If we think this case should raise an error then this should be the behavior in "pip" and "wheel" too, but in order to open an issue for these projects we need a decision in the PyPA issue.

PR #1371, opened 2018-05-17
Fix copy of files from .data dir of wheels if the files already exist

wiggin15:
See issue #1362.
From "rename" docs:
```
On Unix, if dst exists and is a file, it will be replaced silently if the user has permission. [...] On Windows, if dst already exists, OSError will be raised even if it is a file
```
To make the behavior the same on both platforms, this PR manually removes existing files before rename.

### DIFF
--- a/changelog.d/1362.change.rst
+++ b/changelog.d/1362.change.rst
@@ -1,0 +1,1 @@
+Fix copy of files from .data dir of wheels if the files already exist
--- a/setuptools/wheel.py
+++ b/setuptools/wheel.py
@@ -35,6 +35,8 @@ def unpack(src_dir, dst_dir):
         for f in filenames:
             src = os.path.join(dirpath, f)
             dst = os.path.join(dst_dir, subdir, f)
+            if os.path.exists(dst):
+                os.remove(dst)
             os.renames(src, dst)
         for n, d in reversed(list(enumerate(dirnames))):
             src = os.path.join(dirpath, d)


benoit-pierre:
I don't agree with this: if the wheel contains duplicate entries, then it's broken. And the failure on Windows is correct and we should instead make sure the installation fails similarly on Linux.

wiggin15:
But `pip` and `wheel` have no trouble installing such wheel files (e.g. selenium's whl). Since setuptools is the only one that can't, it appears to be a problem with setuptools.

wiggin15:
Umm.. pip is an widely used package manager... And even the "wheel" package works:

```
 wget https://files.pythonhosted.org/packages/57/bc/17164fd471ccdf0df3a992c710c0c3c47743462ff41ab72a02c6ede96e90/selenium-3.12.0-py2.py3-none-any.whl
python -c "import wheel.install; wheel.install.WheelFile('selenium-3.12.0-py2.py3-none-any.whl').install()"

So I can't open a ticket to selenium saying their wheel distribution is broken - because it doesn't look like it is.

pganssle:
At the very least I think this behavior should raise a warning on any OS where files are overwritten.

benoit-pierre:
Pip, setuptools, and wheel have different implementations, so a wheel package can be invalid and still installs successfully with one of those. For example some of the older PyQ5 wheels were created with invalid `RECORD` metadata (wrong checksum calculation), but since pip does not check those, it has not trouble installing such a wheel (the wheel implementation does check those, we don't either).

wiggin15:
Aren't setuptools, pip and wheel all part of the Python packaging _alliance_? Wasn't the whole point of the wheel format to be well-defined and well-documented (as opposed to eggs)? If there are differences between the reference implementation, the de facto package manager and other implementations - shouldn't these differences be reconciled? What is the correct process to resolve this?

jaraco:
> shouldn't these differences be reconciled? What is the correct process to resolve this?

Yes; First step, maybe ask @ncoghlan if he has an opinion on the matter. It may be an easy call for him to make. If it's not clear-cut and obvious, then the proper thing to do is raise an issue with [packaging problems](/pypa/packaging-problems), and it will be settled there.

ncoghlan:
I think a packaging-problems issue would be a good option here, as another case where this might come up is with files that differ only by case, which will work on case sensitive file systems, but fail on case insensitive ones.

wiggin15:
Thanks. Opened https://github.com/pypa/packaging-problems/issues/152

jaraco:
I'm going to close this for now, but happy to re-open if that's what's determined to be the best course in the upstream discussion.

PR #1372, opened 2018-05-17
Stop testing Python 3.3 in Travis

jmbowman:
## Summary of changes

The Python 3.3 job in Travis failed in the last build on master because the new `wheel` release won't install on it.  We planned on dropping 3.3 support soon anyway, so just stop testing it on Travis to avoid further job failures.  We'll release one more version that officially supports Python 3.3 while incorporating all the latest fixes from master, and then the next one will fully remove support for it.

### Pull Request Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 language: python
 python:
 - &latest_py2 2.7
-- 3.3
 - 3.4
 - 3.5
 - &latest_py3 3.6
@@ -43,7 +42,7 @@ install:
 # ensure we have recent pip/setuptools
 - pip install --upgrade pip setuptools
 # need tox to get started
-- pip install tox 'tox-venv; python_version!="3.3"'
+- pip install tox tox-venv
 
 # Output the env, to verify behavior
 - env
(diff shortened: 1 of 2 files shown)


pganssle:
This is already done in #1342 and is deferred to the next major release.

If installing `wheel` is breaking Travis, that is a bug in `wheel`, which should be using `Requires-Python`.

pganssle:
Actually, my bad, it looks like `pip` doesn't have a version that supports Python 3.3 and `Requires-Python`.

I think it's time to land #1342.

jmbowman:
I talked to @jaraco about this at the PyCon sprint earlier this morning, and our tentative plan was:

* Merge this PR so we have clean Travis builds again.
* Release 39.2.0 today with all the fixes already in master or ready to go in today (things that already passed Python 3.3 tests before the new `wheel` release came out).
* Merge #1342 immediately after that, so the next release would be 40.0.0 without Python 3.3 support.

Would you like to handle this differently?  We could immediately merge #1342 and go straight to 40.0.0, just thought it would be nice to make the latest fixes available to any remaining Python 3.3 users before doing that.

pganssle:
I think I have a more surgical PR in #1373.

I made it a single commit so it can be reverted as part of #1342.

pganssle:
Eh, #1373 seems like it's not quite working and I don't think it's worth being fiddly about. This approach works fine.

PR #1373, opened 2018-05-17
Fix wheel build in Python 3.3 tests

pganssle:
Until we're ready to land #1342, this will allow the test suite to pass.

Changelog not necessary, this will be essentially removed when #1342 is merged.


pganssle:
Abandoning this idea.

PR #1379, opened 2018-05-19
Make minor fixes to release docs

jmbowman:
## Summary of changes

I made some tweaks to the release process documentation after using it to create version 39.2.0.  I also updated the 39.2.0 changelog to reflect one item which was omitted because the file was named incorrectly (which prompted one of the doc updates).

The towncrier output is slightly inconsistent in its usage of newlines between preview and release modes, and edits to the template don't seem to affect this.  I manually added one newline and removed another for the 39.2.0 release to keep the formatting consistent; the new output review step is partially to catch things like this until we find a way to get towncrier to automatically get the formatting correct.

The `--allow-dirty` option is needed for bump2version to go ahead even though the changelog updates haven't been committed yet.

### Pull Request Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ v39.2.0
   a text file.
 * #1360: Fixed issue with a mismatch between the name of the package and the
   name of the .dist-info file in wheel files
+* #1364: Add `__dir__()` implementation to `pkg_resources.Distribution()` that
+  includes the attributes in the `_provider` instance variable.
 * #1365: Take the package_dir option into account when loading the version from
   a module attribute.
 * #1353: Added coverage badge to README.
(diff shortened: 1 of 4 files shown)


pganssle:
LGTM. Should we cut a `post` release before landing #1342, or just cut any subsequent patch releases from a branch?

jmbowman:
Not sure exactly what you mean by "cut a post release".  My normal inclination would be to wait a few days to make sure no critical bugs get reported, then go ahead and merge #1342.  Although as you say, we could merge sooner and just release any fixes from a branch if necessary.

pganssle:
Since this PR includes changes to the docs but not the actual software, it *may* be appropriate to make a [post release](https://www.python.org/dev/peps/pep-0440/#post-releases) (e.g. `39.2.0.post1`), but I guess it doesn't matter, since https://setuptools.readthedocs.io/ redirects to the latest, not the stable version of the documentation, so I guess it doesn't make sense to make docs-only releases.

PR #1393, opened 2018-06-21
Fix Town Crier Link

kojiromike:
The project "town-crier", with a hyphen, linked originally, is a blockchain integration. I'm certain the correct one is the one without a hyphen, which is for NEWS.


pganssle:
Oops, that's my fault. Good catch @kojiromike!

PR #1398, opened 2018-06-25
Avoid resource warning from dev scripts installed editable

wimglenn:
Avoid resource warnings on dev scripts installed editable.

    ResourceWarning: unclosed file <_io.TextIOWrapper name='whatever' mode='r' encoding='UTF-8'> 

The `with` statement is supported since Python 2.5, so I assume we are OK to use this. Can use try: finally: instead if proj maintainers prefer?


jaraco:
I can't think of any reason not to accept this.

PR #1404, opened 2018-06-28
fix PEP 518 configuration

benoit-pierre:
## Summary of changes

Add `build-system.requires` key to `pyproject.toml`:
- this is necessary with pip 10.0.1, as otherwise the defaults will be to require both `setuptools` and `wheel` (and we only need the later)
- a `pyproject.toml` with no `build-system.requires` key will be invalid and rejected by a future version of pip (see https://github.com/pypa/pip/issues/5416#issuecomment-399638608)

### Pull Request Checklist


pganssle:
LGTM.

Issue #1381, opened 2018-05-23
Include most recent upload metadata when uploading to PyPI


pganssle:
It seems that for the `upload` command, we're delegating entirely to `distutils`, which has `metadata_version` hard-coded as "1.0", and only uploads the 1.0 metadata. I think this is one reason that Requires-Python data isn't showing up on PyPI.

Although `setup.py upload` is basically deprecated, I think we should probably still make these improvements, even if as part of this we also add `DeprecationWarning` telling people to start using `twine`.

di:
> It seems that for the `upload` command, we're delegating entirely to `distutils`, which has `metadata_version` hard-coded as "1.0", and only uploads the 1.0 metadata. I think this is one reason that Requires-Python data isn't showing up on PyPI.

Indeed, this is true.

However, I'm not sure it's worth doing what it would take to support Metadata 2.1 on upload. I think it makes more sense just to deprecate `setup.py upload` entirely and direct folks to use `twine` instead.

pganssle:
I'll take a look at how difficult it would be to do next week. One big problem is that this is creating a lot of bad and immutable data on PyPI and a *lot* of people are just running scripts they ran a long time ago or copying from some old "how to make a release" guide someone made.

If we're going to say that the new upload metadata is just not supported in setuptools, we should probably start throwing an actual *error* rather than a warning.

I think the options are:

1. Decide we don't support this and throw an error whenever >1.0 metadata is detected on upload.
2. Build whatever basic support for it we can and start raising a warning that you should really be using twine.
3. Try shelling out to `twine` and raising a warning if successful ("Hey, use twine directly!") and throwing an error if unsucccessful.

I think 3 is too clever and from the cursory look I made at the code before I created this issue, I think 2 is a feasible path forward.

benoit-pierre:
I'm for 1: we should deprecate and remove support for `upload`, sooner rather than later. Same with a bunch of other stuff.

benoit-pierre:
I think we should keep this open until the upload/register commands are completely removed.

di:
Fair enough, we should probably revise the original issue in that case.

pganssle:
I'm actually planning on fixing this at some point, because it's currently polluting PyPI with bad packages in a very insidious way, as a major driver of [this issue](https://github.com/pypa/warehouse/issues/3889).

[...]

PR #1410, opened 2018-07-08
Deprecate upload and register commands

di:
## Summary of changes

As discussed in https://github.com/pypa/setuptools/issues/1381, this PR deprecates the `upload` and `register` commands by warning the user not to use these commands.

### Pull Request Checklist

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -48,8 +48,6 @@ Feature Highlights:
 * Command aliases - create project-specific, per-user, or site-wide shortcut
   names for commonly used commands and options
 
-* PyPI upload support - upload your source distributions and eggs to PyPI
-
 * Deploy your project in "development mode", such that it's available on
   ``sys.path``, yet can still be edited directly from its source checkout.
 
@@ -107,11 +105,10 @@ As you can see, it doesn't take much to use setuptools in a project.
 Run that script in your project folder, alongside the Python packages
 you have developed.
 
-Invoke that script to produce eggs, upload to
-PyPI, and automatically include all packages in the directory where the
-setup.py lives.  See the `Command Reference`_ section below to see what
-commands you can give to this setup script. For example,
-to produce a source distribution, simply invoke::
+Invoke that script to produce distributions and automatically include all
+packages in the directory where the setup.py lives.  See the `Command
+Reference`_ section below to see what commands you can give to this setup
+script. For example, to produce a source distribution, simply invoke::
 
     python setup.py sdist
 
@@ -138,7 +135,7 @@ dependencies, and perhaps some data files and scripts::
             'hello': ['*.msg'],
         },
 
-        # metadata for upload to PyPI
+        # metadata to display on PyPI
         author="Me",
         author_email="me@example.com",
         description="This is an Example Package",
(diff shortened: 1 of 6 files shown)


setuptools/command/register.py
```diff
@@ -1,10 +1,24 @@
...
+
     def run(self):
-        # Make sure that we are using valid current name/version info
-        self.run_command('egg_info')
-        orig.register.run(self)
+        try:
+            # Make sure that we are using valid current name/version info
+            self.run_command('egg_info')
+            orig.register.run(self)
+            self._deprecation_notice()
```
pganssle: I'm not sure I understand why the try/catch here.

Can we do this?

```python
self._deprecation_notice()
self.run_command('egg_info')
orig.register.run(self)
```
Is it because `run_command` and `register.run` generate a bunch of text, and you want the deprecation warning to be printed *after* this text?

If so, maybe this would be better:

```
try:
    self.run_command('egg_info')
    orig.register.run(self)
except Exception:
    raise
finally:
    self._deprecation_notice()
```

pganssle:
I'm +1 on this initiative, but I do think it would be nice in the documentation for `register` and `upload` if we actually provide the specific 1-to-1 equivalents in addition to just linking to the `twine` documentation. I often find that it's way easier to move over when I don't have to learn a whole new tool just to figure out how to do the equivalent of what I used to do with the old tool.

Also, ideally this would have tests, though I'm not super familiar with the `announce` mechanism used here and how easy it would be to test that. Maybe the best thing to do is to use a mock of some sort to make sure that `announce` is called with some regex matching something like `.*[Rr]egister.*deprecated.*twine.*` (or matching all three patterns `[Rr]egister`, `[Dd]eprecated`, `twine`).

di:
> I'm +1 on this initiative, but I do think it would be nice in the documentation for register and upload if we actually provide the specific 1-to-1 equivalents in addition to just linking to the twine documentation. 

I kinda disagree with this, I don't think it should be `setuptools`'s job to tell folks how to use `twine`.

pganssle:
I don't think it needs to block this particular PR if you don't want to do it, but the problem is that it has to be *someone*'s job, so either `twine` should have a "migrating from `setuptools`" section or `setuptools` should have a "migrating to `twine`" section, and whichever one has it, the other one should link to it.

Yes we can just say, "Go ahead and read the `twine` documentation", but as a practical matter "If you are trying to do X with `setuptools`, do Y with `twine` instead!" will get people switching over *much* faster than just "Use `twine` for this instead!"

pganssle:
I think we can link to [this](https://twine.readthedocs.io/en/latest/#using-twine) in the setuptools documentation for `upload`. There doesn't seem to be an anchor target for `twine register` (which I guess is unnecessary now anyway).

pganssle:
I think the netlify deployment failure was an intermittent failure. I don't seem to have permissions to restart it, though. Close-and-reopen didn't seem to work. @ewdurbin Do you own the netlify account associated here? Do you know if you can give me and the other setuptools maintainers permissions on it?

ewdurbin:
@pganssle @jaraco was invited as an administrator on the project during PyCon sprints. Can you private message me an email address you'd prefer for me to add yourself as well?

di:
I think it would be reasonable for `twine` to have a "migrating from setuptools" guide, let's see what the maintainers there think.

docs/setuptools.txt
```diff
@@ -2251,25 +2226,16 @@ available:
...
-
-    python -m keyring set <repository> <username>
-    Password for '<username>' in '<repository>': ********
-
-Then, in .pypirc, set the repository configuration as normal,
-but omit the password. Thereafter, uploads will use the
-password from the keyring.
-
 New in 20.1: Added keyring support.
+New in 40.0: Deprecated the upload command.
```
di: @pganssle Since there was a `40.0` release last night, we'll need to tweak this line before releasing. I took a guess at what version this would land in, but seeing as it's a minor change, this should probably be `40.1`.

PR #1412, opened 2018-07-09
Solve race during creation of egg cache (fixes #1256)

### LINKED ISSUE #1256: [Errno 17] File exists when extracting files to egg cache (or creating egg cache directory)
We are experiencing intermittent failures when running tests for one of our Python applications while installing dependencies during test execution in our CI system. The tests are being run with pytest, and we are using the pytest-xdist plugin to run tests in multiple processes in parallel to reduce the execution time.

Here is an example stack trace:
```
.eggs/py-1.5.2-py2.7.egg/py/_path/common.py:377: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
.eggs/py-1.5.2-py2.7.egg/py/_path/common.py:429: in gen
    for p in self.gen(subdir):
.eggs/py-1.5.2-py2.7.egg/py/_path/common.py:419: in gen
    if p.check(dir=1) and (rec is None or rec(p))])
.eggs/pytest-3.2.5-py2.7.egg/_pytest/main.py:737: in _recurse
    ihook = self.gethookproxy(path)
.eggs/pytest-3.2.5-py2.7.egg/_pytest/main.py:641: in gethookproxy
    my_conftestmodules = pm._getconftestmodules(fspath)
.eggs/pytest-3.2.5-py2.7.egg/_pytest/config.py:356: in _getconftestmodules
    mod = self._importconftest(conftestpath)
.eggs/pytest-3.2.5-py2.7.egg/_pytest/config.py:381: in _importconftest
    raise ConftestImportFailure(conftestpath, sys.exc_info())
E   ConftestImportFailure: ExtractionError("Can't extract file(s) to egg cache\n\nThe following error occurred while trying to extract file(s)\nto the Python egg cache:\n\n  [Errno 17] File exists: '/home/travis/.cache/Python-Eggs'\n\nThe Python egg cache directory is currently set to:\n\n  /home/travis/.cache/Python-Eggs\n\nPerhaps your account does not have write access to this directory?\nYou can change the cache directory by setting the PYTHON_EGG_CACHE\nenvironment variable to point to an accessible directory.\n",)
E     File "/home/travis/build/finn/fiaas-deploy-daemon/.eggs/pytest-3.2.5-py2.7.egg/_pytest/assertion/rewrite.py", line 212, in load_module
E       py.builtin.exec_(co, mod.__dict__)
E     File "/home/travis/build/finn/fiaas-deploy-daemon/.eggs/py-1.5.2-py2.7.egg/py/_builtin.py", line 221, in exec_
E       exec2(obj, globals, locals)
E     File "<string>", line 7, in exec2
E     File "/home/travis/build/finn/fiaas-deploy-daemon/tests/fiaas_deploy_daemon/conftest.py", line 9, in <module>
E       from fiaas_deploy_daemon.specs.models import AppSpec, ResourceRequirementSpec, ResourcesSpec, PrometheusSpec, \
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/__init__.py", line 13, in <module>
E       from .crd import CustomResourceDefinitionBindings, DisabledCustomResourceDefinitionBindings
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/crd/__init__.py", line 7, in <module>
E       from .watcher import CrdWatcher
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/crd/watcher.py", line 12, in <module>
E       from ..deployer import DeployerEvent
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/deployer/__init__.py", line 8, in <module>
E       from .deploy import Deployer
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/deployer/deploy.py", line 7, in <module>
E       from .kubernetes.ready_check import ReadyCheck
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/deployer/kubernetes/__init__.py", line 7, in <module>
E       from .adapter import K8s
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/deployer/kubernetes/adapter.py", line 9, in <module>
E       from ...specs.models import ResourcesSpec, ResourceRequirementSpec
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/specs/__init__.py", line 10, in <module>
E       from .app_config_downloader import AppConfigDownloader
E     File "/home/travis/build/finn/fiaas-deploy-daemon/fiaas_deploy_daemon/specs/app_config_downloader.py", line 1, in <module>
E       import yaml
E     File "build/bdist.linux-x86_64/egg/yaml/__init__.py", line 14, in <module>
E     File "build/bdist.linux-x86_64/egg/yaml/cyaml.py", line 5, in <module>
E     File "build/bdist.linux-x86_64/egg/_yaml.py", line 7, in <module>
E     File "build/bdist.linux-x86_64/egg/_yaml.py", line 4, in __bootstrap__
E     File "/home/travis/.local/share/virtualenvs/fiaas-deploy-daemon/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1226, in resource_filename
E       self, resource_name
E     File "/home/travis/.local/share/virtualenvs/fiaas-deploy-daemon/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1726, in get_resource_filename
E       self._extract_resource(manager, self._eager_to_zip(name))
E     File "/home/travis/.local/share/virtualenvs/fiaas-deploy-daemon/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1756, in _extract_resource
E       self.egg_name, self._parts(zip_path)
E     File "/home/travis/.local/share/virtualenvs/fiaas-deploy-daemon/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1293, in get_cache_path
E       self.extraction_error()
E     File "/home/travis/.local/share/virtualenvs/fiaas-deploy-daemon/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1273, in extraction_error
E       raise err
```

While I'm not that familiar with the inner workings of pkg_resources, I took a look at the code referenced at the end of the stack trace, and my theory is that since the tests run in parallel, multiple threads may end up calling [_bypass_ensure_directory](https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L3075-L3082) at the same time, and that the error stems from a file system race condition between `isdir` and `mkdir` when creating the directory, similar to in https://github.com/pypa/setuptools/pull/1083. Perhaps using the same solution could work here?

JonKohler:
## Summary of changes
This patch solves a race condition in get_cache_path in pkg_resource init.py, which generally shows its head: 
- In high volume CICD systems with many cores
- where the build job is configured to use all of said cores
- where there is a lot of concurrent python-y stuff happening around the same time

Closes #1256

## Details of change
Upstreaming a private fix that we have applied to python-setuptools 0.9.8 that solves a race during the creation of an egg cache. We use setuptools 0.9.8 with this fix internally on some of our projects. This fix has been in production for May 2017 and is working great.

The challenge we ran into is that setuptools, by itself, works great, both on development laptops/desktops, our fleet of development VMs, and so on. When put in a high volume CICD pipeline, even the latest version of python-setuptools falls on its face immediately (39.2.0, we haven't tried 40.x since it just came out a few hours ago). there was a recent-ish change (#1083) that (sounded) like it would address this, but did not

We suspect that given the build machines have 32 cores, and we're using make -j32, the concurrency is biting us here. We've added this small patch to work around it, and as mentioned, it has been great. 

For sanity, I did include the most recent failure logging from this morning from the 32 core build system here: https://github.com/pypa/setuptools/issues/1256#issuecomment-403547646

In an effort to get off that ancient version of setuptools, we're submitting this upstream, with the hopes that we can get this mainlined, and can then stay on vanilla "fresh off the branch" versions going forward.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1412.change.rst
+++ b/changelog.d/1412.change.rst
@@ -1,0 +1,1 @@
+Solve race during creation of egg cache
--- a/changelog.d/1412.change.rst
+++ b/changelog.d/1412.change.rst
@@ -1,1 +1,1 @@
-Solve race during creation of egg cache
+Solve race during creation of egg cache by removing bypass_ensure_directory and switching to ensure_directory
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -3030,7 +3030,13 @@ def _bypass_ensure_directory(path):
     dirname, filename = split(path)
     if dirname and filename and not isdir(dirname):
         _bypass_ensure_directory(dirname)
-        mkdir(dirname, 0o755)
+        try:
+          mkdir(dirname, 0o755)
+        except os.error:
+          # Check for a possible race where the directory was already
+          # created.
+          if not isdir(dirname):
+            raise
 
 
 def split_sections(s):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1198,7 +1198,7 @@ class ResourceManager:
         extract_path = self.extraction_path or get_default_cache()
         target_path = os.path.join(extract_path, archive_name + '-tmp', *names)
         try:
-            _bypass_ensure_directory(target_path)
+            ensure_directory(target_path)
         except Exception:
             self.extraction_error()
 
@@ -3021,22 +3021,6 @@ def ensure_directory(path):
     """Ensure that the parent directory of `path` exists"""
     dirname = os.path.dirname(path)
     py31compat.makedirs(dirname, exist_ok=True)
-
-
-def _bypass_ensure_directory(path):
-    """Sandbox-bypassing version of ensure_directory()"""
-    if not WRITE_SUPPORT:
-        raise IOError('"os.mkdir" not supported on this platform.')
-    dirname, filename = split(path)
-    if dirname and filename and not isdir(dirname):
-        _bypass_ensure_directory(dirname)
-        try:
-          mkdir(dirname, 0o755)
-        except os.error:
-          # Check for a possible race where the directory was already
-          # created.
-          if not isdir(dirname):
-            raise
 
 
 def split_sections(s):


JonKohler:
Given that this is a tricky to reproduce on a small scale race condition / edge case, can I get an exception on code coverage? or do you want coverage for this?

pganssle:
I think there are probably ways you can force the "race condition". My initial instinct is patching one of the files involved to create the directory at the right time.

I'm not entirely sure I understand why `_bypass_ensure_directory` exists, though. It seems like it's a deliberate choice, but I don't know what "sandbox" it's referring to that needs to be bypassed. Does `os.makedirs` have the same race condition?

JonKohler:
hey @pganssle - Yes for sure there are ways you can force a race condition, but gut feeling is its a bit more a pain than it might be worth, as you'll get this to bomb out in the same way its bombing out today with the existing logic thats in place. Meaning, you wouldn't have uncaught errors. 

Anyhow - Offhand, I dont know why the bypass exists; however, this is where it fails, this is it. 

I know that the code (from git blame) hasn't been touched in 3-4 years; however, even before that (for example on our 0.9.8) its just about the same, just was in a different place before that refactoring 3-4 years ago.

pganssle:
@JonKohler I think my point is it's hard for me to evaluate if this actually works without tests. You mention it's hard to reproduce on a single machine so I haven't tried.

Can you check if switching the `_bypass_ensure_directories` over to `ensure_directories` causes problems in your system that can reproduce the issue? If the original `SandboxViolation` [referred to in this commit](https://github.com/pypa/setuptools/commit/a239dfea9549314e200d91f65ab50c0009359a56) doesn't apply anymore for whatever reason, we may be able to drop it.

As an aside, this is one very good reason to include tests for whatever behavior you care about. If the SandboxViolation change had associated tests, we'd be able to tell if removing `_bypass_ensure_directories` would cause a regression.

jaraco:
Thanks for the splunking @pganssle . Following that commit leads to [this Distribute issue](https://bitbucket.org/tarek/distribute/issues/13) which links to [this old Setuptools issue](https://bugs.python.org/setuptools/issue23). Probably those issues still exist, although they're probably drastically reduced by the limited use of easy_install these days.

I'm hesitant to accept this change as (a) it is duplicating the behavior in py31compat, (b) it's added inline so difficult to tell if it's important, (c) it adds complication without any hope of ever breaking free, and (d) the sandboxing code isn't relevant except when using easy_install, which is deprecated.

My temptation is to remove the bypass function, rely on ensure_directory, and hope users no longer run into the issue... and if they do see if we can migrate them to pip instead of easy_install.

JonKohler:
Alright, ill give it a whack. I've just upgraded our CICD to 39.2 with this fix, and for sure it works (horray!) but I'll try switching back to ensure_directories and see what happens.

If I'm lucky, that change combined with the previously mentioned #1083 might just fix this by default

JonKohler:
Ok, I've updated the PR with what I've just committed in our side. I've moved from 39.2.0 to 40.0.0 and applied the same fixes on our side.

It takes a few hours for our toolchain to build (its big), then I'll retarget the CICD to use said toolchain. The problem appears immediately in the CICD builds, so I should (knock on wood) know yes/no today

pganssle:
It's a good sign that the tests are passing, but I'm still a bit anxious because we didn't have a test for the `SandboxViolation` problem in the first place, I'm not how to tell if removing the `_ensure_directory_bypass` is going to cause a regression or not. I'm thinking it *is* going to cause a regression, because after quite a bit of digging around I can't see any substantial changes in behavior in the sandbox between now and when the original bypass commit occurred.

I think the best options for now are:

1. Try to write a failing test for the original `SandboxViolation` error so that we can see if we can fix it *without* writing our own `makedirs`.
2. Write a test that exercises something like @JonKohler's original patch and accept that. I'll note that [the upstream version of `makedirs` has a similar implementation](https://github.com/python/cpython/blob/master/Lib/os.py#L209).

I think the main problem here is that because `makedirs` is implemented in terms of `mkdir`, we need to re-implement `makedirs` in terms of the non-sandboxed `mkdir` in order to bypass the sandbox for `makedirs`. We may be able to change the implementation of the sandbox to make it easier to bypass the sandbox when using derived functions like `makedir` (for example, we could have the `os` wrapper add an `os._bypass_sandbox` submodule that exposes all the original functionality).

JonKohler:
Thanks again @pganssle ! Adding in @jaraco and @dirk-thomas to this conversation - as they did the original work and commentary on #1083 as well as https://github.com/pypa/setuptools/commit/925dd35ec77d4cfd79c899f34255e3a753f33ca5

Hey guys - can you help out in this conversation, for some of the historical context on the ensure_directory vs bypass_ensure_directory? We see some good work being done to fight race conditions on https://github.com/pypa/setuptools/commit/925dd35ec77d4cfd79c899f34255e3a753f33ca5 / #1083, but the get_cache_path code wasn't touched at this time, so we're running into races there.

I had two proposed fixes: 
1 - fix the race condition in bypass_ensure_directory, which we've had in production for a year 
2 - nuke bypass_ensure_directory and unify all code paths to go through ensure_directory, which is what the active state of this PR is

dirk-thomas:
I don't have enough context around `WRITE_SUPPORT` and if it is still needed to be handled separately. Besides that the patch looks fine to me. Certainly addressing the same race for this case. :+1:

pganssle:
@dirk-thomas I think the `WRITE_SUPPORT` thing is not so important - that seems to be a check that the platform actually supports the operations you're trying to do. The reason `_bypass_ensure_directory` exists is because the Sandbox modifies `os` with wrappers that prevent certain actions outside of the sandbox. Something related to the cache directory was causing `SandboxViolation` errors, which is why there's a bypass function.

JonKohler:
@pganssle - our toolchain just finished building, and I kicked off the CICD job, and it built up the egg cache perfectly. We're using setuptools 40.0.0 from pypi with the modifications from this PR as of right now.

Once this build finishes and passes all of the smoke/testing/blah blah blah, we'll merge it. Once that's done, all of our CICD builds will run against this, and we'll be able to tell immediately if this is going to die in a fire or not.

net - it's looking really good thus far, and knock on wood, i think this might be a winner

Assuming that it is good, looking at the issue where the sandbox shenanigans originally occurred (back in 2008!) https://bugs.python.org/setuptools/issue23 ... at that time, they're saying the same things ... they can't reproduce it, but this race-generating fix happened to fix it at the time.

I suspect even though they probably tested it a bit, that they just didn't have the processing power 10 years ago to hit this with predictability.

that leaves us in an interesting spot. We might cause a bit of a regression, but the original fix wasn't bulletproof either.

Thoughts?

pganssle:
@JonKohler I think we're still in the same position with respect to the above. We should come up with a regression test for *at least one* of these problems and go from there.

If we really can't reproduce the `SandboxViolation`, I think I'd prefer to play around with the sandboxing logic to give us easier access to a sandbox-bypassing version of `os.makedirs` rather than just removing the sandbox bypass entirely. That seems like the most surgical approach, and we can open a new issue for "write a test that induces the `SandboxViolation` issue" to track whether the sandbox bypass is even necessary.

JonKohler:
Thanks again @pganssle - I do appreciate the commentary and guidance. I've been discussing this internally with the developer who first fixed this (with the original bypass race fix). 

For the race condition: The only test we can think of that might even get close to semi-reliably reproducing this on demand is spawning a ton of processes that import a python library from the same egg file. Even then, it might be a crap shoot depending on what system you run the test on.

For us, since we can hit this 100% of the time with our particular CICD setup, we're happy with the fix as is; however, I dont want to leave you guys high and dry. 

From an "outsider looking in" perspective, the original commit (and others via googling around) have a really hard time reproducing this with certainty when it has come up.

If we do write a unit test, I have a feeling that to get it to do what we want it to do, it would be convoluted and may end up driving inconsistency. That means false negatives and false positives, which is a nightmare.

Truly, I’m against writing unit tests, it's just tricky hard to **deterministically** test against this it seems, which means we'd be doing more harm than good.

As for the sandboxing logic, in reading the previous commits and chatter, my gut feeling is the following: 

@jaraco mentioned:
> the sandboxing code isn't relevant except when using easy_install, which is deprecated.

So, we've had this half-racy code in the code base for the better part of 10 years, and the masses arent out with pitchforks. Given the moves away from easy_install and the moves from python2 (going out to pasture soon enough) to python3; honestly, it might be better to rip the bandaid off here and intentionally open up to a regression / breaking change. 

I say that because, if we do that, we might run into a situation where a user actually has a 100% reproducible use case we could debug and fix properly. 

Otherwise, as @jaraco said, we could end up in a situation where we've got complexity with no light at the end of the tunnel.


All that said, there is a really good chance, since this has been kicking around for 10 years, that fixes in the surrounding code (setuptools, python, or linux/etc) have actually made this a non-issue, so all this discussion could be 100% academic.


Thoughts?

JonKohler:
ninja edited above comment to reflect that I'm NOT against writing the tests (original text didn't have the NOT) :)

pganssle:
I think the fact that we can (eventually) ditch the sandbox stuff entirely is an argument in *favor* of just improving it in this somewhat minor way now, because we know that long term there will be minimal maintenance burden (since that code is going away). I don't think we should be introducing regressions just because it's hard to test fixes.

With regards to the testing, I think the best option to do it deterministically is some sort of mock that deliberately introduces the race condition, though I will note that this relies on implementation details of our solution itself. If we end up using a different solution (like the unpatched `os.makedirs`, or even the patched `os.makedirs`), I think we would need to use a different test. I have created a proposal in #1418.

JonKohler:
Thanks @pganssle - I appreciate it! I looked at #1418  ... LGTM, I appreciate it.

Assuming #1418 get accepted, are we good here? or do you need more of us? Happy to continue the work/conversation here, just want to make sure I know what action items are left :)

JonKohler:
Is this replaced by PR #1418?

PR #1413, opened 2018-07-09
Fix a problem with Jython by using the version_info as a tuple.

datalogics-robb:
Jython 2.7 does not have a named tuple for version_info, so revert to older, gentler tuple syntax.

## Summary of changes

Fix an incompatibility with Jython by using tuple syntax instead of named tuple.

### DIFF
--- a/pkg_resources/py31compat.py
+++ b/pkg_resources/py31compat.py
@@ -15,7 +15,7 @@ def _makedirs_31(path, exist_ok=False):
 #  and exists_ok considerations are disentangled.
 # See https://github.com/pypa/setuptools/pull/1083#issuecomment-315168663
 needs_makedirs = (
-    sys.version_info.major == 2 or
+    sys.version_info[0] == 2 or
     (3, 4) <= sys.version_info < (3, 4, 1)
 )
 makedirs = _makedirs_31 if needs_makedirs else os.makedirs


datalogics-robb:
The latest release of setuptools suddenly broke our automated test suite for our C#/Java interface - which is tested using IronPython and Jython.  Apparently Jython doesn't have sys.version_info as a named tuple, but as a vanilla tuple.  Testing this would require no new code, but a Jython test deployment which would not go in this PR.
IMO this is a terse enough change that it shouldn't require a mention in release notes, but the summary above could be added.

jaraco:
Probably this fix needs to be in bdist_egg also. More importantly, since the issue could easily regress again, a comment, test, or other construct will be necessary to retain this expectation.

datalogics-robb:
A bit more info...

```
robb@robs-mbp:~$ python3
Python 3.6.2 (v3.6.2:5fd33b5926, Jul 16 2017, 20:11:06) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import sys
>>> sys.version_info.major
3
>>> ^D
robb@robs-mbp:~$ python2
Python 2.7.9 (v2.7.9:648dcafa7e5f, Dec 10 2014, 10:10:46) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import sys
>>> sys.version_info
sys.version_info(major=2, minor=7, micro=9, releaselevel='final', serial=0)
>>> sys.version_info.major
2
>>> ^D
robb@robs-mbp:~$ java -jar apdfl-sandbox/DLE/Tools/jython/jython.jar 
Jython 2.7.0 (default:9987c746f838, Apr 29 2015, 02:25:11) 
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java1.8.0_131
>>> import sys
>>> sys.version_info
(2, 7, 0, 'final', 0)
>>> sys.version_info.major
AttributeError: 'tuple' object has no attribute 'major'
>>> robb@robs-mbppython3
Python 3.6.2 (v3.6.2:5fd33b5926, Jul 16 2017, 20:11:06) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
>>> import sys
>>> sys.version_info[0]
3
>>> ^D
```

datalogics-robb:
Hmmm... Any ideas as to how would a useful test be implemented for this?

A comment is certainly do-able.

jaraco:
Hmm. On my system, Jython 2.7.1 has the property.

```
setuptools master $ jython
Jython 2.7.1 (default:0df7adb1b397, Jun 30 2017, 19:02:43)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java1.8.0_45
>>> import sys
>>> sys.version_info
sys.version_info(major=2, minor=7, micro=1, releaselevel='final', serial=0)
>>> sys.version_info.major
2
```

datalogics-robb:
How about pep425tags.py lines 100 and 101? (or do these not get executed in a Jython impl?)

datalogics-robb:
Wow that's bizarre - Looks like your Jython is one micro version ahead of mine

jaraco:
I don't see any note about it in the changelog. I'll bet the stdlib was updated and 2.7.1 absorbed "fixes" from CPython.

jaraco:
What's the chance we can keep the code as-is and ask our Jython users to update their Jython to 2.7.1?

datalogics-robb:
http://www.jython.org/ shows 2.7.0 as the latest version released in 2015

pganssle:
Can we switch it to `six.PY3` and push the testing burden onto `six`?

jaraco:
I found the [source for Jython 2.7.1](https://bitbucket.org/jython/jython/commits/tag/v2.7.1). I got my copy through [Homebrew](http://brewformulas.org/Jython). It looks like it [pulls the jar from here](https://search.maven.org/remotecontent?filepath=org/python/jython-installer/2.7.1/jython-installer-2.7.1.jar).

> Can we switch it to six.PY3 and push the testing burden onto six?

We could, but it feels a little tenuous to rely on six to support Jython 2.7.0 when 2.7.1 works... especially when such a change could be readily optimized away.

I was going to suggest that Jython 2.7.0 users should patch their install to put a compatible object in `sys.version_info` (via something like site.py), but with Java strict type checking, that may not be possible.

```
>>> sys.version_info = collections.namedtuple('versioninfo', 'major minor micro releaselevel serial')(*sys.version_info)
>>> Traceback (most recent call last):
TypeError: can't convert versioninfo(major=2, minor=7, micro=1, releaselevel='final', serial=0) to org.python.core.PyVersionInfo
```

I'm pretty sure the right solution here for Jython users to use old setuptools versions or update Jython.

datalogics-robb:
I found the 2.7.1 distribution, and have updated but I'm still a bit baffled as to why it it's been set to the much older verison.

pganssle:
I actually think independent of the Jython thing, we should probably use `six.PY3` or `six.PY2` if we're using `six` at all. By happenstance it will fix the Jython thing, but it will allow us to not worry about the "best way to detect whether it's Python 3" anyway.

datalogics-robb:
I totally agree with @pganssle here, six is the way to go.  I'm encountering other problems upgrading Jython as well, so for now we're using setuptools 39.x as a work-around

PR #1416, opened 2018-07-10
Switch over to using six.PY{2,3} when possible

### LINKED ISSUE #1415: Use six to determine Python version wherever possible
This issue came up in #1413 - there are a number of places in `setuptools` and `pkg_resources` where we check `sys.version.major`, which is apparently not portable to Jython 2.7.0. While I don't think we have to go out  of our way to support Jython 2.7.0, I think that given that we use `six` in many other places for its compatibility shim, we should also use `six.PY2` and `six.PY3` to determine whether we're in Python 2 or Python 3. We can assume that `six` will do a better job than us at handling edge cases in general for the parts of 2/3 compatibility that they support.

pganssle:
## Summary of changes
Per the reasoning in #1415, I think that we should use `six.PY{2,3}` whenever possible and outsource how to determine the major version to them.

The one concession I made to #1413 is that I changed one `sys.version_info.major` to `sys.version_info[0]` in the version logic *itself*, because that is more consistent with how `six` does it. I'm not saying that we guaranteed that we'll never use `sys.version_info.major` in the future, just that this is another equally good way to check the value that doesn't trigger the Jython bug.

Closes #1415

### Pull Request Checklist
- Changes have tests **N/A**

### DIFF
--- a/setuptools/extern/__init__.py
+++ b/setuptools/extern/__init__.py
@@ -48,7 +48,7 @@ class VendorImporter:
                 # on later Python versions to cause relative imports
                 # in the vendor package to resolve the same modules
                 # as those going through this importer.
-                if sys.version_info.major >= 3:
+                if sys.version_info >= (3, ):
                     del sys.modules[extant]
                 return mod
             except ImportError:
--- a/changelog.d/1416.change.rst
+++ b/changelog.d/1416.change.rst
@@ -1,0 +1,1 @@
+Moved several Python version checks over to using ``six.PY2`` and ``six.PY3``.
--- a/pkg_resources/py31compat.py
+++ b/pkg_resources/py31compat.py
@@ -2,6 +2,8 @@ import os
 import errno
 import sys
 
+from .extern import six
+
 
 def _makedirs_31(path, exist_ok=False):
     try:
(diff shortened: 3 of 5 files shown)


pkg_resources/extern/__init__.py
```diff
@@ -1,6 +1,5 @@
 import sys
 
-
```
pganssle: Oops, I think this was a mistake. I think this is a PEP 8 violation. I'll clean it up.

setuptools/extern/__init__.py
```diff
@@ -48,7 +48,7 @@ def load_module(self, fullname):
                 # on later Python versions to cause relative imports
                 # in the vendor package to resolve the same modules
                 # as those going through this importer.
-                if sys.version_info.major >= 3:
+                if sys.version_info >= (3, ):
```
thisch: Can be changed to  `if not six.PY2`

pganssle: I think not, because this code is for loading the vendored packages and `six` is one of the vendored packages it's loading. I could be wrong, but I think it's not possible to use the vendored packages in `extern/__init__.py`

thisch: Sry, I didn't know that six can't be used in this module.

pganssle:
@jaraco Any objections? I know you didn't want to explicitly handle the Jython case.

pganssle:
I'm going to merge this.

PR #1420, opened 2018-07-11
Add findns: directive to config

### LINKED ISSUE #1419: Support find_packages_ns from setup.cfg
Now that #1312 has been merged, closing #97, we support the `find_packages_ns` function in `setup.py`. We should also enable this functionality in `setup.cfg`.

I suggest adding a `find_namespace:` directive. The current `find:` directive is implemented [here](https://github.com/pypa/setuptools/blob/cca86c7f1d4040834c3265ccecdd9e21b4036df5/setuptools/config.py#L512), so I think that would probably be the place to add support for  namespace packages.

Brought to my attention by @agronholm  in https://github.com/pypa/setuptools/issues/97#issuecomment-404195899

silkentrance:
## Summary of changes

- extend documentation on new `findns:` directive
- realize new `findns:` directive

Closes #1419

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2389,7 +2389,7 @@ Metadata and options are set in the config sections of the same name.
 * In some cases, complex values can be provided in dedicated subsections for
   clarity.
 
-* Some keys allow ``file:``, ``attr:``, and ``find:`` and ``find_namespace:`` directives i
+* Some keys allow ``file:``, ``attr:``, and ``find:`` and ``find_namespace:`` directives in
   order to cover common usecases.
 
 * Unknown keys are ignored.
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2389,8 +2389,8 @@ Metadata and options are set in the config sections of the same name.
 * In some cases, complex values can be provided in dedicated subsections for
   clarity.
 
-* Some keys allow ``file:``, ``attr:``, and ``find:`` directives in order to
-  cover common usecases.
+* Some keys allow ``file:``, ``attr:``, and ``find:`` and ``find_namespace:`` directives i
+  order to cover common usecases.
 
 * Unknown keys are ignored.
 
@@ -2479,7 +2479,7 @@ eager_resources          list-comma
 dependency_links         list-comma
 tests_require            list-semi
 include_package_data     bool
-packages                 find:, list-comma
+packages                 find:, find_namespace:, list-comma
 package_dir              dict
 package_data             section
 exclude_package_data     section
@@ -2489,10 +2489,13 @@ py_modules               list-comma
 
 .. note::
 
-    **packages** - The ``find:`` directive can be further configured
+    **packages** - The ``find:`` and ``find_namespace:`` directive can be further configured
     in a dedicated subsection ``options.packages.find``. This subsection
-    accepts the same keys as the `setuptools.find` function:
+    accepts the same keys as the `setuptools.find_packages` and the
+    `setuptools.find_packages_ns` function:
     ``where``, ``include``, and ``exclude``.
+    
+    **find_namespace directive** - The ``find_namespace:`` directive is supported since Python >=3.3.
 
 
 Configuration API
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -519,25 +519,25 @@ class ConfigOptionsHandler(ConfigHandler):
         trimmed_value = value.strip()
 
         if not trimmed_value in find_directives:
-          return self._parse_list(value)
+            return self._parse_list(value)
 
         findns = trimmed_value == find_directives[1]
         if findns and not PY3:
-          raise DistutilsOptionError('find_namespace: directive is unsupported on Python < 3.3')
+            raise DistutilsOptionError('find_namespace: directive is unsupported on Python < 3.3')
 
         # Read function arguments from a dedicated section.
         find_kwargs = self.parse_section_packages__find(
             self.sections.get('packages.find', {}))
 
         if findns:
-          from setuptools import find_packages_ns as find_packages
+            from setuptools import find_packages_ns as find_packages
         else:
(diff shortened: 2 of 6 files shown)


silkentrance:
@pganssle Have a look and tell me if that is the right direction. I so, I will continue with implementing required tests.

pganssle:
Is there a reason it can't be `find_namespace`? I'm even dubious about the `find_packages_ns` name, since it seems too obscure to me, I'd rather it be `find_namespace_packages` or something.

pganssle:
The `find_packages_ns` stuff is indeed a completely different discussion so don't worry about that.

pganssle:
This seems like it will work, but the implementation might be a bit cleaner if you do this:

```python
directive = value.split(':', 1)[0]
if directive not in {'find', 'find_namespace'}:
    ...
```

silkentrance:
@pganssle good point, but this will be a different commit

silkentrance:
@pganssle note that with the latest commit I refrained from using string.split() and instead used value.strip().

docs/setuptools.txt
```diff
@@ -2389,8 +2389,8 @@ Metadata and options are set in the config sections of the same name.
 * In some cases, complex values can be provided in dedicated subsections for
   clarity.
 
-* Some keys allow ``file:``, ``attr:``, and ``find:`` directives in order to
-  cover common usecases.
+* Some keys allow ``file:``, ``attr:``, and ``find:`` and ``find_namespace:`` directives i
```
agronholm: Typo at the end?

setuptools/config.py
```diff
@@ -515,21 +515,29 @@ def _parse_packages(self, value):
...
         :rtype: list
         """
-        find_directive = 'find:'
+        find_directives = ['find:', 'find_namespace:']
+        trimmed_value = value.strip()
 
-        if not value.startswith(find_directive):
-            return self._parse_list(value)
+        if not trimmed_value in find_directives:
+          return self._parse_list(value)
```
agronholm: Bad indentation

setuptools/config.py
```diff
@@ -515,21 +515,29 @@ def _parse_packages(self, value):
...
+        if findns:
+          from setuptools import find_packages_ns as find_packages
+        else:
+          from setuptools import find_packages
 
         return find_packages(**find_kwargs)
 
     def parse_section_packages__find(self, section_options):
-        """Parses `packages.find` configuration file section.
+        """Parses `packages.find[ns]` configuration file section.
```
agronholm: Does this really use a `packages.findns` section? The tests do not.

setuptools/tests/test_config.py
```diff
@@ -596,6 +601,59 @@ def test_find_directive(self, tmpdir):
...
+        config.write(
+            '[options]\n'
+            'packages = find_namespace:\n'
+            '\n'
+            '[options.packages.find]\n'
+            'exclude =\n'
+            '    fake_package.sub_one\n'
+        )
+        with get_dist(tmpdir) as dist:
+            assert set(dist.packages) == set(
```
agronholm: You can use a set literal here. Python 2.6 support was dropped in v37.

setuptools/config.py
silkentrance: and there was even more where this came from...

silkentrance: done

docs/setuptools.txt
silkentrance: done

setuptools/tests/test_config.py
silkentrance: The other tests also use the non set literal, should I proceed?

silkentrance: done

silkentrance:
If you comply to the changes, I will proceed with fixing #1421 based upon these changes.

setuptools/tests/test_config.py
```diff
@@ -4,18 +4,23 @@
 from mock import patch
 from setuptools.dist import Distribution, _Distribution
 from setuptools.config import ConfigHandler, read_configuration
+from setuptools.extern.six import PY2, PY3
 
+py2_only = pytest.mark.xfail(not PY2, reason="Test runs on Python 2 only")
```
pganssle: I think these should be `pytest.mark.skip`, not `xfail`.

silkentrance: will have to fix the tests for find_packages_ns as well. you would want to fix the other tests that use that pattern.

agronholm: I concur.

silkentrance: six.PY2 seems to fail when marking tests for specific versions

```
SKIP [1] setuptools/tests/test_config.py:604: Test runs on Python 2 only
```

pganssle: I think that's indicating that the test was skipped. When you mark something `xfail`, it actually *runs* the test and ignores the result if it fails (or if you make it a strict xfail, actually fails the test suite if the test passes). `skip` doesn't run the test at all, so I believe that message is expected.

You are right that we should be switching many `xfail`s over to `skip`. I was planning on doing that at some point, but it takes time to figure out which ones are intended as an actual `xfail` (which generally means "this should work but it's a known bug - when this test starts passing drop this decorator") and which are just using the wrong decorator for `skip`.

silkentrance:
currently working on py2_only/py3_only for test_config. It does not seem to work as expected, reporting false results.

pganssle:
I pushed an updated version with the proper decorator to your branch.

silkentrance:
@pganssle thanks. I did find out about skipif in the morning, too. Yesterday I was way too tired...

silkentrance:
I have refactored out the markers for py2/py3 into a separate module since they had been redeclared. Other test modules can be refactored, too, but I leave it at that for the time being.

PR #1427, opened 2018-07-13
Touch 'egg-info' directory

stephenfin:
## Summary of changes

`tox` determines whether a package should be rebuilt by comparing the timestamp of the package's 'egg-info' directory and its `setup.py` or `setup.cfg` files [[1](https://github.com/tox-dev/tox/blob/3.1.0/src/tox/venv.py#L253-L257)] [[2](https://github.com/tox-dev/tox/blob/3.1.0/src/tox/venv.py#L221-L244)]. Unfortunately this checks the `egg-info` directory itself, which is not updated, unlike the contents of that directory. This means that `tox` will always rebuild the package once one of the two setup files has been updated.

While this is clearly a bug in `tox` that should be fixed separately, there is merit in using this as a heuristic so enable it.

Some other random fixes are included as separate patches.

### Pull Request Checklist
- [x] Changes have tests


pganssle:
What is up with all the permissions changes?

pganssle:
OK, the permissions changes are right, looks like it's a duplicate of #1441, which actually missed `easy_install.py`. I've rebased and left that one commit in place.

With regards to the merit of the actual PR, I don't totally understand why we need to do this and it doesn't seem entirely desirable. If we're not actually changing the `.egg-info` directory in any way, spoofing the modified time will make the metadata *inaccurate*. It seems like this might undermine anyone who is actually trying to monitor whether or not the `.egg-info` has changed, and might undermine some caching algorithms (though, admittedly, I don't know of any offhand).

I'm totally willing to be convinced otherwise, though. I feel like I have definitely encountered annoying problems from too-aggressive caches relating to packaging before.

@gaborbernat Any thoughts on this?

stephenfin:
> With regards to the merit of the actual PR, I don't totally understand why we need to do this and it doesn't seem entirely desirable. If we're not actually changing the .egg-info directory in any way, spoofing the modified time will make the metadata inaccurate. It seems like this might undermine anyone who is actually trying to monitor whether or not the .egg-info has changed, and might undermine some caching algorithms (though, admittedly, I don't know of any offhand).

Yes and no. We are actually updating the _contents_ of this folder but, due to how timestamps work on Linux at least, the timestamp of the directory itself is not updated. For example:

    $ ls -la oslo.config.egg-info/
    total 40
    drwxr-xr-x.  2 sfinucan sfinucan 4096 Dec  5  2017 .
    drwxr-xr-x. 12 sfinucan sfinucan 4096 Aug 22 13:56 ..
    -rw-r--r--.  1 sfinucan sfinucan    1 Aug 22 13:56 dependency_links.txt
    -rw-r--r--.  1 sfinucan sfinucan  227 Aug 22 13:56 entry_points.txt
    -rw-r--r--.  1 sfinucan sfinucan    1 Dec  5  2017 not-zip-safe
    -rw-r--r--.  1 sfinucan sfinucan   47 Aug 22 13:56 pbr.json
    -rw-r--r--.  1 sfinucan sfinucan 2041 Aug 22 13:56 PKG-INFO
    -rw-r--r--.  1 sfinucan sfinucan  222 Aug 22 13:56 requires.txt
    -rw-r--r--.  1 sfinucan sfinucan 3459 Aug 22 13:56 SOURCES.txt
    -rw-r--r--.  1 sfinucan sfinucan   12 Aug 22 13:56 top_level.txt

    $ ls -la | grep egg-info
    drwxr-xr-x.  2 sfinucan sfinucan  4096 Dec  5  2017 oslo.config.egg-info

(note Dec 5 2017 vs. Aug 22 2018)

Perhaps there are better heuristics we could be using to determine when an sdist has been regenerated, but this seemed as good as any in my mind. For what it's worth, I have "fixed" this in tox itself but the fix seemed like something others would benefit from.

stephenfin:
Another alternative might be to delete the entire `.egg-info` directory on rebuild, but that seemed overly invasive and I'm unsure what other side effects there might be, e.g. would files ever be placed here that weren't generated as part of the sdist build step?

PR #1429, opened 2018-07-14
Add a space after header, so reST rendering isn't mangled.

benjaminp:


pganssle:
I'm guessing this also needs to be changed in the towncrier template?

pganssle:
LGTM.

PR #1431, opened 2018-07-20
Make install directory if it doesn't exist yet

hroncok:
Problem: In Fedora, `sudo python3 setup.py install` installs to /usr/local/lib/pythonX.Y/site-packages. However that directory is not always there as it is not installed together with the python3 package (Fedora packages should not install stuff to /usr/local).

When the directory does not exist an installation attempt fails with:

    [Errno 2] No such file or directory: '/usr/local/lib/python3.7/site-packages/test-easy-install-11875.write-test'

Solution: Make the directory if it doesn't exists yet. If user has no permissions, the above error is preserved, yet with root (sudo) it works now.

I realize that this is not usefull outside Fedora's specific setup, yet I don't like downstream only patches and this breaks nothing.

- [ ] Changes have tests - not sure how to do that, as this is fairly Fedora specific
- [x] News fragment added in changelog.d

### DIFF
--- a/changelog.d/1431.change.rst
+++ b/changelog.d/1431.change.rst
@@ -1,0 +1,1 @@
+In ``easy_install.check_site_dir``, ensure the installation directory exists.
--- a/changelog.d/1431.chnage.rst
+++ b/changelog.d/1431.chnage.rst
@@ -1,0 +1,3 @@
+Make install directory if it doesn't exist yet to prevent Fedora's
+specific setup to blow up if ``/usr/local/lib/pythonX.Y/site-packages``
+doesn't exist yet.
--- a/changelog.d/1431.chnage.rst
+++ b/changelog.d/1431.chnage.rst
@@ -1,3 +1,0 @@
-Make install directory if it doesn't exist yet to prevent Fedora's
-specific setup to blow up if ``/usr/local/lib/pythonX.Y/site-packages``
-doesn't exist yet.
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -450,6 +450,12 @@ class easy_install(Command):
         instdir = normalize_path(self.install_dir)
         pth_file = os.path.join(instdir, 'easy-install.pth')
 
+        if not os.path.exists(instdir):
+            try:
+                os.makedirs(instdir)
+            except (OSError, IOError):
+                self.cant_write_to_target()
+
         # Is it a configured, PYTHONPATH, implicit, or explicit site dir?
         is_site_dir = instdir in self.all_site_dirs
 


hroncok:
Trying to add tests for this. Locally I fail to run them, so let's see what travis says.

hroncok:
https://github.com/pypa/setuptools/pull/1431/commits/9683b7798cd6940cf14080d6705dc239ec8c8125 this seems to pass even without my patch.

pganssle:
I'm not really sure we want to support `sudo python3.6 setup.py install` *at all*, but if this isn't installed as part of `python3`, are you even sure its in the `PYTHONPATH`?

I am -1 on this for the following reasons:

1. `setup.py install` should be deprecated in favor of `pip install`
2. Using `sudo` to install things is not even supported in `pip` - you should use the system package manager for system packages, and `pip install --user` for user installs, or use virtualenvs.
3. If we *did* want to support this, I don't think creating the directory is the right way to do it. If the directory we're trying to install it in doesn't exist, it seems very likely that we're installing it in the wrong place. The "right place" should probably be inferred from the system configuration.

If we're already inferring the right place to install packages from the system configuration (haven't checked), then Fedora probably needs to be fixed to properly advertise its non-standard package deployment location (if, again, Fedora wants to support this very dangerous workflow).

hroncok:
1. agreed. yet people use this and report this error to us. sudo pip works in this scenario.
2. agreed. yet people still use sudo here and report problems.
3. pip supports it. this is the right place form our system configuration. 

Fedora properly advertises this at https://fedoraproject.org/wiki/Changes/Making_sudo_pip_safe - this regression was a direct result of this change.

That said, I understand if you won't accept this, yet we'll still apply it downstream, becasue otherwise users are rightfully angry.

pganssle:
@jaraco @benoit-pierre At this point we can safely deprecated `setup.py install` in favor of pip, like we've deprecated `upload` and `register`, right?

@pradyunsg @dstufft Thoughts?

pganssle:
If we can deprecated `setup.py install`, then I think that can at least partially help with the "people are complaining that this doesn't work" issue.

That said, if Fedora actually intends for people to deploy things to `/usr/lib/python3.6/site-packages`, why isn't the solution to create that directory as part of the `python3` package?

hroncok:
It has two point of views. We don't support it, but we are trying to prevent sudo pip installers to nuke their systems. Yet they complain we broke their workflow and we try to make it work even when we say: don't do this. We cannot install rpm packaged stuff into /usr/local.

dstufft:
This isn't really specific to ``/usr/local/lib/python3.7/site-packages/``, it just so happens that's the directory where Fedora ran into it. I think it makes sense for setuptools to make sure that the directory it's trying to write to exists.

pganssle:
@dstufft In what other situations would it come up? I'm not deeply opposed to having setuptools create the directory it's trying to write to, but my worry is that the only reason this would happen is if there's been a horrible mistake (in which case you may end up installing somewhere that won't get picked up - or worse, you'll mess up some user's file system), in which case it makes sense to throw an error instead.

If this only affects people using `setuptools` in a dangerous way, I'm not sure we should be facilitating this at the - admittedly small - risk of failing to report an actual error condition.

dstufft:
@pganssle Not sure, I'm pretty sure that pip has this code though, so I assume it came up in some other context at some point.

hroncok:
pip install creates the directory, yes

silkentrance:
@hroncok have you tried installing pyenv + tox in the various environments, say 2.7.x etc.
Then simply run ``tox`` from inside the ``setuptools`` clone and you will be able to run the tests locally.

hroncok:
I've used tox, I don't remember what was the problem, will try now.

hroncok:
The tests work for me now. However I'd rather see an approval of action before spending time figuring out how to test this if it gets rejected at the end.

jaraco:
> At this point we can safely deprecated setup.py install in favor of pip, like we've deprecated upload and register, right?

Correct. `setup.py install` is only recommended with `--single-version-externally-managed`. However, `setup.py develop` is still supported (as pip relies on it for `pip install -e`), and that might also hit this code path.

jaraco:
@hroncok Can you confirm this issue also affects `setup.py develop`?

hroncok:
Thanks.

PR #1441, opened 2018-07-27
Remove spurious executable permissions

hroncok:
## Summary of changes

Remove spurious executable permissions from files that don't need them.

### Pull Request Checklist
- [ ] Changes have tests (not really, but I really doubt the usefulness) 


jaraco:
This change seems useful. Should we give the same treatment to the other files in the root of the project?

jaraco:
And probably all files in the project that don't need to be executable. I see executable bits on, for example, setuptools/namespaces.py.

hroncok:
launcher.c, pytest.ini, ok, I'll try to get them all.

hroncok:
Amended.

jaraco:
Excellent. Thanks!

Issue #1453, opened 2018-08-11
Tests broken on master due to missing setuptools.tests.fixtures


pganssle:
Not sure what's going on here. I can replicate this locally only after doing `git clean -idx` followed by a new `python boostrap.py`.

benoit-pierre:
I can reproduce the issue. `tox-3.2.1` is the problem, latest successful Travis build was using `3.1.2`.

jaraco:
It's something about the way tox 3.2 creates the venv; because after creating the env, invoking `.tox/python/bin/pytest` manually, replicates the error.

pganssle:
For now we can pin to 3.1.2 (#1454) until we can figure out what's going on.

I'm guessing the issue is that `tox` is starting to test by default against an as-installed version of the packages that it uses, rather than as-seen-by-the-repo-root. Since we don't install the tests, `setuptools.tests.fixtures` shouldn't be importable in that case.

I imagine the solution is to start using some sort of relative imports to pull in the fixtures, or something like that.

jaraco:
This is interesting. The tests only fail if invoked with the console script (`pytest`), but don't fail if invoked by runpy (`python -m pytest`).

benoit-pierre:
I think it's because in the newer tox, pip is called with `python -m pip`, and this result in setuptools being already present in the environment.

pganssle:
@jaraco That strengthens my suspicion. If you invoke `pytest` directly, it deliberately removes the current working directory from your `PYTHONPATH`. `python -m pytest` doesn't do that for whatever reason.

benoit-pierre:
```shell
> ls -d .tox.good/python/lib/python3.7/site-packages/setuptools*                                                                      
.tox.good/python/lib/python3.7/site-packages/setuptools.egg-link
.tox.good/python/lib/python3.7/site-packages/setuptools.pth
> ls -d .tox.bad/python/lib/python3.7/site-packages/setuptools*                                                                                                                                            
.tox.bad/python/lib/python3.7/site-packages/setuptools/
.tox.bad/python/lib/python3.7/site-packages/setuptools.egg-link
.tox.bad/python/lib/python3.7/site-packages/setuptools-39.0.1.dist-info/
.tox.bad/python/lib/python3.7/site-packages/setuptools.pth
```

benoit-pierre:
@pganssle: This is to be expected, no? https://docs.python.org/3/library/sys.html?highlight=sys%20path#sys.path

[...]

PR #1455, opened 2018-08-11
tox: fix configuration so newer tox versions are supported

benoit-pierre:
Starting with tox 3.2.0, the commands for installing/listing dependencies changed from `pip` to `python -m pip`, which is problematic as this will leave bits of the previous (from the virtualenv creation) older version of setuptools installed in the site-packages directory because from pip's point of view, the active setuptools version is the one in the current directory (which cannot be uninstalled).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,6 +9,12 @@ envlist=python
 
 [testenv]
 deps=-rtests/requirements.txt
+# Changed from default (`python -m pip ...`)
+# to prevent the current working directory
+# from being added to `sys.path`.
+install_command={envbindir}/pip install {opts} {packages}
+# Same as above.
+list_dependencies_command={envbindir}/pip freeze
 setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 # TODO: The passed environment variables came from copying other tox.ini files
 # These should probably be individually annotated to explain what needs them.


benoit-pierre:
Fix #1453, alternative to #1454.

benoit-pierre:
Do we need the new behaviour? The commit log in tox is:
> Switch pip invocations to use the module ``-m pip`` instead of direct invocation. This could help avoid some of the shebang limitations.

The only case I can see where it would be useful is when pip itself needs to be updated (a well known issue with using `pip` directly on Windows).

I'm not sure there's another way, like with pip, testing setuptools itself is a bit of a corner case (because it's already installed/present in the environment for installing/managing dependencies).

I'll add some comments explaining was this is necessary.

benoit-pierre:
Actually, moving setuptools' source to a subdirectory would fix the issue. But it would also mess with git's history.

jaraco:
> moving setuptools' source to a subdirectory would fix the issue

Agreed, though I'm also opposed to this idea if it can be avoided for other reasons, as it adds complexity to the project layout and distribution mechanics. I'd like as much as possible for projects to maintain a simple, direct presentation, such that the checkout is as similar as possible to what would be installed, so that one could in theory import the packages/modules direct from the repository code without first requiring a transformation step. Adding a subdirectory violates this goal.

benoit-pierre:
This is also really a bug in pip, IMHO: when installing a package, it should always remove a previous install when there's already one present in the target directory, even if it's not the current active version. @pypa/pip-committers

jaraco:
> Do we need the new behaviour?

Probably not. Happy to proceed with this patch given a link back to the issue.

benoit-pierre:
Amended.

gaborbernat:
In theory a direct invocation should be equal to a ``- m`` invocation, and in case of long paths, e.g. Jenkins runs, the shebang method fails. Hence the change made.

benoit-pierre:
@gaborbernat: maybe those calls should be made from another working directory? Is there a way to change that in the tox configuration?

gaborbernat:
@benoit-pierre not directly at the moment, but I can see an argument they should be invoked from within e.g. the site package, so we should expose a way of doing so.

benoit-pierre:
Yeah, it's only used for the test command invocation. The documentation explicitly states that the command to install dependencies is executed from `{toxinidir}`: http://tox.readthedocs.io/en/latest/config.html#confval-deps=MULTI-LINE-LIST

gaborbernat:
we can add a config that would allow to override it then

PR #1460, opened 2018-08-17
Prepare 40.1.0rc1

pganssle:
I think we're ready to cut a new release. I don't see any breaking changes, so I'm calling this the 40.1.0 release.

I think it might be a good idea to release an 40.1.0rc1 for a few days and then cut the final 40.1.0 release.

Thoughts @benoit-pierre @jaraco @jmbowman?

Anyone have any blockers that they wanted to make sure as in the next release?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,18 @@
+v40.1.0
+-------
+
+* #1410: Deprecated ``upload`` and ``register`` commands.
+* #1312: Introduced find_namespace_packages() to find PEP 420 namespace packages.
+* #1420: Added find_namespace: directive to config parser.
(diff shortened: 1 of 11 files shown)


jaraco:
Happy to see frequent releases 👍 

No need to cut an RC release - there are few systems that will pick up pre-releases, so in most cases any regressions will only be observed after the final release anyway. I just try to be prepared to respond if there are issues (fix the issue, revert changes, or pull a release).

No blockers from me.

pganssle:
OK, I'll give the others a few hours to comment and then make the release.

pganssle:
OK cool, release should be done now.

PR #1465, opened 2018-08-20
setuptools: fix regression with `egg_info` command

### LINKED ISSUE #1462: 40.1.0 breaks version detection on some packages
Hello;

I tried updating setuptools to 40.1.0 in the BuildRoot project, only to find a few packages such as
python-cryptography no longer seem to think setuptools is a version high enough to work.
Another package that has the same issue is python-funcsigs as well.

Downgrading to 40.0.0 fixes the issue.  Any idea as to what could be going on?

Thanks!

benoit-pierre:
Ensure version is tagged only once. Fix #1462.

### DIFF
--- a/changelog.d/1465.change.rst
+++ b/changelog.d/1465.change.rst
@@ -1,0 +1,1 @@
+Fix regression with `egg_info` command when tagging is used.
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -160,7 +160,9 @@ class egg_info(InfoCommon, Command):
 
     def initialize_options(self):
         self.egg_base = None
+        self.egg_name = None
         self.egg_info = None
+        self.egg_version = None
         self.broken_egg_info = False
 
     ####################################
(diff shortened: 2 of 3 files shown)


jaraco:
Thank you, especially for the regression test. Nicely done.

benoit-pierre:
Off-topic: I think the fact that I need to create the PR before I can update it with a towncrier entry (because I need the PR number) really annoying. Don't know how to improve the workflow though...

pganssle:
@benoit-pierre Agreed, that's one of my biggest pet peeves with town crier.

For some background, town crier came out of Twisted, which has a [much more formal process](https://twistedmatrix.com/trac/wiki/ContributingToTwistedLabs) for contributions. For both Twisted and CPython, every PR must have an associated ticket in the bug tracker, and that's the number that's supposed to go in the changelog.

I think we were never quite so formal about it because we wanted people to be able to do drive-by PRs without an associated issue. If we want to alleviate the issue (which I agree is a pain), I see two ways forward:

1. All PRs must have an associated issue, and the news fragment is numbered by issue.
2. Change the workflow so that you can commit a file called `#PR_STUB#.change.rst` in the `changelog.d`, and we write a bot to update the PR when a number becomes available.

PR #1479, opened 2018-09-14
Remove compatibility shim for bytes type

jdufresne:
## Summary of changes

The type `bytes` is available on all supported Pythons. Makes the code more forward compatible with Python 3.

### Pull Request Checklist

### DIFF
--- a/setuptools/glob.py
+++ b/setuptools/glob.py
@@ -3,14 +3,12 @@ Filename globbing utility. Mostly a copy of `glob` from Python 3.5.
 
 Changes include:
  * `yield from` and PEP3102 `*` removed.
- * `bytes` changed to `six.binary_type`.
  * Hidden files are not ignored.
 """
 
 import os
 import re
 import fnmatch
-from setuptools.extern.six import binary_type
 
 __all__ = ["glob", "iglob", "escape"]
 
@@ -92,7 +90,7 @@ def _iglob(pathname, recursive):
 
 def glob1(dirname, pattern):
     if not dirname:
-        if isinstance(pattern, binary_type):
+        if isinstance(pattern, bytes):
             dirname = os.curdir.encode('ASCII')
         else:
             dirname = os.curdir
@@ -129,8 +127,8 @@ def glob2(dirname, pattern):
 # Recursively yields relative pathnames inside a literal directory.
 def _rlistdir(dirname):
     if not dirname:
-        if isinstance(dirname, binary_type):
-            dirname = binary_type(os.curdir, 'ASCII')
+        if isinstance(dirname, bytes):
+            dirname = os.curdir.encode('ASCII')
         else:
             dirname = os.curdir
     try:
@@ -149,7 +147,7 @@ magic_check_bytes = re.compile(b'([*?[])')
 
 
 def has_magic(s):
-    if isinstance(s, binary_type):
+    if isinstance(s, bytes):
         match = magic_check_bytes.search(s)
     else:
         match = magic_check.search(s)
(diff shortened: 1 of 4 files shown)


setuptools/glob.py
```diff
@@ -92,8 +90,8 @@ def _iglob(pathname, recursive):
 
 def glob1(dirname, pattern):
     if not dirname:
-        if isinstance(pattern, binary_type):
-            dirname = os.curdir.encode('ASCII')
+        if isinstance(pattern, bytes):
+            dirname = os.curdir.decode('ASCII')
```
jaraco: So while I think you're right about this change (encode->decode), it should probably be in a separate commit and maybe even a separate PR, as it's very different from the stated intention of the commit.

setuptools/glob.py
```diff
@@ -129,8 +127,8 @@ def glob2(dirname, pattern):
 # Recursively yields relative pathnames inside a literal directory.
 def _rlistdir(dirname):
     if not dirname:
-        if isinstance(dirname, binary_type):
-            dirname = binary_type(os.curdir, 'ASCII')
+        if isinstance(dirname, bytes):
+            dirname = os.curdir.decode('ASCII')
```
jaraco: Again, this seems like a change to do more than just replace binary_type with bytes. How did this code work before?

setuptools/glob.py
benoit-pierre: I think this is wrong: the goal of `glob` is to support this usage too:
```python
python -c 'from setuptools.glob import glob; print(glob(b"*.rst"))'
```
no?
in any case, this fails with this PR.

benoit-pierre: On `master`, from setuptools directory, the results are:
```python
[b'README.rst', b'towncrier_template.rst', b'CHANGES.rst']
```

benoit-pierre: This is probably a good time to add some tests.

benoit-pierre: It looks like the intent of the patch was to change to:
```python
dirname = os.curdir.encode('ASCII')
```

jdufresne:
> This is probably a good time to add some tests.

There is no behavioral change. It is a pure internal refactoring. Do you still want to see tests for this PR? Let me know. If so, I'll add them.

benoit-pierre:
> Do you still want to see tests for this PR? Let me know. If so, I'll add them.

Yes, here are some simple ones to prevent future regressions:
```python
import pytest

from setuptools.glob import glob

from .files import build_files


@pytest.mark.parametrize('tree, pattern, matches', (
    ('', b'', []),
    ('', '', []),
    ('''
     appveyor.yml
     CHANGES.rst
     LICENSE
     MANIFEST.in
     pyproject.toml
     README.rst
     setup.cfg
     setup.py
     ''', '*.rst', ('CHANGES.rst', 'README.rst')),
    ('''
     appveyor.yml
     CHANGES.rst
     LICENSE
     MANIFEST.in
     pyproject.toml
     README.rst
     setup.cfg
     setup.py
     ''', b'*.rst', (b'CHANGES.rst', b'README.rst')),
))
def test_glob(monkeypatch, tmpdir, tree, pattern, matches):
    monkeypatch.chdir(tmpdir)
    build_files({name: '' for name in tree.split()})
    assert list(sorted(glob(pattern))) == list(sorted(matches))
```
To be put in `setuptools/tests/test_glob.py`.

jdufresne:
Thanks. I've added the suggested tests (which do fail with the mistake I temporarily had).

PR #1481, opened 2018-09-14
build_meta sdist directory delegate to --dist-dir

gaborbernat:
This avoids a lot of current bugs (e.g. sdists are always generated in ``dist``, and at end moved out; however ``dist`` folder remains around - so if e.g. the version changes you'll get two sdists now); no need to move around just build in the requested location. Discovered this while trying to use this backend with tox's PEP-517/8.

### DIFF
--- a/changelog.d/1481.change.rst
+++ b/changelog.d/1481.change.rst
@@ -1,0 +1,1 @@
+join the sdist ``--dist-dir`` and the ``build_meta`` sdist directory argument to point to the same (this means the build frontend no longer needs to clean manually the dist dir - to avoid multiple sdist presence, and we no longer need to handle conflicts between these two flags)
--- a/changelog.d/1481.change.rst
+++ b/changelog.d/1481.change.rst
@@ -1,1 +1,1 @@
-join the sdist ``--dist-dir`` and the ``build_meta`` sdist directory argument to point to the same (this means the build frontend no longer needs to clean manually the dist dir - to avoid multiple sdist presence, and we no longer need to handle conflicts between these two flags)
+Join the sdist ``--dist-dir`` and the ``build_meta`` sdist directory argument to point to the same target (meaning the build frontend no longer needs to clean manually the dist dir to avoid multiple sdist presence, and setuptools no longer needs to handle conflicts between the two).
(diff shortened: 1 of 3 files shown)


pganssle:
I think the code coverage failure is just because you have simplified the code and the total number of covered lines went down, so I think we can ignore it.

pganssle:
@gaborbernat Can you add some tests for any of the bugs that this fixes?

gaborbernat:
I can try :+1: there are multiple bugs though.

pganssle:
@gaborbernat "This fixes too many bugs" is a pretty nice problem to have. ;)

benoit-pierre:
@pganssle: what's the deal with netlify? I can't reproduce the issue locally when running `tox -e docs` with Python 2.7.15.

gaborbernat:
@pganssle added two of them

pganssle:
@gaborbernat I rebased this against master because I believe the netlify problem is fixed in master, but that obviously wiped out your signatures. [Here's my alias](https://github.com/pganssle/dotfiles/blob/master/dotfiles/git/.gitconfig#L12) for "rebase-and-sign-all-commits" if you want to use it to do a new rebase.

gaborbernat:
@pganssle added the changelog. I would say this is fairly critical as without it the build frontend (in my case tox) is troublesome to use. E.g. the user needs to manually clean the ``dist`` folder between version changes for a package (which when using ``setuptools-scm`` is at every commit).

gaborbernat:
@benoit-pierre @jaraco @pganssle can we get a merge and release on this? thank you very much. 

@pfmoore did you run into this issue while implementing PEP-517 for pip?

gaborbernat:
``build_wheel`` I think has the same issue; though to be fair if you start with a clean folder the issue does not manifest (which pip does I assume). The problem arrises more if you do subsequent builds (e.g. dist already exists and contains an sdist). Or you specify the ``dist`` as the build dist directory.

pfmoore:
Yeah, I don't think we do that (although tbh we use https://github.com/pypa/pep517 so the details are hidden in there). Does tox use its own wrapper code rather than using that library? (I wouldn't be surprised if it did, the pep517 library isn't particularly well publicised)

gaborbernat:
I was aware of that (I pointed it out for you 😁) but given we need a subset of that, decided not to pull it in.

pfmoore:
lol yes you did. Looks like old age is getting to my brain cells :smile:

jaraco:
Looks good to me. Let's give it a spin.

jaraco:
Releasing as 40.4.0.

gaborbernat:
Managed to validate it, works fine for me :+1:

PR #1482, opened 2018-09-14
netlify: fix build

benoit-pierre:
Pin Sphinx to 1.7.9 until https://github.com/sphinx-doc/sphinx/issues/5417 is fixed.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,4 +2,4 @@
 
 [build]
     publish = "docs/build/html"
-    command = "pip install tox && tox -e docs"
+    command = "pip install tox && tox -e docs --notest && ./.tox/docs/bin/pip install 'Sphinx==1.7.9' && tox -e docs"


pganssle:
It seems like this should go in a requirements file, not in the netlify script, right? Did you do it this way deliberately?

benoit-pierre:
It works fine locally with Python 2.7.15, which is why I only patched it for netlify. #1483 is a better alternative if it works.

benoit-pierre:
I don't see how this is relevant, since tox is being used. Just to be clear, I tested locally with a virtualenv using Python 2.7.15, and `tox -e docs`.

pganssle:
@benoit-pierre I think `tox` keeps a cache. Try deleting `.tox` and running it again.

In any case, how is the readthedocs version installed? We should probably try and keep netlify and readthedocs in sync as much as possible. I don't have access to the readthedocs config through my regular account, and they want some sort of write permissions for my account when I try to log in through GH, so I am hesitant to do so.

benoit-pierre:
I did delete .tox before testing, and the output shows Sphinx 1.8.0 is being used.

benoit-pierre:
I have no idea how readthedocs is managed.

benoit-pierre:
And yeah, the site is asking for an awful lot of permissions when trying to login-in with Github...

jaraco:
> I have no idea how readthedocs is managed.

I think it's configured in the app under my account. I'll try to move the configuration to a declarative file in the repo. I can grant access to the other maintainers as well - just share your rtd username.

jaraco:
In the upstream ticket, users suggested instead pinning to `sphinx!=1.8.0`, allowing 1.8.1 and later.

jaraco:
I think #1490 is a cleaner approach.

PR #1483, opened 2018-09-14
netlify: fix build

benoit-pierre:
Use Python 3.6 to work around https://github.com/sphinx-doc/sphinx/issues/5417.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,4 +2,4 @@
 
 [build]
     publish = "docs/build/html"
-    command = "pip install tox && tox -e docs"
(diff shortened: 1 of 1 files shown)


benoit-pierre:
*sigh*

pganssle:
I do think we should use the latest python to build the documentation.

benoit-pierre:
It should not matter, as long as the Python version we're using is supported, since tox is used, no? It should be fixed in the next version of sphinx (see https://github.com/sphinx-doc/sphinx/pull/5443). For now we can either globally pin sphinx in the requirements or merge #1482.

PR #1484, opened 2018-09-14
netlify: fix build

benoit-pierre:
Use Python 3 to work around https://github.com/sphinx-doc/sphinx/issues/5417.


benoit-pierre:
;(

PR #1485, opened 2018-09-14
netlify: fix build

benoit-pierre:
Work around https://github.com/sphinx-doc/sphinx/issues/5417.

Testing: I wonder if the right tox is being used?


benoit-pierre:
Nope...

PR #1492, opened 2018-09-16
WIP: Make safe_name compliant to PEP 503 and behaviour of pip

jaraco:
This PR revives the work begun in #1324 (merged and reverted).

### DIFF
--- a/changelog.d/1324.change.rst
+++ b/changelog.d/1324.change.rst
@@ -1,0 +1,1 @@
+Make safe_name compliant with PEP 503.
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1312,9 +1312,9 @@ def get_default_cache():
 def safe_name(name):
     """Convert an arbitrary string to a standard distribution name
 
-    Any runs of non-alphanumeric/. characters are replaced with a single '-'.
+    Any runs of non-alphanumeric characters are replaced with a single '-'.
     """
-    return re.sub('[^A-Za-z0-9.]+', '-', name)
+    return re.sub('[^A-Za-z0-9]+', '-', name)
 
 
 def safe_version(version):


jaraco:
@torsava: Can you work on the failing tests please?

torsava:
> @torsava: Can you work on the failing tests please?

Sadly I'm completely swamped right now and won't get to it in the foreseeable future. :/

hroncok:
This has risen from the dead and it bites again, see https://bugzilla.redhat.com/show_bug.cgi?id=1791530

jaraco:
This problem is much more insidious than this PR purports.

It's not obvious from this PR what `safe_name` is used for, so it's not clear that PEP 503 even applies. PEP 503 is about normalizing names in a package index, such that these URLs are equivalent:

- `/jaraco-functools`, `/jaraco.functools`
- `/CherryPy`, `cherrypy`

PEP 503 does not declare the scope of these normalized names. It merely declares a normalization routine for use in the context of a package index. It is clear, however, that package names are allowed to have [dots](https://pypi.org/project/jaraco.functools) and [capital characters](https://pypi.org/project/CherryPy) in them.

However, safe-name is intentionally doing something different. It's generating a safe name for a project from an unsafe name. And the [tests for safe name](https://github.com/pypa/setuptools/blob/54b76e63bb7ace4cd1be29d75f6dcfe3227444f9/pkg_resources/tests/test_resources.py#L646-L651) _explicitly allow_ for capital letters and the dot character and _specifically disallow_ other characters like `$` and `🐍`.

The current behavior is more desirable because it's more restrictive (only ascii letters and select symbols) but also more descriptive (allowing for namespaced packages and camel-cased names), even if these descriptive elements are normalized downstream to further constrain the namespace. Setuptools is focused on building safe, descriptive names while pip and PyPI are focused on resolving descriptive names from less-descriptive, normalized ones. They're different operations.

This one tiny change causes 4/5 of the prior assertions (and all non-degenerate assertions) in testSafeName to fail in addition to four other test failures.

I believe the solution to the (largely underdefined) problem here is going to require quite a bit more than mutating a function. You'll need to more clearly define the problem and express what change should address that problem, and if that change involves constraining the namespace for Python packages, that's probably going to require a separate PEP and some extensive transition effort.

In the meantime, Setuptools is working as intended.

PR #1493, opened 2018-09-16
Move to xenial (Ubuntu 16.04) and test against Python 3.7.

jaraco:
In other projects, I've started testing against Python 3.7 using this technique.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,19 @@
 dist: xenial
 sudo: false
 language: python
+
 python:
-- &latest_py2 2.7
-- 3.6
+- 2.7
 - &latest_py3 3.7
 
 jobs:
   fast_finish: true
   include:
-  - python: *latest_py3
-    env: LANG=C
-  - python: *latest_py2
-    env: LANG=C
-  - stage: deploy (to PyPI for tagged commits)
+  - stage: deploy
     if: tag IS present
     python: *latest_py3
     install: skip
     script: skip
-    after_success: true
-    before_deploy: python bootstrap.py
     deploy:
       provider: pypi
       on:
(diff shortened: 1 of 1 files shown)


jaraco:
Annoyingly, the tests are failing. I see this in the logs:

```
Build language: python
Build group: stable
Build dist: xenial
Build id: 429263907
Job id: 429263911
Runtime kernel version: 4.14.12-041412-generic
travis-build version: dc1da7641
Build image provisioning date and time
Tue Dec  5 20:11:19 UTC 2017
Operating System Details
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty
```

Wait, wtf?

On other projects, the same travis config results in Ubuntu 16.04.

jaraco:
Perhaps implicated: travis-ci/travis-ci#10110

benoit-pierre:
Was that with `sudo: false`?

benoit-pierre:
https://github.com/travis-ci/travis-ci/issues/7260#issuecomment-354350596

jaraco:
> Was that with `sudo: false`?

Yes

benoit-pierre:
So it looks like it's not officially supported, and should not be relied upon.

pganssle:
FWIW, [here's how we do it in `dateutil`](https://github.com/dateutil/dateutil/blob/master/.travis.yml#L23), with `xenial` and `sudo` *only* used for the Python 3.7 test.

Hopefully Travis will improve support for `xenial` so we don't need these workarounds, though.

jaraco:
Most of my projects are running fine with `sudo: false`. Here's a [recent example](https://travis-ci.org/jaraco/path.py/jobs/429029758).

jaraco:
> So it looks like it's not officially supported, and should not be relied upon.

Yes, perhaps. And if Travis can't get support for Xenial over two years after its release, there's something broken with their model. It's probably time to migrate to something more nimble, maybe Circle-CI or Microsoft Team Services or GitLab.

PR #1512, opened 2018-10-19
[WIP] setuptools: add support for reproducible source distributions

benoit-pierre:
## Summary of changes

I decided to take a look at making it possible to have reproducible outputs when using `python setup.py sdist`. This involve 2 changes: always sorting each archive entries so the order is deterministic, and supporting the use of the environment variable `SOURCE_DATE_EPOCH` for setting the timestamp of all entries (the same variable is supported by `wheel` to support reproducible .whl files with `bdist_wheel`).

I merged in part of distutils `archive_util.py` code from the Python 3.7 version (as well as backporting the tests).

Note: When `SOURCE_DATE_EPOCH`, all timestamps will set to the current time (instead of using the files actual timestamps). I dropped support for the `compressed tar file` format.  The `xz'ed tar-file` format is now supported with Python 3.4.

TODO:
- [x] tests
- [ ] news fragment
- [ ] fix `--help-formats` output

### DIFF
--- a/setuptools/archive_util.py
+++ b/setuptools/archive_util.py
@@ -4,15 +4,35 @@ import zipfile
 import tarfile
 import os
 import shutil
+import struct
+import time
 import posixpath
 import contextlib
 from distutils.errors import DistutilsError
+from distutils.dir_util import mkpath
+from distutils import log
+
+try:
+    from pwd import getpwnam
+except ImportError:
+    getpwnam = None
+
+try:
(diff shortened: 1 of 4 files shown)


jaraco:
There's a lot going on here. @benoit-pierre, what's the plan for this PR? I'm going to close it for now as it appears to have languished, but feel free to re-open or re-submit at any point.

PR #1517, opened 2018-10-23
Drop use of six.u

### LINKED ISSUE #1516: Drop use of six.u
Since we no longer support Python 2.6, there's really no reason to use `six.u` anymore, but it is still used in [`easy_install`](https://github.com/pypa/setuptools/blob/5b90a0d9d97f52fc91792a8143670ee1a73f51c5/setuptools/command/easy_install.py#L635):

```python
tmpdir = tempfile.mkdtemp(prefix=six.u("easy_install-"))
```

This can be replaced with a u-prefixed string until we're Python 3-only:

```python
tmpdir = tempfile.mkdtemp(prefix=u"easy_install-")
```

zapstar:
## Summary of changes

Drop use of six.u in the code. Trivial changes, I thought you folks had more than one occurrence of six.u here.

Closes #1516 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1517.change.rst
+++ b/changelog.d/1517.change.rst
@@ -1,0 +1,1 @@
+Dropped use of six.u in favor of `u""` literals.
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -632,7 +632,7 @@ class easy_install(Command):
 
     @contextlib.contextmanager
     def _tmpdir(self):
-        tmpdir = tempfile.mkdtemp(prefix=six.u("easy_install-"))
+        tmpdir = tempfile.mkdtemp(prefix=u"easy_install-")
         try:
             # cast to str as workaround for #709 and #710 and #712
             yield str(tmpdir)


setuptools/tests/test_easy_install.py
```diff
@@ -121,6 +121,21 @@ def test_write_exception(self):
...
             cmd.cant_write_to_target()
 
+    def test__tmpdir(self):
+        """
+        Test that a temporary directory gets created and then destroyed.
+        """
+        dist = Distribution()
+        cmd = ei.easy_install(dist)
+        tmp_dir = None
+        with cmd._tmpdir() as tmp_dir_name:
```
pganssle: We don't usually test the private interface of these functions directly, it's better to try to hit the relevant code indirectly. `_tmpdir` seems to have pretty good code coverage already as it's used every time `easy_install` is called.

zapstar: So no tests required?

pganssle: Very good instincts to add a test, though.

I suspect there is *some* way to test this particular functionality by trying to assert that none of the files written into this directory persist when `easy_install` is done being used, but frankly we're trying to get rid of `easy_install` anyway, so it's probably not worth spending much time working out clever tests for it.

pganssle:
I have dropped the unnecessary test, this can be merged when CI passes.

pganssle:
@zapstar Thanks for your PR, really quick on the draw! Once the CI passes myself or one of the other maintainers will merge it. Feel free to check out other issues in the [help wanted](https://github.com/pypa/setuptools/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) label if you want something with a bit more meat to dig into!

pganssle:
@benoit-pierre Did you want to merge this? I don't know what you need to see with respect to the rolling builds, so I don't want you to miss it if I do the merge.

benoit-pierre:
It's unrelated to rolling builds support, but yes, assuming setuptools suffered from the same bug, let's see how many builds are triggered by the merge!

PR #1520, opened 2018-10-23
Add data_files support in setup.cfg

### LINKED ISSUE #1189: data_files support in setup.cfg
Hello,

I work on a Python project that uses [data_files](https://docs.python.org/2/distutils/setupscript.html#installing-additional-files) in its `setup.py`. While I've moved most of my `setup.py` arguments over to a `setup.cfg` file, I still have `data_files` in `setup.py` because `setup.cfg` doesn't seem to support it.

I could take a stab at a PR to add this support in, but I just wanted to check first. Is this feature missing on purpose, or just hasn't been implemented yet?

ssato:
## Summary of changes

These commits add data_files support in setup.cfg, discussed in the issue #1189, such like the following.
```
[options.data_files]
data = data/48x48/logo.png, data/scale/logo.svg
conf =
    site.d/00_default.conf
    host.d/00_default.conf
```

I think it may close #1189.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -703,6 +703,23 @@ class TestOptions:
         with get_dist(tmpdir) as dist:
             assert dist.entry_points == expected
 
+    def test_data_files(self, tmpdir):
+        fake_env(
+            tmpdir,
+            '[options.data_files]\n'
+            'cfg =\n'
+            '      a/b.conf\n'
+            '      c/d.conf\n'
+            'data = e/f.dat, g/h.dat\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            expected = [
+                ('cfg', ['a/b.conf', 'c/d.conf']),
+                ('data', ['e/f.dat', 'g/h.dat']),
+            ]
+            assert sorted(dist.data_files) == sorted(expected)
+
 saved_dist_init = _Distribution.__init__
 class TestExternalSetters:
     # During creation of the setuptools Distribution() object, we call
--- a/changelog.d/1520.change.rst
+++ b/changelog.d/1520.change.rst
@@ -1,0 +1,1 @@
+Added support for ``data_files`` in ``setup.cfg``.
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2367,6 +2367,11 @@ boilerplate code in some cases.
         src.subpackage1
         src.subpackage2
 
+    [options.data_files]
+    /etc/my_package =
+        site.d/00_default.conf
+        host.d/00_default.conf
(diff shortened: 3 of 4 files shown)


setuptools/tests/test_config.py
```diff
@@ -703,6 +703,22 @@ def test_entry_points(self, tmpdir):
...
+            tmpdir,
+            '[options.data_files]\n'
+            'cfg =\n'
+            '      a/b.conf\n'
+            '      c/d.conf\n'
+            'data = e/f.dat, g/h.dat\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            assert dist.data_files == [
```
pganssle: I believe this is failing because `dist.data_files` gets parsed first to a dict (with arbitrary order on Py < 3.6), then turned into this key/value list thing.

I would look at what the other tests do, but I think a quick fix would be this:

```python
expected = {
    'cfg': ['a/b.conf', 'c/d.conf'],
    'data': ['e/f.dat', 'g/h.dat']
}

    # Order independent comparison
    assert dict(*dist.data_files) == expected
```

ssato: I committed the fix based on your suggestion but data type of expected result is kept.

benoit-pierre:
The documentation should be updated.

pganssle:
@benoit-pierre Good catch, I guess we need to add it to [this table](https://deploy-preview-1520--ecstatic-morse-dada75.netlify.com/setuptools.html#options). Anywhere else?

benoit-pierre:
Yes, and I would add a new column to the table mention the minimum supported version.

ssato:
> @ssato If you want to update the documentation, please do a force-pull from your branch, since I've rewritten the PR's history since your last commit, and that will cause conflicts.

@pganssle Thanks a lot for the cleanups! It looks much better ;-)
I'll not update my branch and keep as it is now, just wait for the merge.

pganssle:
@jaraco @benoit-pierre Normally I'd block on getting an update, but given that we're doing a sprint this weekend, I'm tempted to just merge this as-is and leave the documentation update as an easy `[good first issue]`.

If no one takes it at the sprints this weekend, I can do it myself on Monday, but frankly it's nice to have some very simple fixes that you can use as a demo PR for new contributors.

jaraco:
Works for me. @pganssle Would you file that ticket?

pganssle:
Filed #1522.

dirk-thomas:
Since `distutils` on the very lowest level replaces `-` in keys with `_` (https://github.com/python/cpython/blob/31ec52a9afedd77e36a3ddc31c4c45664b8ac410/Lib/distutils/dist.py#L414) this can't be used to install data files into a location which contains dashes.

PR #1521, opened 2018-10-24
Call os.path.normpath to normalize paths for comp

### LINKED ISSUE #1519: False positive when checking for inconsistent setup script path on Windows
```
> python setup.py develop
running develop
Checking .pth file support in C:\project\venv\Lib\site-packages\
C:\project\venv\Scripts\pythonw.exe -E -c pass
TEST PASSED: C:\project\venv\Lib\site-packages\ appears to support .pth files
error: ("Can't get a consistent path to setup script from installation directory", 'c:\\project\\', 'c:\\project')
```

I am not sure what exactly is the cause (I did not have this problem previously, but suddenly sees it when coming back to work on an old project), but it seems to come from a quirk of `realpath()`:

```pycon
>>> import os
>>> os.path.realpath('.')
'C:\\Users\\uranusjr\\'
>>> os.path.realpath('../')
'C:\\Users\\'
>>> os.path.realpath('..')
'C:\\Users'
```

uranusjr:
Closes #1519.

### Pull Request Checklist
- [x] Changes have tests

I’m not sure how the test should be constructred. It seems silly to just test whether the function is identical to calling the three functions, but any fixed strings would be platform-dependant.

### DIFF
--- a/changelog.d/1519.change.rst
+++ b/changelog.d/1519.change.rst
@@ -1,1 +1,1 @@
-Perform additional path normalization to ensure path values to a directory is always the same, preventing false positives when checking scripts have a consistent prefix to set up on Windows.
+In ``pkg_resources.normalize_path``, additional path normalization is now performed to ensure path values to a directory is always the same, preventing false positives when checking scripts have a consistent prefix to set up on Windows.
--- a/changelog.d/1519.change.rst
+++ b/changelog.d/1519.change.rst
@@ -1,0 +1,1 @@
+Perform additional path normalization to ensure path values to a directory is always the same, preventing false positives when checking scripts have a consistent prefix to set up on Windows.
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -236,3 +236,8 @@ class TestDeepVersionLookupDistutils:
         req = pkg_resources.Requirement.parse('foo>=1.9')
         dist = pkg_resources.WorkingSet([env.paths['lib']]).find(req)
         assert dist.version == version
+
+    def test_normalize_path(self, tmpdir):
+        path = str(tmpdir)
+        expected = os.path.normcase(os.path.normpath(os.path.realpath(path)))
+        assert pkg_resources.normalize_path(path) == expected
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -237,7 +237,55 @@ class TestDeepVersionLookupDistutils:
         dist = pkg_resources.WorkingSet([env.paths['lib']]).find(req)
         assert dist.version == version
 
-    def test_normalize_path(self, tmpdir):
-        path = str(tmpdir)
-        expected = os.path.normcase(os.path.normpath(os.path.realpath(path)))
-        assert pkg_resources.normalize_path(path) == expected
+    @pytest.mark.parametrize(
+        'unnormalized, normalized',
+        [
+            ('foo', 'foo'),
+            ('foo/', 'foo'),
+            ('foo/bar', 'foo/bar'),
+            ('foo/bar/', 'foo/bar'),
+        ],
+    )
+    def test_normalize_path_trailing_sep(self, unnormalized, normalized):
+        """Ensure the trailing slash is cleaned for path comparison.
+
+        See pypa/setuptools#1519.
+        """
+        result_from_unnormalized = pkg_resources.normalize_path(unnormalized)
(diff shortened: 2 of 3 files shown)


uranusjr:
I added a (rather trivial) test, and modified the fragment to describe the behavioural change rather than the code modification.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -236,3 +236,8 @@ def test_version_resolved_from_egg_info(self, env):
         req = pkg_resources.Requirement.parse('foo>=1.9')
         dist = pkg_resources.WorkingSet([env.paths['lib']]).find(req)
         assert dist.version == version
+
+    def test_normalize_path(self, tmpdir):
```
jaraco: In 092ef22, I extracted from your commit just the change for the added test, but as you can see in [this build](https://ci.appveyor.com/project/pypa/setuptools/builds/19789289), even on Windows, that test doesn't capture the failure. I think instead of using `tmpdir` here, you should use a contrived path, something where `pkg_resources.normalize_path()` produces an unexpected result before the patch and an expected result after.

changelog.d/1519.change.rst
```diff
@@ -0,0 +1 @@
+Perform additional path normalization to ensure path values to a directory is always the same, preventing false positives when checking scripts have a consistent prefix to set up on Windows.
```
jaraco: Prefix with 

> In pkg_resources.normalize_path, perform additional...

pganssle: Oooh, a chance to use the new "suggest changes" feature in GH!

```suggestion
In ``pkg_resources.normalize_path``, additional path normalization is now performed to ensure path values to a directory is always the same, preventing false positives when checking scripts have a consistent prefix to set up on Windows.
```

pganssle: I also changed the mood and voice from imperative to present passive indicative while I was in there.

I think we haven't been terribly consistent about the tense and voice and such of the news fragments. I think maybe this weekend we should discuss it and I can write up something about the grammar (though generally it's not that important that we ask contributors to do it, because when preparing a release we can edit the `CHANGES.rst` for consistency after all the fragments are assembled).

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -237,7 +237,20 @@ def test_version_resolved_from_egg_info(self, env):
...
 
-    def test_normalize_path(self, tmpdir):
-        path = str(tmpdir)
-        expected = os.path.normcase(os.path.normpath(os.path.realpath(path)))
-        assert pkg_resources.normalize_path(path) == expected
+    @pytest.mark.parametrize(
+        'suffix, parts',
+        [
+            # Trailing sep should be normalized.
+            ('foo', ('foo')),
```
jaraco: I doubt this does what you intended. It's equivalent to `('foo', 'foo'),`.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -237,7 +237,20 @@ def test_version_resolved_from_egg_info(self, env):
...
+            ('foo/', ('foo')),
+
+            # Casing should be consistent.
+            ('MiXeD/CasE', ('MiXeD', 'CasE')),
+        ],
+    )
+    def test_normalize_path(self, tmpdir, suffix, parts):
+        target = tmpdir.join(*parts)
+        target.mkdir()
+        expected = os.path.normcase(os.path.normpath(os.path.realpath(target)))
```
jaraco: This code just repeats what the primary code does. When I read this test, I can't tell what it's asserting without running it. Better would be to materialize what the expected values are... so you probably want to supply that in the parameters, so that a reader can see what 'MiXeD/CasE' or 'foo/' becomes.

pganssle: @uranusjr If that's the case, I think you have two options:

1. Two separate tests for different platforms
2. Give two non-normalized inputs that should normalize to the same thing and check that `pkg_resources.normalize_path(a) == pkg_resources.normalize_path(b)`

pganssle: You can probably write the second test as a parameterized test, and then have some Windows-specific values that normalize things like `C:\\path\\two\\` and `C:\\path\\two` that are under a `skipif`.

I think all *nix-style paths will be accepted by Windows, but not vice-versa.

uranusjr:
Test cases updated.

pganssle:
We're getting a lot of duplicate reports for this, so I think we should go ahead and merge it. I'd like to cut a release soon.

PR #1531, opened 2018-10-27
Mark Py 2/3-only tests as skip instead of xfail

### LINKED ISSUE #1529: Switch all Python 2 or Python 3 specific tests from xfail to skip
The difference between `xfail` and `skip` is that `xfail` is *supposed to work*, whereas conditional skips are tests that only work in specific situations, so for example [this test](https://github.com/pypa/setuptools/blob/5c8ff5a17840790b68d75e47652264f4f9931101/setuptools/tests/test_namespaces.py#L15) is only *intended* to work on python >= 3.5:

```python
    @pytest.mark.xfail(
        sys.version_info < (3, 5),
        reason="Requires importlib.util.module_from_spec",
    )
    def test_mixed_site_and_non_site(self, tmpdir):
```

It should be marked with `skip` rather than `xfail`. Compare that to [this test, which fails on Windows](https://github.com/pypa/setuptools/blob/a94ccbf404a79d56f9b171024dee361de9a948da/pkg_resources/tests/test_resources.py#L790):

```python
    issue591 = pytest.mark.xfail(platform.system() == 'Windows', reason="#591")

    @issue591
    def test_two_levels_deep(self, symlinked_tmpdir):
```

This test *should* pass on windows, but because of issue #591 it doesn't. That test should remain an xfail.

For starters, let's switch over all the `xfail` tests that should be `skip` tests, such as:

- [`test_mixed_site_and_non_site`](https://github.com/pypa/setuptools/blob/5c8ff5a17840790b68d75e47652264f4f9931101/setuptools/tests/test_namespaces.py#L15)
- [`test_tests_are_run_once`](https://github.com/pypa/setuptools/blob/a94ccbf404a79d56f9b171024dee361de9a948da/setuptools/tests/test_test.py#L96) - the xfail can be removed entirely here, because Python 2.6 is no longer supported
- [`py3_only`](https://github.com/pypa/setuptools/blob/a94ccbf404a79d56f9b171024dee361de9a948da/setuptools/tests/test_sdist.py#L24) in `test_sdist`
- [`py3_only`](https://github.com/pypa/setuptools/blob/a94ccbf404a79d56f9b171024dee361de9a948da/setuptools/tests/test_manifest.py#L23) in `test_manifest`

madig:
## Summary of changes

Mark Python 2/3-only tests as skip instead of xfailing them. Also use predefined `py2_only` and `py3_only` decorators where appropriate.

Closes #1529.

### Pull Request Checklist
- [x] Changes have tests -> only modified existing tests?

### DIFF
--- a/changelog.d/1531.misc.rst
+++ b/changelog.d/1531.misc.rst
@@ -1,0 +1,1 @@
+Converted Python version-specific tests to use ``skipif`` instead of ``xfail``, and removed Python 2.6-specific code from the tests.
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -7,6 +7,7 @@ import pytest
 
 from .files import build_files
 from .textwrap import DALS
+from . import py2_only
 
 __metaclass__ = type
 
@@ -143,7 +144,7 @@ def test_prepare_metadata_for_build_wheel(build_backend):
     assert os.path.isfile(os.path.join(dist_dir, dist_info, 'METADATA'))
 
 
-@pytest.mark.skipif('sys.version_info > (3,)')
+@py2_only
 def test_prepare_metadata_for_build_wheel_with_str(build_backend):
     dist_dir = os.path.abspath(str('pip-dist-info'))
     os.makedirs(dist_dir)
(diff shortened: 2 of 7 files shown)


pganssle:
Thanks for your PR @madig!

PR #1532, opened 2018-10-27
Update setup.cfg documentation to include data_files

### LINKED ISSUE #1522: Update setup.cfg documentation to include data_files
After the implementation of `data_files` support in `setup.cfg` in #1520, we still need to [add it to the documentation](https://github.com/pypa/setuptools/pull/1520#issuecomment-432726077). Specifically, [this table](https://deploy-preview-1520--ecstatic-morse-dada75.netlify.com/setuptools.html#options) should be updated in the following ways:

1. Add a `data_files` row, of type `list-comma`
2. Add an "added in" or "minimum version" column, with `data_files` set to `40.5.0`

Unfortunately I think it's not clear whether the next feature version will be 40.5.0 or 41.0.0, but let's go with 40.5.0 for now and if we decide that there will never be a 40.5.0 release, we'll do a search-and-replace for it.

DanGolding:
## Summary of changes

Closes #1522

### Pull Request Checklist
- [x] Changes have tests
No tests or fragment in changelog.d as this is a small change to documentation.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2455,7 +2455,7 @@ keywords                                           list-comma
 platforms                       platform           list-comma
 provides                                           list-comma
 requires                                           list-comma
-obsoletes                                          dict
+obsoletes                                          list-comma
 ==============================  =================  =====
 
 .. note::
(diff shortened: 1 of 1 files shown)


pganssle:
@DanGolding You seem to have undone your first commit with your second commit, and switched `obsoletes` over to `dict` instead of `data_files`

benoit-pierre:
Done: #1535.

PR #1534, opened 2018-10-27
Document building Cython projects instead of Pyrex

### LINKED ISSUE #1395: Documentation front page mentions pyrex but not cython
I don't think its controversial to claim cython has superseded pyrex (which last had a release 8 years ago and afaik has not py3 support), and that the documentation should reflect this.

xrmx:
Refs #1395

## Summary of changes

Update documentation to mention integrations with Cython instead of Pyrex.
I'm no Cython expert and I've used what https://github.com/thearn/simple-cython-example is doing in its setup.py as a reference. Cython documentation uses distutils and not setuptools https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#basic-setup-py

Closes #1395 

### Pull Request Checklist
- [ ] Changes have tests


pganssle:
Has Pyrex been renamed to Cython, or is it still a going concern? I think I'm fine with just doing an `s/Pyrex/Cython/g`, but we can maybe mention Pyrex somewhere if it's still used?

Alternatively, we can have maybe the first mention changed from `Cython` to `Cython (or Pyrex)`

benoit-pierre:
Last release of Pyrex is from 2006, I think it's fine to switch to Cython and don't mention Pyrex anymore.

docs/setuptools.txt
```diff
@@ -1651,29 +1651,34 @@ See the sections below on the `egg_info`_ and `alias`_ commands for more ideas.
...
-``distutils.Extension``.  You must also not import anything from Pyrex in
-your setup script.
+``setuptools`` will detect at build time whether Cython is installed or not.
+If Cython is not found ``setputools`` will ignore pyx files. In case it's
+available you are supposed it will work with just a couple of adjustments.
+``setuptools`` includes transparent support for building Cython extensions, as
+long as you define your extensions using ``setuptools.Extension``.
+Then you should use Cython own ``build_ext`` in ``cmdclass``, e.g.::
+
+   from Cython.Distutils import build_ext
```
pganssle: I can never keep track, is this still the right way to do it?

I think that there's a way to do this that is more compatible with `setup_requires`, for whatever that's worth.

I'm fine with requiring PEP 518, but if we're going to have code snippets that only meaningfully work with PEP 518, we need to mention how to use `pyproject.toml` as well I think.

@jaraco Thoughts?

PR #1536, opened 2018-10-27
Setuptools will install licenses if included in setup.cfg

### LINKED ISSUE #357: setuptools doesn't install licenses
Originally reported by: **notting (Bitbucket: [notting](http://bitbucket.org/notting), GitHub: [notting](http://github.com/notting))**

---
1. Most python software is licensed under MIT, BSD, or similar licenses
2. These licenses require shipping a copy of the license with distributions of the software.

Ergo, setuptools should add the LICENSE or COPYING file to the egginfo/distinfo directory by default, so licenses get distributed automatically.

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/357

dtaneli:
`python setup.py sdist` now includes the license file if `license_file`
is included in `setup.cfg` unless it is explicitly excluded in `MANIFEST.in`.

Co-Authored-By: Poyzan Nur Taneli <31743851+ptaneli@users.noreply.github.com>

## Summary of changes

Closes #357 

### Pull Request Checklist
- [x] Changes have tests


pganssle:
@dtaneli Can you add some tests for this?

dtaneli:
Hi @pganssle, I added some tests but I am not sure if it is better to merge these four test cases into a single test or not. Any preferences?

Also, does this change warrant a changelog update?

pganssle:
And yes, this is a behavior change, it definitely needs a changelog.

dtaneli:
Made the tests parametrized, added a changelog entry and updated the`setup.cfg` documentation.

@pganssle, during the London sprint, we discussed re-using `license = file:<path>` instead of `license_file = <path>` since the usage of these attributes kind of overlap. Or maybe we would like to support both?

dtaneli:
Ok, I just wanted to make sure because we found that you could actually pass a file according to https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata

----

Is there anything else that is missing for the scope of this issue?

setuptools/command/sdist.py
```diff
@@ -198,3 +199,27 @@ def read_manifest(self):
...
+    def check_license(self):
+        """Read the setup configuration file ('setup.cfg') and use it to find
+        if a license is defined with the 'license_file' attribute.
+        If the license is declared and exists, it will be added to
+        'self.filelist'.
+        """
+
+        cfg_file = 'setup.cfg'
+        log.debug("Reading configuration from %s", cfg_file)
+        parser = configparser.RawConfigParser()
```
jaraco: It's a little disconcerting to me that this routine needs to parse the setup.cfg file. I would expect that whatever logic is needed to include this file by default would happen _after_ the setup.cfg is parsed and the 'license_file' is set as an attribute on some class (Distribution, maybe?). It shouldn't be necessary to re-read the configuration file... or if it is, it should re-use the same code.

Did you or could you explore such an approach?

dtaneli: I initially tried `setuptools.config.read_configuration('setup.cfg')`, which does not parse `license_file` from `[metadata]` even if it exists. (Is this expected?)

After your comment, I did manage to find a way to get this without parsing `setup.cfg` again (see 4e4efa77)

docs/setuptools.txt
```diff
@@ -2448,6 +2448,7 @@ maintainer                                         str
 maintainer_email                maintainer-email   str
 classifiers                     classifier         file:, list-comma
 license                                            file:, str
+license_file                                       str
```
gaborbernat: isn't this duplicating ``license=file:str``?

setuptools/tests/test_egg_info.py
```diff
@@ -486,6 +486,57 @@ def test_doesnt_provides_extra(self, tmpdir_cwd, env):
...
+                              """),
+            'MANIFEST.in': DALS("exclude LICENSE"),
+            'LICENSE': DALS("Test license")
+        }, False) # license file is manually excluded
+    ])
+    def test_setup_cfg_license_file(
+            self, tmpdir_cwd, env, files, license_in_sources):
+        self._create_project()
+        build_files(files)
+        environ = os.environ.copy().update(
```
pganssle: `dict.update` mutates an object in place and returns `None`, so this is a no-op.

`os.environ.copy()` creates a copy of `os.environ`, then `update` mutates that copy and returns `None`. `environ` is set to `None` and the ephemeral copy that was mutated is deleted since nothing holds a reference to it. I would do it this way:

```python
environ = os.environ.copy()
environ['HOME'] = env.paths['home']
```

However, I don't understand what this line is intended to do. Is it actually necessary, considering the tests seem to be passing fine without it?

setuptools/tests/test_egg_info.py
```diff
@@ -486,6 +486,57 @@ def test_doesnt_provides_extra(self, tmpdir_cwd, env):
...
+            HOME=env.paths['home'],
+        )
+        environment.run_setup_py(
+            cmd=['egg_info'],
+            pypath=os.pathsep.join([env.paths['lib'], str(tmpdir_cwd)]),
+            data_stream=1,
+            env=environ,
+        )
+        egg_info_dir = os.path.join('.', 'foo.egg-info')
+        with open(os.path.join(egg_info_dir, 'SOURCES.txt')) as sources_file:
```
pganssle: Can you add blank lines around this context manager? It's a bit hard for me to read everything running together like this.

setuptools/tests/test_egg_info.py
dtaneli: Removed these statements from this test.

I originally added them as this block looked to be a common pattern in this test component (e.g. `test_python_requires_egg_info` or `test_long_description_content_type` etc). I copied it over as I thought that not having this may have unintended side effects that I couldn't spot.

setuptools/command/sdist.py
```diff
@@ -198,3 +199,23 @@ def read_manifest(self):
...
             self.filelist.append(line)
         manifest.close()
+
+    def check_license(self):
+        """Checks if license_file' is configured and adds it to
+        'self.filelist' if the value contains a valid path.
+        """
+
+        opts = self.distribution.get_option_dict('metadata')
+        try:
```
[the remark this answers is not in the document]
dtaneli: done

setuptools/command/sdist.py
```diff
@@ -6,6 +6,7 @@
 import contextlib
 
 from setuptools.extern import six
+from setuptools.extern.six.moves import configparser
```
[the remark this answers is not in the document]
dtaneli: It isn't, removed.

setuptools/tests/test_egg_info.py
dtaneli: done

jakirkham:
Thanks for doing this @dtaneli and @jaraco for reviewing!

PR #1537, opened 2018-10-27
Document using setup.cfg for src-layouts

### LINKED ISSUE #1524: Document how to use setup.cfg with find_packages and the src layout
When trying to create an MWE for #1436, I found it very hard to find the right syntax for using the `src` layout from `setup.cfg`. Here's what I landed on:

```
[options]
package_dir=
 =src
packages=find:
        where=src
```

It would be nice to see this in the documentation somewhere as it's a pretty common layout (and I daresay [probably should be the preferred layout](https://hynek.me/articles/testing-packaging/)).

madig:
## Summary of changes

Document how to use setup.cfg with `src/` layouts.

Closes #1524.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2400,10 +2400,11 @@ Metadata and options are set in the config sections of the same name.
 * Unknown keys are ignored.
 
 
-Using a ``src/`` layout (storing packages in a subdirectory)
-============================================================
+Using a ``src/`` layout
+=======================
 
-If you use a ``src/`` layout, as in, you package structure looks like this::
+One commonly used package configuration has all the module source code in a
+subdirectory (often called the ``src/`` layout), like this::
 
     ├── src
     │   └── mypackage
@@ -2412,7 +2413,7 @@ If you use a ``src/`` layout, as in, you package structure looks like this::
     ├── setup.py
     └── setup.cfg
 
-You can set up your ``setup.cfg`` to automatically look up all your packages in
+You can set up your ``setup.cfg`` to automatically find all your packages in
 the subdirectory like this:
 
 .. code-block:: ini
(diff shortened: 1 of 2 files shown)


docs/setuptools.txt
```diff
@@ -2400,6 +2400,34 @@ Metadata and options are set in the config sections of the same name.
 * Unknown keys are ignored.
 
 
+Using a ``src/`` layout (storing packages in a subdirectory)
```
```suggestion
Using a ``src/`` layout
```

I think maybe we don't need the parenthetical part of this.

docs/setuptools.txt
```diff
@@ -2400,6 +2400,34 @@ Metadata and options are set in the config sections of the same name.
...
+============================================================
+
+If you use a ``src/`` layout, as in, you package structure looks like this::
```
```suggestion
One commonly used package configuration has all the module source code in a subdirectory (often called the `src/` layout), like this::
```

docs/setuptools.txt
```diff
@@ -2400,6 +2400,34 @@ Metadata and options are set in the config sections of the same name.
...
+
+    ├── src
+    │   └── mypackage
+    │       ├── __init__.py
+    │       └── mod1.py
+    ├── setup.py
+    └── setup.cfg
+
+You can set up your ``setup.cfg`` to automatically look up all your packages in
```
```suggestion
You can set up your ``setup.cfg`` to automatically find all your packages in
```

pganssle:
Also, I'm happy to merge with just the documentation, but if you want to it would be nice to have a test that this `setup.cfg` configuration works properly.

madig:
Oh right, I can put a test in somewhere.

Not sure about the double-colon, I just copy-pasted it from further up.

pganssle:
Ah, I see, the second one has the `code block` thing.

jaraco:
Nice work @madig. Thanks for the review @pganssle.

madig:
I can come up with a tets when I get to it :)

PR #1541, opened 2018-10-27
Deprecate requires

### LINKED ISSUE #1374: Deprecate requires keyword
See pypa/packaging-problems#3

If I understand correctly, at least the `requires` keyword can be safely deprecated, with its functions distributed (for now) to `setup_requires`, `install_requires` and `test_requires` as appropriate.

smenon8:
## Summary of changes
Will throw a `DeprecatedWarning` whenever `requires` parameter is used. 

Closes #1374 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -2,8 +2,10 @@ from __future__ import absolute_import, unicode_literals
 import io
 import os
 import sys
+import warnings
 from collections import defaultdict
 from functools import partial
+from functools import wraps
 from importlib import import_module
 
 from distutils.errors import DistutilsOptionError, DistutilsFileError
@@ -399,6 +401,20 @@ class ConfigHandler:
 
             section_parser_method(section_options)
 
+    def _deprecated_config_handler(self, func, msg, warning_class):
+        """ this function will wrap around parameters that are deprecated
+        
+        :param msg: deprecation message 
+        :param warning_class: class of warning exception to be raised
+        :param func: function to be wrapped around
+        """
+        @wraps(func)
+        def config_handler(*args, **kwargs):
+            warnings.warn(msg, warning_class)
+            return func(*args, **kwargs)
+        
+        return config_handler
+
 
 class ConfigMetadataHandler(ConfigHandler):
 
@@ -434,7 +450,10 @@ class ConfigMetadataHandler(ConfigHandler):
             'platforms': parse_list,
             'keywords': parse_list,
             'provides': parse_list,
-            'requires': parse_list,
+            'requires': self._deprecated_config_handler(parse_list,
+                "The requires parameter is deprecated, please use " +
+                "install_requires for runtime dependencies.",
+                DeprecationWarning),
             'obsoletes': parse_list,
             'classifiers': self._get_parser_compound(parse_file, parse_list),
             'license': parse_file,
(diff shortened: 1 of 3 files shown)


pganssle:
@smenon8 We actually want to deprecate this parameter, not remove it, that means that when someone specifies `requires` they should see a deprecation warning. I think this will cause `requires` to be silently ignored, or possibly throw an error?

smenon8:
@pganssle I see, sorry, I misunderstood.

setuptools/config.py
```diff
@@ -399,6 +401,20 @@ def parse(self):
 
             section_parser_method(section_options)
 
+    def deprecated_config_handler(self, msg, warning_class, func):
```
pganssle: This is a bit of a nitpick, but I'd prefer if `func` came first in this case.

pganssle: Also can you name this `_deprecated_config_handler`? We do not want this to be considered part of the public interface.

smenon8: Sure, let me make that change.

setuptools/tests/test_config.py
```diff
@@ -391,6 +391,23 @@ def test_classifiers(self, tmpdir):
...
+    def test_deprecated_config_handlers(self, tmpdir):
+        fake_env(
+            tmpdir,
+            '[metadata]\n'
+            'version = 10.1.1\n'
+            'description = Some description\n'
+            'requires = some, requirement\n'
+        )
+
+        with pytest.raises(DeprecationWarning):
```
pganssle: I think this needs to be `pytest.warns`.

pganssle:
```suggestion
        with pytest.warns(DeprecationWarning):
```

smenon8: Oops. saw this now. I did not realize this. I am using pytest.deprecated_call() instead which I believe is equivalent.

setuptools/config.py
```diff
@@ -399,6 +401,20 @@ def parse(self):
...
+    def _deprecated_config_handler(self, func, msg, warning_class):
+        """ this function will wrap around parameters that are deprecated
+        
+        :param msg: deprecation message 
+        :param warning_class: class of warning exception to be raised
+        :param func: function to be wrapped around
+        """
+        @wraps(func)
+        def config_handler(*args, **kwargs):
+            warnings.warn(msg, warning_class)
```
jaraco: Is there a `stacklevel` parameter we could supply here that would make it easier for the the user to know where the deprecated usage lies?

smenon8: I see. I am not super familiar with `pdb` but I will read up and work on this. Thank you for your comments.

pganssle: I'm not sure `stacklevel` is necessary here, but if it is, the right answer is probably 2.

The problem is that because this is a decorator, the "natural" place for the warning to be emitted is inside the function it's wrapping, which is actually a level in the stack deeper than the one where the issue is actually raised.

Considering this warning is generated in response to a keyword argument to `setup` or a keyword in `setup.cfg`, the code that causes this is not present *anywhere* in the stack. Because the `requires` keyword can be specified in either location, I don't think there's a better place for this warning.

I think the wording of the warning itself is specific enough that hopefully people will understand what it means, but if not I don't think we can fix it with `stacklevel`.

PR #1546, opened 2018-10-28
docs: fix setuptools typo

xrmx:
## Summary of changes

Fix typo introduced in ab4e5b9b3c3404f897e12a9fd882fddc9a36cc77


pganssle:
I think we can fix this in #1550.

jaraco:
Thanks for catching this; will follow up in #1550.

PR #1548, opened 2018-10-28
Isolate tests better

### LINKED ISSUE #1428: Better testing isolation
In https://github.com/pypa/setuptools/issues/1425 it was discovered that a test was failing due to a configuration file in the user's home directory. Some tests also populate the global scope and fail the second time they are run.

By using pytest's `tmpdir` fixture and setting some environmental variables, it should be possible to improve the isolation of tests (at the expense of adding a bit more runtime to the test suite).

An example of how it's done from the `pip` test suite: https://github.com/pypa/pip/blob/master/tests/conftest.py#L71-L137

donjar:
## Summary of changes

This pull request isolates tests better by auto-using a global `tmpdir` Pytest fixture, which is copied and slightly modified from https://github.com/pypa/pip/blob/db127330d4e071c84c5e1c107f60f8229e7d68ce/tests/conftest.py#L79-L148.

Closes #1428.

### Pull Request Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,7 +1,6 @@
 import sys
 import pytest
 from path import Path
-import os
 
 
 pytest_plugins = 'setuptools.tests.fixtures'
@@ -32,7 +31,7 @@ if sys.version_info < (3, 6):
     collect_ignore.append('pavement.py')
 
 
-@pytest.yield_fixture
+@pytest.fixture(autouse=True)
 def global_tmpdir(request, tmpdir):
     """
     Return a temporary directory path object which is unique to each test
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -23,3 +30,82 @@ def pytest_addoption(parser):
...
+
+    # Create a directory to use as our home location.
+    home_dir = os.path.join(str(global_tmpdir), "home")
+    os.makedirs(home_dir)
+
+    # Create a directory to use as a fake root
+    fake_root = os.path.join(str(global_tmpdir), "fake-root")
+    os.makedirs(fake_root)
+
+    if sys.platform == 'win32':
```
pganssle: Hm... Do we need all this stuff with `USERPROFILE` and such?

benoit-pierre: I'm not sure we do... Copy-pasting the whole function from pip's testsuite is a bit much, we certainly don't need those pip related changes to the environment (`PIP_DISABLE_PIP_VERSION_CHECK` and `PIP_REQ_TRACKER`).

donjar: In fact, I think I can remove the `isolate` method entirely and just use `global_tmpdir`. What do you think?

conftest.py
```diff
@@ -23,3 +29,23 @@ def pytest_addoption(parser):
...
+    Return a temporary directory path object which is unique to each test
+    function invocation, created as a sub directory of the base temporary
+    directory. The returned object is a ``tests.lib.path.Path`` object.
+
+    This uses the built-in tmpdir fixture from pytest itself but modified
+    to return our typical path object instead of py.path.local as well as
+    deleting the temporary directories at the end of each test case.
+    """
+    assert tmpdir.isdir()
+    yield Path(str(tmpdir))
```
jaraco: You don't need to cast to `str` first, as `path.Path` is a subclass of `str`/`unicode`.

jaraco:
I'm going to close this PR for now, but I welcome a follow-up PR (possibly reviving or re-purposing these commits) once we have a better story for what we're after.

Issue #1549, opened 2018-10-28
Review Cython docs for best practices


jaraco:
In [this comment](https://github.com/pypa/setuptools/pull/1534/files#r228717921), Paul clearly highlights a potential issue with the PR. In my review, it wasn't clear to me that the docs were making a specific recommendation. I'd like to revisit that change - probably remove that paragraph.

PR #1550, opened 2018-10-28
Update recommendation.

jaraco:
Per #1549, in #1534, I realized that the Cython recommendation was introduced following a very old recommendation. I had missed in the review that new recommendation.

In this PR, I supply instead what I believe to be the best practices.

@xrmx, would you review this change and let me know if there's any reason to think this technique doesn't work for your use-case?

@pganssle Does this recommendation match your expectation?

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -1655,17 +1655,26 @@ Distributing Extensions compiled with Cython
 --------------------------------------------
 
 ``setuptools`` will detect at build time whether Cython is installed or not.
-If Cython is not found ``setputools`` will ignore pyx files. In case it's
-available you are supposed it will work with just a couple of adjustments.
-``setuptools`` includes transparent support for building Cython extensions, as
-long as you define your extensions using ``setuptools.Extension``.
-Then you should use Cython own ``build_ext`` in ``cmdclass``, e.g.::
+If Cython is not found ``setputools`` will ignore pyx files.
 
-   from Cython.Distutils import build_ext
+To ensure Cython is available, include Cython in the build-requires section
+of your pyproject.toml::
 
-   setup(...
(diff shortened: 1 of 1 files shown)


xrmx:
@jaraco Not sure i can review this as i never wrote a pyproject.toml or a setup.cfg for that matter :) If that works and is proper modern packaging it's fine for me.

docs/setuptools.txt
```diff
@@ -1655,17 +1655,26 @@ Distributing Extensions compiled with Cython
 --------------------------------------------
 
 ``setuptools`` will detect at build time whether Cython is installed or not.
-If Cython is not found ``setputools`` will ignore pyx files. In case it's
-available you are supposed it will work with just a couple of adjustments.
-``setuptools`` includes transparent support for building Cython extensions, as
-long as you define your extensions using ``setuptools.Extension``.
-Then you should use Cython own ``build_ext`` in ``cmdclass``, e.g.::
+If Cython is not found ``setputools`` will ignore pyx files.
```
```suggestion
If Cython is not found ``setuptools`` will ignore pyx files.
```

docs/setuptools.txt
```diff
@@ -1655,17 +1655,26 @@ Distributing Extensions compiled with Cython
...
 
-   from Cython.Distutils import build_ext
+To ensure Cython is available, include Cython in the build-requires section
+of your pyproject.toml::
```
pganssle: Would be nice to link this to something that shows how to use `pyproject.toml`, since it's not obvious what goes in place of the `...`, but we also don't want to distract from the important part in this example.

jaraco: Yeah, I was trying not to be too prescriptive, but at the same time, it's probably important that 'setuptools' and 'wheel' are included there, as defining the build deps overrides the defaults (IIUC).

PR #1554, opened 2018-10-28
Egg include setup py

### LINKED ISSUE #1506: build_meta does not include setup.py
When someone invokes the build operation via the ``build_meta`` (via https://github.com/pypa/setuptools/blob/master/setuptools/build_meta.py#L170 - https://www.python.org/dev/peps/pep-0517/#build-sdist) the ``setup.py`` is no longer automatically packaged.

In case of ``python setup.py sdist`` what ensured that this will happen is https://github.com/python/cpython/blob/master/Lib/distutils/command/sdist.py#L250. Here ``distribution.script_name`` is set to ``setup.py`` when called via ``python setup.py sdist``. This is not set though when calling it directly via ``build_meta`` so packaging ``setup.py`` is safely ignored. 

Note the files to package is determined while running ``egg_info``, and this is re-used across runs (by reading ``SOURCES.txt``). Therefore this bug is hidden when:
- anyone beforehand calls ``python setup.py sdist``,
- when there's a distutils extension that automatically pulls in setup.py (such as setuptools-scm), 
- when some config value pulls in (one in ``MANIFEST.in``, ``setup.cfg``, ``setup.py``).

shashanksingh28:
## Summary of changes

While creating sdists, we want the source dist .tar.gz to contain `setup.py` because this is what pip would use to install the source distribution. `python setup.py sdist` ensured setup.py was included because distutils by default adds the [script_name](https://github.com/python/cpython/blob/master/Lib/distutils/command/sdist.py#L250) that was executing the build process. However, for someone calling `build_sdist('.')` programatically, the script name is the script callind build_sdist function and not `setup.py`.

The problem is that sdist command uses the `egg-info` command to extract file list to be tar-ziped, which did not include setup.py as default (unless it was called via 'setup.py' as the script name). This fix adds an explicit check in the `egg-info` command for setup.py.

An alternative solution is solving this one level deeper at distutils [add_defaults](https://github.com/python/cpython/blob/master/Lib/distutils/command/sdist.py#L210) method which `egg-info` command [uses](https://github.com/pypa/setuptools/blob/master/setuptools/command/egg_info.py#L570).

Closes #1506 

Will update changelog once we agree on the approach

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1554.change.rst
+++ b/changelog.d/1554.change.rst
@@ -1,0 +1,1 @@
+``build_meta.build_sdist`` now includes ``setup.py`` in source distributions by default
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -575,6 +575,12 @@ class manifest_maker(sdist):
             self.filelist.extend(rcfiles)
         elif os.path.exists(self.manifest):
             self.read_manifest()
+
+        if os.path.exists("setup.py"):
+            # setup.py should be included by default, even if it's not
+            # the script called to create the sdist
+            self.filelist.append("setup.py")
+
         ei_cmd = self.get_finalized_command('egg_info')
         self.filelist.graft(ei_cmd.egg_info)
 
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -2,9 +2,11 @@ from __future__ import unicode_literals
 
 import os
 import shutil
+import tarfile
 
 import pytest
 
+from setuptools.build_meta import build_sdist
 from .files import build_files
 from .textwrap import DALS
 from . import py2_only
@@ -181,3 +183,13 @@ def test_build_sdist_version_change(build_backend):
 
     sdist_name = build_backend.build_sdist("out_sdist")
     assert os.path.isfile(os.path.join(os.path.abspath("out_sdist"), sdist_name))
+
+
+def test_build_sdist_setup_py_exists(tmpdir_cwd):
+    # If build_sdist is called from a script other than setup.py,
+    # ensure setup.py is include
+    build_files(defns[0])
+    targz_path = build_sdist("temp")
+    with tarfile.open(os.path.join("temp", targz_path)) as tar:
+        assert any('setup.py' in name for name in tar.getnames())
+
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -193,3 +193,24 @@ def test_build_sdist_setup_py_exists(tmpdir_cwd):
     with tarfile.open(os.path.join("temp", targz_path)) as tar:
         assert any('setup.py' in name for name in tar.getnames())
 
+
+def test_build_sdist_setup_py_manifest_excluded(tmpdir_cwd):
+    # Ensure that MANIFEST.in can exclude setup.py
+    files = {
+        'setup.py': DALS("""
+    __import__('setuptools').setup(
+        name='foo',
+        version='0.0.0',
+        py_modules=['hello']
+    )"""),
+        'hello.py': '',
+        'MANIFEST.in': DALS("""
+    exclude setup.py
+    """)
+    }
+
+    build_files(files)
+    targz_path = build_sdist("temp")
+    with tarfile.open(os.path.join("temp", targz_path)) as tar:
+        assert not any('setup.py' in name for name in tar.getnames())
+
--- a/setuptools/tests/test_sdist.py
+++ b/setuptools/tests/test_sdist.py
@@ -92,9 +92,8 @@ fail_on_latin1_encoded_filenames = pytest.mark.xfail(
 class TestSdistTest:
     def setup_method(self, method):
         self.temp_dir = tempfile.mkdtemp()
-        f = open(os.path.join(self.temp_dir, 'setup.py'), 'w')
-        f.write(SETUP_PY)
-        f.close()
+        with open(os.path.join(self.temp_dir, 'setup.py'), 'w') as f:
+            f.write(SETUP_PY)
 
         # Set up the rest of the test package
         test_pkg = os.path.join(self.temp_dir, 'sdist_test')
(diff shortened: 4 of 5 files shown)


setuptools/tests/test_sdist.py
```diff
@@ -92,9 +92,10 @@ def latin1_fail():
 class TestSdistTest:
     def setup_method(self, method):
         self.temp_dir = tempfile.mkdtemp()
-        f = open(os.path.join(self.temp_dir, 'setup.py'), 'w')
-        f.write(SETUP_PY)
-        f.close()
+        if not method == self.test_setup_py_missing:
+            f = open(os.path.join(self.temp_dir, 'setup.py'), 'w')
```
pganssle: Can you use a context manager here?

setuptools/tests/test_sdist.py
```diff
@@ -92,9 +92,10 @@ def latin1_fail():
 class TestSdistTest:
     def setup_method(self, method):
         self.temp_dir = tempfile.mkdtemp()
-        f = open(os.path.join(self.temp_dir, 'setup.py'), 'w')
-        f.write(SETUP_PY)
-        f.close()
+        if not method == self.test_setup_py_missing:
```
pganssle: I don't love this, it unnecessarily couples `setup_method` and `test_setup_py_missing`.

I'd rather just have `test_setup_py_missing` in a different class, or create a more general method for registering methods that don't need this setup.

setuptools/tests/test_build_meta.py
```diff
@@ -184,24 +185,10 @@ def test_build_sdist_version_change(build_backend):
...
-    from setuptools.build_meta import build_sdist
     build_files(defns[0])
-    targz_path = build_sdist(".")
-    tar = tarfile.open(targz_path)
+    targz_path = build_sdist("temp")
+    tar = tarfile.open(os.path.join("temp", targz_path))
     assert any('setup.py' in name for name in tar.getnames())
 
-
-@pytest.mark.xfail
```
pganssle: Why the removal? is this not the desired behavior?

pganssle:
@gaborbernat Do you mind taking a look at this? Are there tests that cover your use case?

setuptools/tests/test_build_meta.py
shashanksingh28: That test was my misunderstanding of the [sdist_directory](https://www.python.org/dev/peps/pep-0517/#build-sdist) argument. I thought it meant source directory of the sdist (source-tree) and thus wrote this test that expected `build_sdist("/package/src")` and it should build an sdist from that directory.

On reading PEP517, I understood it means 'destination_directory` for placing the sdist and thus this test stood invalid. I guess the PEP implies that the build_meta commands will always be executed from the source_tree as the cwd() (containing setup.py). 

All other tests are valid though, that test it from the cwd() but with a different script name.

changelog.d/1554.change.rst
```diff
@@ -0,0 +1 @@
+``build_meta.build_sdist`` includes the *setup.py* file by default for scenarios where the invoking script is not "setup.py".
```
gaborbernat: This should be always the case, maybe just say that build sdist ensures that setup.py is always and automatically packaged.

setuptools/command/egg_info.py
```diff
@@ -575,6 +575,12 @@ def add_defaults(self):
             self.filelist.extend(rcfiles)
         elif os.path.exists(self.manifest):
             self.read_manifest()
+        if os.path.exists("setup.py"):
+            """If there exists a setup.py, it should be included in the sdist irrespective
```
gaborbernat: maybe would make sense to check that it's not already included, what happens if we double include stuff? E.g. via manifest and here. If it's ignored is ok as it is. We check for its existence but how and when would it not be available? Isn't setuptools pre requisites to exist?

setuptools/tests/test_build_meta.py
```diff
@@ -181,3 +182,13 @@ def test_build_sdist_version_change(build_backend):
 
     sdist_name = build_backend.build_sdist("out_sdist")
     assert os.path.isfile(os.path.join(os.path.abspath("out_sdist"), sdist_name))
+
+
+def test_build_meta_setup_py_exists(tmpdir_cwd):
+    # if build_sdist is called not from a script other than setup.py, ensure setup.py is included
```
gaborbernat: arguably this is not needed, if it's added to the egg I think the systems ensures it's packaged into the sdist.

setuptools/tests/test_sdist.py
```diff
@@ -135,6 +134,47 @@ def test_package_data_in_sdist(self):
...
+        if os.path.exists("setup.py"):
+            os.remove("setup.py")
+        with quiet():
+            cmd.run()
+
+        manifest = cmd.filelist.files
+        assert 'setup.py' not in manifest
+
+    def test_setup_py_excluded(self):
+        with open("MANIFEST.in", "w") as manifest_file:
```
gaborbernat: @pganssle can it work without setup.py? shouldn't we break the build with an error if it's missing?

shashanksingh28: From [PEP-517](https://www.python.org/dev/peps/pep-0517/)
> Where the build-backend key exists, it takes precedence over setup.py, and source trees need not include setup.py at all. Projects may still wish to include a setup.py for compatibility with tools that do not use this spec.

So there could be cases where someone has explicitly excluded from the Manifest and has a pyproject.toml

setuptools/command/egg_info.py
pganssle: Ideally you'd be able to just specify `setup.cfg` and `setuptools` as a build backend, and you could get away with no `setup.py` at all.

My preference would be a world where the *common case* is that there's no `setup.py` (though that's basically impossible at the moment).

setuptools/tests/test_sdist.py
pganssle: We did talk about this, and I think it's fine. You have to affirmatively remove `setup.py`. At the moment I see no reason to do that, but as long as it's done explicitly I think we have to trust the user that they have a good reason.

pganssle:
@shashanksingh28 Thanks for your PR, and thanks for following up with all the changes!

shashanksingh28:
Thanks @pganssle and @gaborbernat for your help :) This was my first open source PR, will try to be more involved

PR #1556, opened 2018-10-28
Fix upload metadata

### LINKED ISSUE #1381: Include most recent upload metadata when uploading to PyPI

jpurviance:
With the help of @pganssle 

## Summary of changes

Closes #1381

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/command/upload.py
+++ b/setuptools/command/upload.py
@@ -84,7 +84,7 @@ class upload(orig.upload):
             'md5_digest': hashlib.md5(content).hexdigest(),
 
             # additional meta-data
-            'metadata_version': '1.0',
+            'metadata_version': meta.metadata_version,
             'summary': meta.get_description(),
             'home_page': meta.get_url(),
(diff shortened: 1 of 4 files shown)


benoit-pierre:
Isn't the `upload` command deprecated?

pganssle:
@benoit-pierre Yes. Per #1381, we're going to fix this because this bug has been poisoning PyPI with bad metadata, and for better or for worse, people do keep using `setup.py upload`.

pganssle:
@jpurviance Why did you close and delete the branch?

PR #1553, opened 2018-10-28
Remove ez_setup from documentation

varunkamath18:
## Summary of changes

Split up previous "Installing setuptools" section into the following:
1. Use link to pip installation instead of Easy install instructions 
2. Move "Custom installation instructions" to documentation for creating virtual environments.

Closes #1540

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

PR #1547, opened 2018-10-28
Document that rpmbuild is required for bdist_rpm

xrmx:
Refs #1456

## Summary of changes

Document that rpmbuild is required for bdist_rpm, if distutils does not find it it'll use rpm which does not have (anymore?) the -b switch used to build packages.

Closes #1456 at least for the setuptools side

### Pull Request Checklist
- [x] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

Issue #1515, opened 2018-10-23
PEP 518 and the future of custom or directly invoked commands?


pganssle:
I think the current trajectory of `setuptools` is that we're deprecating direct invocation of *most* if not all custom commands, and as such the PEP 518 support is moving entirely to `pip`. This is fine for `install` and `wheel` and presumably `build` and several other commands will *also* be implemented in `pip`, while `test` is switching over to `pytest` and `upload` is switching over to `twine`.

However, people have long been able to write custom commands and have integrated them into their workflow, so they invoke `python setup.py lint` or `python setup.py clean` or something. What are we planning on doing for all of those?

My $0.02 is that I think we can't realistically get rid of these (and, frankly, I'm not sure we want to), but invocation via `setup.py` is a non-starter because the dependencies really have to happen **before** the script is invoked.

As such, I suggest that we create a new entry point script distributed with `setuptools`, like `pysetup`. This would be a thin wrapper around `python setup.py` that is PEP 518-aware. It would install the required dependencies in an isolated build environment and could even address #1201 with a `[tool.setup]` section of `pyproject.toml`.

pfmoore:
I think what's likely needed is a standalone "build environment" library that handles the common task of creating a PEP 517 style isolated build environment (note that it's PEP 517 that discusses isolated builds, rather than PEP 518 - although as this issue proves, it's not something that's limited to PEP 517 implementations...)

The https://github.com/pypa/pep517 project includes a build environment abstraction that might be useful here. Alternatively, pip's own build isolation code might be something we could split out (or merge with the pep517 library).

Build isolation code in pip has proved quite tricky to get right, so it's definitely something that's worth doing once and sharing among projects :-)

pganssle:
@pfmoore I think the question of how the isolated build environment is implemented is important (and I agree that a standalone library for building it would make this *much* easier to implement), but we may want to start by determining whether having a command line script that is just a PEP 518-aware version of invoking `python setup.py` is a good idea in the first place.

pfmoore:
Sure - that's not something I have much of an opinion on, though, so I'll defer to others on that point :-)

[...]

PR #1560, opened 2018-10-28
Update documentation for setuptools distribution

### LINKED ISSUE #1555: What users should know should not refer to easy_install or ez_setup
This is an extension of #1540 where it was found that ez_setup and easy_install are also called in other parts of the documentation [here](file:///C:/Users/varun/Github/setuptools/docs/build/html/setuptools.html#id28)

These parts should also be updated to refer to pip install.

varunkamath18:
## Summary of changes

* update setuptools documentatoin with a quick distribution guide.
* update index to show easy_install and ez_setup to the bottom.
* move ez_setup instructions to its own file.

Closes #1555 

### Pull Request Checklist

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -1223,125 +1223,53 @@ the quoted part.
 Distributing a ``setuptools``-based project
 ===========================================
 
-Using ``setuptools``...  Without bundling it!
----------------------------------------------
+Detailed instructions to distribute a setuptools project can be found at 
+`Packaging project tutorials`_.
+
+.. _Packaging project tutorials: https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives
+
+Before you begin, make sure you have the latest versions of setuptools and wheel::
+
+    python3 -m pip install --user --upgrade setuptools wheel
+
+To build a setuptools project, run this command from the same directory where
+setup.py is located::
+
+    python3 setup.py sdist bdist_wheel
+
+This will generate distribution archives in the `dist` directory.
+
+Before you upload the generated archives make sure you're registered on 
+https://test.pypi.org/account/register/. You will also need to verify your email
+to be able to upload any packages.
+You should install twine to be able to upload packages::
+
+    python3 -m pip install --user --upgrade setuptools wheel
+
+Now, to upload these archives, run::
+
+    twine upload --repository-url https://test.pypi.org/legacy/ dist/*
+
+To install your newly uploaded package ``example_pkg``,  you can use pip::
+
+    python3 -m pip install --index-url https://test.pypi.org/simple/ example_pkg
+
+If you have issues at any point, please refer to `Packaging project tutorials`_
+for clarification.
+
+Distributing legacy ``setuptools`` projects using ez_setup.py
+-------------------------------------------------------------
 
 .. warning:: **ez_setup** is deprecated in favor of PIP with **PEP-518** support.
 
-Your users might not have ``setuptools`` installed on their machines, or even
-if they do, it might not be the right version.  Fixing this is easy; just
-download `ez_setup.py`_, and put it in the same directory as your ``setup.py``
-script.  (Be sure to add it to your revision control system, too.)  Then add
-these two lines to the very top of your setup script, before the script imports
-anything from setuptools:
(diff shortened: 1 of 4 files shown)


pganssle:
I think this branch was maybe based off of the pre-merge version of #1553.

I've cleaned up the history and rebased against master, so @varunkamath18 if you make any further changes please force-pull from your current branch first. Thanks!

varunkamath18:
Right, my bad.

docs/setuptools.txt
```diff
@@ -1223,121 +1223,53 @@ the quoted part.
...
-.. warning:: **ez_setup** is deprecated in favor of PIP with **PEP-518** support.
+.. _Packaging project tutorials: https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives
 
-Your users might not have ``setuptools`` installed on their machines, or even
-if they do, it might not be the right version.  Fixing this is easy; just
-download `ez_setup.py`_, and put it in the same directory as your ``setup.py``
-script.  (Be sure to add it to your revision control system, too.)  Then add
-these two lines to the very top of your setup script, before the script imports
-anything from setuptools:
+Before you begin. make sure you have the latest versions of setuptools and wheel::
```
```suggestion
Before you begin, make sure you have the latest versions of setuptools and wheel::
```

docs/setuptools.txt
```diff
@@ -1223,121 +1223,53 @@ the quoted part.
...
-    ez_setup.use_setuptools()
+To build a setuptools project, run this command from the same directory where
+setup.py is located::
 
-That's it.  The ``ez_setup`` module will automatically download a matching
-version of ``setuptools`` from PyPI, if it isn't present on the target system.
-Whenever you install an updated version of setuptools, you should also update
-your projects' ``ez_setup.py`` files, so that a matching version gets installed
-on the target machine(s).
+    python3 setup.py sdist bdist_wheel
```
pganssle: I'm inclined to leave this for a subsequent PR, but I think it would be preferable not to suggest `python setup.py bdist_wheel` now that `pip wheel` is a thing.

varunkamath18: Would it be better if we create a new issue on both setuptools and [python-packaing-user-guide](https://github.com/pypa/python-packaging-user-guide/blob/master/source/tutorials/packaging-projects.rst#generating-distribution-archives) because I think both places will need changes.
There is also [Issue #539](https://github.com/pypa/python-packaging-user-guide/issues/539) with some existing discussion regarding the suggestion on uploading sdists.

pganssle: We'll just defer this for now. `pip wheel` doesn't do exactly what `python setup.py bdist_wheel` does, so it may be a situation where we solve #1515 and then migrate the guide over to `setuptools <x>` instead.

pganssle:
@benoit-pierre 

> Why even bother duplicating the packaging guide section? Just link to it.

During the sprints, I suggested that we have a brief intro to the overall process on our documentation, just so people know what the general recommendation was without having to read the entire tutorial to figure out what's done.

I think that what's in this PR now is more in-depth than I was thinking, but I didn't want to nit-pick about it in this PR, I'd rather get `ez_setup` out of the main documentation workflow ASAP, then in a subsequent PR tighten up this section of the docs.

> I would also remove the ez_setup section completely, since support stopped at 33.1.1 (released more than 1 year ago).

I'm sympathetic to this, but we don't make archived versions of the `setuptools` documentation available, and I figured a "baby step" would be to move it out of the main documentation - that way, people who want to see how it *used* to be done (perhaps if they are writing a migration guide or something).

An alternative to this would be to change RTD so that in addition to 'stable' and 'latest', there are also builds for all the minor release tags.

I'm open to removing it completely, though again I'd like to resolve this sooner rather than later, so I'm inclined to merge this (which is a clear improvement to the existing documentation), and have a second PR that trims it down.

pganssle:
Hm, when rebasing this against master, I realized that the move didn't include the changes from #1547. I've incorporated them into the latest version.

That may be one reason to keep these docs around for a bit longer, since we probably still do need to include information about things like `bdist_rpm` and `bdist_deb`, even if to just say, "These are deprecated".

pganssle:
Thanks @varunkamath18

PR #1562, opened 2018-10-29
WIP: Add metadata validation

### LINKED ISSUE #1390: Newlines in the `description` field produce a malformed PKG-INFO
We discovered this accidentally by way of https://github.com/zopefoundation/zc.relation/issues/4#issuecomment-397532224: if you pass a string containing newlines to the `description` argument of `setup()`, setuptools will generate a malformed PKG-INFO.

To reproduce:

```
# setup.py
from setuptools import setup
setup(
    name='test-package',
    version='0.1',
    author='Blah Blah',
    author_email='blah@example.com',
    description='description\n\n',
    py_modules=['blah'],
)
```
(The contents of `blah.py` do not matter, but the file should exist.)

Run `python setup.py sdist` and the inspect `test_package.egg-info/PKG-INFO`.  For me, with setuptools 39.1.0, it looks like this:

```
Metadata-Version: 1.0
Name: test-package
Version: 0.1
Summary: description


Home-page: UNKNOWN
Author: Blah Blah
Author-email: blah@example.com
License: UNKNOWN
Description: UNKNOWN
Platform: UNKNOWN
```

The extra newlines lead tools to treat the rest of the PKG-INFO as a long_description.

I would expect `setuptools` to complain about the newlines in the `description` field, or at least escape them properly (i.e. prepend whitespace, like it does for the `long_description` field).

bittner:
## Summary of changes

Adds metadata validation that aborts packaging with invalid package information.

Closes #1390

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -25,6 +25,7 @@ from setuptools.depends import Require
 from setuptools import windows_support
 from setuptools.monkey import get_unpatched
 from setuptools.config import parse_configuration
+from setuptools.validation import Metadata
 import pkg_resources
 from .py36compat import Distribution_parse_config_files
 
@@ -55,17 +56,41 @@ def get_metadata_version(dist_md):
 def write_pkg_file(self, file):
     """Write the PKG-INFO format data to a file object.
     """
+    metadata = Metadata(
+        name=self.get_name(),
+        version=self.get_version(),
+        description=self.get_description(),
+        url=self.get_url(),
+        author=self.get_contact(),
+        author_email=self.get_contact_email(),
+        license=self.get_license(),
+        download_url=self.download_url,
+        project_urls=['%s, %s' % (key, val)
+                      for key, val in self.project_urls.items()],
+        long_description=rfc822_escape(self.get_long_description()),
+        long_description_content_type=self.long_description_content_type,
+        keywords=self.get_keywords(),
+        platforms=self.get_platforms(),
+        classifiers=self.get_classifiers(),
+        requires=self.get_requires(),
+        provides=self.get_provides(),
+        obsoletes=self.get_obsoletes(),
+        python_requires=getattr(self, 'python_requires', None),
+        provides_extras=self.provides_extras,
+    )
+    metadata.validate(throw_exception=True)
+
     version = get_metadata_version(self)
 
     file.write('Metadata-Version: %s\n' % version)
-    file.write('Name: %s\n' % self.get_name())
-    file.write('Version: %s\n' % self.get_version())
-    file.write('Summary: %s\n' % self.get_description())
-    file.write('Home-page: %s\n' % self.get_url())
+    file.write('Name: %s\n' % metadata.name)
+    file.write('Version: %s\n' % metadata.version)
+    file.write('Summary: %s\n' % metadata.description)
+    file.write('Home-page: %s\n' % metadata.url)
 
     if version < StrictVersion('1.2'):
-        file.write('Author: %s\n' % self.get_contact())
-        file.write('Author-email: %s\n' % self.get_contact_email())
+        file.write('Author: %s\n' % metadata.author)
+        file.write('Author-email: %s\n' % metadata.author_email)
     else:
         optional_fields = (
             ('Author', 'author'),
@@ -75,51 +100,50 @@ def write_pkg_file(self, file):
         )
 
         for field, attr in optional_fields:
-            attr_val = getattr(self, attr)
+            attr_val = getattr(metadata, attr, None)
             if six.PY2:
                 attr_val = self._encode_field(attr_val)
 
             if attr_val is not None:
                 file.write('%s: %s\n' % (field, attr_val))
 
-    file.write('License: %s\n' % self.get_license())
+    file.write('License: %s\n' % metadata.license)
     if self.download_url:
-        file.write('Download-URL: %s\n' % self.download_url)
-    for project_url in self.project_urls.items():
-        file.write('Project-URL: %s, %s\n' % project_url)
+        file.write('Download-URL: %s\n' % metadata.download_url)
+    for project_named_url in metadata.project_urls:
+        file.write('Project-URL: %s\n' % project_named_url)
 
-    long_desc = rfc822_escape(self.get_long_description())
-    file.write('Description: %s\n' % long_desc)
+    file.write('Description: %s\n' % metadata.long_description)
 
-    keywords = ','.join(self.get_keywords())
+    keywords = ','.join(metadata.keywords)
     if keywords:
         file.write('Keywords: %s\n' % keywords)
 
     if version >= StrictVersion('1.2'):
-        for platform in self.get_platforms():
+        for platform in metadata.platforms:
             file.write('Platform: %s\n' % platform)
     else:
-        self._write_list(file, 'Platform', self.get_platforms())
+        self._write_list(file, 'Platform', metadata.platforms)
 
-    self._write_list(file, 'Classifier', self.get_classifiers())
+    self._write_list(file, 'Classifier', metadata.classifiers)
 
     # PEP 314
-    self._write_list(file, 'Requires', self.get_requires())
-    self._write_list(file, 'Provides', self.get_provides())
-    self._write_list(file, 'Obsoletes', self.get_obsoletes())
+    self._write_list(file, 'Requires', metadata.requires)
+    self._write_list(file, 'Provides', metadata.provides)
+    self._write_list(file, 'Obsoletes', metadata.obsoletes)
 
     # Setuptools specific for PEP 345
-    if hasattr(self, 'python_requires'):
-        file.write('Requires-Python: %s\n' % self.python_requires)
+    if metadata.python_requires:
+        file.write('Requires-Python: %s\n' % metadata.python_requires)
 
     # PEP 566
-    if self.long_description_content_type:
+    if metadata.long_description_content_type:
         file.write(
             'Description-Content-Type: %s\n' %
-            self.long_description_content_type
+            metadata.long_description_content_type
         )
     if self.provides_extras:
-        for extra in self.provides_extras:
+        for extra in metadata.provides_extras:
             file.write('Provides-Extra: %s\n' % extra)
 
 
--- a/setuptools/validation.py
+++ b/setuptools/validation.py
@@ -1,0 +1,106 @@
+"""
+Validation logic for package metadata, supporting PEP 566.
+See: https://www.python.org/dev/peps/pep-0566/
+"""
+
+
+def single_line_str(val):
+    """A string that does not contain newline characters"""
+    if val is None:
+        return
+    assert isinstance(val, str), \
+        "is of type %s but should be of type str" % type(val)
+    assert '\n' not in val, \
+        "to contain a newline character (not allowed)"
+
+
+def multi_line_str(val):
+    """A string that can contain newline characters"""
+    if val is None:
+        return
+    assert isinstance(val, str), \
+        "is of type %s but should be of type str" % type(val)
+
+
+def list_of_str(val):
+    """A list of single-line text strings"""
+    if val is None:
+        return
+    assert isinstance(val, list), \
+        "is of type %s but should be of type list(str)" % type(val)
+    for item in val:
+        single_line_str(item)
+
+
+def set_of_str(val):
+    """A set (i.e. a list of unique entries) of single-line text strings"""
+    if val is None:
+        return
+    assert isinstance(val, set), \
+        "is of type %s but should be of type set(str)" % type(val)
+    for item in val:
+        single_line_str(item)
+
+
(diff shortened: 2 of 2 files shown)


bittner:
As suggested in https://github.com/pypa/setuptools/issues/1390#issuecomment-433732683 I've added the information about which field should satisfy which validation to the new `validation` module directly.

I still have to write the test for the new module. This is meant as WIP, so I can have the code peer-reviewed more quickly. (Thank you for your understanding!)

bittner:
Looks like I have to make the tests pass for Python 2.

Also, the Exception details son't seem to be printed out in Python 2 -- while they do in Python 3, e.g.
```python
E   setuptools.validation.InvalidMetadataError: [('provides_extras', AssertionError("is of type <class 'set'> but should be of type list(str)",))]
```

versus Python 2:
```python
E   InvalidMetadataError

    setuptools/validation.py:97: InvalidMetadataError
```

setuptools/validation.py
```diff
@@ -0,0 +1,106 @@
...
+        "is of type %s but should be of type str" % type(val)
+    assert '\n' not in val, \
+        "to contain a newline character (not allowed)"
+
+
+def multi_line_str(val):
+    """A string that can contain newline characters"""
+    if val is None:
+        return
+    assert isinstance(val, str), \
```
pganssle: Do not use `assert` for control flow. It is disabled when python is run optimized.

di:
Agree with @pganssle here, there should be no actual validation logic added to `setuptools` unless it is somehow `setuptools`-specific.

bittner:
I agree with you. There are some things I still need to clarify for myself, though:

1. `setuptools` is just a "tool" to generate metadata to publish a package (among other things)
1. `distutils` is just another tool (like setuptools, see above) -- _let's ignore the unlucky monkey-patch inheritance nature of setuptools for now_ 
1. `distutils` is [mentioned as a deprecation candidate](https://github.com/pypa/setuptools/issues/1390#issuecomment-398079945) in other discussions
1. other places, such as Warehouse, may also require validating metadata (i.e. the validation code)
1. if cleanly implemented, there should probably be a separate, independent module, specifically dedicated to just the metadata used in Python or for Python modules, and their validation.

TL;DR

- Is the [`write_pkg_file` function](https://github.com/pypa/setuptools/issues/1390#issuecomment-433731850) the right place to add the _use_ of metadata validation, like implemented in this PR?
- Where should the validation business logic be added -- if not in [`distutils.dist.DistributionMetadata`](https://github.com/python/cpython/blob/master/Lib/distutils/dist.py#L1023) (as of item 3 above)?

setuptools/validation.py
```diff
@@ -0,0 +1,106 @@
...
+        for data in data_dicts:
+            for key in data:
+                setattr(self, key, data[key])
+                self.fields += [key]
+
+        for key in kwargs:
+            setattr(self, key, kwargs[key])
+            self.fields += [key]
+
+    def validate(self, throw_exception=False):
```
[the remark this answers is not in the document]
bittner: Please note that @di [suggested to add validation functionality](https://github.com/pypa/setuptools/issues/1390#issuecomment-432411556) that behaves like `validate().errors`, so that it integrates nicely into Warehouse code.

I felt that throwing an exception by default would make code more clumsy. It's just not beautiful. -- I wouldn't want to miss that possibility in general, though.

setuptools/dist.py
```diff
@@ -55,17 +56,41 @@ def get_metadata_version(dist_md):
 def write_pkg_file(self, file):
     """Write the PKG-INFO format data to a file object.
     """
+    metadata = Metadata(
```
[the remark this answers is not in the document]
bittner: I see. You're right.

On the other hand, consider the [validation functionality suggested](https://github.com/pypa/setuptools/issues/1390#issuecomment-432411556) by @di: it requires that you pass a dictionary to the validation class constructor. How would you solve this with the `DistributionMetadata` constructor?

setuptools/validation.py
bittner: What would be similarly elegant? I don't want to do an `if foo: raise Bar()` in all that many places.

setuptools/dist.py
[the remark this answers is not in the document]
pganssle: I believe @di was referring to more generic functionality for a package other than `setuptools`, so it depends on the details there. One option is to just generate a non-validated PKG-INFO and pass it to a function that validates `PKG-INFO` files.

Another is to have an intermediate format that represents `PKG-INFO` as JSON (thus retaining some type information and avoiding the problem where content and format are indistinguishable) which is convertable by `packaging` into a `PKG-INFO` file. `packaging` could then validate the JSON file before a `PKG-INFO`file is generated.

If we're already trying to validate a `DistributionMetadata` file, then I would just validate the results of the actual getter operations.

setuptools/validation.py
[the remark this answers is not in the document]
pganssle: > Please note that @di suggested to add validation functionality that behaves like validate().errors, so that it integrates nicely into Warehouse code.

He suggested no such thing. He suggested that something like the warehouse functionality go into a *different* project, [pypa/packaging](https://github.com/pypa/packaging), not this one. I think you may have misunderstood him.

> I felt that throwing an exception by default would make code more clumsy. It's just not beautiful. -- I wouldn't want to miss that possibility in general, though.

This is in general not an example of good design, because there are now two different error handling pathways. Additionally, the Pythonic way to signal an error condition is to raise an exception. It is not clumsy and in fact it leads to a lot of clumsy code in downstream consumers. The reason is that in most cases, the functions calling your function can do nothing about an error except abort early and notify the user, if you raise an exception in error conditions, anyone who isn't going to try to recover from the error can just use your function as if it always works because the exception you raise will bubble up the stack until someone either catches it or the program is aborted and the user is notified of the error. If you return an error code instead, you are relying on your consumers to know that your function can either return an error code of some sort or not and raise an error or a handler.

setuptools/validation.py
pganssle: `assert` is not intended to be an "elegant" shorthand for raising an exception, it's a way of programmatically making assertions about how your code should work. You use it for things like enforcing a function's contract or documenting assumptions made by the programmer (which are then validated at runtime).

In this specific case, you cannot use it because it *will not work*. Assertions are to be treated as *optional* statements and if you run Python in optimized mode, they will simply not run at all, for example:

```bash
$ python -Oc "assert False"
$ python -c "assert False"
AssertionError
```

In this particular case, I would either raise `ValueError` or convert validators to functions returning a boolean, depending on the details. I'm not sure it's worth working out the details in this case, though, because as mentioned elsewhere this functionality rightly belongs in another library.

pganssle:
@bittner 
Regarding points 1, 2 and 3,  `setuptools` is an extension of `distutils`. Both are tools for building python packages - i.e. converting the source repository into something distributable/installable. `distutils` is basically deprecated in the sense that it is supplanted by `setuptools` and likely `distutils` will move *into* `setuptools`. PEP 517 adds a mechanism for freeing the tight coupling between installer tools like `pip` and `distutils` and `setuptools`. This will allow for the user of other build backends and in general is a sounder design because it's based on well-defined interfaces.

Regarding point 4, there are many points you may want to validate the metdata:

1. Immediately before or after generation (`setuptools`, `flit`, other build backends)
2. Before upload (`twine`)
3. Before accepting it to a package index (`warehouse`, `bandersnatch`)
4. Before installation (`pip`)

The metadata fields and formats are a standard, and as such it is likely that anyone who uses it may want to be able to know the difference between metadata that is compliant with the standard and not compliant. We could write separate validation logic for each of these (and possibly one or more of them will re-implement validation), but it may be a needless duplication of effort, and you may end up with metadata producers generating something that metadata consumers consider invalid if one is more strict than the other.

Regarding your questions:

> Is the write_pkg_file function the right place to add the use of metadata validation, like implemented in this PR?

This depends on the nature of the common validator, but I think so, yes.

> Where should the validation business logic be added -- if not in distutils.dist.DistributionMetadata (as of item 3 above)? Should I move the validation module I introduced with this PR to the packaging package?

I don't know what you mean by "validation business logic", but I'm guessing you mean the `validate` method? If so, I would say it belongs in another repo. Possibly [pypa/packaging](https://github.com/pypa/packaging), but I would open an issue there to figure out if that's something they want to include or if we should possibly make it a completely separate package, and if so what the details would be.

If it were me designing it, I would probably design an interface that looks like this:

```python
def validate_metadata(metadata: Dict[str, str]):
    """Function to validate PKG-INFO metadata

    Raises `InvalidMetadataError` for invalid metadata.
    """
    metadata_version = metadata.get("Metadata-Version", None)
    if metadata_version not in VALID_VERSIONS:
        raise InvalidMetadataError("No valid Metadata-Version found!")

    for key, value in metadata.items():
        validate = _get_key_validator(key, metadata_version)
        validate(value)
```

`metadata` would be a dictionary mapping the keys from the appropriate version of the [Core Metadata Specification](https://packaging.python.org/specifications/core-metadata/) to a function that checks whether `value` is a valid value for that key in that version of the metadata and throws `InvalidMetadataError` otherwise.

If such an interface were added to a third-party package, `write_pkg_info` would need to be modified such that it generates a dictionary (possibly ordered) of key-value pairs, then writes the key-value pairs to file in the `PKG-INFO` format only at the end of the function. This dictionary could then be passed to the `validate_metadata` function.

Note that in this design, you can easily validate a JSON file containing the metadata by reading it into a dictionary with `json.load`.

ssbarnea:
Any chance to fix this?

bittner:
> Any chance to fix this?

I don't think it will be from this PR. I would have loved to get this fixed. It doesn't depend on me now, I'm afraid. :disappointed: 

@ssbarnea Go ahead in https://github.com/pypa/packaging/issues/147 asking for progress, please!

setuptools/validation.py
ssbarnea: @bittner Please update the code and use raise exception. Assert should be used only inside tests and I am sure that some linters are even able to identify accidental use outside tests.

pganssle: @ssbarnea I think at this point this PR needs a much bigger overhaul, and most of it may need to live in this repo. Probably best to hold off on cleaning up this specific code until the overhaul is done.

jaraco:
This contribution seems to have stalled, at an impasse due to lack of clarity on where the validation should be implemented or whose implementation should be used. I'm uninterested in merging functionality here that's going to be implemented redundantly in another package. As a result, I'm going to defer this merge indefinitely, though I'm happy to revive the PR or review a new one at such a point that a (shared) implementation is accepted.

Thanks @bittner for your contribution, which will remain here even if not merged.

bittner:
Well, at least I tried. TBH, I'm a bit disappointed about how this PR went. I would have continued elsewhere if directed properly.

I'm keen on finding the "right place" and implement an elegant, sustainable and beautiful solution (honoring PEP 20 and Clean Code). Clearly, you all know the code base better but you could have leveraged my motivation. Instead, I found myself blocked with some other's intention of "I'll do it myself" (which doesn't seem to have resulted in a working solution until today, half a year after).

This is not meant as a personal rant. I guess, most of us do that kind of programming in their spare time, sacrificing their families. Honor this! And, please, open yourselves to contributions from people that are not yet part of "your club". (At least the was my, clearly personal, impression. Sorry about that.)

ssbarnea:
With all due respect but I seen passing the dead-cat (metadata-validation) between several projects and this is really sad, for multiple reasons: discourage new contributors and fails to implement a feature that is highly desired and needed.

I can understand that the maintainers are worried about added complexity and they prefer to focus on other things but once they realise that metadata needs to be validated by another project, they should be the first to setup this placeholder project and link to it.

The other approach which is build it and we will use it if we like is not really constructive, as is used in many cases as an excuse to do nothing (not saying is the case here).

di:
@bittner Really sorry that you feel like your time/motivation was wasted here, I definitely feel like that's my fault. Unfortunately I haven't had much time (until now) to work on this myself, or help someone else work on it. If you're interested, there will surely still be work to be done here. I posted an update on https://github.com/pypa/packaging/issues/147 outlining what's been holding this up and the current status.

@ssbarnea I'm not sure what you mean. At least on this issue, I think @pganssle and I have agreed that metadata validation should exist in `packaging`, [and this has been the plan from PyPI's perspective as well for quite a while.](https://github.com/pypa/warehouse/blame/7b33d7245cff242c07965fbc4575e5caf127bfaf/warehouse/forklift/legacy.py#L354-L356). This isn't a case of "`setuptools` maintainers" saying "`packaging` maintainers" should do it instead... we're all working on the same thing here.

ssbarnea:
@di Super! I am happy to hear we reached an agreement regarding where this feature will land.

pganssle:
@ssbarnea To be clear, there was never any disagreement here. The submitter merely misunderstood what was being asked of them.

bittner:
> The submitter merely misunderstood what was being asked of them.

Yeah, sure.

Issue #1535, opened 2018-10-27
Update setup.cfg documentation to mention the minimum version required for each setting to be supported


benoit-pierre:
It's otherwise all too easy to create a `setup.cfg` that (silently) does the wrong thing.

See #1532 for an example.

isaac-philip:
Is this issue closed ?

benoit-pierre:
No.

Gouravchawla:
This issue has a merged PR referencing this. Can this be closed?

PR #1564, opened 2018-10-29
Document setup.cfg minimum version for version and project_urls

0az:
## Summary of changes

Progress on #1535 

### Pull Request Checklist
~~- [ ] Changes have tests~~

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -2471,11 +2471,11 @@ Metadata
     The aliases given below are supported for compatibility reasons,
     but their use is not advised.
 
-==============================  =================  =================  ===============
-Key                             Aliases            Type               Minumum Version
-==============================  =================  =================  ===============
+==============================  =================  =================  =============== =====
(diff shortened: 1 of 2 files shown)


docs/setuptools.txt
```diff
@@ -2475,10 +2475,10 @@ Metadata
 Key                             Aliases            Type               Minumum Version
 ==============================  =================  =================  ===============
 name                                               str
-version                                            attr:, file:, str
+version                                            attr:, file:, str  39.2.0
```
pganssle: This one is tricky, because `version` has almost certainly always been supported, but the `file` attribute has only been supported since version 39.2.0.

I think maybe we should do a footnote under "Minimum Version" with an explanation at the bottom of the table, similar to the way footnotes are handled [here](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior).

pganssle:
Thanks @0az!

PR #1576, opened 2018-11-05
Fix upload metadata

### LINKED ISSUE #1381: Include most recent upload metadata when uploading to PyPI

pganssle:
## Summary of changes

This PR adds `get_metadata_version` to `distutils.dist.DistributionMetadata` and also monkey patches `distutils.dist.DistributionMetadata.read_pkg_file` to populate the `metadata_version` attribute. We also have overridden  `distutils.command.upload.upload_file` so that the metadata version is retrieved from `get_metadata_version` rather than hard-coded as 1.0.

I felt that it would be preferable for `read_pkg_file` to actually store the specified metadata version rather than to infer it from the fields present (as is done when *writing* the metadata file), *however*, I think the bug I'm concerned about would still be fixed if we always had `get_metadata_version` infer the version from the keys present.

Closes #1381

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1576.change.rst
+++ b/changelog.d/1576.change.rst
@@ -1,0 +1,1 @@
+Started monkey-patching ``get_metadata_version`` and ``read_pkg_file`` onto ``distutils.DistributionMetadata`` to retain the correct version on the ``PKG-INFO`` file in the (deprecated) ``upload`` command.
--- a/setuptools/command/upload.py
+++ b/setuptools/command/upload.py
@@ -172,9 +172,11 @@ class upload(orig.upload):
             self.announce('Server response (%s): %s' % (status, reason),
                           log.INFO)
             if self.show_response:
-                text = self._read_pypi_response(result)
-                msg = '\n'.join(('-' * 75, text, '-' * 75))
-                self.announce(msg, log.INFO)
+                text = getattr(self, '_read_pypi_response',
+                               lambda x: None)(result)
+                if text is not None:
+                    msg = '\n'.join(('-' * 75, text, '-' * 75))
+                    self.announce(msg, log.INFO)
         else:
             msg = 'Upload failed (%s): %s' % (status, reason)
             self.announce(msg, log.ERROR)
--- a/setuptools/command/upload.py
+++ b/setuptools/command/upload.py
@@ -1,14 +1,24 @@
+import io
+import os
+import hashlib
 import getpass
+
+from base64 import standard_b64encode
+
 from distutils import log
 from distutils.command import upload as orig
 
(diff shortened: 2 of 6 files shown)


pganssle:
We may want to pull in the unit tests from `distutils` for `read_pkg_file` and `upload_file`, but in the case of `upload_file` the command is deprecated anyway, and for `read_pkg_file`, as long as all the lines are covered I'm OK with waiting until we pull `distutils` in its entirety into `setuptools`.

pganssle:
@jaraco @benoit-pierre This is mostly ready for review. I think I may need a few more tests, I didn't realize that we had the diff coverage required configured to be 81% or something and not 100%. That feels like a mistake, so I'll probably add a few more tests to get the diff coverage to 100% (or at least closer to it), but if y'all can give some feedback on what I have so far I'd appreciate it.

The biggest change (which is why I don't feel comfortable just unilaterally merging this) is that we're now monkey-patching *more* of distutils. We can avoid that in this issue, but I think probably it will be hard to avoid if we also want to solve things like #1578.

pganssle:
OK, so lesson learned about not having 100% diff coverage in a situation like this. Almost everything that I originally thought "maybe I should throw that under a `#pragma: nocover` rather than write a test for it" ended up having a bug in it that I needed to fix.

I think I have 100% coverage on the diff now, though (`codecov` is being super weird about displaying diff coverage, but when I manually run `tox -e py27,py37 -- --cov` I don't think I'm missing anything.

pganssle:
I'd like to get this one in for the next release, but I'm OK with pushing it off to the release after that.

I'd like to cut a release soon, so @jaraco @benoit-pierre if you have no objections to merging this let me know preferably in the next few days.

pganssle:
@jaraco To clarify, the reason why I wanted to improve the deprecated command is because it's currently polluting PyPI with bad metadata, and I think that when `upload` starts to raise an error, a bunch of people will just start pinning to the last version of `setuptools` that "works" rather than switching over to `twine`. I want the version they pin to to be uploading good metadata. This is, to me, the next step in making `setup.py upload` raise an exception.

pganssle:
I've also ported python/cpython#10414 in to this branch, because `platform.dist()` is deprecated and this comment is unnecessary anyway.

PR #1559, opened 2018-10-28
Disallow files for license inputs

RajdeepRao:
## Summary of changes

This pull request disallows the file inputs for license fields in the `setup.cfg`. Devs would have to input a string.

Closes 1551

### Pull Request Checklist
- [ ] Changes have tests
- [ ] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

PR #1584, opened 2018-11-10
Remove erroneous environ commands

pganssle:
In #1559, @RajdeepRao pointed out that these code blocks were in many tests here.

`environ.update()` mutates a list in place and returns `None`, so all of these blocks with `os.environ.copy().update(...)` are creating an ephemeral copy of the environment dictionary, updating it, and then it's deleted since nothing holds a reference to it.

Since these were no-ops anyway, I've removed them. I believe the `env` fixture defined in this file should be used instead.

### Pull Request Checklist

### DIFF
--- a/setuptools/tests/test_egg_info.py
+++ b/setuptools/tests/test_egg_info.py
@@ -454,14 +454,11 @@ class TestEggInfo:
     def test_provides_extra(self, tmpdir_cwd, env):
         self._setup_script_with_requires(
             'extras_require={"foobar": ["barbazquux"]},')
-        environ = os.environ.copy().update(
-            HOME=env.paths['home'],
-        )
+
         code, data = environment.run_setup_py(
             cmd=['egg_info'],
             pypath=os.pathsep.join([env.paths['lib'], str(tmpdir_cwd)]),
             data_stream=1,
-            env=environ,
         )
         egg_info_dir = os.path.join('.', 'foo.egg-info')
         with open(os.path.join(egg_info_dir, 'PKG-INFO')) as pkginfo_file:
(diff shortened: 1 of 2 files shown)


setuptools/tests/test_egg_info.py
```diff
@@ -454,9 +454,7 @@ def test_extras_require_with_invalid_marker_in_req(self, tmpdir_cwd, env):
     def test_provides_extra(self, tmpdir_cwd, env):
         self._setup_script_with_requires(
             'extras_require={"foobar": ["barbazquux"]},')
-        environ = os.environ.copy().update(
-            HOME=env.paths['home'],
-        )
+
```
benoit-pierre: Don't you need to remove the references in calls to `environment.run_setup_py(..., env=environ)` too?

pganssle: Indeed.

pganssle:
Hm, I guess there's `setuptools.tests.environment`.

It's a bit confusing how the `env` fixture is named in this case, we may want to rename that.

Also, presumably this was intended to do a specific thing. Maybe we should be setting `env['HOME']` in `setuptools.tests.environment`? I don't really know what problem this was originally intended to solve.

pganssle:
Actually, I think I can actually just easily make this work as intended, though it's disturbing that this has been broken for 4 years (and that everyone's just been cargo-culting this the whole time - almost all the instances of this are from blind copy-pasting).

Not sure how to test that the fixed version actually works, though. Probably seeing if some configuration in my home directory doesn't get picked up after it's fixed?

pganssle:
Interestingly, the tests actually *fail* if `env` is set up "correctly".

jaraco:
> I think I can fix it by looking for all the egg-info files relative to env.paths['egg-base'], but did you want them to be in the local directory deliberately, even when HOME is set to be somewhere else?

I did not have any particular expectations about the tests. I do agree that it's perhaps surprising and undesirable to be generating files in the cwd unless that directory was explicitly set to some temporary directory for the duration of the test. Since the `egg_info` command does _normally_ generate files in the current directory, I would sort-of expect for tests to use that pattern (except for tests that specifically override where egg-info is generated).

I don't understand the meaning/purpose of HOME. I would expect HOME only to affect the reading of `~/.pydistutils.cfg`... and not the location of a generated egg-info.

> I think I can fix it by looking for all the `egg-info` files relative to `env.paths['egg-base']`.

Perhaps that's the right thing to do, but it feels like a workaround for earlier bad behavior. I'm not seeing the problem clearly, though. Feel free to do whatever you think is right.

jaraco:
I don't have any better information, sorry (and sorry for the delayed response).

PR #1589, opened 2018-11-12
Fix Travis test for pyc files in egg-info

pganssle:
This fixes the issue that prevented the initial `40.6.0` release.

The problem is that [`!` is a special character in YAML syntax](https://stackoverflow.com/questions/9664113/what-does-a-single-exclamation-mark-do-in-yaml/9703472#9703472).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
     after_success: true
     before_deploy:
       - python bootstrap.py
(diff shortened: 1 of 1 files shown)


jaraco:
It'd be nice not to have to do the test twice. Would it be sufficient to do it in just one place?

jaraco:
Regardless, getting the release out is the most important thing.

benoit-pierre:
Yes, we should only keep the test in the install script.

PR #1607, opened 2018-12-13
Don't keep file modes for package data.

### LINKED ISSUE #1424: Preserving read-only flags on package data causing access denied when building
I'm working with a read-only source directory on Windows and running into access denied errors when calling `python setup.py bdist_wheel`. I noticed that it was only files included in package_data that were causing this, which helped me track it down to:

[setuptools/setuptools/command/build_py.py](
https://github.com/pypa/setuptools/blob/89155abb4222cf5a9dc81120e5c71e26b5af68f9/setuptools/command/build_py.py#L123)
`outf, copied = self.copy_file(srcfile, target)`

which is slightly different from its distutils version:

[cpython/Lib/distutils/command/build_py.py](
https://github.com/python/cpython/blob/d6345def68d3a0227253da26213dadb247f786db/Lib/distutils/command/build_py.py#L141)
`self.copy_file(os.path.join(src_dir, filename), target, preserve_mode=False)`

So it looks like it comes down to this preserve_mode flag.

I don't know if there is a good reason to preserve things like read-only flags? There's a [comment](https://github.com/python/cpython/blob/d6345def68d3a0227253da26213dadb247f786db/Lib/distutils/command/build_py.py#L74) about preserving mode further up in the distutils code, which seems to suggest not preserving is the way to go.

jorikdima:
## Summary of changes

Prevent keeping files mode for package_data build. It may break a build if user's package data has read only flag. It will be copied to build folder and set RO flag again. Then at the clean up stage a script will not be able to remove this files, asserting Access denied.

Closes #1424 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/1424.change.rst
+++ b/changelog.d/1424.change.rst
@@ -1,0 +1,1 @@
+Prevent keeping files mode for package_data build. It may break a build if user's package data has read only flag.
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -120,7 +120,7 @@ class build_py(orig.build_py, Mixin2to3):
                 target = os.path.join(build_dir, filename)
                 self.mkpath(os.path.dirname(target))
                 srcfile = os.path.join(src_dir, filename)
-                outf, copied = self.copy_file(srcfile, target)
+                outf, copied = self.copy_file(srcfile, target, preserve_mode=False)
                 srcfile = os.path.abspath(srcfile)
                 if (copied and
                         srcfile in self.distribution.convert_2to3_doctests):


jaraco:
The problem described herein seems legitimate and worth correcting.

I tried to replicate it myself and couldn't. I ran `chmod u-w setup.py` and confirmed that the file was read-only. I then ran `python setup.py sdist` and it ran to completion, even though it copied the setup.py file into the build directory and successfully removed it.

Can you describe the conditions in which the issue occurred, such that someone else could possibly replicate the issue?

Also, I don't see tests for this issue, so I cleared the checkbox. It would be good practice to have a test for this behavior if it's something we wish to guarantee.

pganssle:
I am mildly worried about this sort of thing, there's a lot of tricky stuff that can happen with file modes. Some people may be setting the mode of their files very deliberately (e.g. people deploying something cryptographically sensitive?). It may be that I am being overly cautious about this, but I am not sure how to determine if changing this would break anyone (most people won't notice until something they are doing breaks).

If this is mostly a problem of build environment (where the build is done in a read-only file system), maybe a more incremental change would be to have this be a switch enabled by an environment variable?

jorikdima:
Folks, any reaction. What else should be done to accept this?

jaraco:
@jorikdima Thanks so much for putting together a PR, and sorry for the delay.

Paul said:

> I am not sure how to determine if changing this would break anyone

I share this worry, but one way I've found to work through issues like this is just to push through... but to _require_ tests. Let's say we accept this change and it fixes the behavior in this usage, but breaks another, then someone might submit a PR to revert this change (although likely without referencing it), and we're in an infinite loop of somebody's use-case breaking.

If instead we get a good test for this expected behavior, then another use-case wouldn't be able to revert this behavior without breaking/changing that test... and it will be their responsibility to fix the issue in a way that addresses both use-cases (or we might decide to revert the change, but keep the test as xfail, if the impact is too great).

> What else should be done to accept this?

So that's how I'd like to proceed here. It needs a test... and not just one that tests that the parameter is passed to that call, but one that triggers an error when the files are read-only.

Would you be willing to develop such a test?

pganssle:
@jaraco I agree that we need good tests for this, but I think changing the default for this might cause more subtle problems.

I still haven't seen a reason why having this be an option that you can enable with an environment variable (and/or possibly an entry in the `pydistutils.cfg`) wouldn't work. Generally speaking, clobbering a package's file modes is *not* the right thing to do.

I also think maybe it can be more tightly scoped. As it stands now, this will also remove the `+x` bit from installed package data, which is not relevant to the problem at hand and may cause problems.

jorikdima:
> Would you be willing to develop such a test?

I'd love to, but I have never done it i the past. And briefly looking into it when I was preparing the PR made me think that I have not enough experience with testing system. Don't want to make anyone think that I want to move a responsibility to anyone else shoulders :) , but if someone has time I would appreciate.
> Generally speaking, clobbering a package's file modes is not the right thing to do.

Well... but we do that not for package's files without hesitation. Now there is a discrepancy in working with scripts and package's files which I want to get rid of. And that, BTW, is not package's files as such, it's theirs copy, which have been created by us.
And even better, not there is an exception during building a wheel if you have RO flag in your package's files (always working on Perforce with not checked out files e.g.). Isn't **this** a bad behavior?
Anyway, I really do appreciate your attention to this issue and completely rely on your opinion to accept it or not, because my experience is really too low in the setuptools project.

jaraco:
I started looking into writing a test for the reported issue. It's a little difficult because the reported issue indicates `bdist_wheel` as the command, which isn't a command in setuptools but in the wheel package. However, the "fix" is in build_py, so presumably it should be possible to replicate the undesirable behavior with just that command. I'll continue to work on it an a future time.

jaraco:
See e7f29530 for the merge commit that includes a test.

mattclay:
This change breaks functionality in `ansible-test` (part of [ansible](https://pypi.org/project/ansible/)) which internally uses various executable scripts which are not entry points.

jorikdima:
> This change breaks functionality in `ansible-test` (part of [ansible](https://pypi.org/project/ansible/)) which internally uses various executable scripts which are not entry points.

Does it make sense then to preserve flags but clear read-only flag? Basically in the original report the read-only flag was the one who was a culprit.

mattclay:
@jorikdima That should work for the Ansible project, and I expect for other projects with similar needs.

jorikdima:
Proposed solution:
```
outf, copied = self.copy_file(srcfile, target)
os.chmod(target, stat.S_IWRITE)
```
We can keep the original code preserving all flags, but then remove RO flag. Removing RO should be supported on all platforms by os.chmod
Should I make a separate PR+test for this fix?

jaraco:
@jorikdima If you would please, create a single PR, which should have two commits: the first should be a test capturing that executable bit is retained for package data. This test should fail on the current codebase. Then another commit should include your patch, which should cause the test to pass. The existing test for read-only should continue to pass on Windows also.

PR #1610, opened 2018-12-27
Make sure to sudo for tests

RDIL:
## Summary of changes

Travis recently did some stuff and basically it’s better to have this than not. 

Closes  no issue sorry <3

### Pull Request Checklist
- [X] Changes have tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 dist: trusty
 language: python
+sudo: required
 
 jobs:
   fast_finish: true


pganssle:
@RDIL I'm not sure that this solves any problem, but it's already the default on Travis's CI.

jaraco:
My preference would be to run without sudo. That's what I've done in most other projects, and it seems to be the best option unless absolutely required.

pganssle:
@jaraco If so, I think we need to actively turn sudo off. It may still be required on xenial in order to install the latest Python 3.

PR #1611, opened 2018-12-27
h t t p s

RDIL:
## Summary of changes

Shouldn’t we secure links in the SSL code?? Haha

### DIFF
--- a/setuptools/ssl_support.py
+++ b/setuptools/ssl_support.py
@@ -59,7 +59,7 @@ if not match_hostname:
     def _dnsname_match(dn, hostname, max_wildcards=1):
         """Matching according to RFC 6125, section 6.4.3
 
-        http://tools.ietf.org/html/rfc6125#section-6.4.3
+        https://tools.ietf.org/html/rfc6125#section-6.4.3
         """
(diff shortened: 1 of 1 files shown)


RDIL:
What was the force push for??

pganssle:
@RDIL I rebased your commit against the current master and reworded your commit message. I prefer commit messages to be a clear explanation of what the change is intended to do. The first line should complete the sentence, "If applied, this commit would...", so that someone reading the history can easily see what changes each commit intends to do without looking at the patch.

In that spirit, "h t t p s" was a bit vague - it could have been some change to how we make HTTP requests or HTTPS requests, or it could be a change in the documentation, etc. I just clarified that.

That said, "write a good commit message" is not a requirement of a contribution here and I don't want to nitpick first-time contributors, so I usually just clean up commit messages when I can. I realize that getting feedback about how to make good PRs is a valuable part of open source, but it's a balancing act between providing detailed feedback and overwhelming contributors with petty changes. Hope that explains it to your satisfaction!

RDIL:
Woah lots of words uh ok

Issue #1612, opened 2018-12-28
Test failure in develop on all platforms


pganssle:
The CI for [this trivial fix](https://github.com/pypa/setuptools/pull/1611) is failing on all interpreters for Appveyor and Travis. Fails locally as well with this error:

```
_____ [doctest] setuptools.command.develop.VersionlessRequirement __
200 
201     Adapt a pkg_resources.Distribution to simply return the project
202     name as the 'requirement' so that scripts will work across
203     multiple versions.
204 
205     >>> dist = Distribution(project_name='foo', version='1.0')
UNEXPECTED EXCEPTION: NameError("name 'Distribution' is not defined")
Traceback (most recent call last):

  File "~/.pyenv/versions/3.7.1/lib/python3.7/doctest.py", line 1329, in __run
    compileflags, 1), test.globs)

  File "<doctest setuptools.command.develop.VersionlessRequirement[0]>", line 1,
     in <module>

NameError: name 'Distribution' is not defined
```

Seems like it's a doctest failing.

pganssle:
Looks like this commit is the problem: 0902f02d9d68f18e906e727cbafa4a05fe5c9c91, because `Distribution` is no longer in the namespace of the doctest.

jaraco:
Gah. Sorry about that. Thanks for chasing this down. I probably missed that usage because I was relying on the linter for finding the names.

pganssle:
@jaraco No harm no foul, though I think it would be best to go through the motions of a PR next time, even if you just do a self-merge. It's helpful for catching stuff like this.

pganssle:
Closed by #1613

jaraco:
> I think it would be best to go through the motions of a PR next time.

The problem I have with this is it does increase the burden of making a contribution often to the point of making it infeasible in the time allotted. I usually have an implicit, manual check, where I'll notice if tests start failing after pushing a falsely-presumed error-free commit.

I agree with the concept though.

If I could create a command that could automate this operation, I'd be all in favor of it.

I've started looking into using [hub](https://hub.github.com). It looks like it can create a pull request but not set it for merge automatically. It's also not yet clear to me if it can correctly automatically detect the correct fork/branch for the source and target. But if it can, that almost streamlines the process.

jaraco:
[...]

PR #1613, opened 2018-12-28
Import distribution in doctest

pganssle:
Fixes GH issue #1612, bug introduced in commit 0902f02d9d68f18

No news is required as this fixes a change introduced since the last version.


pganssle:
@jaraco No problem.

I've also added an assertion that a warning is raised in the SVN download case. Probably in the future we should be considering all uncaught warnings failures in the test suite, to give us early warning about issues we need to handle.

PR #1614, opened 2018-12-29
Specify gpg --output option explicitly

marmarek:
Do not rely on obscure default behavior of gpg, which may not work with alternative implementations (like qubes-gpg-client-wrapper).

## Summary of changes

Make it work with gpg that does not have default output file. See https://github.com/QubesOS/qubes-issues/issues/4612 for example.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/command/upload.py
+++ b/setuptools/command/upload.py
@@ -57,7 +57,7 @@ class upload(orig.upload):
 
         # Sign if requested
         if self.sign:
-            gpg_args = ["gpg", "--detach-sign", "-a", filename]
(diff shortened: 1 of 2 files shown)


ThomasWaldmann:
I just tried to fix this on the qubes side, before seeing this PR.
There was no easy solution to do that, so fixing this on the setuptools side would be way easier.

If setuptools relies on getting the signature in `<filename>.asc`, it should be no problem explicitly saying so via `--output ...`.

jaraco:
Thanks again for the contribution, but we really do recommend instead to use twine.

PR #1624, opened 2019-01-02
Ensure .zip sdists are supported in build_meta.

### LINKED ISSUE #1623: AssertionError on len(sdists) == 1 when sdist format is zip
In some Windows projects, users expect sdists to be released in zip form. For that reason, I've added the following to a project's setup.cfg:

```
[sdist]
formats=zip
```

With this setting in place, performing a `pep517.build .` on such a project results in the following traceback:

```
  File "/Users/jaraco/code/main/jaraco.video/.tox/release/lib/python3.7/site-packages/pep517/_in_process.py", line 207, in <module>
    main()
  File "/Users/jaraco/code/main/jaraco.video/.tox/release/lib/python3.7/site-packages/pep517/_in_process.py", line 197, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/Users/jaraco/code/main/jaraco.video/.tox/release/lib/python3.7/site-packages/pep517/_in_process.py", line 170, in build_sdist
    return backend.build_sdist(sdist_directory, config_settings)
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pep517-build-env-rgz_zb4_/lib/python3.7/site-packages/setuptools/build_meta.py", line 181, in build_sdist
    assert len(sdists) == 1
AssertionError
```

Even though the output from building the sdist was seen above. I suspect the builder expects the sdist to have a particular extension.

jaraco:
It's a shame the tests didn't already capture this behavior. Does that mean there isn't a test for this function at all (because it would have failed on Windows CI)?

## Summary of changes

Closes #1623

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -175,8 +175,11 @@ def build_sdist(sdist_directory, config_settings=None):
         ["--dist-dir", sdist_directory]
     _run_setup()
 
-    sdists = [f for f in os.listdir(sdist_directory)
-              if f.endswith('.tar.gz')]
+    sdists = [
+        f for f in os.listdir(sdist_directory)
+        if f.endswith('.tar.gz')
+        or f.endswith('.zip')
+    ]
 
     assert len(sdists) == 1
     return sdists[0]


benoit-pierre:
According to PEP 517, [`build_sdist` should create a `.tar.gz`](https://www.python.org/dev/peps/pep-0517/#build-sdist), so the invocation to `setup.py` should probably explicitly ask for it with `--formats=gztar`.

takluyver:
Some extra background: PyPI used to allow you to upload both `.tar.gz` and `.zip` files for the same release of the same package, until a few years ago when it was restricted to only allow one sdist per release. While `.zip` sdists are still allowed, `.tar.gz` was recommended as the default option. PEP 517 turned that recommendation into a restriction.

So I agree with @benoit-pierre that it should explicitly request a tarball.

Seeing this code, though, I suspect there's another bug. AFAIK, the PEP doesn't specify that the sdist directory for the resulting file must be empty before the hook is called. It's not hard to imagine a scenario where you're building multiple sdists into the same target folder. But this hook relies on finding exactly one `.tar.gz` file there after it is called. Maybe it should be building to a temporary directory and then moving the file to the sdist directory?

jaraco:
I agree with both comments, withdrawing PR.

PR #1634, opened 2019-01-11
Include pyproject.toml in sdist

### LINKED ISSUE #1632: pyproject.toml should be included in sdist by default
An issue came up at work recently where installing [python-daemon](https://pagure.io/python-daemon) with `pip` was hitting the `easy_install` path. I assumed it was because `python-daemon` didn't have a `pyproject.toml`, but it turns out they do - it's just not included in the `Manifest.in`, so it didn't get included in the `sdist`. To verify this I created a super basic example project and it does seem that the default is to exclude `pyproject.toml`.

Given that `pyproject.toml` is one of the most important files to have when executing a build, it should be included in the sdist unless explicitly excluded.

daa:
## Summary of changes
Include ``pyproject.toml`` in source distribution if it is present.
Closes #1632 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/command/py36compat.py
+++ b/setuptools/command/py36compat.py
@@ -76,7 +76,7 @@ class sdist_add_defaults:
                     self.warn("standard file '%s' not found" % fn)
 
     def _add_defaults_optional(self):
-        optional = ['test/test*.py', 'setup.cfg']
+        optional = ['test/test*.py', 'setup.cfg', 'pyproject.toml']
         for pattern in optional:
             files = filter(os.path.isfile, glob(pattern))
             self.filelist.extend(files)
--- a/setuptools/tests/test_sdist.py
+++ b/setuptools/tests/test_sdist.py
@@ -449,6 +449,20 @@ class TestSdistTest:
             except UnicodeDecodeError:
                 filename not in cmd.filelist.files
 
+    def test_pyproject_toml_in_sdist(self):
+        """
+        Check if pyproject.toml is included in source distribution if present
+        """
+        open(os.path.join(self.temp_dir, 'pyproject.toml'), 'w').close()
+        dist = Distribution(SETUP_ATTRS)
+        dist.script_name = 'setup.py'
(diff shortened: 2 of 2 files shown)


setuptools/command/sdist.py
```diff
@@ -121,6 +121,14 @@ def __read_template_hack(self):
     if has_leaky_handle:
         read_template = __read_template_hack
 
+    def _add_defaults_optional(self):
+        if six.PY2:
+            sdist_add_defaults._add_defaults_optional(self)
```
pganssle: You can use `super(sdist_add_defaults, self)._add_defaults_optional()`, it works in both Python 2 and 3. It would also be nice to add a comment like "# Python 2 support" or something to that effect so we know to change it when we drop Python 2.

daa: Actually I can't because in Python-2 ``sdist_add_defaults`` and ``distutils``' ``sdist`` are old-style classes and using ``super()`` results in ``TypeError: super() argument 1 must be type, not classobj``

jaraco:
Gah. This PR seemed like a pretty innocuous request, and since the tests _had_ passed, I merged it. But now I see this change causes tests to [fail in master](https://travis-ci.org/pypa/setuptools/builds/485051828) when pip 19 is present (now that setuptools' own pyproject.toml is now present).

jaraco:
It is not.

jaraco:
This merge is the last thing on master. I'm going to force-push master back to its prior state.

jaraco:
```
setuptools master $ git push -f origin 97e8a:master                                                                                                                                      
Total 0 (delta 0), reused 0 (delta 0)
To https://github.com/pypa/setuptools
 + 7a4a99a3...97e8ad4f 97e8a -> master (forced update)
```

jaraco:
If you'd pulled the latest master, please on master: `git fetch origin master; git reset --hard origin/master`.

Issue #1644, opened 2019-01-25
Tests failing to run with BackendUnavailable since pip 19


jaraco:
It seems that since the release of pip 19, Setuptools' own tests fail to run thus:

```
setuptools fix_889_and_non-ascii_in_setup.cfg_take_2 $ tox -e py27                                                                                                                       
py27 create: /Users/jaraco/code/main/setuptools/.tox/py27
py27 installdeps: -rtests/requirements.txt
py27 develop-inst: /Users/jaraco/code/main/setuptools
ERROR: invocation failed (exit code 2), logfile: /Users/jaraco/code/main/setuptools/.tox/py27/log/py27-2.log
ERROR: actionid: py27
msg: developpkg
cmdargs: '/Users/jaraco/code/main/setuptools/.tox/py27/bin/pip install --exists-action w -e /Users/jaraco/code/main/setuptools'

DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Obtaining file:///Users/jaraco/code/main/setuptools
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
Exception:
Traceback (most recent call last):
  File "/Users/jaraco/code/main/setuptools/.tox/py27/lib/python2.7/site-packages/pip/_internal/cli/base_command.py", line 176, in main
    status = self.run(options, args)
  File "/Users/jaraco/code/main/setuptools/.tox/py27/lib/python2.7/site-packages/pip/_internal/commands/install.py", line 315, in run
    resolver.resolve(requirement_set)
  File "/Users/jaraco/code/main/setuptools/.tox/py27/lib/python2.7/site-packages/pip/_internal/resolve.py", line 131, in resolve
    self._resolve_one(requirement_set, req)
  File "/Users/jaraco/code/main/setuptools/.tox/py27/lib/python2.7/site-packages/pip/_internal/resolve.py", line 294, in _resolve_one
    abstract_dist = self._get_abstract_dist_for(req_to_install)
  File "/Users/jaraco/code/main/setuptools/.tox/py27/lib/python2.7/site-packages/pip/_internal/resolve.py", line 226, in _get_abstract_dist_for
    req, self.require_hashes, self.use_user_site, self.finder,
[...]

PR #1645, opened 2019-01-26
Bugfix/downgrade pip

jaraco:
As discussed in #1644, bootstrapping issues have emerged with the pep517 build support in pip 19. This change wraps the 'install' command from tox to ensure that a requisite version of pip is present for builds, unblocking testing/development until a more permanent solution can be devised.

### DIFF
--- a/pip-wrapper.py
+++ b/pip-wrapper.py
@@ -12,7 +12,13 @@ import subprocess
 
 
 def ensure_pip_version(pip_bin, ver):
-    cmd = [pip_bin, 'install', 'pip ' + ver]
+    """
+    Ensure the pip version meets the specified version.
+
+    Use python -m pip to upgrade/downgrade, because for this operation,
+    on Windows, pip.exe must not be locked.
+    """
+    cmd = [sys.executable, '-m', 'pip', 'install', 'pip ' + ver]
     subprocess.check_call(cmd)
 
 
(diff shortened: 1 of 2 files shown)


jaraco:
As you can see, this patch fails on Windows, due probably to the `bin`/`Scripts` disparity or maybe due to the `.exe` extension.

jaraco:
Oh, no. The cause is that pip can't uninstall itself on Windows due to locking of the executable file.

benoit-pierre:
I was going to suggest maybe adding `setuptools` to the build requirements instead, so setuptools can be installed from source with `pip>=19.0`, but I see that we forget to include `pyproject.toml` in our source dist...

jaraco:
> I was going to suggest maybe adding setuptools to the build requirements instead

Actually, that's a very attractive prospect. It's always been a bit messy trying to ascertain 'which' setuptools is present during a build/install of itself. It'd sure be nice to rely on isolation and explicit declaration so that a particular build of setuptools could rely on perhaps a different version (or even build system).

benoit-pierre:
But the command that install setuptools is separate from the dependencies, so this should just work:
```diff
 tests/requirements.txt | 1 +
 1 file changed, 1 insertion(+)

diff --git i/tests/requirements.txt w/tests/requirements.txt
index d301168a..f6719728 100644
--- i/tests/requirements.txt
+++ w/tests/requirements.txt
@@ -10,3 +10,4 @@ coverage>=4.5.1
 pytest-cov>=2.5.1
 paver; python_version>="3.6"
 futures; python_version=="2.7"
+pip==18.1
```

jaraco:
Superseded by #1647

PR #1646, opened 2019-01-26
Pin pip to <19 until a more permanent solution can be devised.

jaraco:
As @benoit-pierre suggested in #1645, try pinning pip in test requirements... since the build issue is only happening with the build of setuptools itself.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,5 @@
+# workaround for #1644
+pip<19
 mock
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Of course, this we changed the install command from `python -m pip`, this will fail on Windows... *sigh*

benoit-pierre:
See #1647.

jaraco:
Superseded by #1647.

PR #1647, opened 2019-01-26
tests: temporary pin pip to fix the Python 2.7 job

benoit-pierre:
Temporary fix for #1644.


jaraco:
LGTM

jaraco:
We should add a comment to the pip pin so it references the relevant issue.

benoit-pierre:
> We should add a comment to the pip pin so it references the relevant issue.

Done.

jaraco:
Perfect.

Issue #1642, opened 2019-01-23
Add PWD to sys.path as part of the PEP 517 build


fchorney:
in reference to this: https://github.com/pypa/pip/issues/6163

pganssle:
I am amenable to this change, but I think if we start messing around with `sys.path` in the build hooks, I don't see a good way to switch that behavior *back* - we won't necessarily be able to detect that someone is counting on this, and so we cannot give any sort of deprecation warning about it.

Given that you need to opt in to using PEP 517, we have an opportunity now to decide what we want the semantics to be, rather than trying to maintain the existing semantics. What are the downsides of requiring that setup.py be more explicit about where it's importing from? I suppose adding to the difficulty of migration is one.

h-vetinari:
> Given that you need to opt in to using PEP 517

That's not the case, hence the swift response to the pip-release, see https://github.com/pypa/pip/labels/C%3A%20PEP%20517%20impact

pganssle:
Maybe I'm wrong, but this should only be affecting projects that have added `build-backend` into their `pyproject.toml`, which greatly limits the scope of this problem (though the fact that there's a bunch of stuff in the wild getting burned is concerning). If `pip` started using `setuptools.build_meta` as the backend by default then I agree it's not opt-in.

fchorney:
I’ll note that this affects any project with a ‘pyproject.toml’, regardless if there is a ‘build-backend’. 

For example, I have a project that uses a ‘pyproject.toml’ as a config for python ‘black’, and this affects that project.

altendky:
https://github.com/altendky/tempexample/tree/82bd123917bedce24b7e19acbfc4a289f681ca28

There's a pretty minimal example showing the success without `pyproject.toml` and failure with it.

<details> <summary>Click to expand log</summary>

```
python3 -m venv venv
+ python3 -m venv venv
venv/bin/python -m pip install --upgrade pip==19.0.1 setuptools==40.6.3
+ venv/bin/python -m pip install --upgrade pip==19.0.1 setuptools==40.6.3
Collecting pip==19.0.1
  Using cached https://files.pythonhosted.org/packages/46/dc/7fd5df840efb3e56c8b4f768793a237ec4ee59891959d6a215d63f727023/pip-19.0.1-py2.py3-none-any.whl
Collecting setuptools==40.6.3
[...]

PR #1650, opened 2019-01-27
Include pyproject.toml in sdist

### LINKED ISSUE #1632: pyproject.toml should be included in sdist by default

jaraco:
This is a re-submit of #1634, which was backed out due to an unexpected regression.

Closes #1632

### Pull Request Checklist
- [X] Changes have tests


jaraco:
This should be deferred until there's a solution for #1642 and #1644.

pganssle:
We should probably defer this until at least `pip` 19.1, because this will start opting people in to PEP 517 in their sdists who were never opted in before. Right now that's a backwards incompatible change, but when we get things more ironed out it hopefully won't be.

gaborbernat:
Is this still deferred?

pganssle:
@gaborbernat Yeah. As it stands now `pip` doesn't support the bootstrapping parts of PEP 517, so `setuptools` cannot be built with PEP 517.

I also suspect that making this change will break many things because, right now, there are a decent number of projects out there that will break if they use PEP 517 in some situations, and they may be relying on the fact that they are not including the `pyproject.toml` in their `sdist`.

We should probably defer this until PEP 517 is more stable, possibly until PEP 517 is the default for all builds.

jaraco:
In that case, let's close it and we can revive it later if needed.

gaborbernat:
I'm not sure I understand what this means? More stable, default? It's already out for at least a year and the default build mechanism in pip for 6 months. In my view no better time than today.

jaraco:
> As it stands now pip doesn't support the bootstrapping parts of PEP 517, so setuptools cannot be built with PEP 517.

That issue was addressed in #1927.

> > > I also suspect that making this change will break many things because, right now, there are a decent number of projects out there that will break if they use PEP 517 in some situations, and they may be relying on the fact that they are not including the pyproject.toml in their sdist.

> > I'm not sure I understand what this means? More stable, default? It's already out for at least a year and the default build mechanism in pip for 6 months. In my view no better time than today.

> Agreed, if " a decent number of projects [...] may be relying on the fact that they are not including the pyproject.toml in their sdist", then they are relying on faulty behavior. That shouldn't deter setuptools from doing it right.

I agree also. It should be possible to make progress without waiting (essentially forever) for usage of the old behavior to drop to zero.

I notice that [setuptools excludes pyproject.toml](https://github.com/pypa/setuptools/blob/e6bdf25f6ab5bf4d32b0f9affa0ab98ea35f3a29/MANIFEST.in#L15). Is that a viable workaround for projects that wish to retain this undesirable behavior?

jaraco:
> Is that a viable workaround for projects that wish to retain this undesirable behavior?

Tentatively confirmed - when [building setuptools sdist against this PR](https://gist.github.com/f3ffeebaa8cbc0c8a25520b87d23a463), it does not include pyproject.toml, suggesting that the `exclude` directive in the manifest is a sufficient workaround to suppress the default inclusion.

I'm inclined to proceed with this PR.

pganssle:
@jaraco I think we might need more details here. This is yet another case where having a docker container or some other mechanism for making a change and then installing the top 100 packages on PyPI would be very helpful. If we do this and it breaks source installs of one of a popular but infrequently-released package on PyPI, we're going to get a lot of people pinning `setuptools` indefinitely.

One possible "soft launch" version of this is to start by only including `pyproject.toml` if the `[build-system]` key is detected. We're likely going to want to vendor the `toml` library anyway to solve #1688, but if we really don't want to pull it in, we might get away with an even simpler heuristic like `re.sub('\s+', '', line) == '[build-system]'` (this would miss keys defined not in their own section, like `build-system.requires="blah"`, but I think that is a rare situation).

Things that break when installed via PEP 517 are somewhat rare, but most of the complaints come from people who are not using `pyproject.toml` to specify build information but rather are using some tool like `black` or `towncrier`. In the long run I think we need to include it by default, but possibly only including it when it contains the build information would break fewer things?

If we could test both changes against a random smattering of PyPI projects I'd probably be able to form a stronger opinion here.

uranusjr:
Does the filelist logic allow for something like “include pyproject.toml by default, unless the user list it as `exclude` MANIFEST.in”? If so it’d be possible to tell packages to explicitly exclude it if they really want to.

pganssle:
@uranusjr IIRC that is how `exclude` works, and if it doesn't, we can certainly make it work that way for anything new added to the default include list.

pganssle:
So I was thinking about this and I realized that I had the people this could affect wrong. I was thinking that this change might break projects already uploaded to PyPI when built from source distributions, but this will only change *new source distributions built with `setuptools`*.

Since there's no chance of it hitting anyone installing from *existing* PyPI distributions, the most likely "this causes breakage" scenario is that someone cuts a release that includes `pyproject.toml` which then breaks a bunch of people installing from `sdist`. This means that the bugs will mostly be in projects with recent releases, which means that those projects are more likely to be responsive with fixes (and importantly we won't start breaking stuff that worked before - people can just pin to earlier versions of the library).

I think we can probably skip the "only include it if it has a `build-system` logic, even though it may cause some short term pain.

The main thing we really have to worry about is people who blacklist `pyproject.toml` today and then later actually add a build backend and forget to take it off the `MANIFEST.in` excludes list, but I think that's a much smaller cohort than the people who won't know to put it on there in the first place, so I'm +1 for adding it by default.

setuptools/tests/test_sdist.py
```diff
@@ -449,6 +449,20 @@ def test_sdist_with_latin1_encoded_filename(self):
             except UnicodeDecodeError:
                 filename not in cmd.filelist.files
 
+    def test_pyproject_toml_in_sdist(self):
+        """
+        Check if pyproject.toml is included in source distribution if present
+        """
+        open(os.path.join(self.temp_dir, 'pyproject.toml'), 'w').close()
```
pganssle: We should maybe have a rough backport of `Pathlib.touch` like we have in [`pkg_resources`](https://github.com/pypa/setuptools/blob/e6bdf25f6ab5bf4d32b0f9affa0ab98ea35f3a29/pkg_resources/tests/test_pkg_resources.py#L249) for this purpose.

Not a big deal, but I don't love this style to create an empty file.

setuptools/tests/test_sdist.py
```diff
@@ -449,6 +449,20 @@ def test_sdist_with_latin1_encoded_filename(self):
             except UnicodeDecodeError:
                 filename not in cmd.filelist.files
 
+    def test_pyproject_toml_in_sdist(self):
```
pganssle: Should we also add a test that it can be successfully *excluded*?

jaraco: Yes, though that may prove more difficult to test.

setuptools/tests/test_sdist.py
jaraco: Agreed, however doing so would imply that should be done on line 95 and 108 and others, so let's make that change in a separate effort.

setuptools/tests/test_sdist.py
jaraco: Turns out it wasn't hard at all. Fixed in f171cde1.

PR #1656, opened 2019-01-28
Revert "improve encoding handling for `setup.cfg`"

### LINKED ISSUE #417: Adopt distutils
Originally reported by: **jaraco (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

Setuptools builds on the distutils framework of the Python Stdlib. Doing so has benefits of depending on a standard, extensible framework, but also has downsides, such as limiting the advancement of distutils (across many Python versions) but simultaneously adding monkey patches to patch bugs in distutils.

I'd like to consider if Setuptools should adopt distutils (or a version of it). Doing so would decouple Setuptools from distutils, freeing Setuptools from the constraints of distutils and freeing distutils from the expectations of Setuptools.

Doing so would allow distutils to adapt at the same pace as Setuptools, addressing issues and vetting patches that could be ported back to distutils in CPython.

I've [once proposed this idea](http://kafka.dcpython.org/day/pypa/2015-04-15#16.59.20.jaraco) and it was promptly dismissed by @RonnyPfannschmidt. Why is vendoring distutils in Setuptools not an option?

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/417

### LINKED ISSUE #1653: TypeError: 'output_dir' must be a string or None since setuptools 40.7.0
We have a [tricky regression](https://bugs.gentoo.org/676608) in Gentoo with setuptools release 40.7.0. This affects many (all?) packages that involve `build_ext`. These errors aren't present in 40.6.3. I think I managed to have reproducible steps:

```
git clone https://github.com/python-pillow/Pillow.git
cd Pillow
virtualenv -p python2 env
. env/bin/activate
pip install -U setuptools
echo -e "[build]\nbuild-base = /tmp/foo" >> setup.cfg
mkdir /tmp/foo
python setup.py build
[...]
  File "setup.py", line 793, in <module>
    zip_safe=not (debug_build() or PLATFORM_MINGW), )
  File "/home/vdupras/src/Pillow/env/lib/python2.7/site-packages/setuptools/__init__.py", line 145, in setup
  File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib64/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib64/python2.7/distutils/command/build.py", line 127, in run
    self.run_command(cmd_name)
  File "/usr/lib64/python2.7/distutils/cmd.py", line 326, in run_command
  File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib64/python2.7/distutils/command/build_ext.py", line 340, in run
    self.build_extensions()
  File "setup.py", line 695, in build_extensions
    build_ext.build_extensions(self)
  File "/usr/lib64/python2.7/distutils/command/build_ext.py", line 449, in build_extensions
    self.build_extension(ext)
  File "/usr/lib64/python2.7/distutils/command/build_ext.py", line 499, in build_extension
    depends=ext.depends)
  File "/home/vdupras/src/Pillow/mp_compile.py", line 40, in _mp_compile
    output_dir, macros, include_dirs, sources, depends, extra_postargs)
  File "/usr/lib64/python2.7/distutils/ccompiler.py", line 329, in _setup_compile
    raise TypeError, "'output_dir' must be a string or None"
TypeError: 'output_dir' must be a string or None
```

Can you reproduce on your side?

It the fix isn't obvious, I can try myself on a PR soon.

(by the way, this is due to distutils's `isinstance()` call being made on `str` rather than `basestring` and only happens under py2. `output_dir` is unicode when it should be `str`)

benoit-pierre:
Reverts pypa/setuptools#1180, fix #1653. #1180 was initially deferred until #417 is handled, with good reason: I see no clean way to work around the lack of Unicode awareness in `distutils.ccompiler`. We either need to resolve #417, or wait for Python 2 support to be dropped.


jaraco:
This might be our only option here... although if there's a shorter path to stability, I'd like to selectively adopt the portions of distutils that are broken by #1180. In other words, could setuptools adopt distutils.ccompiler and enable that for use, addressing the `str` issue the same way I did in b1615d12435289c21853be2f4f40e523317998da and 249f24a1f04ce390a9e48a4d8a5bff7982714c86?

jaraco:
Another approach might be to, when parsing options, instead of allowing unicode values to pass, on Python 2 convert those to `str` where possible.

benoit-pierre:
It would be simpler to either, disable encoding support, and keep using binary for Python 2, or defer support until Python 2 support is dropped.

benoit-pierre:
Do we plan to drop Python 2 support ahead of the 2020 EOL deadline?

ofek:
This is breaking our master builds too on py2 https://travis-ci.com/DataDog/integrations-core/jobs/173502395#L814

benoit-pierre:
Added a simple regression test.

benoit-pierre:
Which of course is failing on Windows... *sigh*

jaraco:
> Added a simple regression test.

@benoit-pierre If you have a regression test, would you consider pushing it as a branch against 40.7.0 so it might be merged with whatever fix we settle on?

benoit-pierre:
@jaraco: the regression test should be fixed now, I suggest just cherry-picking d70335066c976dc03de749891560666b75822b75 into #1660.

jaraco:
>  I suggest just cherry-picking

Will do. Thanks!

jaraco:
Thanks for working on this, but I think #1660 gets us closer to goal, so I'm going to go with that approach.

PR #1657, opened 2019-01-28
tweak #1644 workaround to fix `test_pip_upgrade_from_source`

benoit-pierre:
Explicitly exclude `pyproject.toml` so it's never added (e.g. because `setuptools_scm` is used).


gaborbernat:
eh maybe also remove the build-backend key because now we hit https://github.com/pypa/pip/issues/6230

PR #1659, opened 2019-01-28
tests: improve `test_pip_upgrade_from_source`

### LINKED ISSUE #1649: Start testing installation against pip on master
I think the problem in #1644 and possibly the problem in #1642 would have been detected *way* earlier and possibly fixed before a release if we had been testing against master on `pip`. I know `pandas` tests against master on all their dependencies, and that's a reliable source of pre-release bugs for `dateutil`. In `dateutil` I also [test against master for the tzdata repo](https://github.com/dateutil/dateutil/blob/master/tox.ini#L52).

Considering that we have fairly unusual build requirements, we're very likely to be missed in `pip`'s normal testing, so I think it's reasonable for us to run some intermittent tests where we're using the master branch of `pip`, just as an early warning.

I suspect this will be a bit more complicated than other "build against master" type builds, particularly because we need to make sure it's built against master in all the virtualenvs, but if we can do this I think we should.

benoit-pierre:
Parametrize the test to check different versions of pip (including master) are correctly supported. Fix #1649.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ install:
 - "! grep pyc setuptools.egg-info/SOURCES.txt"
 
 script:
+  - export NETWORK_REQUIRED=1
   - |
     ( # Run testsuite.
       if [ -z "$DISABLE_COVERAGE" ]
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,7 @@ clone_depth: 50
 environment:
 
   APPVEYOR: True
+  NETWORK_REQUIRED: True
   CODECOV_ENV: APPVEYOR_JOB_NAME
 
   matrix:
--- a/setuptools/tests/test_virtualenv.py
+++ b/setuptools/tests/test_virtualenv.py
@@ -52,10 +52,24 @@ def test_clean_env_install(bare_virtualenv):
     )).format(source=SOURCE_DIR))
 
 
-def test_pip_upgrade_from_source(virtualenv):
+@pytest.mark.parametrize('pip_version', (
+    'pip==9.0.3',
+    'pip==10.0.1',
+    'pip==18.1',
+    'pip==19.0.1',
+    'https://github.com/pypa/pip/archive/master.zip',
+))
+def test_pip_upgrade_from_source(virtualenv, pip_version):
     """
     Check pip can upgrade setuptools from source.
     """
+    # Install pip/wheel, and remove setuptools (as it
+    # should not be needed for bootstraping from source)
+    virtualenv.run(' && '.join((
+        'pip uninstall -y setuptools',
+        'pip install -U wheel',
+        'python -m pip install {pip_version}',
+    )).format(pip_version=pip_version))
     dist_dir = virtualenv.workspace
     # Generate source distribution / wheel.
     virtualenv.run(' && '.join((
--- a/setuptools/tests/test_virtualenv.py
+++ b/setuptools/tests/test_virtualenv.py
@@ -52,24 +52,55 @@ def test_clean_env_install(bare_virtualenv):
     )).format(source=SOURCE_DIR))
 
 
-@pytest.mark.parametrize('pip_version', (
-    'pip==9.0.3',
-    'pip==10.0.1',
-    'pip==18.1',
-    'pip==19.0.1',
-    'https://github.com/pypa/pip/archive/master.zip',
-))
-def test_pip_upgrade_from_source(virtualenv, pip_version):
+def _get_pip_versions():
+    # This fixture will attempt to detect if tests are being run without
+    # network connectivity and if so skip some tests
+
+    network = True
+    if not os.environ.get('NETWORK_REQUIRED', False):  # pragma: nocover
+        try:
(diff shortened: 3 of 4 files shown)


pganssle:
I'm wondering if this might be a bug in `pytest-virtualenv`. It seems that that project is not tested on Windows at all. Maybe we should move away from using it? I think we can probably get away with a much simpler fixture.

pganssle:
The one concern I have here is that this requires an internet connection to succeed, which means that these tests will always fail if you're testing offline. I am not sure how big a concern that has been for us, historically and if there are already other tests that fail under those conditions.

It seems like hitting github specifically is also making the test a bit flakier. Maybe we should refactor this a bit to download all the artifacts we need from PyPI and do a conditional skip (possibly with a warning) for all the tests where it fails.

benoit-pierre:
I'm pretty sure that's not the only test needing network access. We can always mark them (`pytest.mark.network`) so they can easily be skipped when testing offline.

pganssle:
@benoit-pierre I disabled my network and ran `tox -e py37` on this branch, the only tests that failed are these ones, and they failed *super* slowly because of all the retrying.

pganssle:
How about something like this:

```diff
diff --git a/setuptools/tests/test_virtualenv.py b/setuptools/tests/test_virtualenv.py
index bd89fd64..784878ad 100644
--- a/setuptools/tests/test_virtualenv.py
+++ b/setuptools/tests/test_virtualenv.py
@@ -52,24 +52,55 @@ def test_clean_env_install(bare_virtualenv):
     )).format(source=SOURCE_DIR))
 
 
-@pytest.mark.parametrize('pip_version', (
-    'pip==9.0.3',
-    'pip==10.0.1',
-    'pip==18.1',
-    'pip==19.0.1',
-    'https://github.com/pypa/pip/archive/master.zip',
-))
-def test_pip_upgrade_from_source(virtualenv, pip_version):
+def _get_pip_versions():
+    # This fixture will attempt to detect if tests are being run without
+    # network connectivity and if so skip some tests
+
+    network = True
+    if not os.environ.get('NETWORK_REQUIRED', False):
+        try:
+            from urllib.request import urlopen
+            from urllib.error import URLError
+        except ImportError:
+            from urllib2 import urlopen, URLError # Python 2.7 compat
+
+        try:
+            urlopen('https://pypi.org', timeout=1)
+        except URLError:
+            # No network, disable most of these tests
+            network = False
+
+    network_versions = [
+        'pip==9.0.3',
+        'pip==10.0.1',
+        'pip==18.1',
+        'pip==19.0.1',
+        'https://github.com/pypa/pip/archive/master.zip',
+    ]
+
+    versions = [None] + [
+        pytest.param(v, **({} if network else {'marks': pytest.mark.skip}))
+        for v in network_versions
+    ]
+
+    return versions
+
+
+@pytest.mark.parametrize('pip_version', _get_pip_versions())
+def test_pip_upgrade_from_source(pip_version, virtualenv):
     """
     Check pip can upgrade setuptools from source.
     """
     # Install pip/wheel, and remove setuptools (as it
     # should not be needed for bootstraping from source)
+    if pip_version is None:
+        upgrade_pip = ()
+    else:
+        upgrade_pip = ('python -m pip install -U {pip_version} --retries=1',)
     virtualenv.run(' && '.join((
         'pip uninstall -y setuptools',
         'pip install -U wheel',
-        'python -m pip install {pip_version}',
-    )).format(pip_version=pip_version))
+    ) + upgrade_pip).format(pip_version=pip_version))
     dist_dir = virtualenv.workspace
     # Generate source distribution / wheel.
     virtualenv.run(' && '.join((
```

We'd also want to either add `NETWORK_REQUIRED` into the CI build environments, or switch the environment check to an existing indicator that we're in a CI build.

We may also want to do something about the possible flakiness of that github end point, possibly by dropping it entirely.

pganssle:
@benoit-pierre I've pushed network detection logic into my branch at #1676. Do you want me to push them to your branch and we can merge it, or do you disagree with the approach?

benoit-pierre:
I still think it's much simpler to use a pytest mark:
```diff
--- i/setuptools/tests/test_virtualenv.py
+++ w/setuptools/tests/test_virtualenv.py
@@ -52,6 +52,7 @@ def test_clean_env_install(bare_virtualenv):
     )).format(source=SOURCE_DIR))
 
 
+@pytest.mark.network
 @pytest.mark.parametrize('pip_version', (
     'pip==9.0.3',
     'pip==10.0.1',
```
And opt-out network tests when needed, e.g.: `tox -- -m 'not network'`.

pganssle:
@benoit-pierre Certainly it's easier, but it's not a great developer experience, particularly because it takes a super long time for these tests to fail if you're not on network and the tests run towards the end of the suite.

pganssle:
@jaraco Any chance you want to break the tie here on how to handle tests that require the network? I'd like to get these changes merged. My approach for making the network tests opportunistic is at #1676.

pganssle:
Talked to Jason at the HackIllinois event, we decided we'll go with the automatic network detection thing for now, then refactor it into a feature later to make it more general.

PR #1669, opened 2019-02-01
Drop build-backend from pyproject.toml

pganssle:
Per @gaborbernat's suggestion, dropping the `build-backend`. I don't know if it makes sense to keep `build-system` at all, but this is the minimal change required to allow us to use `--no-use-pep517` (which is required on the current version of `pip` when building from the repo).

Bernat discovered this when building the first infrastructure for the integration testing repo.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,5 @@
 [build-system]
 requires = ["wheel"]
-build-backend = "setuptools.build_meta"
 
 [tool.towncrier]
     package = "setuptools"


pyproject.toml
```diff
@@ -1,6 +1,5 @@
 [build-system]
-requires = ["wheel"]
-build-backend = "setuptools.build_meta"
+requires=["wheel"]
```
benoit-pierre: Nitpick: why the removal of spaces around `=`? (which differs from the rest of the file).

benoit-pierre:
Also, is even pip building/installing from the repo supported? Isn't there an issue due to the need to run `bootstrap.py` first?

pyproject.toml
pganssle: Oops, this is an accident, let me fix it.

pganssle:
> Also, is even pip building/installing from the repo supported? Isn't there an issue due to the need to run bootstrap.py first?

Well, regardless of whether it is or isn't, at the moment the `build-backend` field is not accurate, because we aren't declaring a dependency on `setuptools`.

pyproject.toml
pganssle: Fixed.

Issue #1663, opened 2019-01-31
v40.7.1 omits port in index download


kokoska69:
In this commit https://github.com/pypa/setuptools/commit/3a351aa00e2110c3b3df0302b1ed6bcf66a1f8d8 is used host instead of netloc, but host does not contain port. Version 40.6.3 works OK.

jaraco:
Sorry about the regression. After re-reading the diff, I can see the issue, which was probably brought on by the naming of 'host' when it meant 'netloc'. Fix should be straightforward.

Issue #1499, opened 2018-09-21
Python 3.8 urllib.parse.splituser DeprecationWarning


tirkarthi:
Using setuptools on CPython master (Python 3.8) generates below warning. Just wanted to give a heads up on this.

Relevant commit : https://github.com/python/cpython/commit/0250de481

```
/home/karthi/requests-html-env/lib/python3.8/site-packages/setuptools/package_index.py:1049: DeprecationWarning: urllib.parse.splituser() is deprecated as of 3.8, use urllib.parse.urlparse() instead
  auth, host = urllib.parse.splituser(netloc)
```

Relevant code : 

https://github.com/pypa/setuptools/blob/e0433cf32dfdf39c0c14a3740c3920c496f09486/setuptools/package_index.py#L1057

Thanks

pganssle:
PR #1501 solves this, but it is blocked on #1501.

PR #1670, opened 2019-02-03
Revert to using a copy of splituser from Python 3.8.

### LINKED ISSUE #1668: Bugfix #1666 break access to index urls that include auth
The fix for issue #1666 breaks with urls that include a username and password. Here's an illustration of how it breaks:
```
$ easy_install -i https://user:password@pypi.example.org/simple/install example
Searching for example
Reading https://user:password@pypi.example.org/simple/install/example/
Download error on https://user:password@pypi.example.org/simple/install/example/: nonnumeric port: 'password@pypi.example.org' -- Some packages may not be found!
```

jaraco:
Using urllib.parse.urlparse is clumsy and causes problems as reported in #1663 and #1668. Alternative to #1499 and fixes #1668.

## Summary of changes

In #1499, we learned that Python [deprecated splituser here](https://bugs.python.org/issue27485) and that the recommended change was to use `urllib.parse.urlparse`. However, as we saw in the original patch and later in regressions, in both places where `splituser` was previously used, the use of `urllib.parse.urlparse` doesn't satisfy the needs that `splituser` did.

So this PR steals the latest implementation of splituser from the Python 3.8 source and uses it instead, largely reverting the changes from #1501 and #1666.

I note that this change restores a flaw in the logic for `open_with_auth`: if `scheme` is not `http/s`, `address` (formerly `host`) is never defined, but if auth is found in PyPIConfig, a `NameError` could occur on line 1072 because `address` isn't defined. Since this flaw existed in the old code and didn't cause any issues, I'm not interested in fixing it.

Closes #1668

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -1083,7 +1083,7 @@ def open_with_auth(url, opener=urllib.request.urlopen):
         # Put authentication info back into request URL if same host,
         # so that links found on the page will work
         s2, h2, path2, param2, query2, frag2 = urllib.parse.urlparse(fp.url)
-        if s2 == scheme and h2 == parsed.hostname:
+        if s2 == scheme and h2 == address:
             parts = s2, netloc, path2, param2, query2, frag2
             fp.url = urllib.parse.urlunparse(parts)
 


cjerdonek:
FYI, pip also has an implementation of this here (with unit tests elsewhere in the code base): https://github.com/pypa/pip/blob/f048eb7a76eab222a4dfcfc3fd01f1c9e992f49a/src/pip/_internal/utils/misc.py#L941
`distlib` does, too: https://bitbucket.org/pypa/distlib/issues/116/parse_credentials-should-percent-decode
You might want to check to see if you need to unquote at any point, too.

PR #1676, opened 2019-02-06
WIP: Debugging branch for pip upgrades in test_virtualenv

pganssle:
Rather than spamming and force-pushing to #1659, I've created this new PR for debugging the issue with `pip install -U pip` on appveyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ install:
 - "! grep pyc setuptools.egg-info/SOURCES.txt"
 
 script:
+  - export NETWORK_REQUIRED=1
   - |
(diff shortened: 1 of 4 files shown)


benoit-pierre:
It's the usual Windows issue, the code need to be changed to use `python -m pip`, otherwise pip's executable cannot be upgraded (because it's in use).

pganssle:
@benoit-pierre Ah, I didn't realize this, OK.

pganssle:
@benoit-pierre Do you want to push an update to #1659, otherwise I can.

benoit-pierre:
I've just pushed one.

PR #1679, opened 2019-02-10
Add setuptools.get_version(path, field='__version__')

### LINKED ISSUE #1316: Helper function to read __version__
Maintaining a consistent version number across different parts of a package is a burden for authors. It may be needed in different places such as the package/module itself, setup.py or a sphinx conf.py.

The [packaging guide](https://packaging.python.org/guides/single-sourcing-package-version/) describes a total of 7 different approaches.

I propose to support at least a few of them with a new `get_version()` function in setuptools.

~~~
def get_version(filename, rel_to_filname=None, method='regexp'):
    """
    Read the content of the variable __version__ from a file.
    
    Arguments:
        filename: 
            The file containing __version__
        rel_to_filename:
            If given, filname will be interpreted to be relative to
            the directory of this file. Otherwise, it will be relative
            to the working directory.
            In most cases, you will want to pass __file__.
        method:
            If 'regexp', the value of __version__ will be read determined
            via a regular expression. This requires, that it's a string.
            If 'exec', the file will be executed and the local variable
            __version__ will be read out. This if __version__ is defined
            in a more complex way. However, the disadvatage is that the
            file will be executed, including possible side-effects. 
    
    Returns:
        The __version__ value or None, if it could not be determined.
~~~

The basic implementation should support methods 1 (parse the file for a regexp) and 3 (execute the file in a separate context). IMO, these are the best approaches unless you have or need further functionality like extra tools (2) or VCS support (7).

The idea is to have `__version__` set in your source code and fetch that value in helper scripts like setup.py or conf.py with a single line of code.

Example uses:
- `get_version('mypackage.__init__.py', __file__)`
- `get_version('../mypackage._version.py, __file__, method='exec')`

I figure this is a reasonable functionality to provide for setuptools. If there is interest, I can provide a PR.

techtonik:
## Summary of changes

This allows people who use setuptools to maintain project versions
in a single place, for example as attributes in their Python source.

Closes #1316
Closes https://github.com/pypa/pip/pull/6004
Closes https://github.com/pypa/python-packaging-user-guide/pull/571

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ def pypi_link(pkg_filename):
 
 setup_params = dict(
     name="setuptools",
-    version="40.8.0",
+    version=setuptools.get_version("setup.cfg", field="current_version"),
     description=(
         "Easily download, build, install, upgrade, and uninstall "
         "Python packages"
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -27,7 +27,7 @@ __metaclass__ = type
 __all__ = [
     'setup', 'Distribution', 'Feature', 'Command', 'Extension', 'Require',
     'SetuptoolsDeprecationWarning',
-    'find_packages'
+    'find_packages', 'get_version'
 ]
 
 if PY3:
@@ -222,6 +222,26 @@ def findall(dir=os.curdir):
         make_rel = functools.partial(os.path.relpath, start=dir)
         files = map(make_rel, files)
     return list(files)
+
+
+def get_version(path, field='__version__'):
+    """
+    Get version information from a text file. Version is extracted
+    from "key = value" format with key matched at the beginning of
+    a line and specified by `field` parameter. Returns string.
+    """
+    version_file = os.path.abspath(path)
+    # Using binary matching to avoid possible encoding problems
+    # when reading arbitrary text files
+    if type(field) is not bytes:
+        field = field.encode('utf-8')
+    for line in open(version_file, 'rb'):
+        if line.startswith(field):
+            # __version__ = "0.9"
+            # current_version = 4.8.0
+            _, value = line.split(b'=')
+            version = value.strip(b' \'\"').decode()
+            return version
 
 
 # Apply monkey patches
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -240,7 +240,7 @@ def get_version(path, field='__version__'):
             # __version__ = "0.9"
             # current_version = 4.8.0
             _, value = line.split(b'=')
-            version = value.strip(b' \n\t\'\"').decode()
+            version = value.strip(b' \r\n\t\'\"').decode()
             return version
 
 
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -240,7 +240,7 @@ def get_version(path, field='__version__'):
             # __version__ = "0.9"
             # current_version = 4.8.0
             _, value = line.split(b'=')
-            version = value.strip(b' \'\"').decode()
+            version = value.strip(b' \n\t\'\"').decode()
             return version
 
 
--- a/setuptools/tests/test_get_version.py
+++ b/setuptools/tests/test_get_version.py
@@ -26,23 +26,23 @@ class TestFiles:
     def setup_method(self, method):
         self.tmpdir = tempfile.mkdtemp()
 
-        self.file_python = os.path.join(self.tmpdir, 'version.py') 
-        with open(self.file_python, 'w') as fp:
-            fp.write('__version__ = "0.23beta"\n')
-
-        self.file_russian = os.path.join(self.tmpdir, 'russian.py') 
-        with open(self.file_russian, 'wb') as fp:
-            fp.write('# файл в русской кодировке\n\n'.encode('cp1251'))
-            fp.write('__version__ = "17.0"\n')
-
     def teardown_method(self, method):
-        shutil.rmtree(self.dist_dir)
+        shutil.rmtree(self.tmpdir)
 
     def test_python_file(self):
-        version = get_version(self.ver_python)
+        path = os.path.join(self.tmpdir, 'version.py')
+        with open(path, 'w') as fp:
+            fp.write('__version__ = "0.23beta"\n')
+
+        version = get_version(path)
         assert version == '0.23beta'
 
     def test_non_utf8_python_file(self):
-        version2 = get_version(self.ver_russian)
-        assert version2 == '17.0'
+        path = os.path.join(self.tmpdir, 'russian.py')
+        with open(path, 'wb') as fp:
+            fp.write(u'# файл в русской кодировке\n\n'.encode('cp1251'))
+            fp.write('__version__ = "17.0"\n')
+
+        version = get_version(path)
+        assert version == '17.0'
 
--- a/setuptools/tests/test_get_version.py
+++ b/setuptools/tests/test_get_version.py
@@ -1,0 +1,40 @@
+# -*- encoding: utf-8 -*-
+
+"""Tests for setuptools.get_version()."""
+import os
+import codecs
+import tempfile
+
+import pytest
+
+from setuptools import __version__, get_version
+
+
+def test_get_version():
+    version = get_version('setup.cfg', field='current_version')
+    assert version == __version__
+
+
+class TestFiles:
+    def setup_method(self, method):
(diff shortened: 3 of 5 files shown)


pganssle:
I am not sure why this function exists, it seems like it is designed to pull the version string from `setup.cfg`, but you can already specify the version string in `setup.cfg`:

```ini
[metadata]
name=foo
version=0.0.1
```

Why would you specify it in `setup.cfg` and then have a function that parses it from `setup.cfg` in `setup.py`?

In general, I think we are more interested in going the other way - we don't want to encourage people to use `setup.py` for things as trivial as single-sourcing their versions.

Additionally, #1316 is already closed and you can use `importlib_metadata` for it.

techtonik:
The function is designed to pull the version string from any file including various `package/__init__.py` and `package/version.py`. This method is used by `pip` and by [2000+ other](https://github.com/search?q=find_version%28*file_paths%29+filename%3Asetup.py) Python projects:

https://github.com/pypa/pip/blob/fb944ab62f755e71204020bc5b9881367803e5ec/setup.py#L18

And by a will of `setuptools` folks all this copy/paste could be replaced by a simple:

    from setuptools import get_version, setup

It even can save you some time to update version in [both](https://github.com/pypa/setuptools/commit/c1243e96f05d3b13392a792144c97d9471581550) `setup.cfg` and `setup.py` as well as obsolete the list of seven (import this!) ways for single sourcing https://packaging.python.org/guides/single-sourcing-package-version/ If `import_metadata` solves the problem, it deserves to the there. But in the meanwhile this small contribution could make some people happy (including good ones).

setup.py
```diff
@@ -89,7 +89,7 @@ def pypi_link(pkg_filename):
 
 setup_params = dict(
     name="setuptools",
-    version="40.8.0",
```
jaraco: Although you _could_ use `get_version` here, this usage would be discouraged. Better would be to store the value in setup.cfg under `metadata.version` (even if in duplicate), although doing so leads to [a limitation in bump2version](https://github.com/c4urself/bump2version/issues/62). What's more concerning about this change is the issues it introduces if setuptools were to move to a declarative config (something that's definitely desirable)... and this speaks to the resistance to accepting this change--it's encouraging patterns that run contrary to the best practices.

setuptools/tests/test_get_version.py
```diff
@@ -0,0 +1,49 @@
...
+        path = os.path.join(self.tmpdir, 'version.py')
+        with open(path, 'w') as fp:
+            fp.write('__version__ = "0.23beta"\n')
+
+        version = get_version(path)
+        assert version == '0.23beta'
+
+    def test_non_utf8_python_file(self):
+        path = os.path.join(self.tmpdir, 'russian.py')
+        with open(path, 'wb') as fp:
```
jaraco: This should use `io.open(..., encoding='cp1251')`.

setuptools/__init__.py
```diff
@@ -224,5 +224,25 @@ def findall(dir=os.curdir):
...
+    """
+    Get version information from a text file. Version is extracted
+    from "key = value" format with key matched at the beginning of
+    a line and specified by `field` parameter. Returns string.
+    """
+    version_file = os.path.abspath(path)
+    # Using binary matching to avoid possible encoding problems
+    # when reading arbitrary text files
+    if type(field) is not bytes:
+        field = field.encode('utf-8')
```
jaraco: This encode should happen unconditionally.

setuptools/__init__.py
```diff
@@ -224,5 +224,25 @@ def findall(dir=os.curdir):
     return list(files)
 
 
+def get_version(path, field='__version__'):
+    """
+    Get version information from a text file. Version is extracted
+    from "key = value" format with key matched at the beginning of
+    a line and specified by `field` parameter. Returns string.
+    """
+    version_file = os.path.abspath(path)
```
jaraco: I suspect this line has no useful effect.

pganssle:
> I'm still on the fence with this one. On one hand, this functionality provides a convenience for packages already dependent on this behavior. In doing so, it also provides a hook where such behavior could be discouraged or deprecated.

I'm increasingly strongly *against* this change. It seems very fragile compared to many of the other single-sourcing options out there, and it encourages a pattern that I would prefer to discourage. The existing users of this pattern already have something that works for them.

I agree with you that to the extent that a change needs to be made, it should be adding more options for single-sourcing your versions in the `setup.cfg` file.

lmittmann:
Single sourcing the package version from the variable `__version__` in `pkg/__init__.py` is a well known and commonly used pattern. The function `get_version()` would finally provide a handy solution for retrieving that version in `setup.py` analogous to `find_packages()` for retrieving packages.

As of today it is not possible to use `setup.cfg` only to describe the package. Instead one must define different information across `setup.py` and `setup.cfg`. It seems absurd, to store the version number in `version.txt` only to import it from `pkg/__init__.py` and `setup.cfg`, when it could be as simple as calling `get_version()` from within `setup.py` and keep storing the version at `pkg/__init__.py`.

The future might be a pure `setup.cfg` for package description with version extraction mechanisms similar to what `get_version()` does. Until then the function `get_version()` provides a great improvement to a well known and widely accepted pattern.

pganssle:
@lmittmann I don't see how this is a great improvement. Generally speaking we don't want people using the imperative build system for super common patterns because it's unnecessarily powerful and error-prone. This patch *itself* is very error prone and makes some very specific assumptions about how the version will be specified.

For example, if someone specifies their version like this:

```python
VERSION_MAJOR = 1
VERSION_MINOR = 2
VERSION_PATCH = 0

__version__ = '.'.join((VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH))
```

This `get_version` function will not work. Similarly it won't work if the version is set somewhere and then modified, possibly conditionally, like this:

```python
__version__ = '0.2.4'

if is_dev():
    __version__ += ".dev" + get_dev_version()
```

I'll also note that this function doesn't even write *one test* using it to enable the pattern of "pull the version from `__version__`", instead, the test is to pull it from `setup.cfg`, which doesn't really make sense, since you can *already* specify the version in `setup.cfg`. That is not a good sign that it's ready to be the go-to way to fetch versions.

It is very difficult to write truly generic functions that work in a wide variety of circumstances, but it's very easy to write functions that work for your particular situation. Given that we're really hoping to get people away from using the imperative build configuration (`setup.py`) for simple things where they *should* be able to use the declarative build configuration (`setup.cfg`) and this kind of function is pretty easy to write if you know exactly how you are specifying your version, I don't think we should be spending our time building and maintaining such a thing.

jaraco:
Thanks for the contrib, but the consensus is we will decline.

setup.py
techtonik: So what is the ETA to introduce the best practices to `pip`? Why doesn't it use best practices?

setuptools/tests/test_get_version.py
techtonik: The test is called `non_ut8`. Python doesn't know that it is russuan or not - filename can be any.

setuptools/__init__.py
techtonik: What if it is already `bytes`?

setuptools/__init__.py
techtonik: Could be. Just a safe practice, useful for debugging.

techtonik:
@pganssle your story is very far-fetched. It could make sense if `get_version` was a default behavior, but it is not - it is just an opt-in for recommended scenario #1 from Python packaging guide https://packaging.python.org/guides/single-sourcing-package-version/

@jaraco if https://github.com/pypa/setuptools/pull/1687 solves single sourcing problem, why it is not possible to remove `get_version` brother from `pip` codebase. https://github.com/pypa/pip/blob/master/setup.py#L18 ?

setuptools/tests/test_get_version.py
jaraco: What I should have asked is if you could open this file as text with an encoding (cp1251) and then write unicode text to it.

setup.py
jaraco: Support for declarative config is relatively recent and I haven't spent much time evangelizing it, so it's probably no big surprise that some projects haven't adopted it. Even setuptools itself, which because it installs itself, would have had no minimum dependency, only adopted declarative config many months after it was available. It's just a matter of priorities and resources. Perhaps that team would accept a pull request to use declarative config; I don't know.

PR #1687, opened 2019-02-17
Move version to declarative config

jaraco:
This change moves the setuptools version to declarative config. In order to accomplish this, the bump2version config had to be moved to .bumpversion.cfg.

I'm creating this PR in light of #1679, which conflicts with this change, to highlight the direction that I'd prefer the setuptools project should move.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,6 @@
+[bumpversion]
+current_version = 40.8.0
+commit = True
+tag = True
+
+[bumpversion:file:setup.cfg]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,3 @@
-[bumpversion]
-current_version = 40.8.0
-commit = True
-tag = True
-
 [egg_info]
 tag_build = .post
 tag_date = 1
@@ -25,6 +20,3 @@ universal = 1
 [metadata]
 license_file = LICENSE
 version = 40.8.0
-
-[bumpversion:file:setup.cfg]
-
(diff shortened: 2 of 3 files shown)


pganssle:
@jaraco Will `bumpversion` update the `setup.cfg` file for us?

Is this a bug in `bumpversion`? Can we fix this upstream in `bumpversion`?

pganssle:
Upstream issue is peritus/bumpversion#198 I guess.

jaraco:
Sadly, the peritus fork doesn't support annotated tags, so I recommend using bump2version, where [I filed the ticket today](https://github.com/c4urself/bump2version/issues/62).

pganssle:
This is fine with me, we can remove the bumpversion configuration when this is fixed in `bump2version`.

techtonik:
So the best practice for single sourcing version from `setuptools` is not providing the solution, but advising to add another dependency on `bump2version`. :-1:

pganssle:
@techtonik I would ask you to please have more respect when interacting with others on this issue tracker. You have recently made many negative, low-effort comments in several issues and PRs which are not constructive and some of which are simply off-topic. Please try to assume good faith from everyone involved in the project and keep your criticisms constructive.

In this instance, this PR is not an advertisement of best practices, just an improvement to the workflow of `setuptools`. We already had a dependency on `bump2version` and it has worked very well for us, to date, this is just moving more of our configuration into the declarative file.

techtonik:
@pganssle what is wrong with respecting an opinion in a non-personal way?

PR #1690, opened 2019-02-17
Honor declared ordering for Provides-Extra entries in package metadata.

jaraco:
Alternative approach to #1305, this approach simply replaces the set with an ordered set, retaining the order of extras as found in the package's 'extras_require' dictionary. On older Pythons where the dictionary ordering is non-deterministic, that non-determinism will still be present.

## Summary of changes

Closes 

### Pull Request Checklist
- [X] Changes have tests

### DIFF
--- a/setuptools/_vendor/vendored.txt
+++ b/setuptools/_vendor/vendored.txt
@@ -1,3 +1,4 @@
 packaging==16.8
 pyparsing==2.2.1
 six==1.10.0
+ordered-set
--- a/changelog.d/1690.change.rst
+++ b/changelog.d/1690.change.rst
@@ -1,0 +1,1 @@
+When storing extras, rely on OrderedSet to retain order of extras as indicated by the packager, which will also be deterministic on Python 2.7 (with PYTHONHASHSEED unset) and Python 3.6+.
--- a/setuptools/_vendor/ordered_set.py
+++ b/setuptools/_vendor/ordered_set.py
@@ -1,0 +1,488 @@
+"""
+An OrderedSet is a custom MutableSet that remembers its order, so that every
+entry has an index that can be looked up.
+
+Based on a recipe originally posted to ActiveState Recipes by Raymond Hettiger,
+and released under the MIT license.
+"""
+import itertools as it
+from collections import deque
+
+try:
+    # Python 3
+    from collections.abc import MutableSet, Sequence
+except ImportError:
+    # Python 2.7
+    from collections import MutableSet, Sequence
+
+SLICE_ALL = slice(None)
+__version__ = "3.1"
+
+
+def is_iterable(obj):
+    """
+    Are we being asked to look up a list of things, instead of a single thing?
+    We check for the `__iter__` attribute so that this can cover types that
+    don't have to be known by this module, such as NumPy arrays.
+
+    Strings, however, should be considered as atomic values to look up, not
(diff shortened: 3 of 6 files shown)


pganssle:
@jaraco Haven't looked into it too much, do we need to bring in `OrderedSet`? Are there set operations later on that are used on it? I'm assuming the `set` is being used because the requirements are supposed to be unique. If it doesn't have to be a set, maybe we can get away with the unique subset of the list?

```python
def unique_list(it):
    unique = set()
    out = []
    for item in it:
        if item not in unique:
            unique.add(item)
            out.append(item)
    return out
```

pganssle:
Ah, it does use `add`, I guess `OrderedSet` is fine for this.

jaraco:
I agree. Adding a dependency for this change is probably overkill. Especially if it only doesn’t address the root issue - that before Python 3.6, extras are going to be defined in a non-deterministic order. So maybe it makes sense to order them and remove duplicates explicitly.

jaraco:
@stevenengler Given the goal of having reproducible builds, would it be sufficient to solve that only on Python 3.6+ plus Python 2.7 (without PYTHONHASHSEED set)? If the solution could restrict to those environments, an approach similar to this one would be possible without enforcing a order (deferring to the order indicated by the package).

I'm guessing builders will want reproducible builds on all supported Python versions, but it would be simpler and less intrusive not to do so, and sets forth a pattern for giving packagers more control over the ordering of fields in the metadata.

jaraco:
> would it be sufficient to solve that only on Python 3.6+ plus Python 2.7

Let's assume yes.

yan12125:
Hi, what is the status of this pull request?

>> would it be sufficient to solve that only on Python 3.6+ plus Python 2.7
>
>Let's assume yes.

This is definitely sufficient for Arch Linux, as only the latest 3.x and 2.x versions are used for Python packages. Not sure about other major distros like Debian or Fedora, though.

jaraco:
> Hi, what is the status of this pull request?

I've had a shortcut on my desktop nagging me to deal with this since I last commented on it. I just hadn't gotten around to it. I have now and have merged the changes.

setuptools/_vendor/vendored.txt
```diff
@@ -1,3 +1,4 @@
 packaging==16.8
 pyparsing==2.2.1
 six==1.10.0
+ordered-set
```
benoit-pierre: Shouldn't this be pinned to the particular version vendored-in?

jaraco: Perhaps. I realized that the vendored dependencies are implicitly pinned when `paver update_vendored` is run... so I'm okay with leaving this declaration unpinned. But you're right to be concerned - why the inconsistency? Just because I'd rather remove the pins... but those other dependencies have probably advanced since they were last updated. My desire is to unpin the others and run another `update` operation.

jaraco: I do think it would be valuable if the dependency isn't declared as a pinned one to capture the effectively pinned version during the `update` operation.

benoit-pierre: On the other hand, pinning the dependencies make it easier to know which version of each package is vendored and to update only some of those dependencies.

jaraco: I've done that for now in 734d09c5.

PR #1703, opened 2019-02-23
Moved constants from setup.py to setup.cfg

### LINKED ISSUE #1697: Move simple constants from setup.py into setup.cfg
Currently we specify a lot of string literals in our `setup.py`. We're trying to encourage people to use declarative build configurations wherever possible, so we might as well start off at home move as much as we can over to `setup.cfg`. Should be almost everything that is a simple string literal.

So [in our `setup.py`](https://github.com/pypa/setuptools/blob/master/setup.py#L91), I would say we can move everything *except*:

1. `dependency_links` (Though I actually have no idea why those are in there in the first place)
2. `entry_points`: which dynamically reads the possible commands from some location somewhere.
3. `package_data`: Not sure what is going on with [this](https://github.com/pypa/setuptools/blob/master/setup.py#L71), but it actually seems dangerous for our universal wheels produced on Linux... For now let's leave it alone.

hpandeycodeit:
Summary of changes

Removed constants from setup.py to setup.cfg as per the issue #1697 
Closes #1697

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,3 +20,38 @@ universal = 1
 [metadata]
 license_file = LICENSE
 version = 40.8.0
+
+[setuptools]
+description= "Easily download, build, install, upgrade, and uninstall Python packages"
+author="Python Packaging Authority"
+author_email="distutils-sig@python.org"
+long_description=long_description
+long_description_content_type='text/x-rst; charset=UTF-8'
+keywords="CPAN PyPI distutils eggs package management"
+url="https://github.com/pypa/setuptools"
+project_urls= "Documentation": "https://setuptools.readthedocs.io/"
+src_root=None
+packages=setuptools.find_packages(exclude=['*.tests'])
+py_modules=['easy_install']
+zip_safe=True
+
+[classifiers]
+classifiers=
+        Development Status :: 5 - Production/Stable
+        Intended Audience :: Developers
+        License :: OSI Approved :: MIT License
+        Operating System :: OS Independent
+        Programming Language :: Python :: 2
+        Programming Language :: Python :: 2.7
+        Programming Language :: Python :: 3
+        Programming Language :: Python :: 3.4
+        Programming Language :: Python :: 3.5
+        Programming Language :: Python :: 3.6
+        Programming Language :: Python :: 3.7
+        Topic :: Software Development :: Libraries :: Python Modules
+        Topic :: System :: Archiving :: Packaging
+        Topic :: System :: Systems Administration
+        Topic :: Utilities
+python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*'
+extras_require="ssl:sys_platform=='win32'": "wincertstore==0.2", "certs": "certifi==2016.9.26"
+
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -20,3 +20,38 @@ universal = 1
...
+description= "Easily download, build, install, upgrade, and uninstall Python packages"
+author="Python Packaging Authority"
+author_email="distutils-sig@python.org"
+long_description=long_description
+long_description_content_type='text/x-rst; charset=UTF-8'
+keywords="CPAN PyPI distutils eggs package management"
+url="https://github.com/pypa/setuptools"
+project_urls= "Documentation": "https://setuptools.readthedocs.io/"
+src_root=None
+packages=setuptools.find_packages(exclude=['*.tests'])
```
pganssle: This is actually a function call in Python, in `setuptools` you annoyingly have to put it in two places, this line needs to be `packages = find:`

Then you create a new header called `[options.packages.find]`, and put 'exclude = *.tests` under it.

setup.cfg
```diff
@@ -20,3 +20,38 @@ universal = 1
...
+py_modules=['easy_install']
+zip_safe=True
+
+[classifiers]
```
pganssle: `[classifiers]` go under the `[metadata]` header.

setup.cfg
```diff
@@ -20,3 +20,38 @@ universal = 1
...
+author="Python Packaging Authority"
+author_email="distutils-sig@python.org"
+long_description=long_description
+long_description_content_type='text/x-rst; charset=UTF-8'
+keywords="CPAN PyPI distutils eggs package management"
+url="https://github.com/pypa/setuptools"
+project_urls= "Documentation": "https://setuptools.readthedocs.io/"
+src_root=None
+packages=setuptools.find_packages(exclude=['*.tests'])
+py_modules=['easy_install']
```
pganssle: Pretty sure this is not the syntax of a list, I think you want:

```ini
py_modules =
    easy_install
```

setup.cfg
```diff
@@ -20,3 +20,38 @@ universal = 1
...
+        Programming Language :: Python :: 3.4
+        Programming Language :: Python :: 3.5
+        Programming Language :: Python :: 3.6
+        Programming Language :: Python :: 3.7
+        Topic :: Software Development :: Libraries :: Python Modules
+        Topic :: System :: Archiving :: Packaging
+        Topic :: System :: Systems Administration
+        Topic :: Utilities
+python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*'
+extras_require="ssl:sys_platform=='win32'": "wincertstore==0.2", "certs": "certifi==2016.9.26"
```
pganssle: This is not the syntax for key-value pairs.

hpandeycodeit:
@pganssle I am getting these warning while running `python setup.py check` 

```

python setup.py check
running check
warning: check: missing required meta-data: name, url

warning: check: missing meta-data: either (author and author_email) or (maintainer and maintainer_email) must be supplied


```

jaraco:
This PR has been abandoned. Happy to accept a resubmit of this work (or an alternate implementation) when it's working.

PR #1706, opened 2019-02-24
Fix #1664: Include path in error message when version is missing

### LINKED ISSUE #1664: Include location of bad METADATA files in errors
`pkg_resources` should probably be including the location of the errant METADATA (e.g. path to the file) in the `ValueError` message to make troubleshooting easier. Maybe someone can file an issue there.

_Originally posted by @cjerdonek in https://github.com/pypa/pip/issues/6194#issuecomment-458878882_

cjerdonek:
Fixes #1664.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1403,8 +1403,15 @@ class NullProvider:
     def has_resource(self, resource_name):
         return self._has(self._fn(self.module_path, resource_name))
 
+    def _get_metadata_path(self, name):
+        return self._fn(self.egg_info, name)
+
     def has_metadata(self, name):
-        return self.egg_info and self._has(self._fn(self.egg_info, name))
+        if not self.egg_info:
+            return self.egg_info
+
+        path = self._get_metadata_path(name)
+        return self._has(path)
 
     def get_metadata(self, name):
         if not self.egg_info:
@@ -1867,6 +1874,9 @@ class FileMetadata(EmptyProvider):
 
     def __init__(self, path):
         self.path = path
+
+    def _get_metadata_path(self, name):
+        return self.path
 
     def has_metadata(self, name):
         return name == 'PKG-INFO' and os.path.isfile(self.path)
@@ -2663,8 +2673,12 @@ class Distribution:
         except AttributeError:
             version = self._get_version()
             if version is None:
-                tmpl = "Missing 'Version:' header and/or %s file"
-                raise ValueError(tmpl % self.PKG_INFO, self)
+                path = self._get_metadata_path_for_display(self.PKG_INFO)
+                msg = (
+                    "Missing 'Version:' header and/or {} file at path: {}"
+                ).format(self.PKG_INFO, path)
+                raise ValueError(msg, self)
+
             return version
 
     @property
@@ -2721,6 +2735,23 @@ class Distribution:
                     "%s has no such extra feature %r" % (self, ext)
                 )
         return deps
+
+    def _get_metadata_path_for_display(self, name):
+        """
+        Return the path to the given metadata file, if available.
+        """
+        try:
+            # We need to access _get_metadata_path() on the provider object
+            # directly rather than through this class's __getattr__()
+            # since _get_metadata_path() is marked private.
+            path = self._provider._get_metadata_path(name)
+
+        # Handle exceptions e.g. in case the distribution's metadata
+        # provider doesn't support _get_metadata_path().
+        except Exception:
+            return '[could not detect]'
+
+        return path
 
     def _get_metadata(self, name):
         if self.has_metadata(name):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2661,7 +2661,7 @@ class Distribution:
         try:
             return self._version
         except AttributeError:
-            version = _version_from_file(self._get_metadata(self.PKG_INFO))
+            version = self._get_version()
             if version is None:
                 tmpl = "Missing 'Version:' header and/or %s file"
                 raise ValueError(tmpl % self.PKG_INFO, self)
@@ -2726,6 +2726,12 @@ class Distribution:
         if self.has_metadata(name):
             for line in self.get_metadata_lines(name):
                 yield line
+
+    def _get_version(self):
+        lines = self._get_metadata(self.PKG_INFO)
+        version = _version_from_file(lines)
+
+        return version
 
     def activate(self, path=None, replace=False):
         """Ensure distribution is importable on `path` (default=sys.path)"""
@@ -2945,7 +2951,7 @@ class EggInfoDistribution(Distribution):
         take an extra step and try to get the version number from
         the metadata file itself instead of the filename.
         """
-        md_version = _version_from_file(self._get_metadata(self.PKG_INFO))
+        md_version = self._get_version()
         if md_version:
             self._version = md_version
         return self
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -17,6 +17,7 @@ try:
 except ImportError:
     import mock
 
+from pkg_resources import DistInfoDistribution, Distribution, EggInfoDistribution
 from pkg_resources.extern.six.moves import map
 from pkg_resources.extern.six import text_type, string_types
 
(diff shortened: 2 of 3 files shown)


cjerdonek:
@pypa/setuptools-developers On this PR, would you be okay with me making also the argument passed to `ValueError` be a single string instead of the two arguments `ValueError(msg, self)`? 
https://github.com/pypa/setuptools/blob/702f2d9127adc6fb6922a2c164b0a9a2856f203a/pkg_resources/__init__.py#L2666-L2667
This will make the exception easier to read because it won't introduce additional escaping of slashes in the path due to calling `repr()` on the string.

pganssle:
Thanks for this PR; I haven't had a chance to review it carefully though, but it looks very thorough!

> On this PR, would you be okay with me making also the argument passed to `ValueError` be a single string instead of the two arguments `ValueError(msg, self)`?

Hm, I find it a little weird when errors set an args tuple instead of a string as the message anyway, so I don't have a big problem with changing it conceptually, the question is whether anyone is *using* the fact that it currently raises a `tuple`?

Might be a little overboard, but maybe we could do this for "best of both worlds"?

```python
class MissingVersionError(ValueError):
    def __repr__(self):
        return "%s(%s, %s)" % (self.__class__.name + self.args)
```

Haven't looked at this too much, maybe it should be something like `InvalidMetadataError` instead, but you get my drift - keep the result of `.args` the same and make the REPR look however you want.

cjerdonek:
Okay, thanks, @pganssle. I'll think about it what I can do here to preserve that aspect, along the lines you suggest.

I was already planning to propose a new `ValueError` exception type as a follow-on PR (because it would e.g. let pip handle errors more gracefully without having to risk catching all `ValueError` objects). It requires a bit more consideration which is why I wanted to propose it separately, but I can start thinking about that here.

cjerdonek:
Okay, I gave this a bit of thought. To keep the scope of this PR limited, I'll keep everything the same here and _only_ add the path to the string portion of the `ValueError` args (as in my original PR). In a follow-on PR we can discuss and settle on creating a new exception type (since that will involve some bikeshedding, etc).

In the meantime, I adjusted the tests I created to also check that the `args` passed to `ValueError` have the expected type (`str` for the first one, and a particular `Distribution` sub-type for the second one). Those tests will come in handy for the follow-on PR.

How does that sound?

pganssle:
@cjerdonek I'm fine with either thing, tightly scoped PR or bigger one. I think your proposed course of action is prudent.

It seems like there's a test failure on Windows. Hard for me to tell what's going on there, but I think it's related to the fact that windows uses backslashes for path separators, and if you're testing against the `repr`, you're getting something double-escaped.

cjerdonek:
> It seems like there's a test failure on Windows.

Oops, it was working before, and I didn't check after my last force-push. (I had reversed two positional arguments `metadata_path ` and `expected_text` when calling `check_version_missing()`.)

It is ready now.

pkg_resources/__init__.py
```diff
@@ -1403,8 +1403,15 @@ def get_resource_string(self, manager, resource_name):
     def has_resource(self, resource_name):
         return self._has(self._fn(self.module_path, resource_name))
 
+    def get_metadata_path(self, name):
```
pganssle: This changes the public interface, do we really need to add a function to the public interface for this? I think I'd rather use a private method or do it in-line.

pkg_resources/__init__.py
```diff
@@ -2656,15 +2666,23 @@ def _warn_legacy_version(self):
 
         warnings.warn(tmpl.format(**vars(self)), PEP440Warning)
 
+    def _make_missing_version_message(self):
```
pganssle: This only seems to be used in one place, I think we can inline it for now (unless I'm missing something).

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,83 @@ def test_setuptools_not_imported(self):
         subprocess.check_call(cmd)
 
 
+def touch_file(path):
```
pganssle: Can you add a `TODO` comment about dropping this when Python 2 support is dropped? As of Python 3.4 `pathlib.Path` objects have a `touch` method.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,83 @@ def test_setuptools_not_imported(self):
...
+    """
+    Create an empty file.
+    """
+    with open(path, 'w') as f:
+        f.write('')
```
pganssle: I think you can write this more succinctly as:

```
with open(path, 'w'):
    pass
```

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,83 @@ def test_setuptools_not_imported(self):
...
+    touch_file(os.path.join(dist_dir, 'temp.txt'))
+
+    dists = list(pkg_resources.distributions_from_metadata(dist_dir))
+    assert len(dists) == 1
+    dist, = dists
+
+    return dist, dist_dir
+
+
+def check_version_missing(dist, metadata_path, expected_text,
```
pganssle: I think rename this to `assert_version_missing`, since this is essentially a fancy type of assertion.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,83 @@ def test_setuptools_not_imported(self):
...
+
+    # Also check the args passed to the ValueError.
+    msg, dist = excinfo.value.args
+    assert expected_text in msg
+    # Check that the message portion contains the path.
+    assert metadata_path in msg, str((metadata_path, msg))
+    assert type(dist) == expected_dist_type
+
+
+class TestDistribution:
```
pganssle: I'm not totally convinced that these need to be wrapped in classes, but if we stick with this design, for Python 2 support I think it would be preferable to derive them from `object` so they are new-style classes.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,83 @@ def test_setuptools_not_imported(self):
...
+
+    def test_version_missing(self, tmpdir):
```
pganssle: I think it would be best to refactor `TestDistribution.test_version_missing` and `TestDistributionInfo.test_version_missing` as a single parametrized test that takes `expected_dist_type` as a parameter.

That would also allow you to move `check_version_info` inline into the parametrized test itself.

cjerdonek:
Thanks for all of your comments, @pganssle. I've updated the PR with all of your suggested changes.

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -190,6 +191,70 @@ def test_setuptools_not_imported(self):
         subprocess.check_call(cmd)
 
 
+# TODO: remove this in favor of Path.touch() when Python 2 is dropped.
+def touch_file(path):
+    """
+    Create an empty file.
+    """
+    with open(path, 'w') as f:
```
pganssle: Minor nitpick, but `f` is unused, so this can just be:

```python
with open(path, 'w'):
    pass
```

pkg_resources/__init__.py
```diff
@@ -2722,11 +2736,34 @@ def requires(self, extras=()):
...
+        try:
+            # We need to access _get_metadata_path() on the provider object
+            # directly rather than through this class's __getattr__()
+            # since _get_metadata_path() is marked private.
+            path = self._provider._get_metadata_path(name)
+
+        # Handle exceptions e.g. in case the distribution's metadata
+        # provider doesn't support _get_metadata_path().
+        except Exception:
+            return '[could not detect]'
```
pganssle: The coverage is showing this as hit, but I don't see a test that actually hits it, do you know why that would be? Unexplained coverage worries me a bit.

cjerdonek: It looks like there were at least a few tests hitting that line. Here is the simplest example of those tests (`pkg_resources/tests/test_resources.py::TestDistro::testDistroDependsOptions`): https://github.com/pypa/setuptools/blob/16e452a42a3dbbb0ab3d3146ffa3b743cdca2539/pkg_resources/tests/test_resources.py#L326

Specifically, this line triggered it: https://github.com/pypa/setuptools/blob/16e452a42a3dbbb0ab3d3146ffa3b743cdca2539/pkg_resources/tests/test_resources.py#L349-L350

The test "manually" constructs a `Distribution` object here: https://github.com/pypa/setuptools/blob/16e452a42a3dbbb0ab3d3146ffa3b743cdca2539/pkg_resources/tests/test_resources.py#L138-L139

I updated the PR with another test case to explicitly test this code path. The test case does something similar to the example above (constructing a `Distribution` object manually). I also addressed your other review comment in the updated PR.

cjerdonek:
Hi @pganssle, just pinging you on this. I answered the question that you asked before, and also included another test case in response to your question.

cjerdonek:
You might want to reference #6177 or #6283 on pip’s tracker instead.

pganssle:
@cjerdonek Thanks, updated the commit message. Not sure why the pypy job is failing. Seems unrelated?

cjerdonek:
Yes, seems unrelated. Are other PR’s failing?

cjerdonek:
Could the PR need to be rebased? (not sure if master is auto-merged prior to running tests)

pganssle:
@cjerdonek I rebased it when I was messing around with the commit history, though I think Travis does auto-merge the base branch if possible.

I have occasionally seen intermittent multi-threading issues like this in pypy in the past. I seem to remember that they were resolved upstream, though, so maybe this is something different.

pganssle:
Thanks @cjerdonek, sorry for the delays and that last hiccup!

cjerdonek:
Thank you, @pganssle! 🎉This will help users a lot by letting them quickly locate and diagnose problematic installs.

pganssle:
@cjerdonek This is included in the 40.9.0 release - hot off the presses!

Issue #1671, opened 2019-02-04
build_meta doesn't use a fresh dist directory, which causes ValueError unpacking tuple


pganssle:
If you use `pip install .` on a source directory that already has a `dist`, the existing directory structure is used when preparing the wheel. This causes an exception if an existing wheel with a different name exists, because [build_meta assumes there's only one wheel in `dist`](https://github.com/pypa/setuptools/blob/64e60fc32981a1615c35962a60297d264bf16734/setuptools/build_meta.py#L157).

Here's a script to create a MWE repo:

```bash
#!/usr/bin/bash

mkdir /tmp/demo_dist_517
cd /tmp/demo_dist_517
echo "from setuptools import setup; setup()" > setup.py
echo "0.0.1" > VERSION
cat > setup.cfg << EOF
[metadata]
name = foo
version = file: VERSION
EOF
cat > pyproject.toml << EOF
[build-system]
requires = ["setuptools", "wheel"]
build-backend = "setuptools.build_meta"
EOF
```

At this point your repo looks like this:
```
$ tree
├── pyproject.toml
├── setup.cfg
├── setup.py
└── VERSION
```

Create a wheel in `dist`, then change the version:

```bash
pip wheel . --no-deps -w dist
echo "0.0.2" > VERSION
```

Now try to create a wheel from the repo:

```
pip wheel . -w dist
```

> This will trigger an error in `build_meta`:
>   File "/tmp/pip-build-env-plomixa1/overlay/.../setuptools/build_meta.py", line 157, in _file_with_extension
>    file, = matching
> ValueError: too many values to unpack (expected 1)
>  Building wheel for foo (PEP 517) ... error
>   Failed building wheel for foo

This is pretty easy to work around, obviously, since the user can just remove `dist` before invoking any `pip` commands, but it might be better to do the wheel build in a clean directory if possible.

pganssle:
I could be wrong, but I think we can somewhat trivially fix this by changing `build_meta.build_wheel` to directly use `--dist-dir=wheel_directory`, so:

pradyunsg:
FYI https://github.com/pypa/pip/commit/51086b371808c3ce093ef28a317df7b7880efffa

pganssle:
@pradyunsg Is this causing major problems? I put it in my mental backlog and left as a possible sprint item just because it seemed like it would only affect maintainers who leave build artifacts around in their source directories. If it's causing real problems for other people, I'll at least bump it up in priority from "minor".

[...]

PR #1726, opened 2019-03-26
Add failing test for issue #1671 (ValueError when .whl is present)

florisla:
## Summary of changes

Add a unit test for the bug in issue #1671.

This test is now failing (xfail) and should be passing when the bug is fixed.

### Pull Request Checklist
- [x] Changes have tests: this *is* a test, with the actual change still TBD.
- [x] ~News fragment `changelog.d`~: assumed to not be required yet, since no actual production code has been changed so far.

### DIFF
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -157,6 +157,44 @@ class TestBuildMetaBackend:
 
         assert os.path.isfile(os.path.join(dist_dir, wheel_name))
 
+    @pytest.mark.xfail(reason="Known error, see GH #1671")
+    def test_build_wheel_with_existing_wheel_file_present(self, tmpdir_cwd):
+        # Building a wheel should still succeed if there's already a wheel
+        # in the wheel directory
+        files = {
+            'setup.py': "from setuptools import setup\nsetup()",
+            'VERSION': "0.0.1",
+            'setup.cfg': DALS("""
+                [metadata]
+                name = foo
+                version = file: VERSION
+            """),
+            'pyproject.toml': DALS("""
+                [build-system]
+                requires = ["setuptools", "wheel"]
+                build-backend = "setuptools.build_meta
+            """),
+        }
+
+        build_files(files)
+
+        dist_dir = os.path.abspath('pip-wheel-preexisting')
+        os.makedirs(dist_dir)
+
+        # make first wheel
+        build_backend = self.get_build_backend()
+        wheel_one = build_backend.build_wheel(dist_dir)
+
+        # change version
+        with open("VERSION", "wt") as version_file:
+            version_file.write("0.0.2")
+
+        # make *second* wheel
+        wheel_two = self.get_build_backend().build_wheel(dist_dir)
+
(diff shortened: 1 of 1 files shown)


florisla:
The Travis failure is not related to the change; e.g. PR `1721` fails in the same way.

benoit-pierre:
@pganssle: what was the force push for?

pganssle:
@benoit-pierre Minor refactoring (everything else uses the `tmpdir_cwd` fixture instead of `tmpdir`, some minor style issues), plus I rebased from master and squashed the fixup commits.

benoit-pierre:
OK.

setuptools/tests/test_build_meta.py
```diff
@@ -157,6 +157,38 @@ def test_build_wheel(self, build_backend):
...
+            'setup.py': "from setuptools import setup\nsetup()",
+            'VERSION': "0.0.1",
+            'setup.cfg': "[metadata]\nname = foo\nversion = file: VERSION",
+            'pyproject.toml': DALS("""
+                [build-system]
+                requires = ["setuptools", "wheel"]
+                build-backend = "setuptools.build_meta
+            """),
+        }
+        build_files(files, prefix=str(tmpdir))
```
pganssle: Oops, this is going to fail, one moment.

setuptools/tests/test_build_meta.py
```diff
@@ -157,6 +157,38 @@ def test_build_wheel(self, build_backend):
...
+
+        # change version
+        with open("VERSION", "wt") as version_file:
+            version_file.write("0.0.2")
+
+        # make *second* wheel
+        wheel_two = self.get_build_backend().build_wheel(dist_dir)
+
+        assert os.path.isfile(os.path.join(dist_dir, wheel_one))
+        assert wheel_one != wheel_two
```
pganssle: Just tested, this actually fails at this assertion, not with the `ValueError` we are expecting.

benoit-pierre:
FWIW, `tmpdir` is a pytest fixture, I think it's better to use it over our own `tmpdir_cwd` fixture (which we should progressively phase out).

setuptools/tests/test_build_meta.py
pganssle: I'm not quite sure what's going on, but I think it's either not picking up the `setup.cfg` or it's not using it, because I'm seeing it generate wheels like `'UNKNOWN-0.0.0-py3-none-any.whl'`.

benoit-pierre: The original test (prior to force push) seems to work fine.

pganssle: Ah, found the problem, it was the `prefix=tmpdir_cwd` part. I've removed that and it now fails as expected.

pganssle:
@benoit-pierre I think the `tmpdir_cwd` fixture is a convenience fixture that is a wrapper around the `tmpdir` fixture. It just runs the test *in* the `tmpdir`, which is particularly convenient for these tests and anything else where you're expecting to run the whole thing in a clean directory.

Issue #1702, opened 2019-02-23
setup.cfg should standardize on UTF-8 for encoding


jaraco:
In jaraco/configparser#34, I learned that although setuptools v40.7.0 presumably added support for non-ASCII, there are still environments where loading non-ASCII is failing.

```
configparser # easy_install --version
setuptools 40.8.0 from c:\python37\lib\site-packages (Python 3.7)
configparser 3.7.2 # python setup.py egg_info
Traceback (most recent call last):
  File "setup.py", line 5, in <module>
    package_dir={'': 'src'},  
  File "C:\Python37\lib\site-packages\setuptools\__init__.py", line 144, in setup
    _install_setup_requires(attrs)  
  File "C:\Python37\lib\site-packages\setuptools\__init__.py", line 137, in _install_setup_requires
    dist.parse_config_files(ignore_option_errors=True)
  File "C:\Python37\lib\site-packages\setuptools\dist.py", line 702, in parse_config_files
    self._parse_config_files(filenames=filenames)
  File "C:\Python37\lib\site-packages\setuptools\dist.py", line 599, in _parse_config_files
    (parser.read_file if six.PY3 else parser.readfp)(reader)
  File "C:\Python37\lib\configparser.py", line 717, in read_file
    self._read(f, source)
  File "C:\Python37\lib\configparser.py", line 1014, in _read
    for lineno, line in enumerate(fp, start=1):
  File "C:\Python37\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 103: character maps to <undefined>
```

jaraco:
Hmm. On further investigation, I discovered that [encoding detection is employed](https://github.com/pypa/setuptools/blob/6575418283ec605f6bca82c525f6680d897ce282/setuptools/dist.py#L593), meaning that adding `# codec: utf-8` to the file corrected the issue.

jaraco:
In [this comment](https://github.com/pypa/setuptools/issues/1661#issuecomment-465088372), the user suggests the default encoding should be UTF-8. I agree and in fact I'd like to work toward the only encoding to be UTF-8 (drop support for encoding detection).

jaraco:
I see that the encoding detection was added in #1180. @benoit-pierre, do you recall why you chose to have the encoding declared rather than demanding and relying on UTF-8?

[...]

PR #1729, opened 2019-03-30
Preserve encoding with edit_config

itsadok:
PR #1180 and 24be5abd4cb changed the way we read setup.cfg, but when
editing setup.cfg we might still lose the encoding hint on the file,
which can cause failures in subsequent steps.
This change detectes the encoding like in `_parse_config_files` and
writes an encoding hint header if the file encoding was detected.
Fixes https://github.com/jaraco/configparser/issues/37.

### DIFF
--- a/setuptools/command/setopt.py
+++ b/setuptools/command/setopt.py
@@ -2,8 +2,11 @@ from distutils.util import convert_path
 from distutils import log
 from distutils.errors import DistutilsOptionError
 import distutils
+import io
 import os
+from ..unicode_utils import detect_encoding
 
+from setuptools.extern import six
 from setuptools.extern.six.moves import configparser
 
 from setuptools import Command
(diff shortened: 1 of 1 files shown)


itsadok:
Wrapped io.open with try-catch IOError, to better match the behavior of `ConfigParser.read()`

jaraco:
See #1702, where I'm proposing a somewhat different approach. Instead of investing more heavily in encoding declarations on a file, I'd rather standardize on UTF-8 as the sole encoding for the file.

itsadok:
OK, yeah. My code doesn't work, as I couldn't find a way to make Python 2.7's ConfigParser write UTF-8.
It would be much simpler to ignore encoding completely here, and assume UTF-8 in `_parse_config_py`. I'm closing this PR.

Issue #1730, opened 2019-04-02
Issue with PyPy on Travis


pganssle:
All jobs are now failing on `pypy`, see [this job](https://travis-ci.org/pypa/setuptools/jobs/513883062#L834). Locally I cannot replicate this with `pypy2.7-X`, but the same thing *does* fail with "Too many open files" in `pypy3.6-7.1.0`. It's very hard to debug this. @gaborbernat replicated it on a macbook, he similarly can't replicate the 2.7 failure, and the 3.6 failure is a different error message than the one I have.

There *may* be two problems here, possibly with similar causes. I think it's possible that Travis has scaled down the amount of memory available and that's biting us in the `pypy2.7` job? Hard to tell.

If we want a short-term mitigation, we can maybe try:

1. Try updating the `pypy` version in the job.
2. Try migrating the `pypy` job in Azure Pipelines (though that would require getting #1721 working.

This is a major blocking issue IMO.

webknjaz:
You can try `pypy2.7-6.0.0` and `pypy3.5-6.0.0` under `xenial`

webknjaz:
@pganssle it's very poorly advertised, but you can ask Travis support to enable debug on repo: https://docs.travis-ci.com/user/running-build-in-debug-mode/
I've done it today myself and they enabled it almost instantly :)
This gives you ssh (via tmate) into the VM and also there's helpers to run travis steps.

pganssle:
With @webknjaz's suggestion fixing the CI, I have downgraded this from "blocker" to "minor". I'm leaving it open a bit longer because this seems to be a real issue that might bite us later, and apparently has other manifestations on `pypy3.6`.

webknjaz:
`pypy3.6` isn't yet accessible in Travis, is it?

pganssle:
@webknjaz To be honest I don't know, but `pypy2.7` on Travis was failing with a resource error in the same tests that `pypy3.6` is currently failing with *different* resource errors. That suggests to me that the tests we have may be more fragile or resource-intensive than we think. Without knowing *why* this is causing particular stress on `pypy`, it's hard to tell if this is something that will bite us in the future, so I'd like to investigate why the `pypy3.6` problem (which is reproducible) is happening.

webknjaz:
@pganssle can you show me the Travis job? I didn't see any `pypy3.6` jobs on Travis...

pganssle:
[...]

PR #1731, opened 2019-04-02
Update pypy in travis

pganssle:
## Summary of changes

Attempting to update the Travis configuration to use a newer version of `pypy2.7` by updating to xenial, per @webknjaz's suggestion. Not sure this will actually work, so I've marked it WIP.

Trying to debug or solve #1730

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ jobs:
     python: 2.7
   - <<: *latest_py2
     env: LANG=C
-  - python: pypy
+  - python: pypy2.7-6.0.0
+    dist: xenial
     env: DISABLE_COVERAGE=1  # Don't run coverage on pypy (too slow).
   - python: pypy3
     env: DISABLE_COVERAGE=1


.travis.yml
```diff
@@ -8,7 +8,8 @@ jobs:
     python: 2.7
   - <<: *latest_py2
     env: LANG=C
-  - python: pypy
+  - python: pypy2.7
```
webknjaz: You should set a specific version, otherwise you might actually get something older.
```suggestion
  - python: pypy2.7-6.0.0
```

pganssle:
Huzzah, that "fixed" it!

webknjaz:
btw here's a place to complain about 7.0 and 7.1: https://travis-ci.community/t/add-pypy-7-support/2228

pganssle:
I've updated the commit message and the CI is fixed, but I will leave #1730 open a bit longer in case we can figure out the actual root cause.

Thanks for the quick fix @webknjaz!

PR #1735, opened 2019-04-05
When reading config files, require them to be encoded with UTF-8.

### LINKED ISSUE #1702: setup.cfg should standardize on UTF-8 for encoding

jaraco:
Closes #1702

### Pull Request Checklist
- [X] Changes have tests


benoit-pierre:
Shouldn't the `setuptools.unicode_utils.detect_encoding` helper be removed?

jaraco:
@pganssle I'd like to merge and release this today, so if you have any reservations, please let me know.

pganssle:
I don't have a major problem with the idea. My over-cautious instinct is to deprecate first, but this is such an easy thing to fix if it breaks something that maybe it's not worth it?

(Famous last words)

jaraco:
Good point. Here's my thinking: because the behavior was only recently released (January), probably not many project depend on the behavior or those that do are likely already cutting-edge enough to rely on this latest release. Moreover, if we find that releasing this does cause problems, I'll be on standby to release a bugfix release to add backward compatibility if needed (so only a small window of the problemmatic release).

pganssle:
Oh if you are reversing a recently introduced behavior then I think it's even less likely to cause problems, I say go for it.

jaraco:
Thanks both. In the interest of progress, I'm going to move forward with this, but don't hesitate to raise other concerns or thoughts.

PR #1743, opened 2019-04-14
Update super old CHANGES.rst dead links

gpotter2:
I was bumping in dead links a bit, so I git-logged them out of the way :smile:

## Summary of changes

- Update some dead old PR links to their commits counterpart

FTR because the project moved, the previous `PR`s are gone, therefore I've replaced them with commits instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1496,8 +1496,8 @@ v20.6.0
 19.7
 ----
 
-* `Off-project PR <https://github.com/jaraco/setuptools/pull/32>`_:
-  For FreeBSD, also honor root certificates from ca_root_nss.
+* Off-project PR: `0dcee79 <https://github.com/pypa/setuptools/commit/0dcee791dfdcfacddaaec79b29f30a347a147413>`_ and `f9bd9b9 <https://github.com/pypa/setuptools/commit/f9bd9b9f5df54ef5a0bf8d16c3a889ab8c640580>`_
+  For FreeBSD, also `honor root certificates from ca_root_nss <https://github.com/pypa/setuptools/commit/3ae46c30225eb46e1f5aada1a19e88b79f04dc72>`_.
 
 19.6.2
 ------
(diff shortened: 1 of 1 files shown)


gpotter2:
I also had an off-topic side question, and am not very confident with mailing-list :/

Could it be possible to add to each parameters of https://setuptools.readthedocs.io/en/latest/setuptools.html a tag `added in vX.Y.Z` ?

For instance, in [Scapy](https://github.com/secdev/scapy/blob/master/setup.py) we're using

```
extras_require={
    'basic': ["ipython"],
    'complete': [
        'ipython>=7;  python_version>="3"',
        'ipython<6;  python_version<"3"',
        'pyx>=0.14; python_version>="3"',
        'pyx==0.12.1; python_version<"3"',
        'cryptography',
        'matplotlib'
    ]
},
```
which not only contains a `extras_require` but also has `python_version` in it. This fails on pretty much any default installation (which are for the most packed with old setuptools versions), which is quite a burden :/

We could tell people to update their `setuptools` if we knew in which versions those were added :/

pganssle:
@gpotter2 That would be great. I think going forward we are trying to do this, but we would gladly accept PRs to annotate any features. I think there is a table of parameters accepted on setup.cfg that now has a "version added" column, and I'm a big fan of making that information readily accessible.

If you use `pyproject.toml`, you can pin a minimum setuptools version, and for good PEP 517 support I recommend pinning a recent version there.

jaraco:
> off-topic side question,

I don't care for mailing lists either. The workflow and user-interface is too heavy for my workload. Still, this would have been better posed as a separate issue. Almost no-one is going to expect packaging advice in a PR called "Update changelog".

What you're looking for is "environment marker" support for extras_require. There has been marker support going back to setuptools 0.8 (IIRC), but it used a different syntax:

```
'complete:python_version>="3"': [
        'ipython>=7;  python_version>="3"',
        'pyx>=0.14; python_version>="3"',
    ]
'complete': [
        'cryptography',
        'matplotlib'
    ]
'complete:python_version<"3"': [
        'ipython<6',
        'pyx==0.12.1',
    ]
```

Although honestly, it shouldn't be the responsibility of packages like Scapy to track which versions of which environments its dependencies support. It should be the responsibility of iPython and pyx to declare their 'python_requires' and users should use pip 9 or later, which will only pull the compatible versions for a given Python version. In other words, Scapy should be able to declare:

```
'complete': [
        'cryptography',
        'matplotlib'
        'ipython',
        'pyx',
    ]
```

I tested this approach, however, and did find that pyx fails to install on Python 2. That's a bug with pyx.

jaraco:
I [filed a ticket](https://sourceforge.net/p/pyx/bugs/66/).

gpotter2:
Thanks for the tips, I'll update that.

PyX said they would stop supporting Python 2 after the 12.1, which they did.

pganssle:
@jaraco By the way, I don't have a sourceforge account, but you may want to warn `pyx` that unfortunately it's not possible to "pull and replace" a precise module version, once you delete you can't re-publish even the same artifacts on PyPI.

PR #1745, opened 2019-04-21
Fix error with PEP 517 builds when wheel exists (GH #1761)

### LINKED ISSUE #1671: build_meta doesn't use a fresh dist directory, which causes ValueError unpacking tuple

shashanksingh28:
## Summary of changes

``build_meta.build_wheel`` uses a temporary directory for cases where ``wheel_directory`` already has 
existing `.whl` files (so as to not throw unpacking error).

Changes from the patch discussed in the issue:
- pay the cost of creating / cleaning a temp-dir only if necessary
- make it python2.7 compatible

Closes #1671 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1671.change.rst
+++ b/changelog.d/1671.change.rst
@@ -1,0 +1,1 @@
+Fixed issue with the PEP 517 backend that prevented building a wheel when the ``dist/`` directory contained existing ``.whl`` files.
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -35,6 +35,7 @@ import contextlib
 
 import setuptools
 import distutils
+from setuptools.py31compat import TemporaryDirectory
 
 from pkg_resources import parse_requirements
 
@@ -182,14 +183,22 @@ class _BuildMetaBackend(object):
                     metadata_directory=None):
         config_settings = self._fix_config(config_settings)
         wheel_directory = os.path.abspath(wheel_directory)
-        sys.argv = sys.argv[:1] + ['bdist_wheel'] + \
-            config_settings["--global-option"]
-        self.run_setup()
-        if wheel_directory != 'dist':
-            shutil.rmtree(wheel_directory)
-            shutil.copytree('dist', wheel_directory)
 
-        return _file_with_extension(wheel_directory, '.whl')
+        # Build the wheel in a temporary directory, then copy to the target
+        with TemporaryDirectory(dir=wheel_directory) as tmp_dist_dir:
+            sys.argv = (sys.argv[:1] +
+                        ['bdist_wheel', '--dist-dir', tmp_dist_dir] +
+                        config_settings["--global-option"])
+            self.run_setup()
+
+            wheel_basename = _file_with_extension(tmp_dist_dir, '.whl')
+            wheel_path = os.path.join(wheel_directory, wheel_basename)
+            if os.path.exists(wheel_path):
+                # os.rename will fail overwriting on non-unix env
+                os.remove(wheel_path)
+            os.rename(os.path.join(tmp_dist_dir, wheel_basename), wheel_path)
+
+        return wheel_basename
 
     def build_sdist(self, sdist_directory, config_settings=None):
         config_settings = self._fix_config(config_settings)
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -157,7 +157,6 @@ class TestBuildMetaBackend:
 
         assert os.path.isfile(os.path.join(dist_dir, wheel_name))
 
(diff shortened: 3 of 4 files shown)


setuptools/build_meta.py
```diff
@@ -35,6 +35,8 @@
 
 import setuptools
 import distutils
+from setuptools.py31compat import TemporaryDirectory
+
```
benoit-pierre: Drop the extra empty line.

setuptools/tests/test_build_meta.py
```diff
@@ -195,6 +194,13 @@ def test_build_wheel_with_existing_wheel_file_present(self, tmpdir_cwd):
         assert os.path.isfile(os.path.join(dist_dir, wheel_one))
         assert wheel_one != wheel_two
 
+        # and if rebuilding the same wheel?
+        open(os.path.join(dist_dir, wheel_two), 'w').close()
+        wheel_three = self.get_build_backend().build_wheel(dist_dir)
+        assert wheel_three == wheel_two
+
```
benoit-pierre: I'd remove this empty line to keep this part of the test as a whole block.

changelog.d/1671.change.rst
```diff
@@ -0,0 +1 @@
+Fixed error in ``build_meta.build_wheel`` that prevented building a wheel when the ``dist/`` directory contained existing ``.whl`` files.
```
benoit-pierre: This should mention PEP 517, something along those lines:
```
Fix possible issue with PEP 517 backend when building a wheel and the destination directory already contains some wheels.
```

pganssle:
@benoit-pierre Fixed the tweaks, let me know what you think of the new changelog.

Also, for some reason pypy3 is having test failures, not sure if that's related.

pganssle:
I think this is related to #1730, but it's hard to tell because Travis is timing out. I suspect that either PyPy or something we're doing is preventing some files or processes from being closed on PyPy, and that causes problems with this module in particular, because it opens and a ton of files (possibly hundreds?). We should probably try to actually resolve #1730.

pganssle:
Thanks @shashanksingh28 and @benoit-pierre and everyone else who contributed to fixing this issue.

benoit-pierre:
Note that `build_sdist` should be similarly fixed, not sure about `prepare_metadata_for_build_wheel`.

pganssle:
@benoit-pierre Would you mind creating a new issue for `build_sdist`? That one's probably also a good first issue.

gaborbernat:
No release since April 22, @pganssle can we do a release from master?

jaraco:
I'm on it.

jaraco:
[41.1 is going out here](https://travis-ci.org/pypa/setuptools/builds/571329988).

benoit-pierre:
@jaraco: the upload to PyPI failed.

jaraco:
I did notice that. I need to file a ticket for it.

benoit-pierre:
For which project? I have not tested the PyPI token support yet myself, did not know it existed before I saw your commit.

jaraco:
For this project. I successfully used the token approach already. I'll describe in the ticket.

PR #1747, opened 2019-04-22
Limit workers in ProcessPoolExecutor

pganssle:
As a mitigation for #1730, this commit limits the number of workers in the `ProcessPoolExecutor` to 1 (default is the number of CPUs). On PyPy, having a higher number of available workers dramatically increases the number of concurrent processes, leading to some resource exhaustion issues.

This does not address the root issue, but should improve the situation until the root issue is addressed.

### DIFF
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -28,7 +28,7 @@ class BuildBackend(BuildBackendBase):
 
     def __init__(self, *args, **kwargs):
         super(BuildBackend, self).__init__(*args, **kwargs)
-        self.pool = futures.ProcessPoolExecutor()
+        self.pool = futures.ProcessPoolExecutor(max_workers=1)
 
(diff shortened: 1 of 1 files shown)


pganssle:
I'm not entirely sure why more than one build backend has pools executing simultaneously. If that's a *necessary* feature and not something we can fix, maybe having a single shared pool for all backends would help this? Or switching to using just a `subprocess` invocation?

benoit-pierre:
Could be the same issue as when using pytest-xdist: sometimes the number of processor detected is really high, leading to too many concurrent jobs and crash / timeout issue with PyPy.

I agree with you that this "feature" does not seem necessary.

pganssle:
It seems like #1746 has mitigated the problem on Travis, but I think we should still merge this until we figure out why it's spawning even as many processes as it is, because I'm still seeing pypy failures when I run the test suite locally without this change.

Issue #1566, opened 2018-10-29
Start pinning doc and test requirements?


pganssle:
When making the issue #1565, I realized that if we don't do exact-version pins of the documentation requirements, some future person who wants to build old versions of the docs may start running into errors. Same for people trying to run old versions of the tests.

I'm thinking maybe we should start pinning all development requirements, and set up a bot to monitor them for changes and automatically bump them. That will make it much easier for people in the future to reproduce old behavior of the repo.

jaraco:
My (often unpopular) opinion on pinning is that one should only pin if there's a known or expected reason to pin. You've given one

> some future person who wants to build old versions of the docs may start running into errors.

That use case doesn't seem a very likely or strong one (as one could manually install older, compatible verisons)... or simply go back to Travis or RTD and find out what versions were used at that time.

In my experience, pinning versions is a much higher maintenance burden than addressing emergent concerns when they arise. But if you (or anyone) is willing to absorb that burden (or better, mitigate it), then I'm +0.

pganssle:
> My (often unpopular) opinion on pinning is that one should only pin if there's a known or expected reason to pin.

In general I agree with this, and I strongly agree with it **for libraries**, but there are other reasons to pin the test dependencies as well. For example, the first person making a PR after a test dependency breaks something in our build will start getting broken CI, whereas if we have all our test dependencies pinned, the CI breaks on the PR that upgrades the dependency. No need to write a separate PR to fix the build, then rebase all the old PRs against it just to get CI working again.

> That use case doesn't seem a very likely or strong one (as one could manually install older, compatible verisons)... or simply go back to Travis or RTD and find out what versions were used at that time.

I have recently revisited an old project that didn't pin its versions, and the tests were failing. It was very hard to figure out the versions that were used in the passing versions of the tests (though admittedly I did not have CI set up for this project). I still don't like the idea of relying on RTD and Travis to store this information.

> In my experience, pinning versions is a much higher maintenance burden than addressing emergent concerns when they arise.

I can see why this would be the case, which is why I only advocate this if there's an out-of-the-box solution that provides a bot that will do the upgrades for us, but I think there's a bot that will do that. At that point, we'll probably end up merging a lot more PRs, but we can just merge all the "upgrade dependencies" PRs where the CI passes.

Ideally we'd be doing the same thing with our vendored dependencies, actually, so that we keep the vendored dependencies relatively up-to-date.

benoit-pierre:
[...]

PR #1761, opened 2019-05-14
Migrate constants from setup.py to setup.cfg

venthur:
## Summary of changes

Moved most of the constants from setup.py to setup.cfg

Closes: #1697

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,5 +18,46 @@ formats = zip
 universal = 1
 
 [metadata]
-license_file = LICENSE
+name = setuptools
 version = 41.0.1
+description = Easily download, build, install, upgrade, and uninstall Python packages
+author = Python Packaging Authority
+author_email = distutils-sig@python.org
+long_description = file: README.rst
+long_description_content_type = text/x-rst; charset=UTF-8
+license_file = LICENSE
+keywords = CPAN PyPI distutils eggs package management
+url = https://github.com/pypa/setuptools
+project_urls =
+    Documentation = https://setuptools.readthedocs.io/
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Software Development :: Libraries :: Python Modules
+    Topic :: System :: Archiving :: Packaging
+    Topic :: System :: Systems Administration
+    Topic :: Utilities
+
+[options]
+zip_safe = True
+python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*
+py_modules = easy_install
+packages = find:
+
+[options.packages.find]
+exclude = *.tests
+
+[options.extras_require]
+ssl =
+    wincertstore==0.2; sys_platform=='win32'
+certs =
+    certifi==2016.9.26
--- a/changelog.d/1697.change.rst
+++ b/changelog.d/1697.change.rst
@@ -1,0 +1,1 @@
+Moved most of the constants from setup.py to setup.cfg
--- a/setup.py
+++ b/setup.py
@@ -3,10 +3,8 @@
 Distutils setup file, used to install or test 'setuptools'
 """
 
-import io
 import os
 import sys
-import textwrap
 
 import setuptools
 
@@ -49,10 +47,6 @@ def _gen_console_scripts():
     yield tmpl.format(shortver=sys.version[:3])
 
 
-readme_path = os.path.join(here, 'README.rst')
-with io.open(readme_path, encoding='utf-8') as readme_file:
-    long_description = readme_file.read()
-
 package_data = dict(
     setuptools=['script (dev).tmpl', 'script.tmpl', 'site-patch.py'],
 )
@@ -88,25 +82,8 @@ def pypi_link(pkg_filename):
 
 
 setup_params = dict(
-    name="setuptools",
-    description=(
-        "Easily download, build, install, upgrade, and uninstall "
-        "Python packages"
-    ),
-    author="Python Packaging Authority",
-    author_email="distutils-sig@python.org",
-    long_description=long_description,
-    long_description_content_type='text/x-rst; charset=UTF-8',
-    keywords="CPAN PyPI distutils eggs package management",
-    url="https://github.com/pypa/setuptools",
-    project_urls={
-        "Documentation": "https://setuptools.readthedocs.io/",
-    },
     src_root=None,
-    packages=setuptools.find_packages(exclude=['*.tests']),
     package_data=package_data,
-    py_modules=['easy_install'],
-    zip_safe=True,
     entry_points={
         "distutils.commands": [
             "%(cmd)s = setuptools.command.%(cmd)s:%(cmd)s" % locals()
@@ -152,28 +129,6 @@ setup_params = dict(
         "setuptools.installation":
             ['eggsecutable = setuptools.command.easy_install:bootstrap'],
     },
-    classifiers=textwrap.dedent("""
-        Development Status :: 5 - Production/Stable
-        Intended Audience :: Developers
-        License :: OSI Approved :: MIT License
-        Operating System :: OS Independent
-        Programming Language :: Python :: 2
-        Programming Language :: Python :: 2.7
-        Programming Language :: Python :: 3
-        Programming Language :: Python :: 3.4
-        Programming Language :: Python :: 3.5
-        Programming Language :: Python :: 3.6
-        Programming Language :: Python :: 3.7
-        Topic :: Software Development :: Libraries :: Python Modules
-        Topic :: System :: Archiving :: Packaging
-        Topic :: System :: Systems Administration
-        Topic :: Utilities
-        """).strip().splitlines(),
-    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
-    extras_require={
-        "ssl:sys_platform=='win32'": "wincertstore==0.2",
-        "certs": "certifi==2016.9.26",
-    },
     dependency_links=[
         pypi_link(
             'certifi-2016.9.26.tar.gz#md5=baa81e951a29958563689d868ef1064d',
@@ -182,7 +137,6 @@ setup_params = dict(
             'wincertstore-0.2.zip#md5=ae728f2f007185648d0c7a8679b361e2',
         ),
     ],
-    scripts=[],
     setup_requires=[
     ] + wheel,
 )


venthur:
I have no idea how to migrate the `extras_require`, as I can't express the:

```
"ssl:sys_platform=='win32'": "wincertstore==0.2"
```

line properly in the ini format. This does not work:

```ini
[options.extras.require]
ssl:sys_platform=='win32' = wincertstore==0.2
```

Also the two tests that are failing, are currently also failing in the master branch.

benoit-pierre:
Use:
```ini
[options.extras_require]
ssl = wincertstore==0.2; sys_platform=='win32'
certs = certifi==2016.9.26
```

benoit-pierre:
I'd be in favor of simplifying things with respect to `setup_requires`: unconditionally require `wheel`.

benoit-pierre:
I'm also not sure those extras are even useful...

venthur:
Done. Do you have any hints regarding the two failing tests that fail also in master?

benoit-pierre:
OK, so adding `python_requires` to the allowed unknown distribution options fix the issue:
```diff
 setuptools/tests/test_integration.py | 1 +
 1 file changed, 1 insertion(+)

diff --git i/setuptools/tests/test_integration.py w/setuptools/tests/test_integration.py
index e54f3209..90147b58 100644
--- i/setuptools/tests/test_integration.py
+++ w/setuptools/tests/test_integration.py
@@ -141,6 +141,7 @@ def test_build_deps_on_distutils(request, tmpdir_factory, build_dep):
         allowed_unknowns = [
             'test_suite',
             'tests_require',
+            'python_requires',
             'install_requires',
         ]
         assert not match or match.group(1).strip('"\'') in allowed_unknowns
```
As for why it's falling now, this seem related to the recent update to `mock==3.0.5`, this fix the issue too:
```diff
 tox.ini | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git i/tox.ini w/tox.ini
index bb940ac0..61e6a718 100644
--- i/tox.ini
+++ w/tox.ini
@@ -19,7 +19,9 @@ setenv=COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 # TODO: The passed environment variables came from copying other tox.ini files
 # These should probably be individually annotated to explain what needs them.
 passenv=APPDATA HOMEDRIVE HOMEPATH windir APPVEYOR APPVEYOR_* CI CODECOV_* TRAVIS TRAVIS_* NETWORK_REQUIRED
-commands=pytest --cov-config={toxinidir}/tox.ini --cov-report= {posargs}
+commands=
+	pip install mock==2.0.0
+	pytest --cov-config={toxinidir}/tox.ini --cov-report= {posargs}
 usedevelop=True
 
 
```
Now to understand why, investigating...

pganssle:
Another good reason for implementing #1566.

benoit-pierre:
OK, so the test has been broken for some time, but was skipped because `pbr` was installed (as a dependency for `mock==2.0.0`).

setup.cfg
```diff
@@ -18,5 +18,45 @@ formats = zip
...
+description = Easily download, build, install, upgrade, and uninstall Python packages
+author = Python Packaging Authority
+author_email = distutils-sig@python.org
+long_description = file: README.rst
+long_description_content_type = text/x-rst; charset=UTF-8
+license_file = LICENSE
+keywords = CPAN, PyPI, distutils, eggs, package, management
+url = https://github.com/pypa/setuptools
+project_urls =
+    Documentation = https://setuptools.readthedocs.io/"
```
jwodder: The trailing `"` needs to be deleted.

pganssle:
```suggestion
    Documentation = https://setuptools.readthedocs.io/
```

venthur:
Hmpf, don't know what to do about the `codecov` "test" failing.

pganssle:
@venthur I don't see that, maybe you caught the webhook after Travis submitted coverage but before Appveyor?

pganssle:
@venthur I've cleaned up the commit history a bit, so if you need to make any more changes, make sure to force pull the current branch to avoid conflicts.

This looks fine to me from a basic review, but I am deeply uncertain of my ability to understand the intricacies of the `.ini` file syntax, so when I have a moment  I'd like to generate the metadata before and after the change and make sure nothing has gone awry.

@benoit-pierre (or any other maintainer) If you trust yourself on this score feel free to merge.

benoit-pierre:
Diffing the generated metadata (`setuptools.egg-info`) is a good way to check for possible issues:
```diff
diff -Naur setuptools.egg-info.master/PKG-INFO setuptools.egg-info.pr/PKG-INFO
--- setuptools.egg-info.master/PKG-INFO	2019-05-15 15:18:55.331909099 +0200
+++ setuptools.egg-info.pr/PKG-INFO	2019-05-15 15:21:22.945901914 +0200
@@ -50,7 +50,7 @@
         chat rooms, and mailing lists is expected to follow the
         `PyPA Code of Conduct <https://www.pypa.io/en/latest/code-of-conduct/>`_.
         
-Keywords: CPAN PyPI distutils eggs package management
+Keywords: CPAN,PyPI,distutils,eggs,package,management
 Platform: UNKNOWN
 Classifier: Development Status :: 5 - Production/Stable
 Classifier: Intended Audience :: Developers
diff -Naur setuptools.egg-info.master/requires.txt setuptools.egg-info.pr/requires.txt
--- setuptools.egg-info.master/requires.txt	2019-05-15 15:18:55.333909099 +0200
+++ setuptools.egg-info.pr/requires.txt	2019-05-15 15:21:22.947901914 +0200
@@ -2,5 +2,6 @@
 [certs]
 certifi==2016.9.26
 
-[ssl:sys_platform=='win32']
+[ssl]
 wincertstore==0.2
+sys_platform=='win32'
diff -Naur setuptools.egg-info.master/SOURCES.txt setuptools.egg-info.pr/SOURCES.txt
--- setuptools.egg-info.master/SOURCES.txt	2019-05-15 15:18:55.523909090 +0200
+++ setuptools.egg-info.pr/SOURCES.txt	2019-05-15 15:21:23.137901905 +0200
@@ -23,6 +23,7 @@
 tox.ini
 .github/pull_request_template.md
 changelog.d/.gitignore
+changelog.d/1697.change.rst
 changelog.d/1749.change.rst
 changelog.d/1750.change.rst
 docs/Makefile
```

So the extras declaration is still not right, change it to:
```diff
--- i/setup.cfg
+++ w/setup.cfg
@@ -58,5 +58,7 @@ packages = find:
 exclude = *.tests
 
 [options.extras_require]
-ssl = wincertstore==0.2; sys_platform=='win32'
-certs = certifi==2016.9.26
+ssl =
+	wincertstore==0.2; sys_platform=='win32'
+certs =
+	certifi==2016.9.26
```
And for the keywords:
```diff
--- i/setup.cfg
+++ w/setup.cfg
@@ -26,7 +26,7 @@ author_email = distutils-sig@python.org
 long_description = file: README.rst
 long_description_content_type = text/x-rst; charset=UTF-8
 license_file = LICENSE
-keywords = CPAN, PyPI, distutils, eggs, package, management
+keywords = CPAN PyPI distutils eggs package management
 url = https://github.com/pypa/setuptools
 project_urls =
     Documentation = https://setuptools.readthedocs.io/
```
And let's just drop the `wheel` setup requirement: it's already handled by `pyproject.toml`.

venthur:
According to the [docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata) `keywords` should be a comma separated list. Are you sure you want me to remove the commas?

And was is the difference between your `[options.extras_require]` and mine? Is the line break significant?

pganssle:
> And let's just drop the `wheel` setup requirement: it's already handled by `pyproject.toml`.

We use an in-tree backend, and have not added the [backend path](https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends) key yet, so `pyproject.toml` installs still fail. I'm not sure if any front-ends have actually *implemented* this part of the spec yet, but regardless, if we start shipping a `pyproject.toml`, sdists of `setuptools` will start failing with older versions of `pip`, so I think we need to keep `setup_requires` populated for a bit longer.

pganssle:
I see the project coverage difference thing too now, by the way, I think the reason might be that we're removing covered code and moving it into a file for which "coverage" as a concept doesn't make a lot of sense, so that change is to be expected.

benoit-pierre:
> And was is the difference between your `[options.extras_require]` and mine? Is the line break significant?

Yes, as you can see from the generated metadata (otherwise, the declaration is into requirements on `;`).

benoit-pierre:
> According to the [docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata) `keywords` should be a comma separated list. Are you sure you want me to remove the commas?

According to [PEP 314](https://www.python.org/dev/peps/pep-0314/#keywords-optional), we want the final result to be: `Keywords: CPAN PyPI distutils eggs package management`. So the code/doc is/are wrong.

benoit-pierre:
> > And let's just drop the `wheel` setup requirement: it's already handled by `pyproject.toml`.
> 
> We use an in-tree backend, and have not added the [backend path](https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends) key yet, so `pyproject.toml` installs still fail. I'm not sure if any front-ends have actually _implemented_ this part of the spec yet, but regardless, if we start shipping a `pyproject.toml`, sdists of `setuptools` will start failing with older versions of `pip`, so I think we need to keep `setup_requires` populated for a bit longer.

Still, if you want to build a wheel, then it's your responsibility to install the wheel package first, so the `bdist_wheel` command is available. I highly dislike "dynamic" requirements in `setup.py`, if only because you can't represent those when moving to `setup.cfg`.

benoit-pierre:
So regarding keywords, I think the code is wrong:
```diff
--- i/setuptools/dist.py
+++ w/setuptools/dist.py
@@ -165,7 +165,7 @@ def write_field(key, value):
     long_desc = rfc822_escape(self.get_long_description())
     write_field('Description', long_desc)
 
-    keywords = ','.join(self.get_keywords())
+    keywords = ' '.join(self.get_keywords())
     if keywords:
         write_field('Keywords', keywords)
 
```
And we can fix it without impacting existing users of `keywords = keyword1 keyword2` in `setup.cfg` (I'm one of those).

benoit-pierre:
That would be in another PR anyway, let's keep `keywords = keyword1 keyword2` for this one.

venthur:
I removed the `,` from keywords and fixed the `extras_require` -- not sure what to do with the `wheels` though.

pganssle:
@venthur It seems you merged the current branch into your own instead of working from the one that was on the PR, which messed up the history again. I've rewritten the history and force-pushed to include your latest commit, but can you do this:

```
git fetch origin
git checkout fix/1697
git reset --hard origin/fix/1697
```

This will set your branch's head to be the current head of the branch.

pganssle:
> Still, if you want to build a wheel, then it's your responsibility to install the wheel package first, so the `bdist_wheel` command is available. I highly dislike "dynamic" requirements in `setup.py`, if only because you can't represent those when moving to `setup.cfg`.

I'd rather just leave it as-is for now. I think we should be vendoring `wheel` or something of that nature *anyway*, so it will be a moot point soon, and it's better for us to leave the current dependency declarations as they are rather than create some pointless churn for anyone who might be relying on the current behavior.

benoit-pierre:
That's a small user base, as pip won't try to build a wheel if the `wheel` package is not already installed, so it's only people who don't have wheel already installed, and build wheel directly from source without using pip. 

Anyway, let's just move the declaration to `setup.py`, so we avoid installing it when not needed...

venthur:
So, `setup_requires = [wheel]` goes back to setup.py?

benoit-pierre:
With the whole conditional inclusion logic, yes.

pganssle:
@venthur Please stop whatever merge/push workflow you are using that adds a bunch of unnecessary commits into the PR.

If you need to modify this PR further, see [this comment](https://github.com/pypa/setuptools/pull/1761#issuecomment-492668084) on how to get your commit history in line with the cleaned-up version.

I have cherry-picked your latest fixup commit and squashed it into the main commit.

venthur:
@pganssle Not sure what you mean with "Please stop whatever merge/push workflow you are using that adds a bunch of unnecessary commits into the PR.". I forked your repo, created a new branch for the fix and incrementally added the request suggested by you and your team.

If you want a clean git history, why don't you just squash the commits right before the merge? This creates less burden for **contributors**, that are unfamiliar with *your* preferred git workflow.

setup.cfg
```diff
@@ -18,5 +18,46 @@ formats = zip
...
-license_file = LICENSE
+name = setuptools
 version = 41.0.1
+description = Easily download, build, install, upgrade, and uninstall Python packages
+author = Python Packaging Authority
+author_email = distutils-sig@python.org
+long_description = file: README.rst
+long_description_content_type = text/x-rst; charset=UTF-8
+license_file = LICENSE
+keywords = CPAN PyPI distutils eggs package management
```
webknjaz: @pganssle I'm pretty sure this is parsed as a string now while it should be a list
```suggestion
keywords =
    CPAN
    PyPI
    distutils
    eggs
    package management
```

setup.cfg
```diff
@@ -18,5 +18,46 @@ formats = zip
...
+url = https://github.com/pypa/setuptools
+project_urls =
+    Documentation = https://setuptools.readthedocs.io/
```
webknjaz: May I suggests adding more sidebar links to the PyPI page? (CIs, docs etc.)
Examples:
- https://github.com/sanitizers/octomachinery/blob/f056cda/setup.cfg#L11-L15
- https://github.com/cherrypy/cheroot/blob/247e6a2/setup.cfg#L10-L16

setup.cfg
webknjaz: @benoit-pierre ^

benoit-pierre: See discussion above: https://github.com/pypa/setuptools/pull/1761#issuecomment-492663319  and https://github.com/pypa/setuptools/pull/1761#issuecomment-492666140.

webknjaz: Oh... I'm relying on the list interface in my other projects tho

benoit-pierre: If you're putting them on multiple lines, you're okay then. But `keyword1, keyword2` won't generate the right metadata.

webknjaz: Ah... I multiline them, yes.

benoit-pierre: Wait, no, if you're putting the keyword on multiple lines, it should not work (generating a comma separated list in the final PKG-INFO). Does PyPI handle your projects metadata correctly?

webknjaz: Yes, it does. Which probably means that the de-facto implementations may differ from the standard...

benoit-pierre: Yeah, it's not correct, I just tested with https://test.pypi.org/project/plover-plugins-manager/:
```shell
$ wget --quiet -O - https://test.pypi.org/pypi/plover-plugins-manager/json | json_pp | grep keyword
      "keywords" : "plover,plover_plugin",
```

webknjaz: Look:
https://github.com/sanitizers/octomachinery/blob/f056cda/setup.cfg#L65-L73
https://pypi.org/project/octomachinery/

webknjaz:
```console
➜ wget --quiet -O - https://test.pypi.org/pypi/octomachinery/json | json_pp | grep keyword
      "keywords" : "Bot,Framework,Framework for writing GitHub Apps,GitHub,GitHub Actions,GitHub API,GitHub Apps,GitHub Checks API",
```

benoit-pierre: So? How do you know the keywords are correctly handled? In both case the display on the project page will be the same.

webknjaz: Well, I mean that on the web page it's a list as expected.

benoit-pierre:
```python
from distlib.index import PackageIndex
pki = PackageIndex()
pki.search({'keyword': 'Framework for writing GitHub Apps'})
pki.search({'keyword': 'Framework for writing GitHub Apps'}).keys()
sorted([m['name'] for m in  pki.search({'keyword': 'Framework for writing GitHub Apps'})])
'octomachinery' in {m['name'] for m in  pki.search({'keyword': 'Framework for writing GitHub Apps'})}
=> False
```

benoit-pierre: Update code above (was using `keyword` instead of `keywords`. Anyway, the JSON metadata is different, and the PKG-INFO is still not PEP 314 compliant, but looks like PyPI work around it... *sigh*

webknjaz: *sigh* :)

venthur: According to the [spec](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata), `keywords` should be a comma separated list in `setup.cfg`. My original PR was formatted like this but i removed the commas by request of the team.

How should we move on with this issue now?

venthur: I just checked the code. Setuptools actually [does support](https://github.com/pypa/setuptools/blob/master/setuptools/tests/test_config.py#L85) the correct `keywords` as comma separated list and dangling list. It is just that it uses it wrong it's own `setup.cfg`.

webknjaz: Yes, it says `list-comma` which is explained above that table:
```
Type names used below:
...
* list-comma - dangling list or string of comma-separated values
```

webknjaz: Yet, for readability, I'd prefer a multiline list instead of an inline comma-separated one.

pganssle: Either way this is a complete side-issue. @benoit-pierre Already figured out that it's just a bug in the code *and* how to fix it, so we should fix that bug and then remove the commas.

PR #1762, opened 2019-05-14
tests: fix_test_build_deps_on_distutils

benoit-pierre:
Ignore distutils' warning: `Unknown distribution option: 'python_requires'`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@ cache:
 test_script:
   - python --version
   - python -m pip install --disable-pip-version-check --upgrade pip setuptools wheel
-  - pip install --upgrade tox tox-venv
+  - pip install --upgrade tox tox-venv virtualenv
   - pip freeze --all
   - python bootstrap.py
   - tox -- --cov
--- a/setuptools/tests/test_integration.py
+++ b/setuptools/tests/test_integration.py
@@ -141,6 +141,7 @@ def test_build_deps_on_distutils(request, tmpdir_factory, build_dep):
         allowed_unknowns = [
             'test_suite',
             'tests_require',
+            'python_requires',
             'install_requires',
         ]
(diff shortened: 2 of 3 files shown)


venthur:
Thanks for the quick fix! I'll refresh #1761 asap.

benoit-pierre:
*sigh* still broken on Windows...

benoit-pierre:
@venthur: once it's merged, you'll just have to close and re-open the PR, no need to rebase.

venthur:
I usually just refresh the branch with the latest commits from master.

benoit-pierre:
You don't have to, closing and re-opening will re-trigger a test run with the result of merging to master.

benoit-pierre:
And now there's **another** issue on Windows... ;( And I notice there's an `UNKNOWN==0.0.0` package in the tox environment...

benoit-pierre:
So why did `more-itertools==7.0.0` get installed for the Window Python 2.7 job when it's `python_requires` is `>=3.4`?

venthur:
Maybe it is a good idea to pin all the versions in `tests/requirements.txt` and update them once in a while as a conscious decision?

benoit-pierre:
Yes, that's what @pganssle [suggested](https://github.com/pypa/setuptools/pull/1761#issuecomment-492234274), that still does not explain what's happening.

venthur:
Well, you never really know what get's installed as a dependencies dependency when you don't pin the versions.
In order to investigate what's going on, you could temporarily install and run [pipdeptree](https://pypi.org/project/pipdeptree/) in order to see pulled in what.

benoit-pierre:
The problem was the tox virtual environment was using an old version of pip (not sure which one, that's a really important information, but not part of tox's logs). Unfortunately, there's also no way to pin it... Updating virtualenv seems to have fixed it.

PR #1763, opened 2019-05-15
Pinned all requirements in requirements.txt

venthur:
Closes: #1566

## Summary of changes
Pinned *all* requirements in the `requirements.txt` files, as per [offical recommendation](https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files).

This will prevent us from random tests failing whenever one of the unpinned dependency makes an incompatible update. On the downside, one has to update the pinned dependencies every once in a while as a separate commit.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1566.change
+++ b/changelog.d/1566.change
@@ -1,0 +1,1 @@
+Pinned all requirements in tests/doc-requirements.txt
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,4 @@
-sphinx!=1.8.0
-rst.linker>=1.9
-jaraco.packaging>=6.1
-
-setuptools>=34
+sphinx==2.0.1
+rst.linker==1.10
+jaraco.packaging==6.1
+setuptools==40.8.0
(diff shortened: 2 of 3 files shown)


venthur:
Are you suggesting to use `pipenv` or something similar? I have the impression (although they try really hard to become the standard) thery are still very alpha.

pganssle:
@venthur I am absolutely not suggesting `pipenv`. This is both too much (full pins in the "ground truth" document) and too little (does not pin transitive dependencies). @webknjaz 's suggestion in #1566 is more or less what I'm looking for, plus an update mechanism.

webknjaz:
Probably needs `pip-compile` with some wrapper (https://github.com/jazzband/pip-tools)

jaraco:
Sorry I didn't get to this issue sooner.

I'm pretty much opposed to any change that's going to add maintenance burden to the project or reduce fidelity of the requirements as indicated, both of which this change does.

Moreover, this request is now stale due to requirements having been moved into `setup.cfg` as extras.

Let's settle on a design and plan in the issue and revisit a change later.

Thanks again for the contrib, though!

Issue #1557, opened 2018-10-28
Remove reference to "eggsecutable scripts"


varunkamath18:
eggs are deprecated and we should not recommend creating eggsecutable scripts.

https://setuptools.readthedocs.io/en/latest/setuptools.html#eggsecutable-scripts

jaraco:
Probably makes sense to deprecate the functionality (code) and remove the documentation.

PR #1764, opened 2019-05-15
Deprecated Eggsecutable Scripts

venthur:
Closes: #1557

## Summary of changes

Deprecated "Eggsecutable Scripts" feature via warning and removed the Section in the docs.

### Pull Request Checklist
- [x] Changes have tests


venthur:
I'm really struggling writing a proper test for this warning. Adding another test based on the stuff in `setuptools/tests/test_bdist_egg.py` was not successful.

docs/setuptools.txt
```diff
@@ -569,43 +569,6 @@ on "entry points" in general, see the section below on `Dynamic Discovery of
 Services and Plugins`_.
 
 
-"Eggsecutable" Scripts
```
pganssle: As long as it's deprecated and not *removed*, I think we should put a deprecation notice on the documentation rather than removing the documentation.

venthur: Done.

venthur:
Finally got around, writing a test for the new warning.

setuptools/tests/test_bdist_egg.py
```diff
@@ -64,3 +65,18 @@ def test_exclude_source_files(self, setup_context, user_override):
...
+        dist = Distribution(dict(
+            script_name='setup.py',
+            script_args=['bdist_egg'],
+            name='foo',
+            py_modules=['hi'],
+            entry_points={
+                'setuptools.installation':
+                    ['eggsecutable = my_package.some_module:main_func']},
+        ))
+        with contexts.quiet():
```
pganssle: Is this necessary? It seems like it circumvents `pytest`'s stdin/stdout capture. I'm wondering if it was a holdover from the `unittest` days, maybe?

docs/setuptools.txt
```diff
@@ -572,6 +572,8 @@ Services and Plugins`_.
 "Eggsecutable" Scripts
 ----------------------
 
+.. warning:: Eggsecutable Scripts are deprecated.
```
```suggestion
.. deprecated:: 41.3.0
```

Not actually sure if the next release is going to be a minor release or a major release, so I'm just guessing minor release.

setuptools/tests/test_bdist_egg.py
venthur: Well, removing that line still makes the test pass, so I can remove it. However, *all* other tests in this module follow the same pattern and have that line.

jaraco:
I don't know the best replacement. Doing a quick search, I see [zipapp](https://docs.python.org/3/library/zipapp.html) seems to implement something very similar. I haven't personally spent much time creating these apps nor have I seen them in the wild, so I can't speak to the use cases.

venthur:
Any updates on this one?

jaraco:
Thanks for the ping. I can't imagine belaboring this issue any longer. I agree it would be nice to provide a replacement, but one of the goals of the PyPA has been to simplify the tool chains and limit the supported use-cases, and I think this is one of those cases, so there is no recommended replacement, at least from PyPA.

pganssle:
> Thanks for the ping. I can't imagine belaboring this issue any longer.

Agreed. Assuming we give a sufficiently long deprecation period, there's always the opportunity to update the warning message if this niche in the ecosystem really needs to be filled.

Issue #1569, opened 2018-10-30
Deprecate obsoletes and provides keywords


smenon8:
As per [comment](https://github.com/pypa/setuptools/issues/1374#issuecomment-405096721), we also should consider deprecating `obsoletes` and `provides` keywords. 

Please refer to issue #1374 for more details.

pganssle:
To do on this ticket:

1. Understand the difference between `Obsoletes` and `Obsoletes-Dist`
2. Understand how that affects `setuptools`
3. Determine what the correct upgrade path is for both `obsoletes` and `provides`
4. Start raising warnings when either is specified
5. Add documentation to indicate that they are deprecated and how to migrate.

I believe `obsoletes` and `provides` are not currently used by `pip` or, to my knowledge, anything else, so we may want to note that in the documentation.

The  relevant specs for the `-Dist` versions are here (two anchor links to the same page):

- [Provides-Dist](https://packaging.python.org/specifications/core-metadata/#provides-dist-multiple-use)
- [Obsoletes-Dist](https://packaging.python.org/specifications/core-metadata/#obsoletes-dist-multiple-use)

The relevant spec for `Obsoletes` and `Provides` is [found in PEP 314](https://www.python.org/dev/peps/pep-0314/#provides-multiple-use).

PR #1765, opened 2019-05-16
Moved distutils' doc for setup-keywords into our doc to document all supported keywords

### LINKED ISSUE #1700: Document *all* supported keywords, including those from distutils
When setuptools was first developed and its documentation was first written, it was assumed that you were already using `distutils` and that you would have read and understood the `distutils` documentation, which is why [`setuptools` only documents *additional* keywords](https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords), not all the supported options.

However, `distutils` is deprecated now, and and the [`distutils` documentation](https://docs.python.org/3/library/distutils.html) is all in legacy mode, plus it's super annoying when you don't know which one contains the keywords you care about.

To satisfy this ticket, we should reorganize the documentation in the following ways:

1. Move the `keywords` documentation to their own separate documentation page.
2. Start moving all the documentation from the `distutils`. This will be slow going, I imagine, because there does not seem to be a "keyword reference" on `distutils`. I think you'll have to extract a lot of the values from [this prose description](https://docs.python.org/3/distutils/index.html#distutils-index), or possibly read the `distutils` source code itself.

venthur:
## Summary of changes

1. Moved keywords into own document and linked it from index.txt (as suggested in #1700)
2. Fixed the now broken section-cross-references (only works in same document) by making them explicit
3. Migrated docs from distutils: https://docs.python.org/3/distutils/setupscript.html

Closes #1700

### Pull Request Checklist


benoit-pierre:
I think some of those keywords should be marked as deprecated:
* `data_files`: does not work with wheels, so should be avoided
* `dependency_links`: not supported anymore by pip

benoit-pierre:
And for `setup_requires`, we should point people towards PEP 518 instead.

docs/setuptools.txt
```diff
@@ -239,179 +239,6 @@ pre- or post-release tags.  See the following sections for more details:
 * The `egg_info`_ command
 
 
-New and Changed ``setup()`` Keywords
```
pganssle: Can we leave this header in place with link to the new keywords document? People are likely linking directly to some headers on this page, so I like to leave a stub in place pointing to the new location, at least for some time.

venthur:
Should `setup_requires` also marked as deprecated then, or just a link to PEP-518?

pganssle:
@venthur It's not quite deprecated, more discouraged. As long as it's possible to invoke the `setup.py` file without your `setup_requires` dependencies installed, they're actually fine to use, since they're added on to the requirements specified in `pyproject.toml` as part of the PEP 517 build, but it's better to specify the requirements in `pyproject.toml`.

docs/setuptools.txt
venthur: Done.

jwodder:
You're missing the `requires`, `obsoletes`, and `provides` keywords (mentioned [here](https://docs.python.org/2/distutils/setupscript.html#relationships-between-distributions-and-packages) in the Distutils docs) and `use_2to3_exclude_fixers` (mentioned [here](https://setuptools.readthedocs.io/en/latest/python3.html#using-2to3) in the Setuptools docs).  Also missing is `test_runner`, a setuptools keyword that was apparently never documented.

benoit-pierre:
`requires` is superseded by `install_requires`, and as `obsoletes`/`provides` are not supported by pip, I'm not sure we want to mention those.

jwodder:
> `requires` is superseded by `install_requires`, and as `obsoletes`/`provides` are not supported by pip, I'm not sure we want to mention those.

They are still valid `setup()` arguments and should be documented.  At the very least, the documentation should state why you're not supposed to use them so that the few people who currently do use them have an authoritative source to be directed to.

Another keyword missing from this PR: `long_description_content_type` (currently only documented for use in `setup.cfg`, though it can be used in `setup()` as well)

pganssle:
I agree with @jwodder about what should and should not be documented, but given that those things are *already* undocumented, I don't think we need to block @venthur's PR on getting them added. I tend to think that we should merge a PR as long as it's an improvement on the current state of affairs even if it doesn't perfectly fix the problem it's attempting to solve.

That said, I have three concerns here:

1. The way "Supported `setup.py` keywords" sits in the table of contents doesn't seem right to me. It seems like it should be a sub-page of `setuptools.txt` that is just broken out into its own page.
2. I'd like to reduce the separation between `setup.cfg` and `setup.py` keywords, because they are just two ways to specify the same data. Moving the `setup.py` keywords without also moving the `setup.cfg` table (or otherwise specifying the `setup.cfg` equivalent for each keyword) cuts against this effort.
3. If a bunch of this information is copied from `distutils`' documentation, what do we need to do to comply with the license the text is released under? I think the PSF license is similar to a BSD-style license, so we may have some obligation to note somewhere that some of this text is at least a derivative work of the distutils documentation.

It is possible we're already doing the third thing and I just don't know about it, I just thought I'd bring it up in this ticket so we don't accidentally fall out of compliance.

venthur:
I've added the missing keywords (also the ones that are not supported anymore for completeness.

@jwodder How exactly you want me to re-structure the file, heading, etc? I'm open to suggestions but if you have a formulation I can copy-paste -- that would be the easiest :) You can also push to my branch directly if that's the fastest way to proceed.

Regarding the licensing, I wouldn't bother too much. First of all, I didn't just copy paste from distutils, I reformulated the explanation so they match with what you had before. But if you want to be bullet proof, you could add a file somewhere in this repo, explaining that some parts of setuptools are based/copied from distutils which is licensed under ...

docs/keywords.txt
```diff
@@ -0,0 +1,294 @@
...
+
+``obsoletes``
+
+   .. warning::
+      ``obsoletes`` is not supported by ``pip`` and should not be used.
+
+``requires``
+
+   .. warning::
+      ``requires`` is not supported by ``pip`` and should not be used.
```
benoit-pierre: Should be ``provides`` instead, no?

venthur:
So I've re-created the old structure. The keywords are in a separate file but `include`ed in the main document.

docs/keywords.txt
venthur: Ooops, you're right! Thanks!

docs/keywords.txt
```diff
@@ -0,0 +1,294 @@
...
+    Dependencies` for details and examples of the format of this argument.
+
+``python_requires``
+    A string corresponding to a version specifier (as defined in PEP 440) for
+    the Python version, used to specify the Requires-Python defined in PEP 345.
+
+``setup_requires``
+
+    .. warning::
+        Using ``setup_requires`` is discouraged in favour for `PEP-518`_
```
benoit-pierre:
```suggestion
        Using ``setup_requires`` is discouraged in favor of `PEP-518`_
```

docs/keywords.txt
```diff
@@ -0,0 +1,294 @@
...
+
+``obsoletes``
+
+   .. warning::
+      ``obsoletes`` is not supported by ``pip`` and should not be used.
+
+``provides``
+
+   .. warning::
+      ``provides`` is not supported by ``pip`` and should not be used.
```
pganssle: This doesn't explain what these keywords *do*, and I also disagree with the idea that they "should not be used".

`provides` and `obsoletes` are actually valid keywords, they translate to [`Provides-Dist`](https://packaging.python.org/specifications/core-metadata/#provides-dist-multiple-use) and [`Obsoletes-Dist`](https://packaging.python.org/specifications/core-metadata/#obsoletes-dist-multiple-use), respectively.

It is true that `pip` does nothing with them, but this is very distinct from `requires`, which is actually deprecated in favor of `install_requires`. If these fields are included, what they do should be explained, possibly with a note that `pip` will not automatically do anything with this metadata.

benoit-pierre:
We actually use the [deprecated by PEP 314](https://www.python.org/dev/peps/pep-0345/#summary-of-differences-from-pep-314) fields `Provides` / `Obsoletes` when writing metadata, instead of `Provides-Dist` / `Obsoletes-Dist`, the core difference being:
> Each entry contains a string describing a package or **module**...

The `XXX-Dist` version of the fields [only allow a project name](https://www.python.org/dev/peps/pep-0345/#provides-dist-multiple-use).

So we should deprecate the use of `provides` / `obsoletes` and mark them at such in the documentation (in fact we already have a ticket for that: #1569). And if we do support the PEP 314 fields at some points, we should do so with the addition of 2 new setup keywords (`provides_dists` / `obsoletes_dists`?).

pganssle:
@benoit-pierre Is there any way for us to detect whether something is a package or module when constructing the metadata? If so, I think it might be easier for us to detect modules and just deprecate the use of `provides`/`obsoletes` for modules, rather than renaming them.

If not, it seems somewhat excessive to add new keywords and drop the old ones just to notify people not to put module names there (since most people probably aren't doing that anyway).

@venthur I was originally thinking we should put these keywords on their own page, just not with the top-level link to them, but I think it might be better to do a total re-org of the docs at some point, which would mean additional links would have a pretty short lifetime so maybe you are right that having it directly in-line is better?

I'll take a closer look at this soon-ish and see what I think. Sorry for the delay.

venthur:
Any updates on this? This was quite a lot of work and it would be a shame to let it rot.

jaraco:
I've re-synced the PR. Unless there's objection, I recommend to merge this change and iterate from there.

venthur:
this branch gets more and more difficult to maintain. can we please merge?

layday:
This also missing `libraries` which I believe is used by `build_clib` but I can't find where this was ever documented (see https://stackoverflow.com/a/49277244) and I've no idea how `setup` options are derived in the source code.

PR #1767, opened 2019-05-20
Add support for `license_files` option in metadata

kchmck:
## Summary of changes

Support for `license_file` was added to `bdist_wheel` in pypa/wheel#47 / pypa/wheel@0acfb4c19dc2d2f2d067eead3d59a5f3415c5c20 and then to setuptools in #1536. Last year, wheel also added support for a `license_files` option in pypa/wheel#138 / pypa/wheel@59976ab294e1b118f42cab404d95df66ed55f7e4 to enable including more than one license. At the same time, `license_file` was deprecated in that project, and using it leads to a `DeprecationWarning` if warnings are enabled. The [docs](https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file) for wheel also mention that `license_files` is the favored option.

Currently, setuptools only supports `license_file`, which leads to the `DeprecationWarning` code path if wheel is also used when building distributions. 

This PR adds the `license_files` functionality to setuptools, which enables the more versatile form to work across both tools and allows users to take advantage of the functionality without hitting deprecated code.

A few notes about the implementation:

- If `license_file` is specified, the file is also added to the set (same behavior as wheel), which preserves backwards compatibility. I wasn't sure if any sort of deprecation was appropriate here, so I left that part out

- I added `license_files` to `Distribution` and `ConfigMetadataHandler` because it seemed to be the best way to reuse the list option parser, but please advise if it should be done a different way

- Leaving the option empty in wheel disables license auto-including, which AFAICT matches the default behavior of setuptools if either option is empty (and MANIFEST.in doesn't include them)

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -5,7 +5,7 @@ import sys
 import io
 import contextlib
 
-from setuptools.extern import six
+from setuptools.extern import six, ordered_set
 
 from .py36compat import sdist_add_defaults
 
@@ -204,7 +204,7 @@ class sdist(sdist_add_defaults, orig.sdist):
         valid paths to 'self.filelist'.
         """
 
-        files = set()
+        files = ordered_set.OrderedSet()
 
         opts = self.distribution.get_option_dict('metadata')
 
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -226,6 +226,6 @@ class sdist(sdist_add_defaults, orig.sdist):
                 log.warn(
                     "warning: Failed to find the configured license file '%s'",
                     f)
-                continue
+                files.remove(f)
 
-            self.filelist.append(f)
+        self.filelist.extend(files)
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -200,9 +200,11 @@ class sdist(sdist_add_defaults, orig.sdist):
         manifest.close()
 
     def check_license(self):
-        """Checks if license_file' is configured and adds it to
-        'self.filelist' if the value contains a valid path.
+        """Checks if license_file' or 'license_files' is configured and adds any
+        valid paths to 'self.filelist'.
         """
+
+        files = set()
 
         opts = self.distribution.get_option_dict('metadata')
 
@@ -211,11 +213,19 @@ class sdist(sdist_add_defaults, orig.sdist):
 
         if license_file is None:
             log.debug("'license_file' option was not specified")
-            return
+        else:
+            files.add(license_file)
 
-        if not os.path.exists(license_file):
-            log.warn("warning: Failed to find the configured license file '%s'",
-                    license_file)
-            return
+        try:
+            files.update(self.distribution.metadata.license_files)
+        except TypeError:
+            log.warn("warning: 'license_files' option is malformed")
 
-        self.filelist.append(license_file)
+        for f in files:
+            if not os.path.exists(f):
+                log.warn(
+                    "warning: Failed to find the configured license file '%s'",
+                    f)
+                continue
+
+            self.filelist.append(f)
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -409,6 +409,7 @@ class Distribution(_Distribution):
         'long_description_content_type': None,
         'project_urls': dict,
         'provides_extras': ordered_set.OrderedSet,
+        'license_files': list,
     }
 
     _patched_dist = None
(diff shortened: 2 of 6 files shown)


jakirkham:
Would it be possible to get a review?

techdragon:
@jakirkham I cant put proper review comments as I'm not on the setuptools team, but it looks like @kchmck has a pretty good PR here. 👍

jaraco:
An important reference to this effort is [this thread](https://discuss.python.org/t/improving-license-clarity-with-better-package-metadata/2154), where many in the PyPA settle on supporting license files.

setuptools/command/sdist.py
```diff
@@ -200,22 +200,32 @@ def read_manifest(self):
         manifest.close()
 
     def check_license(self):
-        """Checks if license_file' is configured and adds it to
-        'self.filelist' if the value contains a valid path.
+        """Checks if license_file' or 'license_files' is configured and adds any
+        valid paths to 'self.filelist'.
         """
 
+        files = set()
```
jaraco: Use of `set` will lead to non-deterministic builds for projects with more than one license. You'll want to use `orderedset` or a `dict` (ignoring key values) or a `list` (and filter duplicates).

setuptools/command/sdist.py
```diff
@@ -200,22 +200,32 @@ def read_manifest(self):
...
+
+        for f in files:
+            if not os.path.exists(f):
+                log.warn(
+                    "warning: Failed to find the configured license file '%s'",
+                    f)
+                continue
 
-        self.filelist.append(license_file)
+            self.filelist.append(f)
```
jaraco: Use `self.filelist.extend(files)` after filtering out any missing.

setuptools/tests/test_egg_info.py
```diff
@@ -567,6 +567,204 @@ def test_setup_cfg_license_file(
...
 
+    @pytest.mark.parametrize("files, incl_licenses, excl_licenses", [
+        ({
+            'setup.cfg': DALS("""
+                              [metadata]
+                              license_files =
+                                  LICENSE-ABC
+                                  LICENSE-XYZ
+                              """),
+            'LICENSE-ABC': DALS("ABC license"),
```
jaraco: `DALS()` is only needed for multi-line strings.

setuptools/dist.py
```diff
@@ -406,6 +406,7 @@ class Distribution(_Distribution):
         'long_description_content_type': None,
         'project_urls': dict,
         'provides_extras': set,
+        'license_files': list,
```
jaraco: I'm not sure this is necessary; why did `license_file` not require an entry here?

setuptools/command/sdist.py
kchmck: Fixed

setuptools/tests/test_egg_info.py
kchmck: Fixed

setuptools/command/sdist.py
kchmck: Is this for efficiency? I ended up going for mutating the set while iterating in e08ec2b640f6b2bf943fea70e2a7f9881bbe6e91, but I'm not sure that's the best. It seemed like appending filtered files to an intermediate list/set before doing the extend would cancel out any performance benefits. Maybe there's a better way to do it?

setuptools/dist.py
kchmck: Unless I'm missing a better way to do it, it seems like this is required for the list parser to actually extract and store the list into the metadata object. If I don't add this line, then the `license_files` member doesn't exist when `check_license()` is called.

I think since `license_file` is just a simple string value, it can use the `get()` method on the metadata dictionary, but if I inspect that for `license_files`, it just returns the raw string (e.g., `\nLICENSE-ABC\nLICENSE-XYZ\nLICENSE-PQR`)

jakirkham:
Looks like this is ready for another review! 😄

setuptools/command/sdist.py
jaraco: I generally avoid incremental changes to any mutable object, as that leaves the object in lots of intermediate states, so it's more obvious what the end effect is if it happens in a single instruction instead of in a lot of instructions. That said, I'm not staunchly opposed to this approach. Happy to proceed with however this ended up.

setuptools/dist.py
jaraco: Got it. Thanks for the clarification.

jakirkham:
Thanks all for getting this in! This is a really great feature 😄

PR #1776, opened 2019-06-01
Use license classifiers rather than the license field.

Steap:
The license field has a 'free' format, while the classifiers are unique
identifiers, similar to SPDX identifiers. In the documentation, we
should therefore showcase the use of classifiers.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -136,16 +136,18 @@ dependencies, and perhaps some data files and scripts::
         author="Me",
         author_email="me@example.com",
         description="This is an Example Package",
-        license="PSF",
         keywords="hello world example examples",
         url="http://example.com/HelloWorld/",   # project home page, if any
         project_urls={
             "Bug Tracker": "https://bugs.example.com/HelloWorld/",
             "Documentation": "https://docs.example.com/HelloWorld/",
             "Source Code": "https://code.example.com/HelloWorld/",
-        }
+        },
+        classifiers=[
+            'License :: OSI Approved :: Python Software Foundation License'
+        ]
 
-        # could also include long_description, download_url, classifiers, etc.
+        # could also include long_description, download_url, etc.
     )
 
 In the sections that follow, we'll explain what most of these ``setup()``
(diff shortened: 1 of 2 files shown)


docs/setuptools.txt
```diff
@@ -136,16 +136,18 @@ dependencies, and perhaps some data files and scripts::
         author="Me",
         author_email="me@example.com",
         description="This is an Example Package",
-        license="PSF",
```
pganssle: I think maybe we should leave this in? It's actually preferable to use *both* the `license` field *and* the classifiers.

If you look at the [PyPI page for `python-dateutil`](https://pypi.org/project/python-dateutil/), you'll see that the "free form" field populates the "License" field in "Meta" sidebar, whereas the license classifiers are shown lower down on the page, under "Classifiers".

How about just changing this to:

```
    license="PSF",    # Short form description, should also use classifiers
```

jaraco: The [official documentation specifically indicates](https://docs.python.org/3/distutils/setupscript.html#additional-meta-data) that `license` is only for licenses not present in Trove classifiers, so I've been discouraging use of `license` when it's redundant (or contradicting) the classifiers.

Steap: It does not really hurt, but is indeed not necessary.

I'm the author of a tool that tries to guess licenses (among other pieces of information) from the JSON served by PyPI and generated from setup.{cfg,py}, and I think that if we really wanted to give an example of the use of the "license" field, we should mention [SPDX identifiers](https://spdx.org/licenses/). People who use SPDX identifiers/expressions when the classifiers are not precise enough (think of the ":: BSD" classifier, which can refer to BSD-2, BSD-3, BSD-4, BSD-x-with-some-exception, etc.) make it really easy to guess the license.

pganssle: I think this is fine for now. Indeed it seems that the [spec for the license field](https://packaging.python.org/specifications/core-metadata/#license) says that it should only be used when there is no appropriate license in the classifiers, so I'm going to merge this.

PR #1781, opened 2019-06-10
Fix the specifier regexp.

Steap:
PEP345 states that the "Requires-Python" metadata field expects version
numbers in the format specified in Version Specifiers, such as:

	Requires-Python: 2.5
	Requires-Python: >2.1
	Requires-Python: >=2.3.4
	Requires-Python: >=2.5,<2.7

This commit makes sure version specifiers without an explicit operator
are valid.

Closes: packaging/issues#107

### DIFF
--- a/pkg_resources/_vendor/packaging/specifiers.py
+++ b/pkg_resources/_vendor/packaging/specifiers.py
@@ -85,7 +85,7 @@ class _IndividualSpecifier(BaseSpecifier):
             raise InvalidSpecifier("Invalid specifier: '{0}'".format(spec))
 
         self._spec = (
-            match.group("operator").strip(),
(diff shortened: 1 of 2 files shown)


benoit-pierre:
If it's a bug in `packaging`, then it must be fixed in that project, and we'll get the fix by updating our vendored copy. So please PR this to https://github.com/pypa/packaging.

Steap:
Out of curiosity, is the packaging code literally copied into pkg_resources/_vendor? Why is it not used as a library?

jaraco:
@Steap I've updated the developer docs to provide an explanation and links to relevant issues and resources.

Steap:
Thanks!

PR #1788, opened 2019-06-18
stop using deprecated HTMLParser.unescape

methane:
## Summary of changes

HTMLParser.unescape is deprecated since Python 3.4.

### Pull Request Checklist
- **N/A** Changes have tests

### DIFF
--- a/changelog.d/1788.change.rst
+++ b/changelog.d/1788.change.rst
@@ -1,0 +1,1 @@
+Changed compatibility fallback logic for ``html.unescape`` to avoid accessing ``HTMLParser.unescape`` when not necessary. ``HTMLParser.unescape`` is deprecated and will be removed in Python 3.9.
--- a/setuptools/py33compat.py
+++ b/setuptools/py33compat.py
@@ -52,4 +52,8 @@ class Bytecode_compat:
 Bytecode = getattr(dis, 'Bytecode', Bytecode_compat)
 
 
-unescape = getattr(html, 'unescape', html_parser.HTMLParser().unescape)
+unescape = getattr(html, 'unescape', None)
+if unescape is None:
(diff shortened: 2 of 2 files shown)


methane:
Should this type of change require News fragment and test?

Note that this change is executed always when this module is imported...

pganssle:
It took me a second to realize how this was different from the current formulation, but it's because as is, `HtmlParser().unescape` is always executed, even if unused.

I think a changelog entry would be wise, because when people start using Python 3.9, they'll want to know the minimum version of `setuptools` to pin.

pganssle:
Thanks @methane, I've squashed the current commits and marked this for auto-merge, so it should merge as soon as the CI passes.

Glad to have people getting us ahead of the curve on avoiding deprecation warnings.

GegX (commented):
I encounter that issue with python 3.9 while installing 
python3 -m pip install robotframework-debugger
so this would do the trick

methane:
What version of setuptools do you use?
Don't you just using very old setuptools?

joergent:
The file py33compat.py is buggy placing the test for the html module in the bottom of the file probably inside a class where the html module import is hidden and not just after importing html. This is my interpretation as I don't program in python.
However the py33compat.py file is omitted in version 51.0.0 and the wrong test for html.unescape is thus history.

This incorrect test prevented me in using letsencrypt as it was upgrading itself automatically.
I changed the setuptool lines in certbot-auto to this

   ('12/e1/b9a2926a3c5a3fb055b8f85052f5baa890106a0e21b64a977c10affea751/'
    'setuptools-51.0.0.zip',
   '029c49fd713e9230f6a41c0298e6e1f5839f2cde7104c0ad5e053a37777e7688'),
-

PR #1791, opened 2019-06-21
Fix #1790 : Include the file path in get_metadata() UnicodeDecodeErrors

### LINKED ISSUE #1790: Include path in error message when get_metadata() raises e.g. UnicodeDecodeError
This issue is another manifestation of issue #1664 ("Include location of bad METADATA files in errors"), but for the `value.decode('utf-8')` line in `NullProvider.get_metadata()`: https://github.com/pypa/setuptools/blob/8aeff6b2c9a64e47ad2a22533d7e65c08cd4103f/pkg_resources/__init__.py#L1416-L1420

This came up recently in this pip issue (and probably other pip issues -- this is just one example): https://github.com/pypa/pip/issues/6566

Rather than re-opening #1664, it probably makes sense to have a new issue. It's unknown to me how many code paths in `pkg_resources` affect the larger issue. PR #1706 addressed the version-is-missing case.

cjerdonek:
Fixes https://github.com/pypa/setuptools/issues/1790 .

### DIFF
--- a/changelog.d/1790.change.rst
+++ b/changelog.d/1790.change.rst
@@ -1,2 +1,2 @@
-Add to the error message the path to the file if a ``UnicodeDecodeError``
-occurs while reading a metadata file.
+Added the file path to the error message when a ``UnicodeDecodeError`` occurs
+while reading a metadata file.
--- a/changelog.d/1790.change.rst
+++ b/changelog.d/1790.change.rst
@@ -1,0 +1,2 @@
+Add to the error message the path to the file if a ``UnicodeDecodeError``
+occurs while reading a metadata file.
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1416,8 +1416,17 @@ class NullProvider:
     def get_metadata(self, name):
         if not self.egg_info:
             return ""
-        value = self._get(self._fn(self.egg_info, name))
-        return value.decode('utf-8') if six.PY3 else value
+        path = self._get_metadata_path(name)
+        value = self._get(path)
+        if six.PY2:
+            return value
+        try:
+            return value.decode('utf-8')
+        except UnicodeDecodeError as exc:
+            # Include the path in the error message to simplify
+            # troubleshooting, and without changing the exception type.
+            exc.reason += ' in {} file at path: {}'.format(name, path)
+            raise
 
     def get_metadata_lines(self, name):
         return yield_lines(self.get_metadata(name))
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -18,6 +18,7 @@ except ImportError:
     import mock
 
 from pkg_resources import DistInfoDistribution, Distribution, EggInfoDistribution
+from setuptools.extern import six
 from pkg_resources.extern.six.moves import map
 from pkg_resources.extern.six import text_type, string_types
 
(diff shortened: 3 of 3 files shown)


pkg_resources/__init__.py
```diff
@@ -1416,8 +1416,17 @@ def has_metadata(self, name):
...
+        path = self._get_metadata_path(name)
+        value = self._get(path)
+        if six.PY2:
+            return value
+        try:
+            return value.decode('utf-8')
+        except UnicodeDecodeError as exc:
+            # Include the path in the error message to simplify
+            # troubleshooting, and without changing the exception type.
+            exc.reason += ' in {} file at path: {}'.format(name, path)
```
pganssle: Hm... I don't know how I feel about mutating the reason for the exception.

Seems like we could use exception chaining here *but* I doubt that that would give a better user experience. I guess keep it as is unless someone has a better idea.

cjerdonek: Subclassing would be another option (e.g. with an extra attribute and different __str__ method). It’s up to you.

pganssle: Well, the way you've done it has the very nice advantage that no trace of the "we rewrote this exception message" shows up in the Traceback - and that would be noise for any end user anyway.

changelog.d/1790.change.rst
```diff
@@ -0,0 +1,2 @@
+Add to the error message the path to the file if a ``UnicodeDecodeError``
```
pganssle:
```suggestion
Added the file path to the error message file when a ``UnicodeDecodeError``
```

pkg_resources/__init__.py
cjerdonek: Could “from None” be used to address that point?

pganssle: I had only kinda peripherally heard of "from None", so thanks for bringing it up. The PEP is a bit opaque about the problem it's intended to solve, but I think it does the opposite of what I was thinking. `raise SubclassException(...) from None` will make it so that you only see the traceback that led to `SubclassException()`, whereas we mainly care about the traceback that led to the original `UnicodeDecodeError`, I assume.

I haven't actually looked at what this looks like - if you think *that* exception is the one that's noise, then raising reraising a new `UnicodeDecodeError` from `None` is probably the right thing to do.

cjerdonek: > if you think that exception is the one that's noise,

If we raised a new `SpecialUnicodeDecodeError` exception, the only noise IMO would be seeing the original chained `UnicodeDecodeError` because the new one would contain the same information (and more). So the purpose of doing `from None` would be to get rid of that. But you're right that this would also cause both (1) any previous chaining to be lost, and (2) any traceback lines "inside" the `value.decode('utf-8')` to be lost.

For (2), I'm not sure we would actually be losing anything since `decode()` is so low-level. If you look at one of the original pip issue [reports](https://github.com/pypa/pip/issues/6566), you can see that no lines of the traceback originate from inside the call to `decode()` -- it starts at that line:

```
  File "c:\users\shige\test\lib\site-packages\pip\_internal\utils\packaging.py", line 48, in get_metadata
    metadata = dist.get_metadata('METADATA')
  File "c:\users\shige\test\lib\site-packages\pip\_vendor\pkg_resources\__init__.py", line 1420, in get_metadata
    return value.decode('utf-8') if six.PY3 else value
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 185: invalid continuation byte
```

For (1), that could probably be addressed by doing something more like the following, slightly more advanced style (chain the new exception to the original exception's parent, to preserve any previous chain):

```
try:
    return value.decode('utf-8')
except UnicodeDecodeError as exc:
    # Prepare new arguments.
    raise SpecialUnicodeDecodeError(...) from exc.__context__
```

Either way, I'm pretty sure raising a new exception will still show the current stack that led to the function being called.

However, there's a point that might make all of this moot. Including `raise ... from` in Python 2 will cause a `SyntaxError`, and I don't know if setuptools already allows Python 3-only code to exist. If not, it's probably not worth the trouble, in which case adding to the original exception seems like the best / safest bet.

cjerdonek:
FYI, I updated the PR with the suggested change to the changelog entry text.

pkg_resources/__init__.py
pganssle: > However, there's a point that might make all of this moot. Including `raise ... from` in Python 2 will cause a `SyntaxError`, and I don't know if setuptools already allows Python 3-only code to exist.

If we want to do it this way, I believe we can use `six.raise_from` to avoid the syntax error (we don't really have to worry about any differences in behavior, since this code is only ever reached in Python 3 anyway).

I think I'm fine with any of the options - the current commit, the `from None` or the `from exc.__context__` - they all have their own trade-offs, but the final effect is minor.

I'll give this a few days for @benoit-pierre and @jaraco to weigh in if they so choose, then merge.

cjerdonek: One thought in favor of the current form  is that a new subclass exception type can always be added later, but once it’s added the API can’t be undone. So it might be better to wait until after the current commit is deployed to get a better sense of whether a more structured API would be beneficial (and what that would look like).

cjerdonek:
Thanks, @pganssle!

Issue #479, opened 2016-01-06
imp is deprecated


ghost:
Originally reported by: **jaraco (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

According to the docs, [imp is deprecated](https://docs.python.org/3/library/imp.html) since Python 3.4. Yet `pkg_resources` relies on it. In particular, it relies on the `acquire_lock` and `release_lock` functions, which don't have an equivalent in importlib.

The docs do say, "Changed in version 3.3: The locking scheme has changed to per-module locks for the most part. A global import lock is kept for some critical tasks, such as initializing the per-module locks."

What does that mean for the usage in `pkg_resources`? Is it possible to eliminate the usage of locks on Python 3.3+ in declare_namespace and fixup_namespace_packages?

This issue was revealed in [Pull Request 129](https://bitbucket.org/pypa/setuptools/pull-requests/129/try-not-to-use-deprected-imp-module-on-34), but the resolution was to keep it and the deprecation warnings. What's the proper solution here?

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/479

mgedmin:
`pkg_resources` isn't the only place that imports `imp`

```
pkg_resources/__init__.py:    import imp as _imp
setuptools/command/build_ext.py:import imp
setuptools/command/install_lib.py:import imp
setuptools/depends.py:import imp
setuptools/depends.py:from imp import PKG_DIRECTORY, PY_COMPILED, PY_SOURCE, PY_FROZEN
setuptools/site-patch.py:                import imp  # Avoid import loop in Python >= 3.3
```

Python 3.6 upgrades the PendingDeprecationWarning to a full DeprecationWarning, which in practical terms means nothing, only some psychological pressure not to forget about this issue. ;)

5j9:
Is it acceptable to temporarily suppress the warnings in setuptools using `warnings.catch_warnings` or does it need to be a thread-safe or some other solution?

jaraco:
I'm not sure what you're asking. Are you asking if you can suppress the warnings in your code? Well, sure. If you're asking about suppressing the warnings in setuptools, it feels like that might just kick the can down the road. And besides, most applications have these warnings disabled by default anyway, so if a user is enabling them, I would expect them to want to see them.

5j9:
[...]

PR #1798, opened 2019-07-08
Fix some usage of deprecated `imp` module

asottile:
## Summary of changes

Partial resolution of #479 -- the other usage is a little more tricky and will require some more clever fixes

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/command/install_lib.py
+++ b/setuptools/command/install_lib.py
@@ -1,5 +1,5 @@
 import os
-import imp
+import sys
 from itertools import product, starmap
 import distutils.command.install_lib as orig
 
@@ -74,10 +74,10 @@ class install_lib(orig.install_lib):
         yield '__init__.pyc'
         yield '__init__.pyo'
 
-        if not hasattr(imp, 'get_tag'):
+        if not hasattr(sys, 'implementation'):
             return
 
-        base = os.path.join('__pycache__', '__init__.' + imp.get_tag())
+        base = os.path.join('__pycache__', '__init__.' + sys.implementation.cache_tag)
         yield base + '.pyc'
         yield base + '.pyo'
         yield base + '.opt-1.pyc'
--- a/changelog.d/479.change.rst
+++ b/changelog.d/479.change.rst
@@ -1,0 +1,1 @@
+Remove some usage of the deprecated ``imp`` module.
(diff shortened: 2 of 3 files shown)


asottile:
not sure why travis didn't report status -- it is complete and successful :man_shrugging:

setuptools/command/install_lib.py
```diff
@@ -74,10 +74,10 @@ def _gen_exclusion_paths():
         yield '__init__.pyc'
         yield '__init__.pyo'
 
-        if not hasattr(imp, 'get_tag'):
+        if not hasattr(sys, 'implementation'):
             return
 
-        base = os.path.join('__pycache__', '__init__.' + imp.get_tag())
+        base = os.path.join('__pycache__', '__init__.' + sys.implementation.cache_tag)
```
mgedmin: The docs say `sys.implementation.cache_tag` may be `None` (in which case you're supposed to disable module caching).  Maybe check for that too and do an early return?

asottile: this isn't a regression, `imp.get_tag()` has identical semantics and is implemented as just `return sys.implementation.cache_tag`: https://github.com/python/cpython/blob/9b5ce62cac27fec9dea473865d79c2c654312957/Lib/imp.py#L67-L69

pganssle:
@asottile I don't have any objections to this, but when running the test suite against master with `DeprecationWarning` set to raise errors on `master`, I'm not getting any failures, but `coverage.py` is saying all these lines are covered. Any idea what's going on there?

asottile:
> @asottile I don't have any objections to this, but when running the test suite against master with `DeprecationWarning` set to raise errors on `master`, I'm not getting any failures, but `coverage.py` is saying all these lines are covered. Any idea what's going on there?

depending on the version of `pytest` you're using (I've only fixed pytest in 5.x) `imp` will get imported as a side-effect of pytest's initialization (and so imports in user-space code won't trigger that warning) -- this is probably what you're seeing if I had to guess?

pganssle:
Hm.. That seems like a huge problem with `pytest`, that's super annoying.

asottile:
hehe yeah that's why I fixed it 🙃

asottile:
@pganssle @jaraco any updates on this? I think this is ready for merge

PR #1810, opened 2019-07-31
Fix comment in documentation

vanschelven:
Quoting from the same section as the example:

> The data files must be specified via the distutils’ MANIFEST.in file.

It is true that in principle `include_package_data` can also be used to
include files from source control, but to get that behavior you'd need
to configure the appropriate plugin, which is not done in the changed
example.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -970,7 +970,7 @@ to do things like this::
         packages=find_packages('src'),  # include all packages under src
         package_dir={'':'src'},   # tell distutils packages are under src
 
-        include_package_data=True,    # include everything in source control
+        include_package_data=True,  # include package data as per MANIFEST.in
 
         # ...but exclude README.txt from all packages
         exclude_package_data={'': ['README.txt']},


vanschelven:
@jaraco any way I can make help to merge this rather trivial fix?

vanschelven:
Or perhaps @pganssle ?

jaraco:
I'm -1 on this change. In my projects, I use the functionality described here. The intention is for `include_package_data` to include in any package all files that are included by default, which includes the MANIFEST (generated by MANIFEST.in), files in revision control, and other hard-coded common files (like readme, setup.py, and the manifest itself). So while the current description isn't quite correct, changing it to another form that's also incorrect isn't an improvement and in fact sends the signal that source-control files aren't supported, but they very much are.

If I were to accept this change, I could just as easily accept the reverse change, as both forms are equally correct (and incorrect). I'd be happy to accept something that adds information without losing fidelity.

docs/setuptools.txt
```diff
@@ -970,7 +970,7 @@ to do things like this::
         packages=find_packages('src'),  # include all packages under src
         package_dir={'':'src'},   # tell distutils packages are under src
 
-        include_package_data=True,    # include everything in source control
+        include_package_data=True,  # include package data as per MANIFEST.in
```
jaraco: I should say also, `MANIFEST.in` isn't actually the source. `MANIFEST` is. It's just that most people rely on a MANIFEST.in template to generate MANIFEST.

vanschelven:
Meh... I remain unconvinced. I'm well aware of the functionality this comment is referencing, and I'm well aware that it is useful functionality and underlying implementation. I'm simply saying that in the context of the code example this comment is included in, it is quite confusing:

* in this code example no code is included that would make the comment true.
* in the section in which the code example is included, `MANIFEST.in` is mentioned exclusively and explicitly as the mechanism "unless they are listed in the `MANIFEST.in` file."
* the mechanism of including files via source control system plugins is only first mentioned well below the code example.

Anyway, at this point we're well into [someone is wrong on the internet](https://xkcd.com/386/) territory, so I'll leave it at that.

jaraco:
I didn’t realize in my review that this comment was specific to a possibly limited scope example. I’ll take another look.

jaraco:
I looked at this again, and I can see what you're saying. It seems that the code has evolved over time to be inconsistent. In particular:

> in this code example no code is included that would make the comment true.

It used to be the case that Subversion support was integrated with setuptools, so was available by default. Now it's been externalized as a plugin (for separation of concerns), but still no code is required for this to be true (only for plugins to be installed), so the condition still holds.



> in the section in which the code example is included, `MANIFEST.in` is mentioned exclusively and explicitly as the mechanism "unless they are listed in the `MANIFEST.in` file."

Except in the paragraph before, it mentions:

> you may want to include package README files in your revision control
system

And earlier in the same section on L865: 

> This tells setuptools to install any data files it finds in your packages.
The data files must be specified via the distutils' `MANIFEST.in` file.
(They can also be tracked by a revision control system, using an appropriate
plugin.  See the section below on Adding Support for Revision Control
Systems for information on how to write such plugins.)

So it does seem that the documentation is _trying_ to capture that both mechanisms are alternates for one another.

So my earlier assessment holds; changing this comment only confuses the matter more. I would be open to some more invasive edits that more accurately describe the intention, but waffling between two intentions is not a suitable change. Better would be to remove the comment altogether and rely on documentation above to indicate what the effect is.

Do feel free to re-word the documentation to capture the broader behavior.

PR #1811, opened 2019-08-03
Improve Visual C++ 14.X support

JGoutin:
## Summary of changes

* Improve VC++14 support for VS 2017 and 2019.
* Separate VC from VS version (Miss match starting VS15).
* Improve docstrings args and returns information + fixe typos.
* Fix coding style and minor coding issues.
* Remove Microsoft "Windows SDK 7.0" dead link.

Tested on Windows 10 with Python 3.7.4 and "Visual Studio Build Tools 2019" with following setup cases:
* VC++14.2 (VS 2019)
* VC++14.1 (VS 2017)
* VC++14.0 (VS 2015)

This setuptools file cover a wide variety of VS configurations that I cannot test easily. So any code review and other tests are really welcome. 

There is no automated tests for this. This setuptools part did not really had tests before this PR, and it may be very hard to do real relevant automated tests for it.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/1811.change.rst
+++ b/changelog.d/1811.change.rst
@@ -1,0 +1,1 @@
+Improve Visual C++ 14.X support, mainly for Visual Studio 2017 and 2019.
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -11,13 +11,17 @@ Microsoft Visual C++ 9.0:
 Microsoft Visual C++ 10.0:
     Microsoft Windows SDK 7.1 (x86, x64, ia64)
 
-Microsoft Visual C++ 14.0:
+Microsoft Visual C++ 14.X:
     Microsoft Visual C++ Build Tools 2015 (x86, x64, arm)
-    Microsoft Visual Studio 2017 (x86, x64, arm, arm64)
     Microsoft Visual Studio Build Tools 2017 (x86, x64, arm, arm64)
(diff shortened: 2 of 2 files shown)


JGoutin:
Concerning the evolution of this file (`msvc.py`):

The majority of the code was written to support some  VC9 & VC10 old compilers that does not really had a  `vcvarsall.bat` or had a messed up one. It was also done to add support for some early VC14 standalone compilers:
* Since Python 3.4 is deprecated, the VC10 part is deprecated.
* Once Python 2.7 is deprecated in setuptools, the VC9 part can also be deprecated.
* Recent versions on Python does not need the VC14 fix. Only Python 3.5 will require help to detect VC 2017 or 2019 `vcvarsall.bat`.

So, in the end of this year, this file can be easily simplified to remove almost all its content and only keep a Python 3.5/VC compat part that will be kept until Python 3.5 deprecation.

jaraco:

jgsogo:
Hi! There is a change in this PR that breaks Python 2: https://github.com/pypa/setuptools/pull/1811/files#diff-ebddaa1b649ef567c2534a012ba25913R627

```python
with open(state_path, 'rt', encoding='utf-8') as state_file:
```

That `encoding` argument is only available for py3

Is it intentional?

Thanks

PR #1822, opened 2019-08-17
pep425tags: don't append 'm' to ABI tag in Python 3.8

rdb:
This makes changes to the `pep425tags` module following changes made in pip and wheel:
pypa/pip#6874
pypa/wheel#303

It fixes the `get_abi_tag()` function, which appended a `m` flag to the ABI tag on Windows despite this tag no longer existing in Python 3.8.  For more info see:
https://bugs.python.org/issue36707
https://docs.python.org/dev/whatsnew/3.8.html#build-and-c-api-changes

I'm not actually sure whether this function is used anywhere important; my testing was around building and installing wheels and this change didn't seem to be needed, but it is nonetheless good to ensure that the function does the right thing as long as it is there.  As such, I am not sure whether it is critical to get this change into the ensurepip shipped with Python 3.8.

### DIFF
--- a/changelog.d/1822.change.rst
+++ b/changelog.d/1822.change.rst
@@ -1,0 +1,1 @@
+Fixed use of removed 'm' ABI flag in Python 3.8 on Windows
--- a/setuptools/pep425tags.py
+++ b/setuptools/pep425tags.py
@@ -93,7 +93,9 @@ def get_abi_tag():
             d = 'd'
         if get_flag('WITH_PYMALLOC',
                     lambda: impl == 'cp',
-                    warn=(impl == 'cp')):
+                    warn=(impl == 'cp' and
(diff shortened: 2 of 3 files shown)


benoit-pierre:
Thanks, but if we were to backport some fixes in our copy of `pep425tags`, than it should be done by updating the whole files (including tests and `glibc` related changes), since it's better in the long term (we don't want to deviate from pip's versions of those).

But since `packaging` now has a dedicated module for handling PEP 425, I think switching to that is an even better long term solution, see #1829.

jaraco:
Agreed with Benoit. Let's rely on shared libraries where possible.

Thanks for the contrib. This approach was clean and acceptable otherwise and served as a good inspiration to develop a more robust solution.

PR #1828, opened 2019-08-22
Allow (and parse) `[options.cmdclass]` in setup.cfg

lorencarvalho:
## Summary of changes

Adds functionality to `ConfigOptionsHandler` to be able to parse `cmdclass` configurations.

I added a test, and you may notice that in my test I put the cmdclass into it's own file, this is because conventionally folks put their command classes into their setup.py, however if we want to specify one via setup.cfg, we must put it outside setup.py in order to reference it via the `attr:` specifier, or you end up invoking `setup()` twice (once during the attribute lookup, and again when it's actually invoked). I thought that this might be why cmdclass wasn't implemented in the first place, but I'm not sure. Let me know if this feature is problematic or warrants further discussion.

Thanks for looking!

### DIFF
--- a/changelog.d/1828.change.rst
+++ b/changelog.d/1828.change.rst
@@ -1,0 +1,1 @@
+Add support for specifying cmdclass in setup.cfg.
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -654,3 +654,11 @@ class ConfigOptionsHandler(ConfigHandler):
         """
         parsed = self._parse_section_to_dict(section_options, self._parse_list)
         self['data_files'] = [(k, v) for k, v in parsed.items()]
+
+    def parse_section_cmdclass(self, section_options):
+        """Parses `cmdclass` configuration file section.
+
+        :param dict section_options:
+        """
+        parsed = self._parse_section_to_dict(section_options, self._parse_attr)
+        self['cmdclass'] = parsed
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -6,6 +6,7 @@ import pytest
 
 from distutils.errors import DistutilsOptionError, DistutilsFileError
 from mock import patch
+from setuptools import Command
 from setuptools.dist import Distribution, _Distribution
 from setuptools.config import ConfigHandler, read_configuration
 from setuptools.extern.six.moves import configparser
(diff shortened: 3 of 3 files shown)


pganssle:
Hmm.. I am +0 on this feature. For the most common use case I don't see it as terribly useful - if you are writing your *own* custom commands, I don't see why you would want to put your `cmdclass` in `setup.cfg` rather than `setup.py` - usually the point of `setup.cfg` is to avoid custom code in the first place. I am not sure what the import path will be for custom code *in your repository* and what will happen when you use `setuptools.build_meta` with one of these files.

On the other hand, this combined with `build-system.requires` in `pyproject.toml` *could* lead to a convenient plugin system for overriding specific command classes from third-parties, in which case it would still allow some people to avoid writing any custom code. Are there any such third party plugins that people are waiting to use here?

@lorencarvalho What is your use case for this? Did you have something in mind when you added it?

I think this needs further discussion (CC: @jaraco @benoit-pierre), but if it were to go through, I think it would need documentation and an additional test that it works with `setuptools.build_meta`.

lorencarvalho:
hi Paul,

Thanks for such a thoughtful response!

> Are there any such third party plugins that people are waiting to use here?

This is a great question, and I'm not sure the answer myself. However...

> What is your use case for this? Did you have something in mind when you added it?

My use case is that I like to include a little helper command class to set up a venv (e.g. `python3 setup.py venv`) for some projects, and while migrating all the metadata and options out of the `setuptools.setup()` invocation I noticed that cmdclass seemed to be unsupported. Aesthetically, I disliked leaving my `setup` invocation as `setup(cmdclass={...` while the rest of the logic got pushed into setup.cfg. That was really the primary motivator for opening this PR. 

I am interested in a future where setup.py ceases to exist and `build-backend = setuptools:setup` in a pyproject.toml would be enough, but that would require any cmdclasses to be defined outside of a setup.py, they could be put into their own file at the root of the repository, which would work with the `attr:` specifier. But I'll be the first to admit this isn't a fully fleshed out thought.

Thanks again for looking!

jaraco:
I'm also a little reluctant on this one. One one hand, it seems reasonable if not desirable for all setuptools supported build features to be supported using setup.cfg. On the other hand, I'm not sure `cmdclass` is such a feature. In some ways, usage of that feature is discouraged.

I tracked down [one such use-case](https://github.com/linkedin/shiv/blob/7ea3cf14ec709fa5282614d367e79f953bd5c7d5/setup.py#L65-L88). What's the advantage of `python setup.py venv` over `python -m virtualenv venv/shiv; ln -s venv/shiv/bin/activate`? If it's just about brevity, it really feels to me like that functionality should be its own tool, perhaps an alias or a standalone script or routine in developer library, and not integrated into the build system.

May I suggest just replace that setuptools command with this script:

```python
#!/usr/bin/env python3

import os
import contextlib
import venv
import pathlib


def main():
    venv_path = pathlib.Path('venv/shiv')
    print('Creating virtual environment in', venv_path.absolute())
    venv.main(args=[str(venv_path)])
    print(
        'Linking `activate` to top level of project.\n'
        'To activate, simply run `source activate`.'
    )
    with contextlib.suppress(FileExistsError):
        os.symlink(venv_path / 'bin' / 'activate', 'activate')


__name__ == '__main__' and main()
```

And then instead of `python setup.py venv`, run `python mkvenv`?

At this point, I'm -1 on this contribution until there's a more compelling use-case.

jaraco:
Closing until there's a more compelling use-case. Thanks for the contrib, and feel free to re-engage at any point.

PR #1832, opened 2019-08-26
Add RequirementParseError to pkg_resources.__all__.

### LINKED ISSUE #1831: pkg_resources does not include RequirementParseError in `all`
This causes mypy type checks of code importing RequirementParseError as a 3rd party library to fail with:
```
error: Module has no attribute "RequirementParseError"
```

eg:
in some other python project, in a new file:
```
from pkg_resources import RequirementParseError
```

run mypy and watch it choke with
```
error: Module has no attribute "RequirementParseError"
```

TansyArron:
## Summary of changes

Add RequirementParseError to `pkg_resources.__all__`

Closes #1831 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1832.change.rst
+++ b/changelog.d/1832.change.rst
@@ -1,0 +1,1 @@
+Add RequirementParseError to pkg_resources.__all__
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -217,7 +217,7 @@ __all__ = [
 
     # Exceptions
     'ResolutionError', 'VersionConflict', 'DistributionNotFound',
-    'UnknownExtra', 'ExtractionError',
+    'UnknownExtra', 'ExtractionError', 'RequirementParseError',
 
     # Warnings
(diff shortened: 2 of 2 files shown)


jaraco:
What's the motivation for this change? I'm not sure the project is interested in _enhancing_ the exposure of this exception. It does appear to be a simple wrapper around `packaging.requirements.InvalidRequirement` (probably a compatibility behavior).

Tracing the history of RequirementParseError, it was introduced in bd090b8067e77d30325f5b8acc22f9b323c7c222 and in a718819d2849196e902808301c9a95724510c5c1, almost all of the uses of that error were removed. What would happen if we just removed it instead?

jaraco:
I apologize. I think I missed the reference to #1831, which does provide some background on the motivation for the change. Thanks for that. I'll follow-up there.

jaraco:
I'm going to close this one for now, but happy to revisit after exploring more in the issue.

Issue #1725, opened 2019-03-21
Spaces in file names are looking for troubles


alevesely:
What is the purpose of naming a file `script (dev).tmpl`?  Oh, I see, to prove that Setuptools can support funny file names.  Great.

I read bug #134, and especially jaraco's comment on Jan 3, 2016.  There is a third principle, besides [least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and [robustness](https://en.wikipedia.org/wiki/Robustness_principle), and it is the [readability](https://en.wikipedia.org/wiki/Readability).  As an invisible character, a space cannot be distinguished from other invisible characters, such as tab and a plethora of UTF-8 space variations.  How easily can you  tell  `script (dev).tmpl` from  `script   (dev).tmpl`?  (Hint: one of them has multiple spaces, but you need to inspect HTML code to tell).  An accidental misspell can hardly catch your eyeballs.

Certainly, `script-dev.tmpl` would have looked too simple, eh?  So here's a fourth principle, [KISS](https://en.wikipedia.org/wiki/KISS_principle).

The URI syntax ([rfc3986](https://tools.ietf.org/html/rfc3986#section-2.3)) provides for a set of unreserved characters (`ALPHA / DIGIT / "-" / "." / "_" / "~"`), whereby the rest, including spaces, has to be percent-encoded.  What doddering fools!

Email syntax allows for spaces in the local-part of an email address, provided that spaces are quoted.  However, [SMTP](https://tools.ietf.org/html/rfc5321#section-4.1.2) wording is "*a host that expects to receive mail SHOULD avoid [them]*".

Some SQL implementation provide for [spaces in variable names](https://dojo.domo.com/t5/Beast-Mode-ETL-Dataflow/SQL-variables-with-spaces-in-them/td-p/17494).  I'd guess you're looking forward for the next version of Python to support a similar feature too, possibly varying quotation styles from a version to the next.  Real progress, eh?

I hope I didn't annoy you by re-opening a bug that you consider a feature.  I guess I won't be the last one, until you fix it.

Best
Ale

jaraco:
Sorry, but spaces are valid in filenames and systems should support them.

jaraco:
By the way, I agree with you about "looking for trouble". I think that's a mostly fair characterization. And given that using spaces has shown little trouble over the years, it's now a strength.

RuRo:
Have you considered renaming the file to

[...]

Issue #767, opened 2016-08-29
python setup.py bdist --format=rpm


0xHM:
running install_scripts
Installing easy_install script to /home/builder/tmp/setuptools-26.0.0/build/bdist.linux-x86_64/rpm/BUILDROOT/setuptools-26.0.0-1.x86_64/usr/bin
Installing easy_install-2.7 script to /home/builder/tmp/setuptools-26.0.0/build/bdist.linux-x86_64/rpm/BUILDROOT/setuptools-26.0.0-1.x86_64/usr/bin
writing list of installed files to 'INSTALLED_FILES'
+ /usr/lib/rpm/find-debuginfo.sh --strict-build-id -m --run-dwz --dwz-low-mem-die-limit 10000000 --dwz-max-die-limit 110000000 /home/builder/tmp/setuptools-26.0.0/build/bdist.linux-x86_64/rpm/BUILD/setuptools-26.0.0
/usr/lib/rpm/sepdebugcrcfix: Updated 0 CRC32s, 0 CRC32s did match.
find: ‘debug’: No such file or directory
+ /usr/lib/rpm/check-rpaths /usr/lib/rpm/check-buildroot
+ /usr/lib/rpm/redhat/brp-compress
+ /usr/lib/rpm/redhat/brp-strip-static-archive /usr/bin/strip
+ /usr/lib/rpm/brp-python-bytecompile /usr/bin/python 1
Bytecompiling .py files below /home/builder/tmp/setuptools-26.0.0/build/bdist.linux-x86_64/rpm/BUILDROOT/setuptools-26.0.0-1.x86_64/usr/lib/python2.7 using /usr/bin/python2.7
+ /usr/lib/rpm/redhat/brp-python-hardlink
+ /usr/lib/rpm/redhat/brp-java-repack-jars
Processing files: setuptools-26.0.0-1.noarch
error: File must begin with "/": (dev).tmpl

RPM build errors:
    File must begin with "/": (dev).tmpl
error: command 'rpmbuild' failed with exit status 1

0xHM:
The reason is the space you have space in file name:

setuptools/script (dev).tmpl

jaraco:
Setuptools does have spaces in some of the filenames in the package. I believe it generates everything properly.

I'm a little surprised that rpm can't handle spaces properly. I'm pretty sure this issue would have been reported before if RPM had issues handling spaces in package filenames.

[...]

Issue #134, opened 2014-01-08
Spaces in file names make it impossible to package for Solaris 9 and 10


ghost:
Originally reported by: **automatthias (Bitbucket: [automatthias](http://bitbucket.org/automatthias), GitHub: [automatthias](http://github.com/automatthias))**

---

The `script template.py` file name causes packaging of setuptools to fail, because the Solaris 8, 9, and 10 package manager does not support file names with spaces: spaces are the column delimiter in its internal data files.
### References
1. http://docs.oracle.com/cd/E23824_01/html/821-1473/pkgmap-4.html
2. https://mail.python.org/pipermail/distutils-sig/2013-October/022803.html
3. https://gar.svn.sourceforge.net/svnroot/gar/csw/mgar/pkg/lang-python/pysetuptools/trunk/Makefile
4. https://gar.svn.sourceforge.net/svnroot/gar/csw/mgar/pkg/lang-python/pysetuptools/trunk/files/0001-Solaris-package-mge-does-not-handle-spaces.patch

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/134

jayeye:
The purpose of any tool is to make life easier for the tool user, not to be a vehicle for the ego of its writer.

There is a tremendous number of Unix tools that do not handle spaces in filenames well. It would be far easier to remove the space from that file (which, btw, is the only filename with a space in the entire package).

jaraco:
> There is a tremendous number of Unix tools that do not handle spaces in filenames well.

I acknowledge that many would just prefer to admit defeat and not support spaces in filenames. I think we can rise to the challenge and do better.

Thusfar, there have been two concrete reports of easily surmountable issues over the several years that this naming has been in place.

aixtools:
Hello Jaraco,

Finally getting around to replying to your comment from https://github.com/pypa/pip/issues/3989#issuecomment-249711001

I scanned above - and shall give as short a reply as possible.

While an OS filesystem may support names with spaces it does not make them suddenly better than names without spaces.
Wonderful that setuptools supports installations that want to use that feature. But forcing it on all is dogma. It becomes your opinion of what is best for all becoming the law.

In - what I would call 'UserSpace' these may be useful (e.g., what the python developer is developing as a User application) - but for normal 'binary' packaging (what platform packagers are doing to make a language such as python more easily available) they may be undesired - aka a hindrance for normal systems administration.

[...]

PR #1837, opened 2019-08-29
filename fix for building/packaging

13bscsaamjad:
## Summary of changes
Removed spaces from filenames, spaces create problems when building from sources, e.g., using rpmbuild to make RedHat package. Its better to have these fixes in the upstream rather than applying a patch.

### Pull Request Checklist
- [ ] Changes have tests


jaraco:
Thanks for the PR. In #134 and #1725, I describe the motivations for retaining spaces in filenames. Spaces are valid characters in all modern file systems and packaging tools. I was unaware that Redhat doesn't support spaces in filenames and I'm surprised to hear after 8 years that Redhat struggles with these.

Aah. I do see in #767 and #1846, an issue was reported, but since then no one has taken the time to determine where the underlying defect lies. I'd much rather see the underlying defect (RPM tools don't support legitimate filenames in packages) addressed rather than having Setuptools adapt to that weakness (and leave it for downstream packages to be subject to the same limitation).

I encourage you to delve into the underlying cause, find where in the RPM toolchain that the name is parsed incorrectly, and address that. I'd be happy to provide guidance or to accept patches to improve support for such filenames in setuptools/distutils.

PR #1855, opened 2019-10-03
Upgrade setuptools.depends to importlib from depracated imp

isidentical:
### Summary
Fix;
```console
(.venv) batuhan@x200-trisquel:~/astor$ python setup.py develop
/home/batuhan/.venv/lib/python3.8/site-packages/setuptools/depends.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
```

### Pull Request Checklist
- [x] Changes have tests
- [x] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details (already added about same subject

### DIFF
--- a/setuptools/_imp.py
+++ b/setuptools/_imp.py
@@ -73,4 +73,4 @@ def _resolve(spec):
 
 def get_module(module, paths, info):
     spec = importlib.util.find_spec(module, paths)
-    return importlib.util.module_from_spec(_resolve(spec))
+    return importlib.util.module_from_spec(spec)
--- a/setuptools/_imp.py
+++ b/setuptools/_imp.py
@@ -4,6 +4,7 @@ from the deprecated imp module.
 """
 
 import os
+import sys
 import importlib.util
 import importlib.machinery
 
@@ -70,7 +71,12 @@ def _resolve(spec):
         else spec
     )
 
+def _module_from_spec(spec):
+    if sys.version_info >= (3, 5):
+        return importlib.util.module_from_spec(spec)
+    else:
+        return spec.loader.load_module(spec.name)
 
 def get_module(module, paths, info):
     spec = importlib.util.find_spec(module, paths)
-    return importlib.util.module_from_spec(spec)
+    return _module_from_spec(spec)
--- a/setuptools/_imp.py
+++ b/setuptools/_imp.py
@@ -60,13 +60,17 @@ def find_module(module, paths=None):
 
 def get_frozen_object(module, paths):
     spec = importlib.util.find_spec(module, paths)
-    if hasattr(spec, 'submodule_search_locations'):
-        spec = importlib.util.spec_from_loader('__init__.py', spec.loader)
-    return spec.loader.get_code(module)
+    return spec.loader.get_code(_resolve(module))
+
+
+def _resolve(spec):
+    return (
+        importlib.util.spec_from_loader('__init__.py', spec.loader)
+        if hasattr(spec, 'submodule_search_locations')
+        else spec
+    )
 
 
 def get_module(module, paths, info):
     spec = importlib.util.find_spec(module, paths)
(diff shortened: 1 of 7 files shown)


isidentical:
apparently the error is related with https://github.com/pytest-dev/pytest/issues/2506

isidentical:
I'll change the `get_code` api calls to manually compiling / marshalling.

setuptools/tests/test_setuptools.py
```diff
@@ -64,6 +65,7 @@ def f1():
         # recognized name, not assigned
         dep.extract_constant(fc, 'z', -1) is None
 
+    @py2_only
```
pganssle: `dep.find_module` appears to be a tested part of the public API, we cannot make this test `py2_only` unless we intend to *remove* it for Python 3 users.

If we're planning to remove it for Python 3 users, we should probably remove it for Python 2 users as well, and we should go through a deprecation cycle.

setuptools/tests/test_integration.py
```diff
@@ -143,6 +143,7 @@ def test_build_deps_on_distutils(request, tmpdir_factory, build_dep):
             'tests_require',
             'python_requires',
             'install_requires',
+            'long_description_content_type'
```
pganssle: Why is this test changed?

setuptools/tests/test_setuptools.py
isidentical: If you want to keep it, i can make it working with importlib on py3+, i thought the `find_spec` solves the problem so this is not needed on py3+

isidentical:
I dont know why but when i force pushed my branch, github auto closed this PR for me? I can't reopen it either so i'll make another one
image

isidentical:
All done, it is working as expectedly. No API changed.

jaraco:
I really hate how entangled these behaviors are... and how messy the code will continue to be even after python 2.7 support is dropped. I've got a patch incoming that I think helps.

jaraco:
Almost an hour ago, I thought I'd found a solution, but I realized used of `importlib.import_module` probably isn't the right usage... that the `paths` wasn't being honored. So I re-did the work and found that tests were failing, specifically with `setuptools._imp.get_module()`.

I think things are working in that last commit.

@isidentical can you review this latest change and in particular explain what the code in _imp._resolve is meant to do? ...and whether it should be called in _imp.get_module?

Thanks.

isidentical:
now it passes py27 tests (a little fix about `imp.load_module`

isidentical:
py34 tests are passing (use `module_from_spec` if py35+)

isidentical:
> @isidentical can you review this latest change
done

> explain what the code in _imp._resolve is meant to do?
when i was testing my initial patch i encountered with a case but i dont think it is needed here.

setuptools/_imp.py
```diff
@@ -0,0 +1,70 @@
...
+    else:
+        path = None
+        suffix = mode = ''
+
+    return file, path, (suffix, mode, kind)
+
+
+def get_frozen_object(module, paths):
+    spec = importlib.util.find_spec(module, paths)
+    return spec.loader.get_code(_resolve(module))
```
jaraco: `_resolve` was removed, but it's still referenced here. Is it needed?

isidentical: nop, 
```py
 $ python
Python 3.8.0 (default, Oct 14 2019, 21:29:03)
[GCC 7.4.0] on linux
>>> from setuptools._imp import get_frozen_object
>>> get_frozen_object("__hello__")
<code object <module> at 0x7f5c5e76bbe0, file "Tools/freeze/flag.py", line 1>
```

isidentical:
Thanks

jaraco:
This change was implicated in #1896

Issue #1296, opened 2018-03-17
setuptools should not rely on pkg_resources vendored packages


jaraco:
A prerequisite of separating setuptools and pkg_resources (#863), setuptools must vendor the dependencies it uses directly and not import them from pkg_resources' vendoring.

Issue #229, opened 2014-07-05
Add support for dependencies (such as six)


ghost:
Originally reported by: **jaraco (Bitbucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

I'd like to use the [six](/gutworth/six) library for Python 2 compatibility.

As it currently stands, setuptools does not have any non-extra dependencies, so introduction of any dependency has the potentially to be disruptive, so this ticket is created primarily around that discussion.

Here's what I have in mind:
1. Vendor six in the sdist to be used when setuptools installs itself (but never install from that verison).
2. Declare those dependencies as proper dependencies in install_requires.
3. Have the code reference those libraries naturally.

This change would require bootstrap installers to either:
- Have the dependencies installed in advance.
- Have network access to install the dependencies from PyPI.
- Configure an index-url from which to install the dependencies.

What I want to avoid is for setuptools to vendor the dependencies for any purpose other than bootstrapping. I'd also like to avoid having conditional imports in the code. I especially want to avoid creating complicated vendoring scenarios that violate the expectations of platform package maintainers (namely Debian and Fedora).

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/229

jaraco:
In the aforementioned "those changes", here's an updated verison of [those changes](https://github.com/pypa/setuptools/compare/9a531b50020849193a55262b3f9cfc375ad3f37c...2dc55bc1a9904ffd44d4cd5a3c83c7a1f12c75ec) in Github.

PR #1890, opened 2019-10-27
Remove module importing hack

### LINKED ISSUE #1888: package metadata is no longer picklable
Since switching to OrderedDict, package metadata can no longer be pickled.

This can cause problems if you try to use setuptools with multiprocessing.

```
  File "/Users/dan/Documents/colcon_ws/src/colcon-python-setup-py/colcon_python_setup_py/package_identification/python_setup_py.py", line 257, in get_setup_information
    'stop_after': 'config'
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 261, in apply
    return self.apply_async(func, args, kwds).get()
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
multiprocessing.pool.MaybeEncodingError: Error sending result: '<distutils.dist.DistributionMetadata object at 0x10c1fcd50>'. Reason: 'PicklingError("Can't pickle <class 'setuptools._vendor.ordered_set.OrderedSet'>: it's not the same object as setuptools._vendor.ordered_set.OrderedSet")'
```

_Originally posted by @rotu in https://github.com/pypa/setuptools/pull/1690#issuecomment-545992670_

rotu:
## Summary of changes

Remove import hack that prevents objects from pickling. This was to support relative imports between vendored dependencies, but such relative imports do not (and should not) exist anyway.
 
Closes #1888

### Pull Request Checklist
- [x] Changes have tests


jaraco:
The comment around this hack is clearly there to protect this code - to indicate its purpose so it's not capriciously removed. I don't see in this PR or in the associated bug an analysis of why this code is no longer necessary. I do agree it's undesirable that the Distribution object isn't pickleable due to this hack, but we'll want to assess and weigh the costs of removing this hack with keeping it. Could you try tracing this code to its inception and see if there's any hint there as to what problem this code solves?

rotu:
> Could you try tracing this code to its inception and see if there's any hint there as to what problem this code solves?

This seems to be its origin, from when `extern` and `_vendor` both had their own `VendorImporter`: https://github.com/pypa/setuptools/commit/a76f5c03136f0c69e8c366bf4069ec8d89d85da7#diff-c6950cefad8b244938b76f24a0db9a6a
commit message: "Pop the module off the stack, preventing the 'Version' class from having a different manifestation in packaging than in pkg_resources."

Below is my understanding and may be wrong:

It would have been a problem at that time, since, depending which importer it goes through, you would have a different module object and a different class object, causing `isinstance` checks to fail.

https://github.com/pypa/setuptools/blob/de5caaee3c9a23600b96f512ed2054b4e27357a8/pkg_resources/_vendor/packaging/specifiers.py#L149-L152

There still is some weirdness which I didn't address and exists both before and after removing this hack. Namely that, while immediate descendants of `_vendor` (e.g. `setuptools.extern.packaging`) report (via `__name__`) as being in the `_vendor` package, their children (e.g. `setuptools.extern.packaging.version`) report as being in the `extern` package instead.

jaraco:
Here are [my thoughts](https://github.com/pypa/setuptools/issues/229#issuecomment-202914063) around when that hack originated.

jaraco:
> Could you please contribute a test that demonstrates the continued need for this code?

A fair and reasonable request. Reading my thoughts from four years ago, it appears to me as if tests were failing without this change. The fact that they're not now suggests that maybe the hack is unnecessary. I tried drafting a test, but I'm not able to ascertain what behavior was faulty without that effect.

I think the goal was that when using this importer, there should be exactly one instance of the imported modules in place... that there shouldn't be both `setuptools._vendor.packaging` and `setuptools.extern.packaging` in `sys.modules`.

I observe that in #1296, I copied the VendorImporter code. I definitely want to avoid allowing those implementations to deviate (and perhaps I should revert d0f7a33daba39f60aa402c094cd4276b8b6ce462 unless revisiting that issue). Ugh. Those two have already deviated (949bd166692af0694631542b02e3a2f26a48b455), and on this line of code that's being removed.

jaraco:
In this latest commit, I'm applying the same change to both implementations. Tests are passing for me locally. If they pass CI, then my plan is to move forward with this change. If it causes problems, we can capture those issues and add tests around those use-cases, and thanks to the tests you've added here, that expectation should be able to be retained as well.

PR #1898, opened 2019-10-31
Remove "upload" and "register" commands.

pganssle:
## Summary of changes
The upload and register commands were deprecated over a year ago, in July 2018 (PR #1410, discussed in issue #1381). It is time to actively remove them in favor of twine.

I think the most controversial part of this might be that I've added `setuptools.errors`. It wasn't clear what the best error class to use would be. `distutils` will use `raise SystemExit` for a missing command; that didn't seem quite like the right thing to do, but I'm willing to be convinced otherwise.

CC @pypa/setuptools-developers @di 

### Pull Request Checklist
- [X] Changes have tests


docs/setuptools.txt
```diff
@@ -2087,16 +2087,8 @@ New in 41.5.0: Deprecated the test command.
...
-    <https://pypi.org/p/twine>`_
-
-The ``upload`` command is implemented and `documented
-<https://docs.python.org/3.1/distutils/uploading.html>`_
-in distutils.
-
-New in 20.1: Added keyring support.
-New in 40.0: Deprecated the upload command.
+The ``upload`` command was deprecated in version 40.0 and removed in version
+42.0. Use `twine <https://pypi.org/p/twine>`_ instead.
```
di: I think it'd be helpful to show folks exactly what command they should use instead here, something like "Instead of `python setup.py upload`, use `twine upload dist/*`"

jaraco: Agreed, except those commands aren't equivalent. `setup.py upload` would have no effect unless preceded with a `_dist` command... whereas `twine` requires rethinking your workflow (first run your `_dist` command or a `pep517` builder, then run twine on the output of that command). Do we really want to try to describe all those details, or leave it vague and let the user's rely on the twine docs? I lean toward the latter.

pganssle: Given how infrequently we update this documentation, I think a link is probably better than an explanation (especially since we actually already have an explanation on how to use `twine` earlier in this same document). I updated the documentation to point at the Python Packaging User Guide.

Sounds good?

di: I think that's a good compromise.

pganssle:
Oops, turns out that if you name your classes `UploadTest` instead of `TestUpload` it doesn't get detected by `pytest`. I guess that's why they were originally named `Test(Upload|Register)Test`...

Yet another time when I am appreciative that we include the tests in the coverage, and that we enforce diff coverage...

pganssle:
OK, merging this, which means that next release should be a breaking one!

PR #1899, opened 2019-10-31
Error on warnings.

### LINKED ISSUE #1823: CI should fail on deprecation warnings

jaraco:
Fixes #1823


pytest.ini
```diff
@@ -6,5 +6,7 @@ flake8-ignore =
     setuptools/py*compat.py F811
 doctest_optionflags=ELLIPSIS ALLOW_UNICODE
 filterwarnings =
+	# Fail on warnings
+	error
 	# https://github.com/pypa/setuptools/issues/1823
 	ignore:bdist_wininst command is deprecated
```
hugovk: The file also uses spaces, should these tabs be spaces too?

jaraco: Yes, indentation should be consistent.

pytest.ini
```diff
@@ -6,5 +6,7 @@ flake8-ignore =
     setuptools/py*compat.py F811
 doctest_optionflags=ELLIPSIS ALLOW_UNICODE
 filterwarnings =
-	# https://github.com/pypa/setuptools/issues/1823
-	ignore:bdist_wininst command is deprecated
+    # Fail on warnings
+    error
+    # https://github.com/pypa/setuptools/issues/1823
+    ignore:bdist_wininst command is deprecated
```
webknjaz: Shouldn't this be something like 
```suggestion
    ignore:bdist_wininst command is deprecated:DeprecationWarning
```

jaraco: It works either way. It seems to me unnecessary to additionally specify `DeprecationWarning` unless we could imagine that a warning that's not a `DeprecationWarning` would be used to indicate "bdist_wininst command is deprecated" and that we would want to trap _that_ warning but not this one.

jaraco: Plus, this PR isn't concerned with that line; it's just getting reindented to follow the convention I use elsewhere.

jaraco:
I see this error on Python 2. I'm just going to suppress it.

```
___________________________________________________ TestExceptionSaver.test_sandbox_violation_raised_hiding_setuptools ___________________________________________________

self = <setuptools.tests.test_sandbox.TestExceptionSaver instance at 0x10c5f5560>
tmpdir = local('/private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pytest-of-jaraco/pytest-30/test_sandbox_violation_raised_0')

    def test_sandbox_violation_raised_hiding_setuptools(self, tmpdir):
        """
        When in a sandbox with setuptools hidden, a SandboxViolation
        should reflect a proper exception and not be wrapped in
        an UnpickleableException.
        """
    
        def write_file():
            "Trigger a SandboxViolation by writing outside the sandbox"
            with open('/etc/foo', 'w'):
                pass
    
        with pytest.raises(setuptools.sandbox.SandboxViolation) as caught:
            with setuptools.sandbox.save_modules():
                setuptools.sandbox.hide_setuptools()
                with setuptools.sandbox.DirectorySandbox(str(tmpdir)):
>                   write_file()

setuptools/tests/test_sandbox.py:125: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/contextlib.py:35: in __exit__
    self.gen.throw(type, value, traceback)
setuptools/sandbox.py:166: in save_modules
    saved_exc.resume()
setuptools/sandbox.py:141: in resume
    six.reraise(type, exc, self._tb)
setuptools/sandbox.py:154: in save_modules
    yield saved
setuptools/tests/test_sandbox.py:125: in test_sandbox_violation_raised_hiding_setuptools
    write_file()
setuptools/tests/test_sandbox.py:118: in write_file
    with open('/etc/foo', 'w'):
setuptools/sandbox.py:418: in _open
    self._violation("open", path, mode, *args, **kw)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <setuptools.sandbox.DirectorySandbox instance at 0x10c5f5a70>, operation = 'open', args = ('/etc/foo', 'w'), kw = {}

    def _violation(self, operation, *args, **kw):
>       from setuptools.sandbox import SandboxViolation
E       RuntimeWarning: Parent module 'setuptools' not found while handling absolute import

setuptools/sandbox.py:406: RuntimeWarning
```

Issue #1851, opened 2019-09-26
Update docs for required compilers for Windows


letmaik:
https://wiki.python.org/moin/WindowsCompilers has Visual C++ 14.X listed for Python 3.5, 3.6, 3.7, 3.8, which to me means I can use any 14.x compiler for these Python versions. This seems incorrect though as I'm getting the following error for Python 3.5 where I only have > 14.0 available:
> error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++ Build Tools": https://visualstudio.microsoft.com/downloads/

It would be great if the table could be fixed.

zooba:
The table is fine - setuptools needs to be updated to detect newer versions than 14.0.

This function from distutils can be cloned, and possibly even enhanced to use the newer `-find` argument to vswhere.

https://github.com/python/cpython/blob/61691d833631fed42b86605b09e1535e3e8d40e5/Lib/distutils/_msvccompiler.py#L58-L88

letmaik:
Please have a look at https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2019 in the comments sections:
> Binary compat between 2015 and 2017 is guaranteed only when the consuming code is built with a toolset/compiler version that is the same or newer than the toolset/compiler used to build the library. So, a VS2015 exe + VS2017 DLL is not guaranteed to be compatible.

PR #1904, opened 2019-11-11
Use CPython 3.8.0 mechanism to find msvc 14+

### LINKED ISSUE #1903: setuptools fails to build a windows python 3.5 C extension on GitHub Actions (msvc)
setuptools fails to build a windows python 3.5 C extension on GitHub Actions using provided msvc.

python 3.6, 3.7 & 3.8 do succeed meaning the proper msvc compilers are present.

Per @zooba [comment](https://github.com/pypa/setuptools/issues/1851#issuecomment-537714885), cloning the detection mechanism from CPython 3.8.0 distutils does indeed work as can be seen [here](https://github.com/joerick/cibuildwheel/pull/194) where all CI tests pass (windows: TravisCI, AppVeyor, Azure Pipelines, GitHub Actions).

I do have a branch of setuptools with such a patch and will open a PR. Please note that I can't test it thoroughly cause I don't have any Windows environment at hand (except for CI which requires very long iterations).

mayeut:
## Summary of changes

Update `msvc.py` to use CPython 3.8.0 mechanism to find msvc 14+ as recommended in this [comment](https://github.com/pypa/setuptools/issues/1851#issuecomment-537714885).

Closes #1903 

This PR has been tested as part of [cibuildwheel tests using various CI providers](https://github.com/joerick/cibuildwheel/pull/194).
Please note that I can't test it more thoroughly cause I don't have any Windows environment at hand (except for CI which requires very long iterations).

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,15 @@ environment:
   CODECOV_ENV: APPVEYOR_JOB_NAME
 
   matrix:
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      APPVEYOR_JOB_NAME: "python35-x64-vs2015"
+      PYTHON: "C:\\Python35-x64"
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      APPVEYOR_JOB_NAME: "python35-x64-vs2017"
+      PYTHON: "C:\\Python35-x64"
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      APPVEYOR_JOB_NAME: "python35-x64-vs2019"
+      PYTHON: "C:\\Python35-x64"
     - APPVEYOR_JOB_NAME: "python36-x64"
       PYTHON: "C:\\Python36-x64"
     - APPVEYOR_JOB_NAME: "python37-x64"
--- a/changelog.d/1904.change.rst
+++ b/changelog.d/1904.change.rst
@@ -1,0 +1,1 @@
+Update msvc.py to use CPython 3.8.0 mechanism to find msvc 14+
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -26,6 +26,7 @@ from os.path import join, isfile, isdir, dirname
 import sys
 import platform
 import itertools
+import subprocess
 import distutils.errors
 from setuptools.extern.packaging.version import LegacyVersion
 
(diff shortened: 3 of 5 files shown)


JGoutin:
Using a Python 3.8 back port instead of the current mechanism seem to be the good way to do.

This also allow to completely drop the old mechanism that become unused. This will make this file far more easy to maintain. Note that this will remove the following:
- VC 2010 support, only used by Python 3.4 that is deprecated.
- Microsoft Windows SDK 6.1 and 7.0 support, tools that are deprecated and where removed from download by Microsoft. (And there is still the "VC for python 2.7" that replace them)

I sent you a PR with the following changes:
* Drop the old mechanism.
* Fixed a missing import.
* PEP8 + Add python 3.8 backport comment.

I do not have a Windows right now to do more tests on it.

mayeut:
Thanks @JGoutin, this is much cleaner indeed. All CI tests of cibuildwheel are passing with your version of `msvc.py`. I will correct the missing import in the meantime (itertools.count() right ?)

If @pypa/setuptools-developers are ok with your notes (Microsoft Windows SDK 6.1 and 7.0, Python 3.4 support dropped), I'll merge that in this PR.

mayeut:
> There are maybe tests with Python 3.8 that can be back-ported too.

Indeed, I'll check if I can add those:
https://github.com/python/cpython/blob/master/Lib/distutils/tests/test_msvccompiler.py

JGoutin:
> (itertools.count() right ?)
It's exactly that.

JGoutin:
Concerning the drop of the VC2010 support, Python 3.4 will be dropped with #1908.

JGoutin:
> If @pypa/setuptools-developers are ok with your notes (Microsoft Windows SDK 6.1 and 7.0, Python 3.4 support dropped), I'll merge that in this PR.

#1908 is merged, Python 3.4 support is dropped from setuptools.

mayeut:
I will drop python 3.4 support in `msvc.py` from the PR

JGoutin:
> Edit: I won't drop python 3.4
> 
> Since python 3.4 uses the same code base as python 2.7, I'd rather not change it. It'll be easier to drop as a whole once python 2.7 support is really removed from setuptools.

Removing the python 3.4 support will not remove the Python 2.7 support. It will only remove the
Microsoft Windows SDK 6.1 and 7.0 support, that are deprecated and not available anymore.

Python 2.7 requires `msvc9_find_vcvarsall` & `msvc9_query_vcvarsall` functions. 
In the PR code, theses functions are still available: https://github.com/JGoutin/setuptools/blob/msvc-python3.5/setuptools/msvc.py#L56.

mayeut:
@JGoutin,
I'd rather not change python 2.7 support at all to prevent any regressions, even if unlikely. So, in my opinion, removing  Microsoft Windows SDK 6.1 and 7.0 support is not an option.

mayeut:
@jaraco,
can you review this PR please ?

jaraco:
Released as 45.3.0

Issue #1896, opened 2019-10-30
TestDepends.testRequire regression in 41.6.0 (py3)


mgorny:
While trying to build the new release of setuptools, I get the following test failure:

```
==================================================================== FAILURES =====================================================================
_____________________________________________________________ TestDepends.testRequire _____________________________________________________________

self = <setuptools.tests.test_setuptools.TestDepends object at 0x7fbfae31d710>

    @needs_bytecode
    def testRequire(self):
        req = Require('Json', '1.0.3', 'json')
    
        assert req.name == 'Json'
        assert req.module == 'json'
        assert req.requested_version == '1.0.3'
        assert req.attribute == '__version__'
        assert req.full_name() == 'Json-1.0.3'
    
        from json import __version__
        assert req.get_version() == __version__
        assert req.version_ok('1.0.9')
        assert not req.version_ok('0.9.1')
        assert not req.version_ok('unknown')
    
        assert req.is_present()
        assert req.is_current()
    
        req = Require('Json 3000', '03000', 'json', format=LooseVersion)
        assert req.is_present()
        assert not req.is_current()
        assert not req.version_ok('unknown')
    
        req = Require('Do-what-I-mean', '1.0', 'd-w-i-m')
        assert not req.is_present()
        assert not req.is_current()
    
        req = Require('Tests', None, 'tests', homepage="http://example.com")
        assert req.format is None
        assert req.attribute is None
        assert req.requested_version is None
        assert req.full_name() == 'Tests'
        assert req.homepage == 'http://example.com'
    
        from setuptools.tests import __path__
        paths = [os.path.dirname(p) for p in __path__]
>       assert req.is_present(paths)
E       AssertionError: assert False
E        +  where False = <bound method Require.is_present of <setuptools.depends.Require object at 0x7fbfae0d0b38>>(['/tmp/portage/dev-python/setuptools-41.6.0/work/setuptools-41.6.0-python3_5/setuptools'])
E        +    where <bound method Require.is_present of <setuptools.depends.Require object at 0x7fbfae0d0b38>> = <setuptools.depends.Require object at 0x7fbfae0d0b38>.is_present

setuptools/tests/test_setuptools.py:120: AssertionError
```

I can reproduce it reliably with at least pypy3.6 (7.2.0) & python3.5 (3.5.7). I haven't tested other versions yet.

[...]

PR #1905, opened 2019-11-15
Fix _imp module behaviour if is defined paths in find_spec call

mimi1vx:
## Summary of changes

Closes <#1896>

### Pull Request Checklist
- [x] Changes have tests


changelog.d/1905.change.rst
```diff
@@ -0,0 +1 @@
+Fixed test failure introduced in 41.6.0 when the 'tests' directory is not present.
```
mimi1vx: test fail? original code in _imp.py was broken.

jaraco: But you're right, that message is a little unclear, implicating the test and not the code. I can improve that.

PR #1909, opened 2019-11-17
Revert "drop easy_install script and associated documentation"

benoit-pierre:
This reverts commit 6e1838a9fb5feb000ba9b6a3c37c8b39d7e872b3.

### DIFF
--- a/changelog.d/1830.breaking.rst
+++ b/changelog.d/1830.breaking.rst
@@ -1,0 +1,7 @@
+Mark the easy_install script and setuptools command as deprecated, and use `pip <https://pip.pypa.io/en/stable/>`_ when available to fetch/build wheels for missing ``setup_requires``/``tests_require`` requirements, with the following differences in behavior:
+ * support for ``python_requires``
+ * better support for wheels (proper handling of priority with respect to PEP 425 tags)
+ * PEP 517/518 support
+ * eggs are not supported
+ * no support for the ``allow_hosts`` easy_install option (``index_url``/``find_links`` are still honored)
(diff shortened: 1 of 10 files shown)


benoit-pierre:
Also added changelog entries for this and #1830.

jaraco:
Thanks @benoit-pierre for putting this together. I wonder if we can avoid re-adding this functionality by instead moving it to another package (called easy_install), and having setuptools depend on that package. That decoupling would make it obvious what is deprecated and provide a path to backward compatibility (those requiring/expecting this functionality could retain it by installing `easy_install` separately).

My main reluctance for doing this is that by adding a dependency, that dependency becomes non-optional (`pkg_resources` will expect the `easy_install` package to be present and crash if it's not).

Let me put together a draft.

pganssle:
@jaraco I think that might overly complicate things. Even in the now-reverted change, all the old functionality was preserved, we just had removed one of the public interfaces. We can just add a new wrapper script that warns that it's deprecated and in say 1 year replace it with one that actually raises an exception, same way we've done for `upload` and `register `.

I think there will be very little ongoing support burden to ship a script that just raises an exception, so we should probably just do the lowest-effort thing that gets us to that endpoint without introducing any footguns for end users.

jaraco:
You make a good argument. I think I agree.

I did throw together the [easy_install package](/pypa/easy_install). All it needs is a tag to cut a release, but I'll refrain from doing that if we're not going to use it.

jaraco:
In #1913, I drafted my concept, but right away it reveals issues (with bootstrapping), at least one of the foot guns. Let's proceed with this approach for now.

Issue #1920, opened 2019-11-26
Handling of find_links in setup.cfg in 42.0.0


neuroid:
Hi,

I have a couple of `setup_requires` dependencies in my setup.py and my setup.cfg looks like this:

```
[easy_install]
find_links = https://example.com
```

Looking at: https://github.com/pypa/setuptools/blob/e84f616a6507ec9115fad68b221cbf5333d9d2d9/setuptools/installer.py#L119

The `cmd` ends up being:

```python
['python', '-m', 'pip', '--disable-pip-version-check', 'wheel', '--no-deps', '-w', '/tmp/tmpWyNzjZ', '--quiet', '--find-links', 'h', '--find-links', 't', '--find-links', 't', '--find-links', 'p', '--find-links', 's', '--find-links', ':', '--find-links', '/', '--find-links', '/', '--find-links', 'e', '--find-links', 'x', '--find-links', 'a', '--find-links', 'm', '--find-links', 'p', '--find-links', 'l', '--find-links', 'e', '--find-links', '.', '--find-links', 'c', '--find-links', 'o', '--find-links', 'm', 'babel; extra == "i18n"']
```
It seems the assumption was that `find_links` is a list but it ends up being a string.

Additionally, since pip is unable to build/fetch a wheel, the following line throws an IndexError:
https://github.com/pypa/setuptools/blob/e84f616a6507ec9115fad68b221cbf5333d9d2d9/setuptools/installer.py#L122

Perhaps this could be caught and a more user-friendly error displayed instead.

benoit-pierre:
See #1921 for the `find_links` issue.

As for the `IndexError`, this should not happen, as the expectation is that the pip command will return with non-zero exit code in case of failure. This is probably related to the `extra == "i18n"` marker, which I think should have been stripped.

Can you provide the full `setup.py`?

neuroid:
Here's a relevant excerpt:

```python
from setuptools import find_packages
from setuptools import setup

setup(name='templates',
      version='0.0.1',
      description='Templates',
      packages=find_packages(),
      setup_requires=[
          'setup[i18n]',
      ])
```

The `setup` package has a very similar setup.py, but with the following extras definition:

```python
      ...
      extras_require={
          'i18n': ['babel'],
      },
      ...
```

benoit-pierre:
[...]

PR #1922, opened 2019-11-26
fix possible issue with transitive build dependencies

benoit-pierre:
Handle the case where a missing transitive build dependency is required by an extra for an already installed build dependency.

Example:
* a test package with a `setup.py` containing: `setup_requires=['dep[extra]']`
* `dep` is already installed, and the dependencies for `extra` are: `['extra_dep']`
* `extra_dep` is not already installed, and `fetch_build_egg('extra_dep; extra =="extra"')` is called

If the marker is not stripped, `pip wheel [...] 'extra_dep; extra =="extra"'` ignores the requirement, and does not build a wheel, but still returns with no error.

## Summary of changes

Closes the second issue found in #1920.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/installer.py
+++ b/setuptools/installer.py
@@ -64,8 +64,11 @@ def fetch_build_egg(dist, req):
         pkg_resources.get_distribution('wheel')
     except pkg_resources.DistributionNotFound:
         dist.announce('WARNING: The wheel package is not available.', log.WARN)
-    if not isinstance(req, pkg_resources.Requirement):
-        req = pkg_resources.Requirement.parse(req)
+    # Ignore environment markers: if we're here, it's needed. This ensure
+    # we don't try to ask pip for something like `babel; extra == "i18n"`,
+    # which would always be ignored.
+    req = pkg_resources.Requirement.parse(str(req))
+    req.marker = None
     # Take easy_install options into account, but do not override relevant
     # pip environment variables (like PIP_INDEX_URL or PIP_QUIET); they'll
     # take precedence.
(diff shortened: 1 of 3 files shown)


frivoire:
It seems I'm impacted by this bug too, with one difference : my use case is dependencies in `tests_require`  and not in `setup_requires`.

So, the proposed fix should be ok (it's in a generic code path that is used for both `tests_require`  and `setup_requires` as far as I understand) :)

setuptools/tests/test_easy_install.py
```diff
@@ -744,6 +745,49 @@ def test_setup_requires_with_python_requires(self, monkeypatch, tmpdir):
         eggs = list(map(str, pkg_resources.find_distributions(os.path.join(test_pkg, '.eggs'))))
         assert eggs == ['dep 1.0']
 
+    def test_setup_requires_with_transitive_extra_dependency(self, monkeypatch):
```
frivoire: The addition of this test seems very relevant 😃   
You should probably add another (very similar) to test the same behaviour in `tests_require`, or maybe modify your existing function to do the 2 situations in 1 function.

benoit-pierre: Good point.

setuptools/installer.py
```diff
@@ -66,6 +66,10 @@ def fetch_build_egg(dist, req):
         dist.announce('WARNING: The wheel package is not available.', log.WARN)
     if not isinstance(req, pkg_resources.Requirement):
         req = pkg_resources.Requirement.parse(req)
+    # Ignore environment markers: if we're here, it's needed. This ensure
+    # we don't try to ask pip for something like `babel; extra == "i18n"`,
+    # which would always be ignored.
+    req.marker = None
```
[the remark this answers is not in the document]
benoit-pierre: Good point.

setuptools/tests/test_easy_install.py
benoit-pierre: Yeah, but what all those `test_setup_requires...` tests really test is implementation of `fetch_build_egg`.

PR #1925, opened 2019-12-01
Simplify mechanized releases through tox

jaraco:
This approach adopts the release process from jaraco/skeleton. This approach gives the project more control over the release process, decoupling from DPL (which has its own issues) and relies on tokens as installed to environment variables (as travis has issues with that).

This approach makes it possible to cut manual releases with `tox -e release` (assuming TWINE_PASSWORD is set to a valid token) as well as control other things about the release process (such as using pep517 to build the dists).

I know in d012529d, Paul removed the `build-backend` from `pyproject.toml`, but when I tested this approach just now, it worked fine (and actually cut a post release), suggesting that through some advancements, Setuptools _can_ now be built by pep517.

Is this approach viable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ jobs:
     env: DISABLE_COVERAGE=1
   - python: 3.4
   - python: 3.5
-  - &default_py
-    python: 3.6
+  - python: 3.6
   - python: 3.7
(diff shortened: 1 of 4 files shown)


jaraco:
Since the tests are failing, I guess that answers my question.

PR #1938, opened 2019-12-12
Update setuptools.txt

maggyero:
This PR:

- uses the long option `--upgrade` consistently;
- uses the built-in `exec` function instead of the Python 2-only `execfile`;
- fixes a command typo installing `setuptools `and `wheel `twice instead of `twine`;
- removes inconsistent `--user` options in some `pip install` commands;
- calls `pip install` from the PATH consistently instead of from the module namespace (`python -m pip install`) as we are not upgrading `pip` itself on Windows anywhere;
- fixes a URI.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -62,7 +62,7 @@ Installing ``setuptools``
 
 To install the latest version of setuptools, use::
 
-    pip install -U setuptools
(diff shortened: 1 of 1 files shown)


jaraco:
Thank you very much for this gift.

Issue #1943, opened 2019-12-21
Docs: dead link


gpotter2:
Hi,
https://github.com/pypa/setuptools/blob/1d03fdc94c3676a5b675ec7d818d48c6a772fb49/docs/setuptools.txt#L1007
This link is dead.

jaraco:
Thanks for reporting this. Can you update the link to point to [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources)?

gpotter2:
See https://github.com/pypa/setuptools/pull/1954

PR #1944, opened 2019-12-21
Accessing Package Resources dead link

gpotter2:
## Summary of changes
Fixed a dead link.
The [new link](https://github.com/getnikola/nikola/issues/1229#issue-31382857) isn't perfect for sure: it's very short. But IMHO the examples provided are very clear and explicit.

Feel free to provide a better alternative 

Closes https://github.com/pypa/setuptools/issues/1943


jaraco:
Thanks for this, but I think the best recommendation is to use [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources).

PR #1949, opened 2019-12-27
Patched sdist in PEP 517 to include pyproject.toml

uranusjr:
## Summary of changes

A flag is added to `setuptools.command.sdist.sdist` to include `pyproject.toml` if present. This flag is False by default, but set to True in PEP 517 build hooks, so the built sdist is compliant to the PEP.

Non-PEP 517 sdist builds remain the same as before (not including `pyproject.toml` unless explicitly included in `MANIFEST.in`).

Ref #1632

### Pull Request Checklist

- [x] Changes have tests

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -36,6 +36,7 @@ import contextlib
 import setuptools
 import distutils
 from setuptools.py31compat import TemporaryDirectory
+from setuptools.command.sdist import sdist
 
 from pkg_resources import parse_requirements
 from pkg_resources.py31compat import makedirs
@@ -209,9 +210,13 @@ class _BuildMetaBackend(object):
                                          wheel_directory, config_settings)
 
     def build_sdist(self, sdist_directory, config_settings=None):
-        return self._build_with_temp_dir(['sdist', '--formats', 'gztar'],
-                                         '.tar.gz', sdist_directory,
-                                         config_settings)
+        orig, sdist.include_pyproject_toml = sdist.include_pyproject_toml, True
+        try:
+            return self._build_with_temp_dir(['sdist', '--formats', 'gztar'],
+                                             '.tar.gz', sdist_directory,
+                                             config_settings)
+        finally:
+            sdist.include_pyproject_toml = orig
 
 
 class _BuildMetaLegacyBackend(_BuildMetaBackend):
--- a/changelog.d/1632.breaking
+++ b/changelog.d/1632.breaking
@@ -1,0 +1,1 @@
+Include ``pyproject.toml`` in sdist built with PEP 517 build hooks.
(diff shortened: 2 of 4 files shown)


setuptools/build_meta.py
```diff
@@ -209,9 +210,13 @@ def build_wheel(self, wheel_directory, config_settings=None,
                                          wheel_directory, config_settings)
 
     def build_sdist(self, sdist_directory, config_settings=None):
-        return self._build_with_temp_dir(['sdist', '--formats', 'gztar'],
-                                         '.tar.gz', sdist_directory,
-                                         config_settings)
+        orig, sdist.include_pyproject_toml = sdist.include_pyproject_toml, True
```
jaraco: This monkeypatching style is not threadsafe. Overlapping calls could leave the original unset. I don't believe thread safety is needed here.

jaraco: I'd love for this not to be monkey-patched, but assuming that's how it must be, this patch should be a one-line call to encapsulate the monkeypatch behavior and decouple it from the underlying call. Something like `with self._enable_pyproject_toml():`.

jaraco:
> Given that this change will cause the pyproject.toml to be _present_ in some sdists, how can we be confident that this change won't have the same regression as #1634?

Looking at [this comment](https://github.com/pypa/setuptools/issues/1644#issuecomment-554677756), the plan for setuptools is to start supporting PEP 517 builds of itself, which would in all likelihood cause the same regression. That comment specifically recommends for setuptools to exclude pyproject.toml from its source distribution, so probably all that's needed is to provide a specialized exclusion for Setuptools (but otherwise include the file for all other projects even on non-PEP 517 builds).

uranusjr:
> Given that this change will cause the pyproject.toml to be present in some sdists, how can we be confident that this change won't have the same regression as #1634?

This patch is different from #1634 in that it only affects packages built through PEP 517 hooks, not `setup.py sdist` (because it does not go through the hooks; nothing in Setuptools does). So nobody using the usual build commands would be affected.

> Looking at this comment, the plan for setuptools is to start supporting PEP 517 builds of itself, which would in all likelihood cause the same regression. That comment specifically recommends for setuptools to exclude pyproject.toml from its source distribution, so probably all that's needed is to provide a specialized exclusion for Setuptools (but otherwise include the file for all other projects even on non-PEP 517 builds).

I would be OK if Setuptools decide to include `pyproject.toml` by default even for non-PEP 517 build. Support for `backend-path` has been added to pip in pypa/pip#7394 and will be released in 21.0, but an exception would still be needed for Setuptools for quite a while, until people upgrades.

jaraco:
I think we want to go with the approach described in #1650, but I'd like to include your test, so I'll cherry-pick that.

jaraco:
In 8a7a6272, I've cherry-picked your test onto #1650.

PR #1959, opened 2020-01-08
Fix for Python 4: replace unsafe six.PY3 with PY2

hugovk:
## Summary of changes

We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule) and work begins on Python 3.9+1.

There's some code which uses `six.PY3`:

```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code"
```

Where:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code! Instead, use `six.PY2`.

Found using https://github.com/asottile/flake8-2020

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/1959.change.rst
+++ b/changelog.d/1959.change.rst
@@ -1,0 +1,1 @@
+Fix for Python 4: replace unsafe six.PY3 with six.PY2
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -113,7 +113,7 @@ class build_ext(_build_ext):
         if fullname in self.ext_map:
             ext = self.ext_map[fullname]
             use_abi3 = (
-                six.PY3
+                not six.PY2
                 and getattr(ext, 'py_limited_api')
                 and get_abi3_suffix()
             )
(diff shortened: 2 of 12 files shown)


jaraco:
Perhaps this issue should be fixed in six rather than in every package on PyPI. I'm pretty sure PY3 was never meant to mean PY<4.

hugovk:
It's a WONTFIX in six. 

https://github.com/benjaminp/six/pull/297

Harmon758:
Note, it's almost certainly going to be 3.10. That's [the collective agreement from the Python-Dev mailing list](https://mail.python.org/archives/list/python-dev@python.org/thread/M5N3PZP4IUMKRZJKKJQA4IXK5VDX5MOP/) and that's what's being used to reference the future release in PEPs and CPython. There's even references to 3.11 and 3.12 already.

hugovk:
Thanks for the link!

webknjaz:
> Found using [asottile/flake8-2020](https://github.com/asottile/flake8-2020)

Maybe this should also be added to `tests/requirements.txt`

hugovk:
Good idea, please see PR https://github.com/pypa/setuptools/pull/1968.

PR #1961, opened 2020-01-12
Restore testing on Python 2, bypassing the requires-python check when installing Setuptools.

jaraco:
Follow-up to #1955, this change allows the tests to run on Python 2 even though the packages declares a requirement on Python 3.5 or later. This change will prevent the risk of bit-rot on the master and maintain the expectation that Python 2 compatibility is retained for those installers not honoring PEP 345.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ language: python
 jobs:
   fast_finish: true
   include:
+  - &latest_py2
+    python: 2.7
+  - <<: *latest_py2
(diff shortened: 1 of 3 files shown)


jaraco:
This technique worked for me locally. I'm not yet sure why it's not working here.

jaraco:
aha. I know why it's failing - because I'm invoking tox with `tox -e py27`, but travis is using `tox -e python` (on Python 2.7).

PR #1966, opened 2020-01-13
Improve setuptools.txt

maggyero:
This PR will make the following changes:

- fix some typos;
- remove the `python` command from `setup.py` calls;
- uniformise quotation marks.

### DIFF
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -1800,7 +1800,7 @@ to support "daily builds" or "snapshot" releases.  It is run automatically by
 the ``sdist``, ``bdist_egg``, ``develop``, and ``test`` commands in order to
 update the project's metadata, but you can also specify it explicitly in order
 to temporarily change the project's version string while executing other
-commands.  (It also generates the``.egg-info/SOURCES.txt`` manifest file, which
+commands.  (It also generates the ``.egg-info/SOURCES.txt`` manifest file, which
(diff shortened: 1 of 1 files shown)


maggyero:
Would you like to review this PR @jaraco?

jaraco:
Hey @maggyero. How about instead of adding `python setup.py` everywhere, write one section that explains (or links to an explanation on) how to execute a Python script and to avoid proliferating the otherwise obvious execution semantics? In other words, I'd like to use the minimum text/code that achieves the desired purpose.

The use of `python setup.py` only works in some scenarios, such as when `python` points to the preferred interpreter (which it often doesn't, especially in today's split Python 2/3 world). In practice, the user will need to execute `python3 setup.py` or `py -3 setup.py` or some other invocation.

Moreover, `*.py` works on Windows (invoking the default Python interpreter) and `./setup.py` works on Unix OSs, assuming the 'executable' bit is set. All of these nuances are difficult to capture in each and every execution, which is why the documentation leaves it as an exercise to the user to figure out the best way to execute the python script, and why simply `setup.py` is preferable to `python setup.py`.

How about simply removing the `python3 ` from the `setup.py invocation, and fix the other typos?

maggyero:
@jaraco

> How about simply removing the `python3 ` from the `setup.py invocation, and fix the other typos?

That makes sense. Updated.

PR #1973, opened 2020-01-18
Remove pkg_resources.py31compat.makedirs() in favor of the stdlib

jdufresne:
As setuptools is now python 3.5+, this compatibility shim is no longer
necessary.

Closes 

### Pull Request Checklist
- [ ] Changes have tests


pganssle:
@jdufresne Before you go too far in this, we are still maintaining support for Python 2, despite the change in the *declared* support (note the CI matrix).

We won't be ready to merge these sorts of PRs for some months to come.

jdufresne:
Depending on if `pkg_resources.py31compat` is considered a public interface, this could be considered a breaking change. If it should go through a deprecation cycle, please let me know. My ultimate goal is to remove `six.py` from the project, this is just a first step.

jdufresne:
> We won't be ready to merge these sorts of PRs for some months to come.

Got it! Thanks for the heads up. I'll hold off until later then.

pganssle:
@jdufresne That said, we may want to actually go ahead with deprecation warnings on all these things if it's not too much effort.

Just move all our internal uses of the compat stuff to another pointer and make public wrappers saying to migrate away from the compatibility shims, maybe?

jaraco:
I don't expect any users to be relying on these compatibility shim modules. I would expect not to have a deprecation period for these modules.

PR #1997, opened 2020-02-15
Avoid installing setuptools 45 on Python 2.

### LINKED ISSUE #1996: Tests failing on master on Python 2.7
In #1961, I implemented a workaround for allowing the Python-3-only version of setuptools to continue to be tested on Python 2, and at the time, the [tests passed](https://travis-ci.org/pypa/setuptools/builds/636029104).

However, since then, the tests [started failing](https://travis-ci.org/pypa/setuptools/builds/646690528), I suspect due to the Python 2 incompatible releases now being present and being required for setuptools' build dependencies.

jaraco:
Because Requires-Python is disabled for Python 2 to allow this declared-incompatible version to install on Python 2, the build-requirements must exclude installing setuptools 45 during the build-system setup.

Because Python 2 tests are only needed for internal validation, and because the latest pip is already installed, the build-system can rely on `backend-path = ["."]`.

Fixes #1996.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,9 @@
 [build-system]
-requires = ["setuptools >= 40.8", "wheel"]
+requires = [
+    # avoid self install on Python 2; ref #1996
+    "setuptools >= 40.8; python_version > '3'",
+    "wheel",
+]
 build-backend = "setuptools.build_meta"
 backend-path = ["."]
 


jaraco:
I note this test also illustrates that the `backend-path` is working as intended and it should be possible to remove setuptools from the `build-requires` altogether once Setuptools can drop support for sdist installs from older pips.

jaraco:
I don't understand why the tests are failing ([example](https://travis-ci.org/pypa/setuptools/jobs/650822411?utm_medium=notification&utm_source=github_status)). "no module named posixpath" is the error reported  by 3 failing tests. These tests don't fail on my local macOS runs.

jaraco:
The 3 failures [also exist on maint/44.x](https://travis-ci.org/pypa/setuptools/jobs/6508413140), indicating a regression in some aspect independent of this effort, so I've filed #1998 to track that separately.

PR #2009, opened 2020-02-27
Add mapping for long_description_content_type

RDIL:
I believe this is the right place to put this so here

## Summary of changes

Puts `long_description_content_type` in the mappings so IDEs can see it as a keyword argument

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -486,6 +486,7 @@ class ConfigMetadataHandler(ConfigHandler):
             'license_files': parse_list,
             'description': parse_file,
             'long_description': parse_file,
+            'long_description_content_type': parse_file,
             'version': self._parse_version,
             'project_urls': parse_dict,
         }


di:
This doesn't seem right, `long_description_content_type` is a string, not a file. It shouldn't need parsed.

Could you file an issue describing the problem that you're experiencing?

RDIL:
In my IDE, when typing kwargs into setuptools.setup, long description content type isn’t attempted to be autocompleted

pganssle:
@RDIL Hard to tell where that is coming from, maybe some type stubs or something?

It seems unlikely to me that any sort of static analysis is able to pick this up.

RDIL:
Okay, I’ll try Typeshed.

pganssle:
@RDIL I think you may want to maybe file a bug report with your IDE maker first? I don't see any type hints for setuptools in typeshed, so that might not be what your IDE is using.

PR #2014, opened 2020-03-07
Try building docs with python3.

jaraco:


jaraco:
While trying to troubleshoot this issue, Pradyun led me to [this setting](https://github.com/netlify/build-image/blob/2dbd444fcdce00cf06325060a8238d5ae3e86774/run-build-functions.sh#L167), where it seems that if a `runtime.txt` file exists in the root of the project, that will be used to select the Python version used for Netlify builds. It's sort-of problematic that the file is not netlify specific and can't accept any comments to explain its purpose, but if it does the trick, that'll have to do.

Issue #308, opened 2014-12-19
Add option to package build to disable normalization of version identifier


ghost:
Originally reported by: **mklein0 (Bitbucket: [mklein0](http://bitbucket.org/mklein0), GitHub: [mklein0](http://github.com/mklein0))**

---

Could an option be added to package builds to disable the use of normalized version identifiers in the final package file name?

Under previous versions of setup tools a version identifier of `20140510.003458` would generate a package name of:

```
dated_release-20140510.003458-cp27-none-macosx_10_9_intel.whl
```

After setuptools 8:

```
dated_release-20140510.3458-cp27-none-macosx_10_9_intel.whl
```

My primary concern is for older installation tool/devpy versions, the package version matching will fail if I start using the new normalized package names. And I do not want to have to test it.

Aesthetics is another aspect as in this case, the version identifier is actually an encoding of an ISO8601 timestamp.:

```
date -u +'%Y%m%d.%H%M%S'
```

This request is derived from issue #302

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/308

kdubezerra:
I also noticed that, if I set the version number as, say, 1.2.3-rc1, setuptools normalizes it to 1.2.3rc1. This created several problems for me when generating debian packages. Allowing such kinds of normalization do be disabled would be a major improvement.

jobevers:
Would also like to be able to disable normalization, 1.2.3-rc1 being normalized to 1.2.3rc1 is causing problems for packaging for me as well.

jaraco:
@dstufft, do you have any opinion on this? IIRC, it was your contribution that added the normalization. What would be the ramifications of allowing bypassing of this normalization?

dstufft:
@jaraco I think as long as you still validate that it is a PEP 440 version (or warn if not.. I don't think we have it set to hard fail), not doing the normalization shouldn't be a big deal. I think my only reason to normalize was just to get everyone using the normal forms as the default since that makes things a bit saner I think. However allowing a flag to bypass normalization should not have any negative affect.

jaraco:
Thanks @dstufft. Glad to hear this wouldn't be a blocker. @mklein0 or others, would you consider putting together a pull request?

nomadlogic:
Hi there, I was wondering if there was a status update or further discussion on this issue.  I've just run into this when our CI pipeline built a new package (which appends a short git has to the version string) was changed to a different hash.

```
Version is: 2.4.5.5513c055
[...]

PR #2026, opened 2020-03-15
Add 'sic' function to bypass version number normalization

jaraco:
## Summary of changes

As discussed in #308, this proposed change adds a simple API to bypass version number normalization, forcing the user to explicitly opt into this undesirable behavior. Now a user can specify:

```
from setuptools import setup, sic

setup(version=sic('1.0.0a'))
```

And version number normalization will be skipped.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -438,30 +438,35 @@ class Distribution(_Distribution):
                 value = default() if default else None
             setattr(self.metadata, option, value)
 
-        if isinstance(self.metadata.version, numbers.Number):
-            # Some people apparently take "version number" too literally :)
-            self.metadata.version = str(self.metadata.version)
+        self.metadata.version = self._validate_version(self.metadata.version)
+        self._finalize_requires()
 
-        if self.metadata.version is not None:
+    @staticmethod
+    def _validate_version(version):
+        if isinstance(version, numbers.Number):
+            # Some people apparently take "version number" too literally :)
+            version = str(version)
+
+        if version is not None:
             try:
-                ver = packaging.version.Version(self.metadata.version)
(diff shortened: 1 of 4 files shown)


setuptools/dist.py
```diff
@@ -438,9 +439,22 @@ def __init__(self, attrs=None):
...
             setattr(self.metadata, option, value)
 
-        self.metadata.version = self._validate_version(self.metadata.version)
+        self.metadata.version = self._normalize_version(
+            self._validate_version(self.metadata.version))
         self._finalize_requires()
 
+    @staticmethod
+    def _normalize_version(version):
+        if isinstance(version, setuptools.sic) or version is None:
```
pombredanne: What happens if `version=''` as opposed to `None` or is not a string?

jaraco: I would expect in those cases for the behavior to be unchanged with this PR.

jaraco:
> Because I like to nitpick a bit, `verbatim` may be a more common Latin word than `sic` though? ... but that's purely cosmetic. An alternative could be `exactly()` or `as_is()` ?

I do strongly prefer a single-word names for functions where possible. I think `verbatim` or `exactly` would be suitable, but as I thought about it more, I like `sic` better because (a) the fact that it's a little esoteric in source code is jarring, and that's good for this case because it is a discouraged use and (b) the word `sic` inherently carries a connotation of "abnormal", so more than just "I want it this way,", it forces the author to acknowledge "I want it this way and that's not normal."

Good suggestions though, and compelling enough to warrant serious consideration and reflection.

pombredanne:
@jaraco thank you++ :bowing_man:

PR #2037, opened 2020-03-19
A bit "cosmetic" changes though needed for implementing more advanced stuff. 

KOLANICH:
Also when an entry point now fails to be loaded, it doesn't disrupt building process.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/2037.change.rst
+++ b/changelog.d/2037.change.rst
@@ -1,0 +1,2 @@
+A bit "cosmetic" changes though needed for implementing more advanced stuff.
+* Renamed args names params for some static methods and marked them with ''@staticmethod''.
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -285,6 +285,7 @@ def check_requirements(dist, attr, value):
 
 def check_specifier(dist, attr, value):
     """Verify that value is a valid version specifier"""
+
     try:
         packaging.specifiers.SpecifierSet(value)
     except packaging.specifiers.InvalidSpecifier as error:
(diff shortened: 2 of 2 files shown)


jaraco:
Per the request in #2031, I'm re-opening this for selective consideration.

changelog.d/2037.change.rst
```diff
@@ -0,0 +1,4 @@
+A bit "cosmetic" changes though needed for implementing more advanced stuff.
+* Renamed args names params for some static methods and marked them with ''@staticmethod''.
```
jaraco: What's the purpose of this change? What value does it provide? What issue does it address?

changelog.d/2037.change.rst
```diff
@@ -0,0 +1,4 @@
...
+* Moved ''0'' into ''default_order_value" variable.
```
jaraco: I'd agree with this change to provide more context. The problem is that the purpose of the default value is already documented in the docstring, so with this change, it's now documented redundantly, opening up the opportunity for the two definitions to diverge. Especially if the code for the function remains small (< 10 lines), I find it unnecessary to make the code self-documenting here.

changelog.d/2037.change.rst
```diff
@@ -0,0 +1,4 @@
...
+* When an entry point now fails to be loaded, it doesn't disrupt building process.
```
jaraco: I can see value in failing rather than suppressing errors. One should be cautious about suppressing errors, as it might be the case that the failed operation might be critical. The warning is helpful, and given the 'fail on warning' behavior of the setuptools test suite, this warning would prevent the core functionality from being suppressed (at least for the duration of the test suite). I think this change encapsulates the main purpose of this PR and probably addresses the underlying bug (which unfortunately isn't referenced here).

setuptools/dist.py
```diff
@@ -702,30 +703,45 @@ def finalize_options(self):
...
-            return getattr(hook, 'order', 0)
+            return getattr(hook, 'order', default_order_value)
+
         eps = pkg_resources.iter_entry_points(hook_key)
+
         for ep in sorted(eps, key=by_order):
-            ep.load()(self)
+            try:
+                f = ep.load()
+            except Exception as ex:
```
jaraco: This branch probably needs a test. And it probably deserves to be its own function with a docstring describing its purpose.

changelog.d/2037.change.rst
KOLANICH: 1. Warnings in #2034 about legacy calling convention (the arg is not named `dist`)
2. it turned out that the methods are static ones, but they look like bound ones. So I have marked them as static, no make this explicit and also allow static analysis tools to capture that.

changelog.d/2037.change.rst
KOLANICH: It was done to just make [the diff](https://github.com/pypa/setuptools/pull/2038/files#diff-0f33d7d69d9f143736a252a75373261bR791) of  #2038  a bit cleaner.

setuptools/dist.py
KOLANICH: I am not sure how to properly test this. If we move that into a separate function and test only it, it will turn out we are actually testing `EntryPoint.load`. To test it properly I guess we need to mock `iter_entry_points` into a one returning our not really existing entry point.

changelog.d/2037.change.rst
KOLANICH: Yeah, I have added it as one of the ways to deal with #2030.

jaraco:
#2047 is a much more surgical way of addressing this issue, so I'll close this in favor of that change, although more important is going to be refining the problem description in #2030 to justify that a change is needed at all.

Issue #1988, opened 2020-02-09
Deprecate bdist_dumb, bdist_rpm, bdist_dmg, bdist_msi, bdist_wininst


hugovk:
Warehouse is planning on removing upload support for `bdist_dumb`, `bdist_rpm`, `bdist_dmg`, `bdist_msi`, and `bdist_wininst`:

* https://www.python.org/dev/peps/pep-0527/
* https://github.com/pypa/warehouse/issues/6792

CPython's `distutils` deprecated `bdist_wininst` in 3.8:

* https://github.com/python/cpython/pull/14553
* https://discuss.python.org/t/deprecate-bdist-wininst/1929

And is planning on removing `bdist_wininst` in 3.9:

* https://github.com/python/cpython/pull/18329
* https://discuss.python.org/t/remove-distutils-bdist-wininst-command/3115

In this latter discussion, @pganssle said:

> I think the right thing to do here would be to propose removing this [`bdist_wininst`] *in `setuptools`*, and drop the idea of removing it in `distutils`. The same goes for `bdist_rpm` and all the other weird non-wheel binary distributions.

---

What's necessary to deprecate these in `setuptools`?

PR https://github.com/pypa/setuptools/pull/1764 deprecates `bdist_egg`. A similar pattern could be followed to add warnings in `run()` in `bdist_rpm.py` and `bdist_wininst.py`. Does this sound right?

But I'm less clear on `bdist_dumb`, `bdist_dmg` and `bdist_msi`. Are these handled by `setuptools`, and if so, where?

pganssle:
Thanks for kicking off this process, Hugo!

I think we can follow the road map from [`setup.py test`](https://github.com/pypa/setuptools/pull/1878) and [`setup.py upload`/`register`](https://github.com/pypa/setuptools/pull/1898):

1. Add deprecation warnings for 9-12 months (with tests).
2. Document the deprecation.
3. During removal time, add new sub-commands that actively raise errors (rather than simply removing the commands from `setuptools`).



> But I'm less clear on `bdist_dumb`, `bdist_dmg` and `bdist_msi`. Are these handled by `setuptools`, and if so, where?

[...]

PR #2040, opened 2020-03-22
Deprecate bdist_wininst

hugovk:
## Deprecate bdist_wininst

For #1988.

Already deprecated in CPython 3.8's `distutils`.

The deprecation method, message and tests are modelled on the recent https://github.com/pypa/setuptools/pull/1764 (deprecate "eggsecutable" scripts).

I had a go at the other bdists mentioned in #1988, but had a hard time with the tests, so here's just one to be getting on with.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/2040.change.rst
+++ b/changelog.d/2040.change.rst
@@ -1,0 +1,1 @@
+Deprecated the ``bdist_wininst`` command. Binary packages should be built as wheels instead.
--- a/setuptools/command/bdist_wininst.py
+++ b/setuptools/command/bdist_wininst.py
@@ -1,4 +1,7 @@
 import distutils.command.bdist_wininst as orig
+import warnings
+
+from setuptools import SetuptoolsDeprecationWarning
 
 
 class bdist_wininst(orig.bdist_wininst):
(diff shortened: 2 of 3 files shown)


pganssle:
@hugovk I made a change to the news fragment and I've changed the test in two ways:

1. Flattened it so that it is no longer part of a class (I've noticed we don't usually do this in other test suites, and this one seems particularly unnecessary).
2. Added the assertion that the original `bdist_wininst.run` command is invoked as a sanity check.

Let me know if you disagree with these changes, otherwise I'll go ahead and squash my fixup commits into your commits and merge.

hugovk:
Thanks, both look good!

Do you use some tool for fixing up (I noticed the "fixup! XXX" comments), or just vanilla Git tools?

pganssle:
> Do you use some tool for fixing up (I noticed the "fixup! XXX" comments), or just vanilla Git?

This is vanilla git, you can use `git commit --fixup` to mark a commit as a fixup for another commit, and then `git rebase --autosquash` to automatically squash the fixups into the relevant commits. Here's an article I found on it: https://thoughtbot.com/blog/autosquashing-git-commits

There is a fancy tool that I've been starting to use lately (I got hooked on it when using `hg` at work), called `git-absorb`: https://github.com/tummychow/git-absorb

This will automatically determine what parts of whatever you've staged belongs with what commits and create fixup commits for it. Very handy for my workflow.

Issue #2041, opened 2020-03-22
Packaged script execute permissions lost from v46.1.0 onwards


dhallam:
## Issue

We use pipenv and, as a result, are forced to use the latest version of setuptools as it internally pins to the latest version.

We observed that executable scripts that are part of python packages have lost their execute flag from setuptools v46.1.0 onwards.  The example to demonstrate this bug uses pyspark which includes a number of executable scripts in its package.

The issue was introduced by commit https://github.com/pypa/setuptools/commit/7843688bc33dd4e13e10130bc49da4c290fe7d7f where the `copy_file()` function is now called with `preserve_mode=False`. The change log states the reason for the change as:

> Prevent keeping files mode for package_data build. It may break a build if user's package data has read only flag.

Unfortunately, this has the side effect of stripping all execute permissions from files, meaning users can't use the scripts "out of the box" - they have to set the execute permissions manually.

## Demonstration Script

```
#!/bin/bash
set -eu

wget -nc https://files.pythonhosted.org/packages/9a/5a/271c416c1c2185b6cb0151b29a91fff6fcaed80173c8584ff6d20e46b465/pyspark-2.4.5.tar.gz

for version in "46.0.0" "46.1.0"; do
    rm -rf .venv pyspark-2.4.5
    tar xzf pyspark-2.4.5.tar.gz
    virtualenv -q -p /usr/bin/python3.7 .venv
    . .venv/bin/activate
    python3 -m pip install --upgrade setuptools="==${version}" wheel
    pushd pyspark-2.4.5
    python3 setup.py -q bdist_wheel
    pushd dist
    unzip -q pyspark-2.4.5-py2.py3-none-any.whl
    echo -e "\n\n${version}: Here are the permissions for spark-submit:\n"
    ls -l ./pyspark/bin/spark-submit
    echo -e "\n\n"
    popd
    popd
done
```

## Expected result

```
-rwxr-xr-x 1 dave dave 1040 Feb  2 19:35 ./pyspark/bin/spark-submit
```

## Actual Result

```
-rw-rw-r-- 1 dave dave 1040 Feb  2 19:35 ./pyspark/bin/spark-submit
```

tibdex:
I'm facing this issue too on macOS 10.15.3 and Pipenv 2018.11.26. This code was working fine in 46.0.0  but started failing in 46.1.0:

```python
from pyspark.sql import SparkSession # pyspark==2.4.5

[...]

PR #2043, opened 2020-03-24
Undo preserve_mode=False in order to preserve execute permission

leekillough:
Fixes [#2041](https://github.com/pypa/setuptools/issues/2041)

@dhallam @tibdex @tgaddair @mattclay @pselden @jaraco @KOLANICH

### DIFF
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -120,8 +120,7 @@ class build_py(orig.build_py, Mixin2to3):
                 target = os.path.join(build_dir, filename)
                 self.mkpath(os.path.dirname(target))
                 srcfile = os.path.join(src_dir, filename)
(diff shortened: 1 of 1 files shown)


jorikdima:
I suggest not to undo, but make this fix better. I am working on the PR

jaraco:
Thanks jorikdima. As is revealed by the Appveyor tests, this change causes a test to fail on Windows, a test we put in place to protect the requirement that inspired the change. I would similarly like to see a test to capture the (untested) expectation that this PR is attempting to restore.

leekillough:
I'm fine with a "correct" fix -- this was just a short-term stop-gap solution.

Let's fix the original issue(s), as well as the regression caused by the lack of execution bits being preserved.

PR #2044, opened 2020-03-25
Preserve file modes during pkg files copying, but clear read only flag for target afterwards

### LINKED ISSUE #2041: Packaged script execute permissions lost from v46.1.0 onwards

jorikdima:
## Summary of changes

This PR is supposed to fix a regression caused by PR #1607 and initially explained in the issue #1424
Originally the issue #1424 was fixed by not preserving flags for package data, but this caused issues in many other projects described e.g. in #2043 
The new approach is that we again keep file flags as is, but set write permission flag for the target file afterwards. This should not clear eXecutable bit, but allow file deleting.
An updated test also sets eX bit for package file and checks that target file still has it. The test fails for current version of setuptools and passes for the proposed PR.
Closes #2041 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/2041.bugfix.rst
+++ b/changelog.d/2041.bugfix.rst
@@ -1,0 +1,1 @@
+Preserve file modes during pkg files copying, but clear read only flag for target afterwards.
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -7,6 +7,7 @@ import textwrap
 import io
 import distutils.errors
 import itertools
+import stat
 
 from setuptools.extern import six
 from setuptools.extern.six.moves import map, filter, filterfalse
@@ -120,8 +121,10 @@ class build_py(orig.build_py, Mixin2to3):
                 target = os.path.join(build_dir, filename)
                 self.mkpath(os.path.dirname(target))
                 srcfile = os.path.join(src_dir, filename)
-                outf, copied = self.copy_file(
-                    srcfile, target, preserve_mode=False)
+                outf, copied = self.copy_file(srcfile, target)
+                cur_mode = os.stat(target)[stat.ST_MODE]
+                target_mode = cur_mode | stat.S_IWRITE
+                os.chmod(target, target_mode)
                 srcfile = os.path.abspath(srcfile)
                 if (copied and
                         srcfile in self.distribution.convert_2to3_doctests):
--- a/setuptools/tests/test_build_py.py
+++ b/setuptools/tests/test_build_py.py
@@ -43,7 +43,11 @@ def test_read_only(tmpdir_cwd):
     open('pkg/__init__.py', 'w').close()
     open('pkg/data.dat', 'w').close()
     os.chmod('pkg/__init__.py', stat.S_IREAD)
-    os.chmod('pkg/data.dat', stat.S_IREAD)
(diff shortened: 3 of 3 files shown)


leekillough:
LGTM

jorikdima:
One comment about tests. Currently I updated read_only test which have been prepared by @jaraco . That script tests the initial solution. But it may work (really test the solution) only on Windows platform, since on Linux there is no such issue with read only files at all. My add on checks eX flags, which do not exist on Windows but on Linux only. My test needs to be checked on Windows, I don't have access to Win platform to run tox.
I think the best solution (I am talking about tests only) would be to split the test between testing read only flags and eX flags (two test functions). One test should be run only on Win, the other one only on Linux.
If that makes sense, let me know, please. I am too inexperienced in setuptools to make such decisions.

jaraco:

> One test should be run only on Win, the other one only on Linux.

I do aim to test behavior as uniformally as possible. So for example, the previous test would run on both Windows and Linux, and even though it only failed on Windows, it still asserted the expectation on all platforms uniformally.

In some cases, it's not possible to assert a behavior across all platforms, and in that case, it's okay to either skip or xfail a test on that platform (skip if it has no value, xfail if monitoring the failed behavior is more appropriate).

I can see the tests are failing on Windows, so I can take it from here. Thanks @jorikdima.

PR #2047, opened 2020-03-27
When an entry point now fails to be loaded, it doesn't disrupt building process.

KOLANICH:
## Summary of changes
Splitted from #2037

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/distr_opts_improvements_graceful_load_fallback.change.rst
+++ b/changelog.d/distr_opts_improvements_graceful_load_fallback.change.rst
@@ -1,0 +1,2 @@
+* When an entry point now fails to be loaded, it doesn't disrupt building process.
+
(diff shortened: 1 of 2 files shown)


changelog.d/distr_opts_improvements_graceful_load_fallback.change.rst
```diff
@@ -0,0 +1,2 @@
+* When an entry point now fails to be loaded, it doesn't disrupt building process.
```
jaraco, line 1: This changelog entry needs work. First, the filename shouldn't have any text in it, only an issue number and scope (i.e. "change"). Second, the text references "an entry point", but doesn't talk about the context. This change only affects the loading of `setuptools.finalize_distribution_options`.

jaraco:
It would be nice if the description above led me to the problem description that this issue tries to solve, rather than referencing another pull request.

The problem description is in #2030, and there are still outstanding questions about the problem that this solves. Therefore, I'm going to close this until the problem description can be refined a bit more.

Thanks for the continued effort.

PR #2061, opened 2020-04-10
Use Visual Studio setup configuration API instead of vswhere.exe

### LINKED ISSUE #2028: msvc.py fails with VS 2019 Build Tools 14.24
Hi,

I installed VS 2019 with Build Tools 14.24 but my "setup.py build" still fails with the below error in msvc.py file:
error: Microsoft Visual C++ 14.1 is required. Get it with "Build Tools for Visual Studio": https://visualstudio.microsoft.com/downloads/

And AFAIK, the msvc.py should support Build Tools 14.24
What I'm missing? anyone could help?

svanheulen:
## Summary of changes

The current version of `_msvc14_find_vc2017` uses the `vswhere.exe` utility to locate an installation of the Visual C++ compiler, but not all types of Visual Studio installs include that utility. For example, Visual Studio 2019 C++ Build Tools does not include it.

The `vswhere.exe` utility uses the [Visual Studio setup configuration API](https://github.com/microsoft/vs-setup-samples) to locate installs of Visual Studio 2017 and above. These changes use that API directly instead of relying on the `vswhere.exe` utility.

Since it uses the same API, it should work exactly as the current version, except that it will also work in cases where the `vswhere.exe` utility is not available on the system. But the changes do add quite a bit of code since it's accessing a COM interface using `ctypes`.

Since I don't own any version of Windows, my only way to test has been in Wine. Also due to bugs in Wine, running `vcvarsall.bat` causes a page fault... so I really haven't been able to do any extensive testing. What I *have* been able to test is that my version of `_msvc14_find_vc2017` does return the correct path with Visual Studio 2019 C++ Build Tools installed.

Closes #2028

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -285,10 +285,11 @@ def _vs_setup_config_find_vc():
         https://github.com/microsoft/vs-setup-samples
     """
     try:
-        ole32 = ctypes.oledll.ole32
+        ole32 = ctypes.windll.ole32
+        ole32_hr = ctypes.oledll.ole32
 
         # initialize the COM library
-        ole32.CoInitialize(None)
+        ole32_hr.CoInitializeEx(None, 0)  # COINIT_MULTITHREADED
     except OSError:
         return None, None
 
@@ -297,7 +298,7 @@ def _vs_setup_config_find_vc():
     try:
         # create an instance of the ISetupConfiguration interface
         setup_configuration = ctypes.c_void_p()
-        ole32.CoCreateInstance(
+        ole32_hr.CoCreateInstance(
             ctypes.byref(CLSID_SetupConfiguration),
             None,
             1,  # CLSCTX_INPROC_SERVER
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -182,13 +182,14 @@ def _vs_setup_config_check_vc(packages):
     """
 
     try:
-        oleaut32 = ctypes.oledll.oleaut32
+        oleaut32 = ctypes.windll.oleaut32
+        oleaut32_hr = ctypes.oledll.oleaut32
 
         # lock the SafeArray
-        oleaut32.SafeArrayLock(packages)
+        oleaut32_hr.SafeArrayLock(packages)
     except OSError:
         # destroy the SafeArray and release all elements
-        oleaut32.SafeArrayDestroy(packages)
+        oleaut32_hr.SafeArrayDestroy(packages)
 
         return None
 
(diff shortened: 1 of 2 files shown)


setuptools/msvc.py
```diff
@@ -173,39 +174,247 @@ def _msvc14_find_vc2015():
...
-    path = join(path, "VC", "Auxiliary", "Build")
-    if isdir(path):
-        return 15, path
+        setup_instance = ctypes.c_void_p()
+        setup_instance2 = ctypes.c_void_p()
+        installation_path = ctypes.c_wchar_p()
+        packages = ctypes.c_void_p()
+
+        try:
+            while _next(enum_setup_instances, 1, ctypes.byref(setup_instance), None) == 0: # S_OK
```
svanheulen, line 362: I wasn't sure how do wrap this line to 79 characters without making it look like a mess. Any suggestions?

jaraco:
This is amazing. Thanks for doing all this work.

@zooba Would you mind giving it a review?

@svanheulen Did you consider using [comtypes](https://pypi.org/project/comtypes/)?

jaraco:
> `@zooba` Would you mind giving it a review?

In particular, I'm interested in knowing if the approach is a sound one (using the API instead of vswhere).

svanheulen:
@jaraco Hello and thanks for the reply :)
Yeah, I did run across the comtypes module when doing research but I assumed it was best to avoid third party modules. I don't mind adapting it to use comtypes though, if that's preferable.

jaraco:
If using comtypes handles most of the array allocation deallocation and traversal, I'd say it's worth it. Otherwise, probably not.

Having a third-party, pure-python module is easy for setuptools using the vendoring technique. If you experiment with it and think it's a likely outcome, let me know if I can help with vendoring comtypes, but just use a regular dependency for now.

svanheulen:
I did some testing with comtypes.

Unfortunately the COM object we're using doesn't have the required metadata for comtypes to automatically generate the interface classes. They can be manually defined but of course there was no information about how to do that in the documentation. Although I was able to figure it out for the most part, after digging through the code.

The only thing I can't figure out is how to get a SafeArray as an output parameter. Again there's no documentation and the code dealing with SafeArrays is very convoluted so I'm having a hard time wrapping my head around it.

It seems like the code could be a lot more concise using comtypes if I can figure out the SafeArrays so I'll keep trying.

svanheulen:
@jaraco I figured out my issue and have everything working with comtypes. Should I make a separate pull request so you can decide which version you prefer? Or should I just make the changes here?

zooba:
Honestly, I don't like this PR. And it's mostly because I don't like ctypes (_especially_ for COM), and the rest because I don't like the setup API.

I originally used the setup API in [my own package](https://pypi.org/project/pyfindvs/) and then in CPython 3.7 (in native modules), and then switched to vswhere.exe because it's just easier to deal with. I was _not_ aware that it's missing from Build Tools, and tbh that doesn't even make sense because it shouldn't even be an optional component (I'm familiar with how VS Setup is architected).

I've pinged some of the setup team to find out why it may be missing. Ideally, we'd just get it put back in VS and keep the simpler code.

You could also bundle it - that tool is MIT licensed and available at https://github.com/microsoft/vswhere (including pre-built releases).

jaraco:
Wow. So many options. 

Thanks @svanheulen for putting together a comtypes implementation. If it's not too much trouble to put it together as a separate PR, that'd be great so that it'll be easy to compare the two solutions for posterity. Good idea.

Given the response from zooba, I'll wait to hear back before moving forward with a ctypes-based design. I don't hold the same reservations that Steve does about ctypes/comtypes. In my experience, the usage is reliable and fast and usually stable. Still, I'll take his reservations under advisement.

zooba:
So the VS setup team agree that vswhere should always be there, and indeed I just started setup again (haven't even installed anything) and there it is:

image

For reference, I got the download from https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019

This has been the design since vswhere.exe was added, so if it's missing on some machines then it probably indicates corruption in their install. It is certainly not missing by design. Perhaps in those cases, it would be better to raise a more informative error than to fall back on unsuitable behaviour?

zooba:
> In my experience, the usage is reliable and fast and usually stable.

Careful, I'll start tagging you on ctypes bugs on the tracker ;)

svanheulen:
@zooba Ah, sorry, I made a bad assumption then. I found issue #2028 that sounded like my same problem, using the same version of Build Tools. But now I'm guessing that's a different problem and the reason my install is missing `vswhere.exe` is probably because of how I installed it.

The way I understand it, the installer you get from the website is like an installer for the installer... ? But that thing doesn't work in Wine, so I have to manually download the *actual* installer and run that instead.

```
wget https://aka.ms/vs/16/release/installer -O vs_installer.opc
unzip vs_installer.opc
wine ./Contents/vs_installer.exe --channelId VisualStudio.16.Release --channelUri https://aka.ms/vs/16/release/channel --productId Microsoft.VisualStudio.Product.BuildTools
```

Anyway, thank you for reviewing my code and looking into the issue.

Also @jaraco if you're still interested in the comtypes version just for fun, I have that in this branch: https://github.com/svanheulen/setuptools/tree/vc2017-detection-comtypes

PR #2071, opened 2020-04-15
Use importlib instead of imp in __bootstrap__ functions

alexhenrie:
When the PYTHONWARNINGS environment variable is set, I often see the message `DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses` in conjuction with bootstrap files for compiled modules. This pull request fixes the code that triggers that warning.

### DIFF
--- a/setuptools/command/bdist_egg.py
+++ b/setuptools/command/bdist_egg.py
@@ -55,10 +55,11 @@ def write_stub(resource, pyfile):
     _stub_template = textwrap.dedent("""
         def __bootstrap__():
             global __bootstrap__, __loader__, __file__
-            import sys, pkg_resources, imp
+            import sys, pkg_resources
+            from importlib.machinery import ExtensionFileLoader
             __file__ = pkg_resources.resource_filename(__name__, %r)
             __loader__ = None; del __bootstrap__, __loader__
-            imp.load_dynamic(__name__,__file__)
+            ExtensionFileLoader(__name__,__file__).exec_module()
         __bootstrap__()
         """).lstrip()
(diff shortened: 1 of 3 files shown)


setuptools/command/bdist_egg.py
```diff
@@ -55,10 +55,11 @@ def write_stub(resource, pyfile):
     _stub_template = textwrap.dedent("""
         def __bootstrap__():
             global __bootstrap__, __loader__, __file__
-            import sys, pkg_resources, imp
+            import sys, pkg_resources
+            from importlib.machinery import ExtensionFileLoader
```
jaraco, line 59: `importlib.machinery` is not available on Python 2.7, and although this package has nominally dropped support for Python 2.7, it still maintains compatibility with it. This change will need to be deferred until Python 2.7 support is fully dropped in the codebase... and even then, I'm not sure if this might need to be retained to support building packages that target 2.7.

jaraco: Can we say that it's impossible to invoke `build_ext` on Python 3 to produce artifacts (builds) that are run on Python 2? I think that's the case, but I'm unsure. If we can say, now that the setuptools codebase no longer supports Python 2, that this code will never be run on Python 2, then I'm inclined to accept this change.

jaraco: It seems unlikely to me that this code could appear on Python 2. Let's get it out and see what happens. Worst case, we have to revert it.

alexhenrie:
@jaraco Changelog entry added.

PR #2078, opened 2020-04-28
WIP prototype pep517-style editable

dholth:
See https://discuss.python.org/t/third-try-on-editable-installs/3986/22

A new editable hook does an in-place build, returns `{ "src_root" : "where .pth file should point to" }`

Apologies for the reformatting.


pganssle:
Closing, since this is not intended to be merged, and isn't really consistent with the approach we agreed upon at the packaging summit in 2019 anyway.

PR #2090, opened 2020-05-06
improve pkg_resources.Distribution.hashcmp performance

### LINKED ISSUE #2089: pkg_resources.Distribution.hashcmp is slow
As it removes md5 fragment on the fly, `pkg_resources.Distribution.hashcmp` is slow:

    @property
    def hashcmp(self):
        return (
            self.parsed_version,
            self.precedence,
            self.key,
            _remove_md5_fragment(self.location),
            self.py_version or '',
            self.platform or '',
        )

Would there be side effects if `location` is changed into a property where the setter would not only store the location but also store the location without fragment ?

gotcha:
by storing location without md5.

Turns `Distribution.location` into a property to make `hashcmp` more performant.

Closes #2089

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2613,14 +2613,14 @@ class Distribution:
             self.platform or '',
         )
 
-    def get_location(self):
+    @property
+    def location(self):
         return self._location
 
-    def set_location(self, l):
-        self._location = l
-        self._location_without_md5 = _remove_md5_fragment(l)
-
-    location = property(get_location, set_location)
+    @location.setter
+    def location(self, location):
+        self._location = location
+        self._location_without_md5 = _remove_md5_fragment(location)
 
     def __hash__(self):
         return hash(self.hashcmp)
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2608,10 +2608,19 @@ class Distribution:
             self.parsed_version,
             self.precedence,
             self.key,
-            _remove_md5_fragment(self.location),
+            self._location_without_md5,
             self.py_version or '',
             self.platform or '',
         )
+
+    def get_location(self):
+        return self._location
+
+    def set_location(self, l):
+        self._location = l
+        self._location_without_md5 = _remove_md5_fragment(l)
+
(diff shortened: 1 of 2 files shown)


gotcha:
As the fix is not supposed to change any semantics, I wonder which sort of tests should be added.

pkg_resources/__init__.py
```diff
@@ -2608,11 +2608,20 @@ def hashcmp(self):
...
         )
 
+    def get_location(self):
+        return self._location
+
+    def set_location(self, l):
+        self._location = l
+        self._location_without_md5 = _remove_md5_fragment(l)
+
+    location = property(get_location, set_location)
```
McSinyx, line 2623: I wonder if this is to provide some sort of compatibility, or the following is actually OK:
```suggestion
    @property
    def location(self):
        return self._location

    @location.setter
    def location(self, location):
        self._location = location
        self._location_without_md5 = _remove_md5_fragment(location)
```

gotcha: Let's test it as well. I do not have a preference.

jaraco:
Although this approach achieves the desired result (only calculate _location_without_md5 once), it causes quite a bit of entanglement with the `location` property, unnecessarily adding a new `._location` property. How about instead having `_location_without_md5` be a cached property or method to avoid it being calculated more than once for any given Distribution?

gotcha:
I make `location` a property as there is afaik nothing that avoids setting `location` twice.

jaraco:
That's fair, and I had the same concern.

After inspecting the code, it appears it's only ever set once during the constructor. I'd hate to add code that explicitly adds support for setting the value later when it will only ever be set once during construction.

I'll go with #2098 until a use-case demands a setter. Thanks for the contrib!

jaraco:
I think more important than before accepting either PR will be to define a problem that needs to be solved, so I'll await an answer in #2089

jaraco:
I prefer #2108. Thanks again for the contrib.

gotcha:
It is always better to remove code when possible.

Issue #2086, opened 2020-05-03
lib2to3 will be removed in Python 3.10


hroncok:
See also https://github.com/pypa/setuptools/issues/2081

The standard library module lib2to3 is deprecated and will be removed in Python 3.10. 

https://bugs.python.org/issue40360

lib2to3 will likely become an external library available from PyPI, but my guess is that setuptools don't want to have more dependencies.

hroncok:
Here is what I think might be a path forward:

In the next setuptools release, using [use_2to3](https://setuptools.readthedocs.io/en/latest/python3.html) will become deprecated as well (with no set removal date). On Python 3.10, setuptools will have optional dependency (e.g. via `setuptools[use_2to3]`) on the 3rd party lib2to3 module to keep supporting installing old packages.

jaraco:
> On Python 3.10, setuptools will have optional dependency (e.g. via `setuptools[use_2to3]`) on the 3rd party lib2to3 module to keep supporting installing old packages.

I'd prefer simply to drop support for building with 2to3 support and to require those old packages to be built on old versions of setuptools (or not at all). Regardless, let's get the deprecation in ASAP as possible to signal to users to stop using it.

graingert:
I'm getting build failures because of this lib2to3 use https://github.com/twisted/ldaptor/runs/1205572060#step:12:103 so I'm in favour of it being removed.

note, you could use `fissix` instead of lib2to3, if you still wanted to support on the fly translation

graingert:
lib2to3 is still in Python3.10 as of `Python 3.10.0b4 (default, Jul 12 2021, 22:16:23) [GCC 9.3.0] on linux`

PR #2096, opened 2020-05-10
Emit deprecation warning when 2to3 is used.

jaraco:
As discussed in #2086, and as planned for years, deprecate the 2to3 feature. Instead, packagers are encouraged to write single-source code or build a tool chain outside of setuptools to perform conversions.

### DIFF
--- a/setuptools/lib2to3_ex.py
+++ b/setuptools/lib2to3_ex.py
@@ -13,6 +13,7 @@ from distutils import log
 from lib2to3.refactor import RefactoringTool, get_fixers_from_package
 
 import setuptools
+from ._deprecation_warning import SetuptoolsDeprecationWarning
 
 
 class DistutilsRefactoringTool(RefactoringTool):
@@ -40,7 +41,7 @@ class Mixin2to3(_Mixin2to3):
             "requires Python 2 support, please migrate to "
             "a single-codebase solution or employ an "
             "independent conversion process.",
-            DeprecationWarning)
+            SetuptoolsDeprecationWarning)
         log.info("Fixing " + " ".join(files))
         self.__build_fixer_names()
         self.__exclude_fixers()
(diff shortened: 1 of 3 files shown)


setuptools/lib2to3_ex.py
```diff
@@ -33,6 +34,12 @@ def run_2to3(self, files, doctests=False):
             return
         if not files:
             return
+
+        warnings.warn(
+            "2to3 support is deprecated. Please migrate to "
+            "a single-codebase solution or roll your own "
+            "conversion process.",
```
pganssle, line 41: Possibly "if you still need to support a python 2"?

jaraco: Done.

setuptools/lib2to3_ex.py
```diff
@@ -33,6 +34,13 @@ def run_2to3(self, files, doctests=False):
             return
         if not files:
             return
+
+        warnings.warn(
+            "2to3 support is deprecated. If the project still "
+            "requires Python 2 support, please migrate to "
+            "a single-codebase solution or employ an "
+            "independent conversion process.",
+            DeprecationWarning)
```
pganssle, line 43: Oh, actually, just noticed - don't we have our own deprecation warning that is visible by default?

I don't think too many people will see this.

pganssle: This one: https://github.com/pypa/setuptools/blob/master/setuptools/_deprecation_warning.py

jaraco: Good point - unless it's caught through a test runner than enables these warnings (or fails on them), this could be missed.

jaraco: Fixed in a354d7bc.

hroncok:
Should more test be decorated? Like `test_2to3_user_mode`.

Issue #2093, opened 2020-05-09
Doc: thoughts on documentation overhaul?


alvyjudy:
Hi,

I feel currently the setuptools documentation is quite cluttered and it makes it hard to improve the documentation, as one had to think really hard where to put the new documentation pieces (e.g. the 8 open issues that has stalled). 

Take the index for example, the most commonly used features are all clustered into one section, while the relatively trivial ones take up like 4/5 of the top entries. Also, the main section still mentions a lot about python eggs, which I suppose is considered deprecated? The order of each section is less than ideal and the actual content seems to assume a big deal of knowledge from distutils. It would be nicer if they could be organized in such fasion: 1) quick start, 2) user guide, 3) reference, 4) deprecated/discouraged. 

Especially now with PEP 517 being pushed forward, it might be a good idea to overhaul the documentation to 1) encourage the adaptation of this new standard and 2) inform developers of the new role setuptools will be playing in this new landscape and most importantly 3) make it more reader friendly.

I would like to take on this challenge and improve the documentation if that's desired:)

jaraco:
@alvyjudy I'd be delighted if you could help with that. Thanks for volunteering!

In order to not waste your time, I'd like to provide feedback incrementally as you work on it. I'd hate for you to overhaul the documentation only to have to revert half of it based on an understanding that I can't convey quickly. To that end, I recommend one of two approaches. Either,

- Break down the effort into small, incremental steps, editing only a few sections at a time, or
- Migrate most/all of the existing documentation into a "legacy" section and start from scratch in the index with a link to the legacy docs.

Regardless how you approach it, rest assured that we're eager to improve the documentation and your desire to help there warms my heart. I'll do everything I can to help keep you from being stalled on the effort. Feel free to ping me directly if you need attention on something.

alvyjudy:
Hi @jaraco , thanks so much for your support I really appreciate it!

Regarding the approach, I would prefer to update the documentation incrementally considering that most of the content remain relevant, given how ubiquitous setuptools is and how many libraries still depend on it. Re-writing it from scratch and push everything to a legacy section might actually upset readers. 

I think the first step is to reorganize the content order and index, with minimal editing, to make it more reader-friendly so that we can find a better angle to paint the PEP 517 landscape down the road.

I will open a new branch and PR and start editing the docs slowly. This whole process might take a while so for the feedback please take your time. I'll also adjust my own pace depending on how fast the review is taking place.

jaraco:
[...]

PR #2097, opened 2020-05-10
doc overhaul step 1: simplify index and group deprecated files

alvyjudy:
This is the first step on #2093 . Quite a few links might be broken for the time being but I would like to make sure I'm on the right track before moving forward

## Summary of changes
1) simplified top index into three entires: user guide, development guide, and deprecated (depth is now 1, before I can clean up the entries for sub index)
2) renamed doc for python eggs (formats.txt --> python_egg.txt)
3) move python_eggs, easy_intall, python2to3 docs into dedicated folder (deprecated)
4) create index inside "deprecated" folder and reference it in the top index

### Pull Request Checklist
- [ ] Changes have tests


alvyjudy:
If possible, I would like to have this PR merged into a separate branch rather than master because it is an ongoing process and might (and already) break some links until it's completed:)

docs/index.txt
```diff
@@ -14,12 +14,8 @@ designed to facilitate packaging Python projects, where packaging includes:
...
-   :maxdepth: 2
+   :maxdepth: 1
 
-   setuptools
-   pkg_resources
-   python3
-   development
-   roadmap
-   Deprecated: Easy Install <easy_install>
-   history
```
webknjaz, line 25: Was the removal of the changelog deliberate?

jaraco: No. The changelog should be retained, and it perhaps should even have a forwarding link so historical links can still reach it.

webknjaz: FWIW I've sent a few recovery PRs like https://github.com/pypa/setuptools/pull/2425

PR #2098, opened 2020-05-10
Cache the _location_without_md5.

### LINKED ISSUE #2089: pkg_resources.Distribution.hashcmp is slow

jaraco:
Fixes #2089. Alternate to #2090.

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -77,6 +77,7 @@ except ImportError:
     importlib_machinery = None
 
 from . import py31compat
+from .py37compat import cached_property
 from pkg_resources.extern import appdirs
 from pkg_resources.extern import packaging
 __import__('pkg_resources.extern.packaging.version')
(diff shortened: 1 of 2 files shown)


gotcha:
This relies on the fact that `location` should be set only once. Is there anything that avoids to set it twice ?
This is why #2090 makes `location` a property.

jaraco:
There's nothing that prevents it from being set again, but there's also no reason to think that it will be set again.

gotcha:
@jaraco While reading the code to answer to your question on #2089, I notice the following:

- `hashcmp` is mainly slow because of `_remove_md5_fragment` but it uses a lot of properties
- `key` and `parsed_version` are also cached, with two very simple but differing implementations, `key` uses `try except AttributeError`; `parsed_version` uses `hasattr`

This leads me to the following questions:
- Should the full `hashcmp` be cached ? I tend to answer yes because it already relies on other cached properties.
- Could it be cached simply, like `key` or `parsed_version` ? I also tend to answer positively.
- Which implementation should we choose ? `hasattr` or `try except` ? I am not deep enough in Python internals to know which is quicker.
- Is there a reason to use a different implementation of caching for `key`and `parsed_version` ? Should we unify on the quickest ?

jaraco:
I prefer #2108. I'll answer the questions in #2089.

PR #2134, opened 2020-05-24
Crash under Python 2

### LINKED ISSUE #2094: Removing compatibility for Python 2
In #1458 and the Setuptools 45 release, this project dropped declared support for Python 2, adding a warning when a late version was invoked on Python 2. This warning helped address many of the systemic uses of Setuptools 45+ on Python 2, but there continue to be users (presumably) reporting that they've [encountered the warning](https://github.com/pypa/setuptools/issues?q=is%3Aissue+in%3Atitle+%22incompatible+install%22+).

I say presumably because most of them have submitted a blank template without providing any information.

Since March, these users have been directed to the template via bit.ly, so I have metrics on the number of users encountering and following the link.

image

It seems there have been 50-100 clicks per day since Apr 11. I'm guessing bit.ly doesn't give me data older than 30 days.

To put that in perspective, Setuptools received over 45M downloads in the last month, so the number of people that followed that link (3.3k) is 0.007% of the downloads.

Still, that's upwards of 100 people per day whose workflow would be broken until they could fix their environment.

Let's also consider that each of these users encountering this issue are following discouraged if not deprecated workflows and are creating new or updated environments (new since setuptools 45 was released in January).

It seems to me we have two options - support Python 2 until the incidents of users encountering this error message reduces to a trickle (what is that threshold) or bite the bullet and drop support for Python 2.

I'd like to review the outstanding issues relating to this issue, but my inclination is to move forward with dropping support.

jaraco:
This change fully drops support for Python 2 on the master branch, causing an explicit crash if it happens to be installed and executed on Python 2.

Adds new documentation (to which the bit.ly link will point), giving users the best guidance we have today (and in the future) and ultimately leading to the issue template.

There's a part of me that's tempted to release this not as a backward-incompatible release, as the backward-incompatible release was 45.0 and this change is only an enforcement of that incompatibility that's been guarded by a warning since the 45 release.

Fixes #2094.


jaraco:
@McSinyx I'd be interested in your feedback as well.

.github/ISSUE_TEMPLATE/setuptools-warns-about-python-2-incompatibility.md
```diff
@@ -28,6 +28,7 @@ Your first course of action should be to reason about how you managed to get an
 <!-- These are the recommended workarounds for the issue. Please
 try them first. -->
 
+- [ ] Read [Python 2 Sunset docs](https://setuptools.readthedocs.io/en/latest/python%202%20sunset.html).
```
McSinyx, line 31: I think the new doc filename should use hyphen instead of spaces to avoid URL escaping here.

docs/python 2 sunset.txt
```diff
@@ -0,0 +1,35 @@
+:orphan:
+
+Python 2 Sunset
+===============
+
+Since January 2020 and the release of Setuptools 45, Python 2 is no longer supported by the most current release (`discussion <https://github.com/pypa/setuptools/issues/1458>`_). Setuptools as a project continues to support Python 2 with bugfixes and important features on Setuptools 44.x.
```
McSinyx, line 6: I've just checked some other doc files and it seems they wrap lines.  BTW, may I ask why setuptools use `txt` extension for docs for reST?

docs/python 2 sunset.txt
```diff
@@ -0,0 +1,35 @@
...
+It's still not working
+----------------------
+
+If after trying the above steps, the Python environment still has incompatible versions of Setuptools installed, here are some things to try.
+
+1. Uninstall and reinstall Setuptools. Run ``pip uninstall -y setuptools`` for the relevant environment. Repeat until there is no Setuptools installed. Then ``pip install setuptools``.
+
+2. If possible, attempt to replicate the problem in a second environment (virtual machine, friend's computer, etc). If the issue is isolated to just one unique enviornment, first determine what is different about those environments (or reinstall/reset the failing one to defaults).
+
+3. End users who are not themselves the maintainers for the package they are trying to install should contact the support channels for the relevant application. Please be respectful by searching for existing issues and following the latest guidance before reaching out for support. When filing an issue, be sure to give as much detail as possible to help the maintainers understand what factors led to the issue after following their recommended guidance.
```
McSinyx, line 31: English is not my native language, but *please be respectful* somehow sounds a bit odd here.  In addition, I think enumerate list elements should not be separated, i.e.
```rst
1. Uninstall and reinstall Setuptools...
2. If possible, attempt to replicate the problem...
```

.github/ISSUE_TEMPLATE/setuptools-warns-about-python-2-incompatibility.md
jaraco: Acknowledged, but I prefer the spaces, especially as it appears nicely for users:

image

Too many systems avoid spaces in filenames, even though they're well supported in almost all environments. I plan to continue to break the convention of replacing spaces with some other character, as spaces have their purpose.

docs/python 2 sunset.txt
jaraco: Happy for someone to wrap the lines. I don't have a tool that does it for me automatically and I don't wish to invest the time to do it myself. Happy to accept a follow-on PR if someone wishes to expend the effort.

docs/python 2 sunset.txt
jaraco: The reason I say "Please be respectful" is I'm directing users at other maintainers, and I want to be cautious not to create a deluge of support requests to other projects. Perhaps it would help to say "Please be respectful to those projects" or "Please be considerate".

> list elements should not be separated

I separated them because it makes it easier for me to parse visually when in text form. I don't think it makes a difference for rendering. If it's just a matter of opinion, I'll stick with what's there. If there's guidance that not having the intervening lines is better, I'm happy to align with best practices.

jaraco:
> @jaraco, personally I believe this would give users better guidance. I also spot that in the issue template
> 
> > To ease the transition, Setuptools 45 was released to continue to have Python 2 compatibility, but emit a strenuous warning that it will stop working.
> 
> and wonder if you want to remove this given setuptools will crash on purpose and that we're on version 46+ now. If this PR is to be merged with the crashing part, I think it is fair to bump the version to 47 for backward compatible distinction.

Good feedback. I was leaning away from make it an incompatible release because the guidance is unchanged (use `Setuptools<45`), but after you shared this, I agree, a breaking release would be valuable. I'll do that. I'll also update that guidance in the template. Thanks!

docs/python 2 sunset.txt
McSinyx: I can take the job, it's literally 3 key stokes in my editor :smile:

docs/python 2 sunset.txt
McSinyx: I think giving more context (either of *Please be respectful/considerate to those projects*) might be better.

If the markup language is reST (I took a look at `conf.py` and saw no language set, so I think it's the default), I'll remove the lines in the follow-up PR since it will not look bad in plain text, e.g.
```rst
1. One two three
   four five six
2. Un deux trois
   quart cinq six

docs/python 2 sunset.txt
jaraco: TIL, SublimeText does that for me too. I'll do that.

docs/python 2 sunset.txt
jaraco: Indeed, that looks great. I'll incorporate it.

jaraco: Done in dbdd19d4 and f7b7ee25.

jaraco:
Thanks for the feedback. I'll plan to release this tomorrow unless there are other emergent concerns.

jaraco:
Okay. So today I learned that Bit.ly link targets can't be updated, at least not without a paid plan, and even then seemingly not without a custom domain. I was expecting that was one main advantage to using bit.ly (right after shorter urls). I guess that means that users on setuptools 45 and 46 will always be directed to the bug or bug template :/.

.github/ISSUE_TEMPLATE/setuptools-warns-about-python-2-incompatibility.md
hugovk: Please see PR https://github.com/pypa/setuptools/pull/2150.

pganssle:
Sorry I never responded to this, I'm fine with fully dropping support. Doesn't look like there's much progress in any of the stuff where I was worried that adoption would be held back by anyway, and we can't wait forever.

Issue #2133, opened 2020-05-24
support for SOURCE_DATE_EPOCH in sdist. 


Carreau:
SOURCE_DATE_EPOCH is useful for reproducible build, when set, no timestamp should be greater than this value. 

It seem that setuptools sdist does not support SOURCE_DATE_EPOCH, I've traced it to the following: 

sdit inherit from Commands, which leads to these successives calls. 

```
Lib/distutils/cmd.py:Command.make_archive
Lib/distutils/archive_util.py:make_archive
Lib/distutils/archive_util.py:ARCHIVE_FORMATS
Lib/distutils/archive_util.py:make_tarball
```

Make tarball seem to be the right place to monkeypatch to look for SOURCE_DATE_EPOCH as it itself can pass a filter to `tarfile.add()`, which will ensure the mtime is bounded (it already pass a filter to set uid/gid). 

With this most sdist (except tgz) are reproducibles. TGZ has this last problem that `GzipFile` adds `time.time()` in the header and that's a bit harder to patch.

joshuagl:
There's some excellent work towards this started in #2136, thanks @Carreau! Are you planning to pick this up? If not, perhaps I could help finish up this work? 

We would like to be able to produce reproducible sdists for [python-tuf](https://github.com/theupdateframework/tuf). (Curious readers can see: https://github.com/theupdateframework/tuf/issues/1269)

Carreau:
> Are you planning to pick this up? If not, perhaps I could help finish up this work?

At some point;  but I don't have much time these days; feel free to take over.

tiran:
I'm interested in reproducible sdists, too. Reproducible artifacts make it much easier to verify the provenance of code.  

* BPO https://bugs.python.org/issue31526 is the Python bug report for gzip timestamping issue.
* you can find my working implementation of reproducible ``tar.bz2`` at https://src.fedoraproject.org/rpms/python-cryptography/blob/rawhide/f/vendor_rust.py

dalcinl:
Just in case this is useful to others, I paste below a self-contained hunk of monkeypatching that allowed me to get reproducible (same sha256 hash) sdist tarballs. This hunk of code can be dumped in setup.py, for example.

```python
# Support for Reproducible Builds
# https://reproducible-builds.org/docs/source-date-epoch/

timestamp = os.environ.get('SOURCE_DATE_EPOCH')
if timestamp is not None:
    import distutils.archive_util as archive_util
    import stat
    import tarfile
    import time

[...]

PR #2136, opened 2020-05-24
Respect SOURCE_DATE_EPOCH when taring sdist.

Carreau:
This pulls just enough of distutils' and modify the make_tarball
function in order to respect SOURCE_DATE_EPOCH; this will ensure that
_when set_ no timestamp in the final archive is greater than timestamp.

This allows (but is not always sufficient), to make bytes for bytes
reproducible build for example:

 - This does not work with `gztar`, and zip does embed a timestamp in
 the header which currently is `time.time()` in the standard library.

 - if some fields passed to setup.py have on determinstic ordering (for
 example using sets for dependencies).

 Partial work toward #2133, with this I was able to make two bytes-identical
 sdist of IPython.

## Summary of changes

pull many function as-is from distutils with the exception of make_tarball to which a couple of lines are added. 

### Pull Request Checklist
- [ ] Changes have tests

No tests yet, nor changelog, I want to make sure this is the right approach first.

### DIFF
--- a/setuptools/archive_util.py
+++ b/setuptools/archive_util.py
@@ -202,8 +202,8 @@ def make_tarball(base_name, base_dir, compress="gzip", verbose=0, dry_run=0,
     # flags for compression program, each element of list will be an argument
     if compress is not None and compress not in compress_ext.keys():
         raise ValueError(
-              "bad value for 'compress': must be None, 'gzip', 'bzip2', "
-              "'xz' or 'compress'")
+            "bad value for 'compress': must be None, 'gzip', 'bzip2', "
+            "'xz' or 'compress'")
 
     archive_name = base_name + '.tar'
     if compress != 'compress':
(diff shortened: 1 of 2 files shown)


jaraco:
This is great. Thanks for putting this together.

I would recommend one change to your approach - if you could separate the commits of (a) copying code from distutils from (b) tweaking that code, it will make it easier to later reconcile any differences that might need to be merged from the original.

As is, if I try to review this, I can't tell what functionality was copied and what was changed. If you can separate those changes in the commit history, I'm pretty sure the changes here are acceptable.

As a side note, there is a larger effort to [incorporate all of distutils](https://github.com/pypa/setuptools/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Adopt+Distutils+and+stop+monkeypatching+stdlib%22), which if implemented would have made this change more straightforward.

Carreau:
Yes, I can do this, no problem. The one issue though will be flake 8 will complain about the raw `distutils` code, but it should be pretty straitforward.

Carreau:
Force-pushed with proper commit descriptions. I'd appreciate a pointer to a test that check a tar (not tar.gz as those are not yet reproducible), and can modify it to make sure the hash can be reproduced.

benoit-pierre:
Have you checked #1512?

Carreau:
I have not, but thanks, you definitely found the same issues that I did, though I went for much smaller and only tackle the minimal amount in one PR. 

I'll likely get inspiration from your tests.

kushaldas:
Any update on this?

jaraco:
@Carreau From your last comment, it wasn't clear to me whether you wanted to do more work on this or not. Can you clarify what the status is of this request?

jaraco:
I'm closing this for now, but happy to revive the effort at any point.

PR #2143, opened 2020-05-25
Adopt distutils

jaraco:
As discussed in #417, this change brings in the distutils code base with the following behaviors:

- This local copy of distutils is used exclusively when Setuptools is imported. The stdlib version is ignored completely.
 - As a result, when pip imports setuptools before building legacy (pre pep517) packages, those distutils-only packages will continue to get the setuptools treatment.
- Setuptools will refuse to import when distutils has been previously imported.
- If stdlib removes distutils or a downstream packager removes distutils from the standard library, the distutils supplied by setuptools becomes the default.
- If setuptools is not imported, the stdlib behavior is available.
- Removing the line from `setuptools.__init__` will disable the patch-by-default behavior but allow explicit patching (this behavior may require moving `distutils_patch` outside of the `setuptools` namespace.

I believe this behavior honors [the objectives](https://github.com/pypa/setuptools/issues/417#issuecomment-392298401) that Nick outlined and furthermore enables the following future enhancements:

- Setuptools can merge implementations of shared functionality, either by pushing patches from setuptools into distutils or by hosting the functionality natively in Setuptools but reflecting (importing) that functionality in the distutils namespace.
- Monkeypatching of distutils can be replaced by the unified implementation.

Tests are passing locally. There are some failures on Windows, likely due to the distutils code being older.

Questions:

- What use-cases does this approach break?
- Is there another approach that would be break fewer things?
- How can we minimize breakage and adopt something suitable?

### Pull Request Checklist
- [n/a] Changes have tests


jaraco:
As the distutils code is stale here, I've recreated xoviat's work and pushed a couple of branches to pypa/distutils@xoviat/setuptools-merge and pypa/distutils@jaraco/setuptools-merge with pypa/distutils@master pointing to the distutils commits from cpython, which I created by running `git filter-branch --prune-empty --subdirectory-filter Lib/distutils master` on the cpython master. I've started work on a new version of this branch that applies the setuptools-local changes first and then merges in the 'distutils' code from the upstream commits.

jaraco:
I've pushed the branch derived from xoviat's work to `refs/archive/xoviat-distutils`.

jaraco:
I had hoped by updating to the latest distutils code (which this latest commit does), the tests would pass, but I see the [Windows tests are failing](https://dev.azure.com/jaraco/setuptools/_build/results?buildId=739&view=logs&j=f2aa1cb7-ddd8-5e13-2793-3e55ac100c6d&t=0158daa4-66d3-5648-b813-c154c6e360f2). @zooba, would you be willing to take a peek at that failure and let me know if you've seen this failure before or if you have any other insight?

jaraco:
The error is this:

```
---------------------------- Captured stdout call -----------------------------
extension.c

LINK : fatal error LNK1104: cannot open file 'python38.lib'

---------------------------- Captured stderr call -----------------------------
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Enterprise\\VC\\Tools\\MSVC\\14.16.27023\\bin\\HostX86\\x64\\link.exe' failed with exit code 1104
```

When trying to run `python setup.py bdist_wheel` or similar.

I suspect there's something about the monkeypatching that's sensitive to distutils being in the stdlib.

jaraco:
Same results on [appveyor Python 3.6](https://ci.appveyor.com/project/jaraco/setuptools/builds/33108253/job/rpcqg6iqdu8joof6), yet when I run the tests locally, they pass, at least for test_build_ext and test_wheel.

jaraco:
I've been able to replicate the issue locally, which has helped in troubleshooting.

I've learned something interesting - if I build a virtualenv on master, then switch the code to the distutils branch, setuptools is able to build the package, but if I let the virtualenv build from the distutils branch, the extension builds fail. Here's the output without any pytest involved:

```
../setuptools/.tox/python/Scripts/python setup.py build
running build
running build_ext
building 'foo' extension
creating foo_build
creating foo_build\temp.win-amd64-3.8
creating foo_build\temp.win-amd64-3.8\Release
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.25.28610\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\Users\jaraco\local\setuptools\.tox\python\include -IC:\Users\jaraco\local\setuptools\.tox\python\include -IC:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.25.28610\include -IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\ucrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared -IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\winrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\cppwinrt /Tcfoo.c /Fofoo_build\temp.win-amd64-3.8\Release\foo.obj
foo.c
creating C:\Users\jaraco\local\ext\foo_build\lib.win-amd64-3.8
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.25.28610\bin\HostX86\x64\link.exe /nologo /INCREMENTAL:NO /LTCG /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\Users\jaraco\local\setuptools\.tox\python\libs /LIBPATH:C:\Users\jaraco\local\setuptools\.tox\python\PCbuild\amd64 /LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.25.28610\lib\x64 /LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.18362.0\ucrt\x64 /LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.18362.0\um\x64 /EXPORT:PyInit_foo foo_build\temp.win-amd64-3.8\Release\foo.obj /OUT:foo_build\lib.win-amd64-3.8\foo.cp38-win_amd64.pyd /IMPLIB:foo_build\temp.win-amd64-3.8\Release\foo.cp38-win_amd64.lib
LINK : fatal error LNK1104: cannot open file 'python38.lib'
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Tools\\MSVC\\14.25.28610\\bin\\HostX86\\x64\\link.exe' failed with exit code 1104
```

jaraco:
It looks as if use of older virtualenvs may be implicated, so merging the latest master appears to help.

jaraco:
I see that the latest distutils is incompatible with Python 3.5 (f-strings, `sysconfig_data_*` modules, maybe more). If Setuptools is going to retain support for Python 3.5, so will distutils. I'm thinking it will prove too difficult to try to keep distutils compatible with older Python versions, so I'm just going to punt and drop support for Python 3.5.

jaraco:
So I learned from that test that even dropping support for Python 3.5 is insufficient as PyPy (as found in github actions) doesn't have the modules to be compatible.

I also tried patching the behavior in `setuptools.distutils_patch`:

```diff
diff --git a/setuptools/distutils_patch.py b/setuptools/distutils_patch.py
index 43b64c0f..649acd36 100644
--- a/setuptools/distutils_patch.py
+++ b/setuptools/distutils_patch.py
@@ -28,4 +28,21 @@ def ensure_local_distutils():
         importlib.import_module('distutils')
 
 
+def repair_sysconfig():
+    """
+    Python 3.6.x adds a private ``_sysconfig_*`` modules. Bypass
+    this behavior where not available.
+    """
+    sysconfig = importlib.import_module('distutils.sysconfig')
+    import pdb; pdb.set_trace()
+    try:
+        sysconfig.get_config_vars()
+        return
+    except ImportError:
+        pass
+
+    os.environ['_PYTHON_SYSCONFIGDATA_NAME'] = '_sysconfigdata'
+
+
 ensure_local_distutils()
+repair_sysconfig()
```

But that technique is insufficient for other callers, such as virtualenv, that run into the same issue if distutils isn't patched.

```
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/cli/base_command.py", line 186, in _main
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/commands/install.py", line 404, in run
    use_user_site=options.use_user_site,
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/req/__init__.py", line 71, in install_given_reqs
    **kwargs
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/req/req_install.py", line 787, in install
    prefix=prefix,
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/locations.py", line 186, in get_scheme
    dist_name, user, home, root, isolated, prefix
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/pip/_internal/locations.py", line 109, in distutils_scheme
    d.parse_config_files()
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/py35/lib/python3.5/site-packages/_virtualenv.py", line 21, in parse_config_files
    result = old_parse_config_files(self, *args, **kwargs)
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/dist.py", line 398, in parse_config_files
    filenames = self.find_config_files()
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/dist.py", line 350, in find_config_files
    check_environ()
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/util.py", line 182, in check_environ
    os.environ['PLAT'] = get_platform()
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/util.py", line 107, in get_platform
    return get_host_platform()
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/util.py", line 93, in get_host_platform
    distutils.sysconfig.get_config_vars(),
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/sysconfig.py", line 487, in get_config_vars
    func()
  File "/Users/jaraco/code/public/pypa/setuptools/distutils/sysconfig.py", line 447, in _init_posix
    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
ImportError: No module named '_sysconfigdata_m_darwin_'
```

So now I'm trying another technique that mutates the actual distutils code.

jaraco:
Okay! So this approach seems to work. At least tests pass (without dropping 3.5 support). Surely there's more to be done, but I'd like now to get some reviews on this proposed change.

pganssle:
Thanks for your work on this Jason! I haven't had time to look at any of the code or the approach yet, but I'm pretty on board with dropping Python 3.5 given we've already dropped Python 2.7.

In my experience, usually whenever someone (often @jdufresne, keeping us modern 😅) proposes dropping a version of Python, it summons @hugovk who shows up with some numbers on current usage. Possibly Hugo can correct my methodology, but based on [pypistats.org](https://pypistats.org/packages/setuptools), it seems like Python 3.5 is holding steady at ~8.67% of all downloads.

Having those people pinned to an older setuptools is probably not _too_ bad, though, particularly considering that something like 25-40% of our downloads are Python 2.7, and we've already dropped support for 2.7 (plus it's way easier to upgrade 3.5 → 3.6 than 2.7 → 3.6.

hugovk:
> In my experience, usually whenever someone (often @jdufresne, keeping us modern 😅) proposes dropping a version of Python, it summons @hugovk who shows up with some numbers on current usage. Possibly Hugo can correct my methodology, but based on [pypistats.org](https://pypistats.org/packages/setuptools), it seems like Python 3.5 is holding steady at ~8.67% of all downloads.

🧙

It's a bit more than ~8.67%, but not a lot more. For all of April it was 13.24%, and for May (up to yesterday) it was 10.41%.

# April 2020

| category | percent | downloads  |
|----------|--------:|-----------:|
| 2.7      |  31.54% | 14,446,049 |
| 3.7      |  25.50% | 11,677,207 |
| 3.6      |  20.56% |  9,417,320 |
| 3.5      |  13.24% |  6,064,133 |
| 3.8      |   6.58% |  3,012,953 |
| null     |   1.88% |    862,835 |
| 3.4      |   0.53% |    241,038 |
| 2.6      |   0.12% |     53,023 |
| 3.9      |   0.04% |     17,927 |
| 3.3      |   0.01% |      3,787 |
| 3.2      |   0.00% |        101 |
| 3.10     |   0.00% |         13 |
| 2.8      |   0.00% |          5 |
| 2.4      |   0.00% |          2 |
| 3.1      |   0.00% |          2 |
| Total    |         | 45,796,395 |

Source: Using https://pypistats.org API: `pip install -U pypistats && pypistats python_minor setuptools --last-month`


# Most of May 2020

| category | percent | downloads  |
|----------|--------:|-----------:|
| 3.7      |  29.52% | 13,819,473 |
| 2.7      |  28.65% | 13,412,660 |
| 3.6      |  21.23% |  9,940,228 |
| 3.5      |  10.41% |  4,875,182 |
| 3.8      |   7.80% |  3,649,642 |
| null     |   1.80% |    841,790 |
| 3.4      |   0.46% |    213,278 |
| 2.6      |   0.07% |     34,440 |
| 3.9      |   0.05% |     22,742 |
| 3.3      |   0.01% |      4,233 |
| 3.10     |   0.01% |      2,481 |
| 3.2      |   0.00% |         80 |
| 2.8      |   0.00% |          2 |
| Total    |         | 46,816,231 |

Source: `pypistats python_minor setuptools --this-month`

# Since 2016

image
Source: pypistats for recent months, pypinfo for older: 
https://github.com/hugovk/pypi-tools#setuptools

---

Note these are downloads of _all versions_ of setuptools. You could do what [pip did](https://github.com/pypa/pip/issues/6148#issuecomment-616213532) and check stats for the _latest version (that's been out for a month)_, as people pinning to old versions will probably keep pinning so it doesn't really affect them.

I've not figured out the BigQuery for that yet, or if you've https://github.com/ofek/pypinfo set up, `pypinfo setuptools pyversion version` might get that info, but I'm out of BigQuery quota at the moment.

This [PePy chart](https://pepy.tech/project/setuptools?versions=*&versions=47.*&versions=46.*&versions=45.*&versions=44.*) suggests that people who were using setuptools 45-47 are keeping up with the latest version, and there's a bunch pinned to v44:

image

I checked v37-43 as well, they're pretty flat.

jaraco:
Sorry if it wasn't clear, this change retains support for Python 3.5, so that's no longer a consideration.

pfmoore:
Can I confirm what you want from me in the way of a review? I don't know the distutils or setuptools code bases that well, and it's a pretty large PR to work through.

If there's something specific you want my review on, let me know and I'll try to make the time (I'm pretty busy right now).

pradyunsg:
> I've not figured out the BigQuery for that yet

```
SELECT
  details.python,
  file.version,
  COUNT(*)
FROM `the-psf.pypi.downloads20200531`
WHERE
   file.project = "setuptools"
GROUP BY
   details.python,
   file.version
```

This is for a single day, you might want to use a range of days (especially once PyPI stops populating the daily-tables). IDK how pypiinfo does things, and the command suggested by @hugovk might be better than this in more ways than one. :)

pradyunsg:
I'm super excited about this. I don't think I know enough to review this... BUT, I know that this would be a great thing to fix. :)

ncoghlan:
I'm happy to hear we'll still have Py3.5 support - while there's definitely no obligation to offer this, it's certainly *nice* to be able to run the latest setuptools in the Debian "oldstable" system Python (currently Py3.5 in Debian 9).

The general approach seems solid to me, so I'll try to take a look on a real computer some time this week, but don't feel obliged to wait for a detailed review from me (the main thing I'd be looking at is any docs merging - the diff is big enough that I can't tell on my phone if that's even in scope for this PR)

jaraco:
> Can I confirm what you want from me in the way of a review?

I'm mainly looking to make sure there aren't any concerns that weren't raised in the issues or that have emerged since then.

It's a complicated PR, but the changes excluding those to distutils can be viewed with `git --no-pager diff master... --name-only | grep -v '^distutils/' | xargs git diff master... --`:

```diff
diff --git a/changelog.d/2143.breaking.rst b/changelog.d/2143.breaking.rst
new file mode 100644
index 00000000..185542ad
--- /dev/null
+++ b/changelog.d/2143.breaking.rst
@@ -0,0 +1 @@
+Setuptools adopts distutils from the standard library as a top-level ``distutils`` package and no longer depends on distutils in the standard library. This new ``distutils`` package will be masked by ``distutils`` in the standard library unless that library has been stripped by a downstream package manager or gets removed in a future version. Although this change is not expected to break any use-cases, it will likely affect tool-chains that are monkey-patching distutils or relying on Setuptools' own monkey patching of distutils.
diff --git a/conftest.py b/conftest.py
index 1746bfb5..3f7e59b4 100644
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,7 @@ def pytest_addoption(parser):
 collect_ignore = [
     'tests/manual_test.py',
     'setuptools/tests/mod_with_constant.py',
+    'distutils',
 ]
 
 
diff --git a/setuptools/__init__.py b/setuptools/__init__.py
index 833e20ea..7a1f9f4f 100644
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -2,6 +2,9 @@
 
 import os
 import functools
+
+import setuptools.distutils_patch  # noqa: F401
+
 import distutils.core
 import distutils.filelist
 import re
diff --git a/setuptools/distutils_patch.py b/setuptools/distutils_patch.py
new file mode 100644
index 00000000..43b64c0f
--- /dev/null
+++ b/setuptools/distutils_patch.py
@@ -0,0 +1,31 @@
+"""
+Ensure that the local copy of distutils is preferred over stdlib.
+
+See https://github.com/pypa/setuptools/issues/417#issuecomment-392298401
+for more motivation.
+"""
+
+import sys
+import importlib
+import contextlib
+from os.path import dirname
+
+
+@contextlib.contextmanager
+def patch_sys_path():
+    orig = sys.path[:]
+    sys.path[:] = [dirname(dirname(__file__))]
+    try:
+        yield
+    finally:
+        sys.path[:] = orig
+
+
+def ensure_local_distutils():
+    if 'distutils' in sys.path:
+        raise RuntimeError("Distutils must not be imported before setuptools")
+    with patch_sys_path():
+        importlib.import_module('distutils')
+
+
+ensure_local_distutils()

setuptools distutils $ git --no-pager diff master... --name-only | grep -v '^distutils/' | xargs git --no-pager diff master... --
diff --git a/changelog.d/2143.breaking.rst b/changelog.d/2143.breaking.rst
new file mode 100644
index 00000000..185542ad
--- /dev/null
+++ b/changelog.d/2143.breaking.rst
@@ -0,0 +1 @@
+Setuptools adopts distutils from the standard library as a top-level ``distutils`` package and no longer depends on distutils in the standard library. This new ``distutils`` package will be masked by ``distutils`` in the standard library unless that library has been stripped by a downstream package manager or gets removed in a future version. Although this change is not expected to break any use-cases, it will likely affect tool-chains that are monkey-patching distutils or relying on Setuptools' own monkey patching of distutils.
diff --git a/conftest.py b/conftest.py
index 1746bfb5..3f7e59b4 100644
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,7 @@ def pytest_addoption(parser):
 collect_ignore = [
     'tests/manual_test.py',
     'setuptools/tests/mod_with_constant.py',
+    'distutils',
 ]
 
 
diff --git a/setuptools/__init__.py b/setuptools/__init__.py
index 833e20ea..7a1f9f4f 100644
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -2,6 +2,9 @@
 
 import os
 import functools
+
+import setuptools.distutils_patch  # noqa: F401
+
 import distutils.core
 import distutils.filelist
 import re
diff --git a/setuptools/distutils_patch.py b/setuptools/distutils_patch.py
new file mode 100644
index 00000000..43b64c0f
--- /dev/null
+++ b/setuptools/distutils_patch.py
@@ -0,0 +1,31 @@
+"""
+Ensure that the local copy of distutils is preferred over stdlib.
+
+See https://github.com/pypa/setuptools/issues/417#issuecomment-392298401
+for more motivation.
+"""
+
+import sys
+import importlib
+import contextlib
+from os.path import dirname
+
+
+@contextlib.contextmanager
+def patch_sys_path():
+    orig = sys.path[:]
+    sys.path[:] = [dirname(dirname(__file__))]
+    try:
+        yield
+    finally:
+        sys.path[:] = orig
+
+
+def ensure_local_distutils():
+    if 'distutils' in sys.path:
+        raise RuntimeError("Distutils must not be imported before setuptools")
+    with patch_sys_path():
+        importlib.import_module('distutils')
+
+
+ensure_local_distutils()
```

pfmoore:
Reviewing the patch you quoted above, this looks OK to me.

One question, did you consider making distutils a *subpackage* of setuptools, and then exposing that subpackage as a top-level `distutils` in the `ensure_local_distutils` function? I feel slightly nervous when a project exposes multiple top-level packages (*ahem* `pkg_resources` 🙂).

But it's a minor point and given that we *already* have `pkg_resources`, I'm not going to worry about it. Just wanted to make sure the option had been considered.

So I'm OK with this.

pganssle:
> One question, did you consider making distutils a _subpackage_ of setuptools, and then exposing that subpackage as a top-level `distutils` in the `ensure_local_distutils` function? I feel slightly nervous when a project exposes multiple top-level packages (_ahem_ `pkg_resources` ).

I agree with this, and I'd go a step further: we should probably aim to integrate `distutils` much more directly into `setuptools`, so that all `from distutils import x` statements can be translated more-or-less directly into `from setuptools import x`, and then we should raise deprecation warnings whenever anyone imports `distutils`.

I think we can do this in stages, though. I'd say we should move the top-level `distutils` to `setuptools._distutils`, then in `ensure_local_distutils` expose that as `distutils`.

changelog.d/2143.breaking.rst
```diff
@@ -0,0 +1 @@
+Setuptools adopts distutils from the standard library as a top-level ``distutils`` package and no longer depends on distutils in the standard library. This new ``distutils`` package will be masked by ``distutils`` in the standard library unless that library has been stripped by a downstream package manager or gets removed in a future version. Although this change is not expected to break any use-cases, it will likely affect tool-chains that are monkey-patching distutils or relying on Setuptools' own monkey patching of distutils.
```
pganssle, line 1: A few things about the hyphenation feels a bit off to me. I would not hyphenate use-case or tool-chain (I would say "use case" and "toolchain" or "tool chain", respectively), and "monkey-patching" and "monkey patching" both appear here. I think either one is fine (I'm leaning towards "monkey-patching", I guess), but it should be consistent.

pganssle:
A few comments:

> * Setuptools will refuse to import when distutils has been previously imported.

I suspect that there's no good way around this, but I think that unless we migrate all `distutils` commands into the `setuptools` namespace (making it so that no one ever *needs* to import `distutils`), this is going to be a source of annoyance for anyone not using `pip` or a similar workflow that does the monkey patching *for you*. A common import sort order (and the default one for `isort`) is to import all standard library modules first, then third party modules. Even if `distutils` is considered third-party, usually people sort alphabetically (or possibly by length of the import name) — both of these will have `distutils` come first.

I don't think it's a big deal, but it suggests to me that *most* `setup.py` files that import both `distutils` and `setuptools` will probably import `distutils` first. It sounds like anyone using PEP 517 who does it this way will start seeing their packages break.

> * Setuptools can merge implementations of shared functionality, either by pushing patches from setuptools into distutils or by hosting the functionality natively in Setuptools but reflecting (importing) that functionality in the distutils namespace.
> * Monkeypatching of distutils can be replaced by the unified implementation.

Are these two different things, or is the second one just a consequence of the first?

Either way, I think at least the second point is the logical next step here. Once this is merged, for anything we're *currently* monkey patching, we should just merge that functionality directly into `distutils`. I personally would prefer it if we made the `setuptools` namespace the canonical namespace, and have `distutils` just import symbols from `setuptools` and possibly wrap them in deprecation warnings.

setuptools/distutils_patch.py
```diff
@@ -0,0 +1,31 @@
...
+    orig = sys.path[:]
+    sys.path[:] = [dirname(dirname(__file__))]
+    try:
+        yield
+    finally:
+        sys.path[:] = orig
+
+
+def ensure_local_distutils():
+    if 'distutils' in sys.path:
```
zooba, line 25:
```suggestion
    if 'distutils' in sys.modules:
```

setuptools/distutils_patch.py
```diff
@@ -0,0 +1,31 @@
...
+        raise RuntimeError("Distutils must not be imported before setuptools")
+    with patch_sys_path():
+        importlib.import_module('distutils')
```
zooba, line 28: Apparently this works, since you've tested it, but I'd have expected some import failures once you take the standard library off `sys.path`. Probably by now, everything has been imported already.

An alternative that should not require patching sys.path at all could be:
```python
spec = importlib.util.spec_from_file_location("distutils", PATH_TO_DISTUTILS_INIT_PY)
mod = importlib.util.module_from_spec(spec)
mod.__loader__.exec_module(mod)
```

changelog.d/2143.breaking.rst
zooba: I agree. "Use case" (n), "toolchains/tool chains" (n) and "monkey-patching distutils" (v). For the last, if you were talking about monkey patching in general it would be different, but because you're qualifying the type of patching being applied to distutils, it gets the hyphen.

(Though tbh, I don't know the formal rules here. I've just learned that trusting my gut on grammar works well for me.)

setuptools/distutils_patch.py
jaraco: yikes! Good catch.

setuptools/distutils_patch.py
jaraco: > I'd have expected some import failures once you take the standard library off `sys.path`.

Actually, I would expect this code to work as well when distutils is removed from stdlib, as long as the `distutils` python package is on `sys.path`, which it is in this proposal (it's exposed as a top-level package).

> Probably by now, everything has been imported already.

Can you phrase that in another way. I'm not sure what you're trying to say.



> An alternative that should not require patching sys.path at all could be

Hmm. Yes, perhaps that's better.

jaraco:
> did you consider making distutils a _subpackage_ of setuptools

Yes, but one key part of [the objectives](https://github.com/pypa/setuptools/issues/417#issuecomment-392298401) (bullet 2) is that setuptools _supplies_ `distutils` (i.e. `import distutils`).

In that sense, setuptools _must_ expose a top-level `distutils` package. I don't see any way around that unless we relax that objective.

This first draft does bring in distutils wholesale, but sets the stage to eventually consolidate the behavior (in compatible and managed-incompatible stages).

changelog.d/2143.breaking.rst
jaraco: Good suggestions, committed in a8081a19.

jaraco:
> I think we can do this in stages, though. I'd say we should move the top-level `distutils` to `setuptools._distutils`, then in `ensure_local_distutils` expose that as `distutils`.

This was my instinct also. Since there seems to be consensus around this approach, I'll implement it that way... and we can worry about support for a plain `import distutils` later.

jaraco:
Interesting. In the three weeks since I last submitted this, something has changed about the environment such that `ensure_local_distutils` is now failing. I suspect `pytest` or `virtualenv` or another dependency has started importing `distutils` :(.

jaraco:
> > I think we can do this in stages, though. I'd say we should move the top-level `distutils` to `setuptools._distutils`, then in `ensure_local_distutils` expose that as `distutils`.
> 
> This was my instinct also. Since there seems to be consensus around this approach, I'll implement it that way... and we can worry about support for a plain `import distutils` later.

Attempting to make `setuptools._distutils` appear as `distutils`, even just for setuptools' use is proving difficult. It opens up new questions, like:

- Is it okay for `distutils.__name__` to be `setuptools._distutils`?
- How to ensure that `setuptools._distutils` is loadable in a zip archive?
- Do all of the imports in setuptools that reference distutils need to be adapted to import the private version?

setuptools/distutils_patch.py
jaraco: I tried implementing the alternative, but I don't think `spec_from_file_location` is viable, as that assumes file-system-based packages, and setuptools could be in a zip archive or loaded through another loader.

jaraco:
> Interesting. In the three weeks since I last submitted this, something has changed about the environment such that `ensure_local_distutils` is now failing. I suspect `pytest` or `virtualenv` or another dependency has started importing `distutils` :(.

I can't figure out what's going on. I added some debugging to stdlib `distutils.__init__` so it records the stack when it loads (identifying the module that imported it). I then ran the tests and the stack implicates `pytest-virtualenv` as importing distutils:

```
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.8/site-packages/_pytest/assertion/rewrite.py", line 152, in exec_module
    exec(co, module.__dict__)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.8/site-packages/pytest_virtualenv.py", line 5, in <module>
    from distutils import sysconfig
```

So I went back to [one of the builds that succeeded](https://travis-ci.org/github/pypa/setuptools/jobs/693182216) and grabbed the tox env dependencies there (`apipkg==1.5,appdirs==1.4.4,attrs==19.3.0,contextlib2==0.6.0.post1,coverage==5.1,distlib==0.3.0,execnet==1.7.1,filelock==3.0.12,flake8==3.8.2,flake8-2020==1.6.0,mccabe==0.6.1,mock==4.0.2,more-itertools==8.3.0,packaging==20.4,path==13.1.0,path.py==12.4.0,Paver==1.3.4,pip==20.1.1,pluggy==0.13.1,py==1.8.1,pycodestyle==2.6.0,pyflakes==2.2.0,pyparsing==2.4.7,pytest==5.4.2,pytest-cov==2.9.0,pytest-fixture-config==1.7.0,pytest-flake8==1.0.6,pytest-shutil==1.7.0,pytest-virtualenv==1.7.0,six==1.15.0,termcolor==1.1.0,virtualenv==20.0.21,wcwidth==0.1.9,wheel==0.34.2`) and installed those manually into the virtualenv, yet still the tests fail.

If I check out db7ff2e1b, the tests are passing. So it does seem like something changed since then to break the tests.

jaraco:
Ah, right. a8081a19 made the code actually do what it was supposed to do, so the fact that the tests were passing was a false pass. :(

jaraco:
Well, this is annoying. I've managed to get tests passing on CPython tests, but not on PyPy3 tests. When the tests are invoked through the pytest script wrapper, the tests fail to load, but when invoked through runpy (`python -m pytest`), the tests run.

I suspect there's some interaction between the pytest "assert rewrite" feature, the "pytest-virtualenv" behavior, and the distutils patching behavior.

jaraco:
I'm able to trigger the same behavior under `python -m pytest` by adding `del sys.path[0]` to `pytest/__main__.py`.

jaraco:
It seems the issue is that `setuptools` isn't getting installed in an editable way on PyPy3 with the latest `distutils` in place. This may be an actual regression of `distutils` as found on Python 3.9.

I notice that there exists `.tox/pypy3/lib/python3.6/site-packages` with only setuptools editable-installed to it, and a separate `.tox/pypy3/site-packages` dir with everything else installed to it::

```
setuptools distutils $ tree .tox/pypy3/lib/python3.6/site-packages
.tox/pypy3/lib/python3.6/site-packages
├── easy-install.pth
├── setuptools.egg-link
└── setuptools.pth

0 directories, 3 files
setuptools distutils $ tree .tox/pypy3/site-packages | (head; echo '...'; tail)
.tox/pypy3/site-packages
├── Paver-1.3.4.dist-info
│   ├── DESCRIPTION.rst
│   ├── INSTALLER
│   ├── LICENSE.txt
│   ├── METADATA
│   ├── RECORD
│   ├── WHEEL
│   ├── entry_points.txt
│   ├── metadata.json
...
├── zipp-3.1.0.dist-info
│   ├── INSTALLER
│   ├── LICENSE
│   ├── METADATA
│   ├── RECORD
│   ├── WHEEL
│   └── top_level.txt
└── zipp.py

324 directories, 2057 files
setuptools distutils $ ls .tox/pypy3/site-packages/easy*
ls: .tox/pypy3/site-packages/easy*: No such file or directory
```

Compare that with what I find when creating the environment without the latest distutils (setuptools master):

```
setuptools master $ tox -r -e pypy3 --notest -q
pypy3: pip_version is pip
_____________________________________________________________________ summary _____________________________________________________________________
  pypy3: skipped tests
  congratulations :)

setuptools master $ cat .tox/pypy3/site-packages/easy-install.pth
/Users/jaraco/code/public/pypa/setuptools
```

If my hunch is correct, attempting to run Setuptools tests at master on Python 3.9 should fail similarly.

jaraco:
Right, so Python 3.9 is irrelevant here. Probably the difference in the `distutils` module for PyPy versus CPython. PyPy probably has some customizations to the distutils implementation that are lost when adopting distutils from CPython.

Tracing the `install_dir` from `develop` to `easy_install` to `install_lib`, I find where the difference resides:

```
setuptools master $ .tox/pypy3/bin/python -c "import setuptools; cmd = setuptools.dist.Distribution().get_command_obj('install_lib'); cmd.ensure_fi
nalized(); print(cmd.install_dir)"
/Users/jaraco/code/public/pypa/setuptools/.tox/pypy3/site-packages/
setuptools master $ git checkout distutils
Switched to branch 'distutils'
Your branch is up to date with 'origin/distutils'.
setuptools distutils $ .tox/pypy3/bin/python -c "import setuptools; cmd = setuptools.dist.Distribution().get_command_obj('install_lib'); cmd.ensure
_finalized(); print(cmd.install_dir)"
/Users/jaraco/code/public/pypa/setuptools/setuptools/distutils_patch.py:29: UserWarning: Setuptools is replacing distutils
  warnings.warn("Setuptools is replacing distutils")
/Users/jaraco/code/public/pypa/setuptools/.tox/pypy3/lib/python3.6/site-packages/
```

Interestingly, I also learned that distutils is imported as a matter of course on PyPy:

```
$ pypy3 -c "import sys; print('distutils' in sys.modules)"
True
```

jaraco:
The PyPy docs have a [differences document](https://doc.pypy.org/en/latest/cpython_differences.html#differences-between-pypy-and-cpython), but it doesn't obviously reference anything relevant about distutils.

jaraco:
I think I've found the relevant commit where that code diverged:

image

But when I try to locate [that commit in heptapod](https://foss.heptapod.net/pypy/pypy/-/commit/4b8c42f02d1ca5211748959b7ad4a2ad165d104f), it's 404.

jaraco:
After patching that one issue, the symptom goes away, but then there are other failures in the test suite.

I recognize that there are several other changes in distutils as found in PyPy that would need to be applied.

I think it may be necessary to merge the distutils fork in PyPy before Setuptools can adopt distutils. Either that or PyPy will lose setuptools support.

jaraco:
With the help of mattip (huge thanks), the tests are now passing on PyPy. As an added bonus, pypa/distutils now has a test suite that runs under tox (and also passes).

Next, I'm working on moving the logic into `setuptools/_distutils` instead of exposing a top-level `distutils` module.

jaraco:
I'm ready to move forward with this effort. I'll merge and release later today after I've had some sleep. If it causes unexpected regressions, it can be readily disabled by removing/disabling the line in `setuptools.__init__`.

PR #2149, opened 2020-05-26
Step 5 on 2093: Package discovery and namespace package userguide now ready

alvyjudy:
Step 5 on #2093 
## Summary of changes
Created user guide for automatic package discovery and namespace package supported by setuptools. It covers the following aspects in both ``setup.py`` and ``setup.cfg`` style:

- automatic package discovery with ``find_packages``
- overview on namespace package
- automatic package discovery with ``find_namespace_package``
- legacy use of namespace package

The quickstart guide's package discovery section references this file

### Pull Request Checklist
- [ ] Changes have tests

If these looks good I can do the rebasing myself to make merging easier.

### DIFF
--- a/docs/userguide/package_discovery.txt
+++ b/docs/userguide/package_discovery.txt
@@ -222,7 +222,8 @@ the previous section.
 ``pkgutil`` style namespace package
 -----------------------------------
 This method is almost identical to the ``pkg_resource`` except that the
-``__init__.py`` file contains the following:
+``namespace_packages`` declaration is omitted and the ``__init__.py``
+file contains the following:
 
 .. code-block:: python
 
--- a/docs/userguide/package_discovery.txt
+++ b/docs/userguide/package_discovery.txt
@@ -22,6 +22,7 @@ included manually in the following manner:
 .. code-block:: ini
 
     [options]
+    #...
     packages =
         mypkg1
         mypkg2
@@ -29,6 +30,7 @@ included manually in the following manner:
 .. code-block:: python
 
     setup(
+        #...
         packages = ['mypkg1', 'mypkg2']
     )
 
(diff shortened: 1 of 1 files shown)


jaraco:
This is great, but it misses a third technique for namespace packages, the pkgutil-based design, and I would say that the pkgutil-based design is preferable to the pkg_resources-based design. It might be best to minimize the guidance here to anything setuptools-specific and otherwise defer to the [packaging users guide on namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/).

alvyjudy:
thanks for pointing that out @jaraco. In fact I omitted the ``pkgutil``style considering that it isn't part of ``setuptools``. Should I briefly mention it and defer to the linked article? 

Also just out of curiosity, was there a reason pkg_resource style became so wide-spread when pkgutil seems to offer better compatibility?

jaraco:
> Also just out of curiosity, was there a reason pkg_resource style became so wide-spread when pkgutil seems to offer better compatibility?

If I recall correctly, the pkg_resources technique was more compatible with earlier Python versions. I see pkgutil was added in Python 2.3 (2003), so it could be the case that pkg_resources added support for older Pythons. I do believe there were other levels of sophistication that pkg_resources offered, such as support for multi-install environments (on-demand package inclusion), so that may have been another use-case it supported.

I wasn't around at the time and the first technique I encountered was that in setuptools.

> thanks for pointing that out @jaraco. In fact I omitted the `pkgutil`style considering that it isn't part of `setuptools`. Should I briefly mention it and defer to the linked article?

That sounds good to me.

alvyjudy:
@jaraco I have made the changes and it should be ready for merging:)

docs/userguide/package_discovery.txt
```diff
@@ -1,181 +1,231 @@
...
+        setup.cfg #or setup.py
 
-    __import__("pkg_resources").declare_namespace(__name__)
+Repeat the same for other packages and you can achieve the same result as
+the previous section.
+
+``pkgutil`` style namespace package
+-----------------------------------
+This method is almost identical to the ``pkg_resource`` except that the
+``__init__.py`` file contains the following:
```
jaraco, line 225:
```suggestion
This method is almost identical to the ``pkg_resource`` except that the
``namespace_packages`` declaration is omitted and the ``__init__.py``
file contains the following:
```

PR #2151, opened 2020-05-27
DOC: fix link missing >

Carreau:
The `decalrative config`_ link present higher in the document is also
unresolved, I'm not quite sure what the intended target was.


jaraco:
I'm pretty sure declarative config is fixed in the docs revamp, underway (#2093).

PR #2154, opened 2020-05-28
Step 6 on 2093: userguide on entry points now available

alvyjudy:
Step 6 on #2093
## Summary of changes
This user guide covers:
- console_script
- plugin support
- optional dependencies

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/docs/userguide/entry_point.txt
+++ b/docs/userguide/entry_point.txt
@@ -1,12 +1,14 @@
+.. _`entry_points`:
+
 ==========================================
 Entry Points and Automatic Script Creation
 ==========================================
 
-When installing a package, you may realize you can invoke some commands without
-explicitly calling the python interpreter. For example, instead of calling
-``python -m pip install`` you can just do ``pip install``. The magic behind
-this is entry point, a keyword passed to your ``setup.cfg`` or ``setup.py``
-to create script wrapped around function in your libraries.
+After installing some packages, you may realize you can invoke some commands
+without explicitly calling the python interpreter. For example, instead of
+calling ``python -m pip install`` you can just do ``pip install``. The magic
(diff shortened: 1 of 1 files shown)


docs/userguide/entry_point.txt
```diff
@@ -1,115 +1,200 @@
...
+
+    def helloworld():
+        print("Hello world")
+
+After installing the package, you can invoke this function in the following
+manner, without applying any magic:
+
+.. code-block:: bash
+
+    python -m mypkg.helloworld
```
jaraco, line 37: Wouldn't this be `timmins.helloworld`? Also, I don't think that works. `-m` (implemented in the stdlib `runpy` module) support only invokes a module or package (with a `__main__` module). Don't worry too much about correctness here. I'll be happy to update this section myself after it's merged.

alvyjudy: must have been an oversight :(two actually). It would be great if you could help out with it! But I do intend to test out the examples after i complete all sections. much appreciated for the understanding

jaraco:
Now that I've gone through and amended the docs, I see that my changes actually bring the docs back closer to how they were originally. I had to undo lots of mistaken descriptions. I think for the future, let's try to avoid re-writing whole documents and instead make more surgical revisions to correct false behaviors but otherwise retain the value from the previous author. I'm still unsure if I want to accept these revisions or go back to the original and just amend what was there. I'm going to think this over some more.

alvyjudy:
Hi @jaraco  either way I will be happy to continue contributing so don't worry about your decision upsetting me:) Personally the learning process is the most important and valuable for myself so even if we were to remove this branch altogether I'm still glad I had put in the efforts and appreciate your time!

PR #2159, opened 2020-05-29
Defer finalize options

### LINKED ISSUE #2158: Plugins using setuptools entry points resolve to system and not setup_requires
Assume that you have e.g. setuptools_scm 3 installed system-wide, and try to to run the following setup.py:
```python
from setuptools import setup
setup(name="foo", setup_requires=["setuptools_scm>=4"])
```
you get the warning
```
/usr/lib/python3.8/site-packages/setuptools/dist.py:701: UserWarning: Module setuptools_scm was already imported from /usr/lib/python3.8/site-packages/setuptools_scm/__init__.py, but /tmp/foo/.eggs/setuptools_scm-4.1.1-py3.8.egg is being added to sys.path
```
(i.e. the system-wide setuptools_scm was loaded, even though it is too old).  One can confirm that the "correct" newer setuptools_scm indeed doesn't get used by changing setup.py to
```python
from setuptools import setup
setup(name="foo",
      setup_requires=["setuptools_scm>=4"],
      use_scm_version={"parentdir_prefix_version": "foo-"})
```
in which case one gets the same warning and then a traceback
```
TypeError: __init__() got an unexpected keyword argument 'parentdir_prefix_version'
```

I guess a solution may be e.g. to move the entry point (which is what gets eagerly loaded by setuptools) to a separate package (let's say `_setuptools_scm_helper`) which itself imports setuptools_scm only when actually needed with a pass-through shim so that setuptools only imports `_setuptools_scm_helper` and then the correct version of setuptools_scm can be loaded when needed.

jaraco:
Closes #2158

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/2158.misc.rst
+++ b/changelog.d/2158.misc.rst
@@ -1,0 +1,1 @@
+Avoid loading working set during ``Distribution.finalize_options`` prior to invoking ``_install_setup_requires``, broken since v42.0.0.
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -129,10 +129,27 @@ if PY3:
 def _install_setup_requires(attrs):
     # Note: do not use `setuptools.Distribution` directly, as
     # our PEP 517 backend patch `distutils.core.Distribution`.
-    dist = distutils.core.Distribution(dict(
-        (k, v) for k, v in attrs.items()
-        if k in ('dependency_links', 'setup_requires')
-    ))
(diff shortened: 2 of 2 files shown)


jaraco:
I'm struggling to think how I could create a suitable test to capture the failure, so I'm going to proceed with rolling this out without a test (other than the manual one I performed).

PR #2169, opened 2020-05-31
Allow specifying the location of setup.cfg files

ofek:
## Summary of changes

The file path resolver will first check for the `SETUPTOOLS_CFG_FILE` env var

### Motivation

The files in my case are dynamically generated at build time and this would remove the need to temporarily write them to the project directory

### DIFF
--- a/setuptools/command/setopt.py
+++ b/setuptools/command/setopt.py
@@ -16,6 +16,9 @@ def config_file(kind="local"):
 
     `kind` must be one of "local", "global", or "user"
     """
+    env_location = os.environ.get('SETUPTOOLS_CFG_FILE')
+    if env_location:
+        return os.path.normpath(env_location)
     if kind == 'local':
         return 'setup.cfg'
     if kind == 'global':


jaraco:
The `config_file` function there has a lot of interactions with other behaviors (`alias.format_alias` uses it directly, and `alias` and `saveopts` use the functionality through subclassing). Moreover, I don't see how this behavior affects today's most common use, that of reading config from those files, which happens in `distutils.dist.Distribution.find_config_files` which doesn't honor this setting.

Have you confirmed that this change actually addresses your use-case?

This change would also need a test to capture the desired use-case.

jaraco:
I also noticed there was an option `--filename` (or similar). Have you explored using that option?

ofek:
@jaraco Hey! Actually I don't need this anymore. I was attempting to build atop setuptools but I'm writing a proper pep 517 backend for Hatch instead.

jaraco:
Well, that's great news, that the standards are working and your project is unblocked!

I welcome you or others to re-engage this issue in the future if needed.

PR #2194, opened 2020-06-14
Decrease start-up time of editable-installed entry points on newer versions of Python

ofek:
## Summary of changes

When `importlib.metadata` is present, the generated scripts will use that rather than the significantly slower `pkg_resources`.

Closes https://github.com/pypa/setuptools/issues/510 completely as editable installs are the final affected part (that is actionable) brought up in that issue and what I've been experiencing the last 2 weeks (it's indeed annoying)

The issue is locked, so here are some pings: @jaraco @scopatz @ninjaaron @untitaker @asottile @pganssle @gaborbernat @pfmoore @pradyunsg

### Benchmark

Before and after, using  [hyperfine](https://github.com/sharkdp/hyperfine):

Capture

### Pull Request Checklist

- [X] Changes have tests

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -2070,9 +2070,7 @@ class ScriptWriter:
     gui apps.
     """
 
-    try:
-        from importlib.metadata import distribution  # noqa: F401
-
+    if sys.version_info >= (3, 8):
         template = textwrap.dedent(r"""
             # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
             __requires__ = %(spec)r
@@ -2086,7 +2084,7 @@ class ScriptWriter:
                     if entry_point.group == %(group)r and entry_point.name == %(name)r:
                         sys.exit(entry_point.load()())
         """).lstrip()  # noqa: E501
-    except ImportError:
+    else:
         template = textwrap.dedent(r"""
             # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
             __requires__ = %(spec)r
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -2070,19 +2070,36 @@ class ScriptWriter:
     gui apps.
     """
 
-    template = textwrap.dedent(r"""
-        # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
-        __requires__ = %(spec)r
-        import re
-        import sys
-        from pkg_resources import load_entry_point
+    try:
+        from importlib.metadata import distribution  # noqa: F401
 
-        if __name__ == '__main__':
-            sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-            sys.exit(
-                load_entry_point(%(spec)r, %(group)r, %(name)r)()
-            )
-    """).lstrip()
+        template = textwrap.dedent(r"""
+            # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
+            __requires__ = %(spec)r
+            import re
+            import sys
+            from importlib.metadata import distribution
+
+            if __name__ == '__main__':
+                sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+                for entry_point in distribution(%(spec)r).entry_points:
+                    if entry_point.group == %(group)r and entry_point.name == %(name)r:
+                        sys.exit(entry_point.load()())
+        """).lstrip()  # noqa: E501
+    except ImportError:
+        template = textwrap.dedent(r"""
+            # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
+            __requires__ = %(spec)r
+            import re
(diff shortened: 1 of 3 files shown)


setuptools/tests/test_easy_install.py
```diff
@@ -71,19 +71,37 @@ def test_install_site_py(self, tmpdir):
...
-            from pkg_resources import load_entry_point
-
-            if __name__ == '__main__':
-                sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-                sys.exit(
-                    load_entry_point('spec', 'console_scripts', 'name')()
-                )
-            """)  # noqa: E501
+        try:
+            from importlib.metadata import distribution  # noqa: F401
```
asottile, line 75: I'd suggest instead `if sys.version_info >= (3, 8):`

setuptools/command/easy_install.py
```diff
@@ -2070,19 +2070,36 @@ class ScriptWriter:
...
+        template = textwrap.dedent(r"""
+            # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
+            __requires__ = %(spec)r
+            import re
+            import sys
+            from importlib.metadata import distribution
+
+            if __name__ == '__main__':
+                sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+                for entry_point in distribution(%(spec)r).entry_points:
```
asottile, line 2085: should this use `entry_points()['console_scripts']` perhaps instead?

ofek: Wouldn't that enumerate every script for every package?

asottile: iirc `distribution` already enumerates every package but I might be misremembering

setuptools/tests/test_easy_install.py
ofek: Done, thanks!

setuptools/command/easy_install.py
ofek: I think that's `distributions`:

https://github.com/python/cpython/blob/811e040b6e0241339545c2f055db8259b408802f/Lib/importlib/metadata.py#L498-L512

setuptools/command/easy_install.py
```diff
@@ -2070,19 +2070,34 @@ class ScriptWriter:
...
-        if __name__ == '__main__':
-            sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-            sys.exit(
-                load_entry_point(%(spec)r, %(group)r, %(name)r)()
-            )
-    """).lstrip()
+    if sys.version_info >= (3, 8):
+        template = textwrap.dedent(r"""
+            # EASY-INSTALL-ENTRY-SCRIPT: %(spec)r,%(group)r,%(name)r
+            __requires__ = %(spec)r
```
jaraco, line 2076: This line is no longer necessary as it only has value when `import pkg_resources` is used (that's what verifies that `__main__.__requires__` is installed and activated).

ofek: Ah I see, thanks. Removed!

setuptools/command/easy_install.py
```diff
@@ -2070,19 +2070,34 @@ class ScriptWriter:
...
-        import sys
-        from pkg_resources import load_entry_point
-
-        if __name__ == '__main__':
-            sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-            sys.exit(
-                load_entry_point(%(spec)r, %(group)r, %(name)r)()
-            )
-    """).lstrip()
+    if sys.version_info >= (3, 8):
```
jaraco, line 2073: I'm trying to imagine a way this could rely also on `importlib_metadata` and thus eliminate the dependency on pkg_resources. I dislike that the code has to be forked to support the legacy behavior (and until Python 3.7 support is dropped). Perhaps that can't be solved now.

jaraco, line 2073: I also wonder if it would make sense to unify these two scripts into one which tries importlib-metadata then falls back to pkg_resources. That approach would avoid the fork and would also avoid the suppressed linter errors.

setuptools/command/easy_install.py
jaraco: I was surprised this interface isn't nicer :/. Yes, I believe this usage is correct. Probably importlib-metadata should provide some helpers to make this usage simpler, though that won't help for this implementation.

setuptools/tests/test_easy_install.py
```diff
@@ -71,19 +71,34 @@ def test_install_site_py(self, tmpdir):
...
-            import sys
-            from pkg_resources import load_entry_point
-
-            if __name__ == '__main__':
-                sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-                sys.exit(
-                    load_entry_point('spec', 'console_scripts', 'name')()
-                )
-            """)  # noqa: E501
+        if sys.version_info >= (3, 8):
```
jaraco, line 74: I'll plan to consolidate this fork later... and just make some assertions about the result that are common to both behaviors.

ofek:
Thank you!

ofek:
Needs small fix https://github.com/pypa/setuptools/issues/2196

PR #2195, opened 2020-06-15
Fix entry point scripts

ofek:
## Summary of changes

Fix to https://github.com/pypa/setuptools/pull/2194 @jaraco 

Is there a better way to inject just the package name? I thought that is what spec was.

```
  File "/opt/hostedtoolcache/Python/3.8.3/x64/bin/ddev", line 9, in <module>
    for entry_point in distribution('datadog-checks-dev==3.11.0').entry_points:
  File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/importlib/metadata.py", line 504, in distribution
    return Distribution.from_name(distribution_name)
  File "/opt/hostedtoolcache/Python/3.8.3/x64/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: datadog-checks-dev==3.11.0
```

### Pull Request Checklist
- [ ] Changes have tests


setuptools/command/easy_install.py
```diff
@@ -2079,7 +2079,7 @@ class ScriptWriter:
 
             if __name__ == '__main__':
                 sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-                for entry_point in distribution(%(spec)r).entry_points:
+                for entry_point in distribution(%(spec)r.split('==')[0]).entry_points:
```
cooperlees, line 2082: I have no context on setuptools test situation etc., but can a test be added to ensure this does not happen again?

jaraco: I'm going to defer the tests for now, as I'm not confident I can build one reliably and quickly. I do agree it would be good to have, although the long-term solution will be for editable installs to have their scripts generated with distlib, same as pip, without the package resources requirement.

PR #2225, opened 2020-07-03
Programmatically disable coverage when running on PyPy.

jaraco:
Building on #2224 and using the advice from pytest-dev/pytest-cov#418, programmatically disable coverage.


jaraco:
The change did not have the intended effect.

PR #2240, opened 2020-07-09
Restore Debian support

### LINKED ISSUE #2232: Debian-specific behavior lost when Setuptools adopts distutils
`pip install -e $PACKAGE_PATH` installs commands to /usr/bin and not /usr/local/bin on Ubuntu as it did in the past.

`pip install $PACKAGE_PATH` continues to install to /usr/local/bin as expected.

Openstack downstream has temporarily capped setuptools until we know if this is expected behaviour or not.  Please see http://lists.openstack.org/pipermail/openstack-discuss/2020-July/015779.html for more context.

jaraco:
As reported in #2232, this patch incorporates the patch from Debian to distutils and detects Debian to apply that behavior at run time.

Closes #2232.

### DIFF
--- a/setuptools/_distutils/_platform.py
+++ b/setuptools/_distutils/_platform.py
@@ -1,0 +1,21 @@
+import re
+import pathlib
+import contextlib
+
+
+# from jaraco.context
+class suppress(contextlib.suppress, contextlib.ContextDecorator):
+    """
+    A version of contextlib.suppress with decorator support.
+
+    >>> @suppress(KeyError)
+    ... def key_error():
+    ...     {}['']
+    >>> key_error()
+    """
+
+
+@suppress(Exception)
+def is_debian():
+    issue = pathlib.Path('/etc/issue').read_text()
+    return bool(re.search('(debian|buntu|mint)', issue, re.IGNORE_CASE))
--- a/setuptools/_distutils/command/install.py
+++ b/setuptools/_distutils/command/install.py
@@ -14,6 +14,7 @@ from distutils.file_util import write_file
 from distutils.util import convert_path, subst_vars, change_root
 from distutils.util import get_platform
 from distutils.errors import DistutilsOptionError
+from distutils._platform import is_debian
 
 from site import USER_BASE
 from site import USER_SITE
@@ -32,6 +33,20 @@ INSTALL_SCHEMES = {
         'purelib': '$base/lib/python$py_version_short/site-packages',
         'platlib': '$platbase/$platlibdir/python$py_version_short/site-packages',
         'headers': '$base/include/python$py_version_short$abiflags/$dist_name',
+        'scripts': '$base/bin',
+        'data'   : '$base',
+        },
+    'unix_local': {
+        'purelib': '$base/local/lib/python$py_version_short/dist-packages',
+        'platlib': '$platbase/local/lib/python$py_version_short/dist-packages',
+        'headers': '$base/local/include/python$py_version_short/$dist_name',
+        'scripts': '$base/local/bin',
+        'data'   : '$base/local',
+        },
+    'deb_system': {
+        'purelib': '$base/lib/python3/dist-packages',
+        'platlib': '$platbase/lib/python3/dist-packages',
+        'headers': '$base/include/python$py_version_short/$dist_name',
         'scripts': '$base/bin',
         'data'   : '$base',
         },
@@ -145,6 +160,9 @@ class install(Command):
 
         ('record=', None,
          "filename in which to record list of installed files"),
+
+        ('install-layout=', None,
+         "installation layout to choose (known values: deb, unix)"),
         ]
 
     boolean_options = ['compile', 'force', 'skip-build']
@@ -165,6 +183,7 @@ class install(Command):
         self.exec_prefix = None
         self.home = None
         self.user = 0
+        self.prefix_option = None
 
         # These select only the installation base; it's up to the user to
         # specify the installation scheme (currently, that means supplying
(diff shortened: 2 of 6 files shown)


pganssle:
I really don't like putting distro-specific code in upstream like this. If debian wants non-standard behavior they should carry the patch. Otherwise if there is a feature they need they should propose it like anyone else and it should be judged on the normal criteria that we judge features.

pganssle:
Also this is all stuff that only gets triggered when you do setup.py install instead of installing from a wheel, right? That seems firmly in "unsupported" territory.

jaraco:
Thank you Paul for the sanity check.

> I really don't like putting distro-specific code in upstream like this. If debian wants non-standard behavior they should carry the patch. Otherwise if there is a feature they need they should propose it like anyone else and it should be judged on the normal criteria that we judge features.

I agree with you and had the same reaction initially.

But as I'm thinking about #2232 and what recommendation I could make, I struggled with saying that Setuptools should just break Debian/Ubuntu installs. Debian has _already_ patched distutils, but the patch is bypassed by the fact that Setuptools is carrying an unpatched copy of distutils. I don't see how Debian could reasonably expect to patch an implementation that's typically installed by pip or other package managers outside of their control, across multiple Python versions, and across many (sometimes virtual) environments.

Setuptools' adoption of distutils is currently blocked on this issue, so I'm seeking some solution that can be implemented in the near term and this approach seemed like the most straightforward. In addition to unblocking distutils adoption, it also captures in a single codebase the commonality and divergence of Debian behaviors, and by being in a high velocity project like Setuptools, can be rapidly iterated/improved. In other words, acceptance of this patch should facilitate the broader adoption or deprecation and ultimate convergence of these divergent behaviors. I do hope/imagine that Debian might straight-up adopt pypa/distutils instead of stdlib distutils in order to avoid patching.

I will say that Debian holds some culpability for this issue, but I don't think the Debian users should bear the cost of the somewhat hostile monkeypatch. I don't know the history of this patch in Debian and I wouldn't be surprised if it was the PyPA or similar that recommended or forced Debian's approach.

> Also this is all stuff that only gets triggered when you do setup.py install instead of installing from a wheel, right?

Unfortunately, no. This behavior affects editable installs (in particular where scripts are installed).

I can imagine an alternate scenario where Setuptools could first start honoring the the Debian expectation (use unix-local by default), though I don't fully understand the implications of that change. My instinct is that ultimately that's the right thing to do, but I was hoping to defer reconciling that divergence until after the distutils code was adopted and the distutils/setuptools codebases were largely unified.

I do believe the `setup.py install` case may also be relevant in the situation where the PyPA deprecates distutils, removes it from stdlib, and declares setuptools as the replacement. Hopefully with PEP 517 and a solution for editable installs, that's a non-issue.


Where can we go from here?

jaraco:
On other small point - I want to avoid the situation where the distutils adoption is blocked again by some other challenging feature (i.e. editable install revamp). This effort stalled once and the work got stale. I'd like to come out of this having distutils in Setuptools in O(weeks) so that the work doesn't grow stale and so that more progress can be made on merging the functionality.

pganssle:
> But as I'm thinking about #2232 and what recommendation I could make, I struggled with saying that Setuptools should just break Debian/Ubuntu installs. Debian has _already_ patched distutils, but the patch is bypassed by the fact that Setuptools is carrying an unpatched copy of distutils. I don't see how Debian could reasonably expect to patch an implementation that's typically installed by pip or other package managers outside of their control, across multiple Python versions, and across many (sometimes virtual) environments.

I don't think they should be patching this in the first place, but the question is why does debian need this at all? Why can't they do what every other distro is doing?

Also, this change only applies to the system python. I realize end users use the system python with non-system-installed packages all the time, but they *shouldn't*, because it causes all kinds of problems. If they're using system packages, they'll use a Debian-supplied `setuptools` which carries the patch. I would really hate it if I were forced to use a Debian-derived distro and I couldn't create a normal Python environment by using `pyenv` etc because `setuptools` now has some Debian-specific behavior in it.

> I will say that Debian holds some culpability for this issue, but I don't think the Debian users should bear the cost of the somewhat hostile monkeypatch. I don't know the history of this patch in Debian and I wouldn't be surprised if it was the PyPA or similar that recommended or forced Debian's approach.

I understand that, but this is what got us into this mess in the first place. Debian wanted some other default behavior and so they patched distutils for whatever reason rather than letting it be fixed in the right place. Now we're going to carry some Debian-specific behavior (which *should* increase our test matrix to include debian and non-debian flavored tools) instead of fixing it in the right place and in the right way — whatever that is.

> Unfortunately, no. This behavior affects editable installs (in particular where scripts are installed).

I'm not certain of the details here, but from what I can tell this will only affect things like `sudo pip install -e package`, which is just a recipe for breaking your Python installation *anyway*. If for some reason *package managers* are using `pip install -e`, they should probably stop, and again they would use the system package, which can carry whatever patches they want.

At the end of the day, I am not convinced that #2232 needs to be fixed on our end, and it doesn't really need to block our releases.

> On other small point - I want to avoid the situation where the distutils adoption is blocked again by some other challenging feature (i.e. editable install revamp). This effort stalled once and the work got stale. I'd like to come out of this having distutils in Setuptools in O(weeks) so that the work doesn't grow stale and so that more progress can be made on merging the functionality.

I don't think you need to worry about editable install revamp blocking anything on the timescale of weeks. I was so thoroughly disappointed by the last round of conversations about this that I'm not totally convinced it will happen at all.

jaraco:
> I would really hate it if I were forced to use a Debian-derived distro and I couldn't create a normal Python environment by using `pyenv` etc because `setuptools` now has some Debian-specific behavior in it.

That's a really good point that I hadn't considered, though I would imagine this risk is at least an order of magnitude less problematic than the one at hand.

jaraco:
After further consideration, I'm going to decline this request. I'm also going to go back to pypa/distutils and reset master to the state without this patch.

PR #2249, opened 2020-07-11
Change exec_module to load_module

alexhenrie:
I originally wrote this code using `load_module` and then changed it to `exec_module` because [the documentation](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) says that `load_module` is deprecated, but I didn't notice that `exec_module` requires an argument. Since it's not clear from the documentation how `exec_module` is supposed to be used, and `load_module` doesn't actually emit a DeprecationWarning (fixing the warning was the original motivation for changing this code), I think we should just use `load_module` for now.

### DIFF
--- a/changelog.d/2249.misc.rst
+++ b/changelog.d/2249.misc.rst
@@ -1,0 +1,1 @@
+Fix extension loading technique in stubs.
--- a/setuptools/command/bdist_egg.py
+++ b/setuptools/command/bdist_egg.py
@@ -59,7 +59,7 @@ def write_stub(resource, pyfile):
             from importlib.machinery import ExtensionFileLoader
             __file__ = pkg_resources.resource_filename(__name__, %r)
             __loader__ = None; del __bootstrap__, __loader__
-            ExtensionFileLoader(__name__,__file__).exec_module()
+            ExtensionFileLoader(__name__,__file__).load_module()
         __bootstrap__()
         """).lstrip()
     with open(pyfile, 'w') as f:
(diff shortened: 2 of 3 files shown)


hugovk:
Thanks for the quick update!

Is it possible to include tests for this code?

And do you have an idea why we only hit this on Python 3.9 and not 3.5-3.8?

alexhenrie:
@hugovk Unfortunately, I don't know the answer to either of those questions.

pangloss:
This issue hit me today on both 3.7.5 and 3.7.7.

grlee77:
> Since it's not clear from the documentation how exec_module is supposed to be used, and load_module doesn't actually emit a DeprecationWarning (fixing the warning was the original motivation for changing this code), I think we should just use load_module for now.

By strange coincidence, I just happened to be looking at this same `load_module` vs `exec_module` usage for our `setup.py` in PyWavelets. I think in your case, the code would be something like the following (untested, adapated from an answer in [this stack overflow post](https://stackoverflow.com/questions/19009932/import-arbitrary-python-source-file-python-3-3)):

```Python
 import types
 from importlib.machinery import ExtensionFileLoader
 loader = ExtensionFileLoader(__name__, __file__)
 module = types.ModuleType(loader.name)
 loader.exec_module(module)
```

danielgordon10:
Can you please push a patch to PYPI? Many projects that don't pin the version are starting to get the bug.

PR #2258, opened 2020-07-13
Remove issue templates

pganssle:
The "setuptools warns about incompatibility" template has generated no useful reports and is creating a lot of noise and clutter on the issue tracker. I believe that removing it will break the link that people are using to generate these reports, which will hopefully deter further erroneous reports.


jaraco:
I'm generally in favor of this change. I was going to make this change myself, but I was convinced that [this bit.ly link](https://bit.ly/setuptools-py2-warning) and the [sunset docs](https://setuptools.readthedocs.io/en/latest/python%202%20sunset.html) referred to the template. Now I see that both of those embed the template, so the template is in fact not needed. Happy to see it gone.

At some point, it may be worthwhile to create templates for general bug/feature reports, soliciting some of this same information, but no reason for that to block this change.

jaraco:
FYI, the issues continue to come in - they just come in without the template giving users guidance in the template.

Issue #2261, opened 2020-07-13
Errors with v49.2.0's distutils adoption


hugovk:
Related to #2256 / #2259.

When upgrading to `setuptools>=49.1.3` for the https://github.com/pypa/setuptools/pull/2249 fix for Pillow, we get the new warning in v49.2.0:

> /home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/setuptools/distutils_patch.py:25: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.

This also causes some C build issues and this sanity test failure; Pillow checks the version in the lower-level C code ("Core version") matches that in the user-facing Python code ("Pillow version"):

> /home/travis/virtualenv/python3.9-dev/lib/python3.9/site-packages/Pillow-7.3.0.dev0-py3.9-linux-x86_64.egg/PIL/Image.py:115: RuntimeWarning: The _imaging extension was built for another version of Pillow or PIL:
Core version: None
Pillow version: 7.3.0.dev0

---

Moving the setuptools import before distutils in our setup.py removes the warning but not the error.

https://travis-ci.org/github/hugovk/Pillow/builds/707730154

I've not dug into this further, but it's happening at least on all the Linux/Python versions tested on Travis (3.5-3.9), and wanted to let you know.

More details at https://github.com/python-pillow/Pillow/pull/4784#issuecomment-657727387.

mattip:
also see numpy/numpy#16851 where we are trying to work around this. The problem is there is no way of knowing where exactly distutils is being imported.

pganssle:
@mattip I think it would be prudent to hold off on trying to fix the warnings at the moment. I think #2259 / #2260 will likely be resolved before the distutils adoption is re-enabled by default.

mattip:
> Though replacing distutils with setuptools wherever possible is worthwhile 

Numpy has distutils extensions, and I hope nothing will break on esoteric combinations like f2py + (mingw or intel compilers)

pganssle:
> Numpy has distutils extensions, and I hope nothing will break on esoteric combinations like f2py + (mingw or intel compilers)

[...]

PR #2297, opened 2020-08-03
Change load_module to exec_module

radarhere:
#2249 changed back from `exec_module` to the deprecated `load_module` (so I'll cc @alexhenrie). This PR moves forward to `exec_module` again. Credit to posts like https://stackoverflow.com/a/41595552/4093019

Also fixes the Pillow compilation error reported in #2261.


alexhenrie:
Thanks Andrew!

PR #2306, opened 2020-08-06
Don't install setup_requires when run as a PEP-517 backend.

### LINKED ISSUE #2303: Don't install setup_requires when run as a PEP 517 backend
PEP 517 separates the responsibilities around build requirements: the backend is responsible for saying what is required, and the frontend is responsible for ensuring they're available to the build.

In setuptools, build requirements are defined by `setup_requires`, and these get passed through to PEP 517's `get_requires_for_build_*` hooks. There is a monkeypatch to return them from these hooks and prevent setuptools trying to install them itself:

https://github.com/pypa/setuptools/blob/5e60dc50e540a942aeb558aabe7d92ab7eb13d4b/setuptools/build_meta.py#L56-L75

But something similar to that - preventing installation - should really be in place for all the PEP 517 hooks, because a PEP 517 backend isn't responsible for installing dependencies.

Why does this matter? Pip has the `--no-build-isolation` option, with which the caller can declare that they have taken care of build dependencies and pip should try to build the package in the current environment. This is useful for downstream packagers, and for experimenting with different versions of your build dependencies. But setuptools doesn't know about this, so it charges ahead and attempts to install things when that's not what you want.

The workaround I'm looking at is to only specify `setup_requires` if `'egg_info' in sys.argv`, as is the case when the `get_requires_for_build_*` hooks are called. But this is clearly not ideal.

takluyver:
## Summary of changes

Under PEP 517, installing build dependencies is up to the frontend. `setup_requires` are already reported as build dependencies, but setuptools should not attempt to install them itself when used as a PEP 517 backend.

Closes #2303

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -380,7 +380,7 @@ class TestBuildMetaBackend:
                 setup(
                     name="qux",
                     version="0.0.0",
-                    py_modules=["hello.py"],
+                    py_modules=["hello"],
                     setup_requires={setup_literal},
                 )
             """).format(setup_literal=setup_literal),
@@ -406,6 +406,36 @@ class TestBuildMetaBackend:
         actual = get_requires()
 
         assert expected == sorted(actual)
+
+    def test_dont_install_setup_requires(self, tmpdir_cwd):
+        files = {
+            'setup.py': DALS("""
+                        from setuptools import setup
+
+                        setup(
+                            name="qux",
+                            version="0.0.0",
+                            py_modules=["hello"],
+                            setup_requires=["does-not-exist >99"],
+                        )
+                    """),
+            'hello.py': DALS("""
+                    def run():
+                        print('hello')
+                    """),
+        }
+
+        build_files(files)
+
+        build_backend = self.get_build_backend()
+
+        dist_dir = os.path.abspath('pip-dist-info')
+        os.makedirs(dist_dir)
+
+        # does-not-exist can't be satisfied, so if it attempts to install
+        # setup_requires, it will fail.
+        build_backend.prepare_metadata_for_build_wheel(dist_dir)
+
 
     _sys_argv_0_passthrough = {
         'setup.py': DALS("""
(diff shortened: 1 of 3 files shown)


setuptools/tests/test_build_meta.py
```diff
@@ -380,7 +380,7 @@ def test_setup_requires(self, setup_literal, requirements, use_wheel,
                 setup(
                     name="qux",
                     version="0.0.0",
-                    py_modules=["hello.py"],
+                    py_modules=["hello"],
```
pganssle, line 383: Why this change?

takluyver: The test was failing locally without it. All the other samples in this file with `py_modules` don't have the `.py` extension, and the [packaging guide](https://packaging.python.org/guides/distributing-packages-using-setuptools/#py-modules) says it doesn't belong there.

pganssle: Hm.. It's disturbing that the tests are currently passing if they fail for you locally without that change, no?

takluyver: I think they may have been passing on master and only failed when I made the change I was interested in. As this seemed like a simple mistake and nothing else was failing, I confess I didn't try to understand what was happening.

jaraco:
I'm inclined to accept this change, especially since it brings the codebase one step closer to fully deprecating setup_requires.

You've flagged the change not as `breaking`. Are we confident there aren't any use cases that will break with this change? In other words, are there users that rely on `setup_requires` in a PEP 517 environment?

takluyver:
The build dependencies are reported to the frontend which is responsible for installing them in a build environment. You can disable that, e.g. with the `--no-build-isolation` flag for pip, in which case the caller is responsible for ensuring the build dependencies are installed in the environment where pip is running.

So I don't believe that this breaks any scenario which is meant to work. But as with any change, someone could have relied on something that wasn't meant to work. For instance:

- If someone uses `pip install --no-build-isolation` without all build dependencies installed, and is used to setuptools attempting to install build dependencies anyway.
- If there are packages which can be installed by `easy_install` but not by `pip`, they may not work with PEP 517.
- If setup scripts sniff what command is being run and specify `setup_requires` for some commands but not others, the dependencies may not be reported to the frontend.

Would you prefer that I called that `breaking`?

jaraco:
Based on that description, I'm inclined to say it's not a breaking change. Thanks for adding that detail so those that are affected can understand better the context.

jaraco:
Releasing as 49.5.

takluyver:
Thanks Jason!

PR #2315, opened 2020-08-08
Add tests capturing expected distutils behavior.

jaraco:
Add three tests to capture the expected and anticipated behavior for distutils adoption.

Ref #417

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,7 @@ tests =
     paver; python_version>="3.6"
     futures; python_version=="2.7"
     pip>=19.1 # For proper file:// URLs support.
+    jaraco.envs
 
 docs =
     # Keep these in sync with docs/requirements.txt
--- a/setuptools/tests/requirements.txt
+++ b/setuptools/tests/requirements.txt
@@ -10,3 +10,4 @@ pytest-cov>=2.5.1
 paver; python_version>="3.6"
 futures; python_version=="2.7"
 pip>=19.1 # For proper file:// URLs support.
+jaraco.envs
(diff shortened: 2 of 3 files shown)


jaraco:
The tests are failing on Windows presumably due to a bug in pytest-virtualenv or subprocess where `virtualenv.run(['pip'])` or `virtualenv.run(['python'])` doesn't execute in the virtualenv.

jaraco:
While working on a test to capture the failure on Windows, I run into [another failure on pytest-virtualenv](https://github.com/man-group/pytest-plugins/issues/165).

jaraco:
The test is failing due to man-group/pytest-plugins#166. I think probably I'll need to find another library to support virtualenv-based testing reliably.

jaraco:
Hmm. Now there are new errors when executing on Windows:

```
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized
```

WTH?

pganssle:
I'm wondering if it might be easier to get "test as installed" working for at least a subset of the tests?

Thanks for all this effort. I'd try to help but the power at my house has been out since Tuesday (and I'm at an out of town event today).

jaraco:
Some searching led me to [others who've had the problem](/appveyor/ci/issues/1995). It seems that `SYSTEMROOT` must be set in the environment or Python will fail to start.

jaraco:
If with the SYSTEMROOT fix, the tests are still failing for me locally on Windows (in particular, test_distutils_local_with_setuptools). I'm trying to figure out why.

PR #2316, opened 2020-08-10
Fix issue with distutils warning

### LINKED ISSUE #2259: Reconsidering distutils replacement strategy

pganssle:
## Summary of changes

Closes #2259 

### Pull Request Checklist

### DIFF
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -16,9 +16,12 @@ def warn_distutils_present():
         # https://foss.heptapod.net/pypy/pypy/-/blob/be829135bc0d758997b3566062999ee8b23872b4/lib-python/3/site.py#L250
         return
     warnings.warn(
-        "Distutils was imported before Setuptools. This usage is discouraged "
-        "and may exhibit undesirable behaviors or errors. Please use "
-        "Setuptools' objects directly or at least import Setuptools first.")
+        "Distutils was imported before Setuptools, but importing Setuptools "
+        "also replaces the `distutils` module in `sys.modules`. This may lead "
+        "to undesirable behaviors or errors. To avoid these issues, avoid "
+        "using distutils directly, ensure that setuptools is installed in the "
+        "traditional way (e.g. not an editable install), and/or make sure that "
+        "setuptools is always imported before distutils.")
 
 
 def clear_distutils():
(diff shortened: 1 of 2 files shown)


jaraco:
Thanks Paul!

PR #2319, opened 2020-08-10
Remove superfluous 'v' in CHANGES.rst

pganssle:
This was an artifact of the fact that I did not use `tox -e finalize` to do the latest release in order to make a patch release.


jaraco:
Thanks Paul. Feel free to push changes like this without a PR (as you wish).

PR #2333, opened 2020-08-16
fix find_packages glob doc

bnavigator:
## Summary of changes

The current documentation of `find_packages` incorrectly includes dots in the `exclude=` patterns. These are `fnmatch` patterns, not regexes.

See also https://github.com/pycroscopy/pyUSID/issues/50#issuecomment-674522758

I am not even sure if the statement "*but not 'foo' itself*" is true.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/changelog.d/2333.doc.rst
+++ b/changelog.d/2333.doc.rst
@@ -1,0 +1,1 @@
+remove dots from `find_package(exclude=['*tests*'])` glob pattern examples
--- a/docs/setuptools.txt
+++ b/docs/setuptools.txt
@@ -256,14 +256,14 @@ patterns are applied to remove matching packages.
 
 Inclusion and exclusion patterns are package names, optionally including
 wildcards.  For
-example, ``find_packages(exclude=["*.tests"])`` will exclude all packages whose
-last name part is ``tests``.   Or, ``find_packages(exclude=["*.tests",
-"*.tests.*"])`` will also exclude any subpackages of packages named ``tests``,
+example, ``find_packages(exclude=["*tests"])`` will exclude all packages whose
+last name part is ``tests``.   Or, ``find_packages(exclude=["*tests",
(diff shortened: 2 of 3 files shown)


McSinyx:
Hello there, I have not checked any deeper, but I think e.g. `*.test` is supposed to unmatch `foo.test` but not `footest`.  If that doesn't work that way, there might be a recent regression prevent such behavior.

bnavigator:
> I think e.g. `*.test` is supposed to unmatch `foo.test` but not `footest`.

In any case the surrounding documentation text does not match. There is no reference to `foo.test`. Of course `*test` also (un)matches foo.test. Note that these get passed to `fnmatchcase`, so they match against the filepath `' foo/test'`, not for `foo.test`. This is not `find_namespace_packages`.

bnavigator:
> Then only `foo` (`foo/__init__.py`) and `foo.spam` (`foo/spam.py`) will be included.

... which is contrary to the statement. `fnmatchcase('foo*')` also matches `foo`!

McSinyx:
> … Note that these get passed to `fnmatchcase`, so they match against the filepath `foo/test`, not `foo.test`. This is not `find_namespace_packages`.

To facilitate the conversation, could you please provide a minimal reproducible case?  What I found doesn't seem to agree your statement:

https://github.com/pypa/setuptools/blob/f991fbb3c9d0e10a0a78ae2b508b3fd99f9cdef2/setuptools/__init__.py#L79-L82

> ... which is contrary to the statement.  `fnmatchcase('foo*')` also matches `foo`!

I sense some confusion here, the original docs say that excluding `foo.*` will match `foo` but not e.g. `foo.bar`.

bnavigator:
Okay I missed that replace line. It changes everything.

bnavigator:
My mistake

PR #2370, opened 2020-09-02
Fix the distutils hacking behavior with the environment variable SETU…

gavincyi:
## Summary of changes

The change is to ensure the distutils hatching behavior works by default, except the environment variable SETUPTOOLS_USE_DISTUTILS is specified as "stdlib". 

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ class install_with_pth(install):
     _pth_contents = textwrap.dedent("""
         import os
         var = 'SETUPTOOLS_USE_DISTUTILS'
-        enabled = os.environ.get(var, 'local') == 'local'
+        enabled = os.environ.get(var, 'local') != 'stdlib'
         enabled and __import__('_distutils_hack').add_shim()
         """).lstrip().replace('\n', '; ')
 
--- a/setuptools/tests/test_distutils_adoption.py
+++ b/setuptools/tests/test_distutils_adoption.py
@@ -67,3 +67,12 @@ def test_distutils_local(venv):
     preferred.
     """
     assert venv.name in find_distutils(venv, env=dict()).split(os.sep)
+
+
+@pytest.mark.xfail('IS_PYPY', reason='pypy imports distutils on startup')
+def test_distutils_invalid_option(venv):
+    """
+    Ensure local distutils is used even when an invalid option is provided.
+    """
+    env = dict(SETUPTOOLS_USE_DISTUTILS='xxx')
+    assert venv.name in find_distutils(venv, env=env).split(os.sep)
--- a/changelog.d/2370.change.rst
+++ b/changelog.d/2370.change.rst
@@ -1,0 +1,1 @@
+Fix the default distutils hatching behavior when environment variable SETUPTOOLS_USE_DISTUTILS is specified


setup.py
```diff
@@ -100,7 +100,7 @@ class install_with_pth(install):
     _pth_contents = textwrap.dedent("""
         import os
         var = 'SETUPTOOLS_USE_DISTUTILS'
-        enabled = os.environ.get(var, 'local') == 'local'
+        enabled = os.environ.get(var, 'local') != 'stdlib'
```
jaraco, line 103: This behavior exists in two places, here and in _distutils_hack. Probably both places would need to be addressed.

setuptools/tests/test_distutils_adoption.py
```diff
@@ -67,3 +67,12 @@ def test_distutils_local(venv):
...
     """
     assert venv.name in find_distutils(venv, env=dict()).split(os.sep)
+
+
+@pytest.mark.xfail('IS_PYPY', reason='pypy imports distutils on startup')
+def test_distutils_invalid_option(venv):
+    """
+    Ensure local distutils is used even when an invalid option is provided.
+    """
+    env = dict(SETUPTOOLS_USE_DISTUTILS='xxx')
```
jaraco, line 77: The problem here is that the test above (and also in `_distutils_hack`) doesn't check against all valid values. Probably, there should be a check that the allowed values are "stdlib" and "local" and the behavior with any other value is undefined.

gavincyi: Undefined means to throw an exception if any other values is provided?

jaraco: Right. I was unclear.

What I mean to say is when the value is anything but "stdlib" or "local", there's no guaranteed behavior, so the test should assert that `find_distutils` could contain venv.name or not. I don't want "xxx" to be a synonym for "local" as that encumbers all those possible values. If there was ever a need for "xxx" to mean a third option, that would not be possible without changing the meaning of "xxx", which would be a backward-incompatible change ("xxx" would mean "local" before the change and something else after).

I guess what I'm realizing now is that I don't want this change.

jaraco:
I specifically created this variable to be a non-boolean variable so that the value signals the purpose (rather than the presence/absence of the variable). I suspect you're coming to this variable expecting it to be a boolean value when it is not. The two supported values are currently "stdlib" and "local". It may make sense to raise a warning if any other values are specified, but I've been treating it as anything other than "stdlib" or "local" is the same as not setting the variable.

This PR is making some assumptions, fixing something that I don't think needs fixing. If you feel strongly about it, please file an issue describing the issue that this PR attempts to fix and we can explore possible fixes in the issue.

Thanks for your patience and for taking the time to prepare this PR.

gavincyi:
When I upgraded to setuptools 50.0, my understanding of the intention was to hatch the distutils module by default, unless "stdlib" is specified (yes similar to a boolean value). It was bizzare to see the behavior falling back to the "stdlib" one even a typo / empty string was specified. So I thought it's a minor bug and set the PR without creating a ticket.

Now I saw the change in 50.1 to fall back the 49.6 behavior so I think you're right to close the PR as it's no longer valid anymore. Not sure if I understand what you meant of "I've been treating it as anything other than "stdlib" or "local" is the same as not setting the variable". I guess it's equivalent to falling to the default behavior if any other values are specified, isn't it?

jaraco:
> I guess it's equivalent to falling to the default behavior if any other values are specified, isn't it?

Yes, that's right.

PR #2379, opened 2020-09-03
Adding changes to support PPC64LE

ezeeyahoo:
## Summary of changes
Hi,

I have added few jobs to support PPC64LE architecture, request you to please review.

Enhancement

### Pull Request Checklist
- [No] Changes have tests
- [No] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,9 +39,12 @@ jobs:
 cache: pip
 
 before_install:
-# Fix for PPC64LE, where permissions are fixed to allow tox/pip to install packages.
-# As travis base images for PPC64LE except for bionic, do not have this fix.
   - |
+    # Except on bionic, Travis Linux base image for PPC64LE
+    # platform lacks the proper
+    # permissions to the directory ~/.cache/pip/wheels that allow
+    # the user running travis build to install pip packages.
+    # TODO: is someone tracking this issue? Maybe just move to bionic?
     if [[ "$TRAVIS_CPU_ARCH" == "ppc64le" ]]; then
       sudo chown -Rfv $USER:$GROUP ~/.cache/pip/wheels
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ jobs:
   - python: 3.9-dev
   - <<: *latest_py3
     env: TOXENV=docs
-    #PPC64LE
   - arch: ppc64le
     python: pypy3
   - arch: ppc64le
@@ -28,8 +27,6 @@ jobs:
     env: LANG=C
   - arch: ppc64le
     python: 3.9-dev
-  - <<: *latest_py3_ppc
-    env: TOXENV=docs
   allow_failures:
   # suppress failures due to pypa/setuptools#2000
   - python: pypy3
@@ -37,14 +34,12 @@ jobs:
     python: pypy3
   - <<: *latest_py3
     env: TOXENV=docs
-  - <<: *latest_py3_ppc
-    env: TOXENV=docs
 
 
 cache: pip
 
 before_install:
-# Fix for PPC64LE, where permissions are fixed to allow tox/pip to install packages. 
+# Fix for PPC64LE, where permissions are fixed to allow tox/pip to install packages.
 # As travis base images for PPC64LE except for bionic, do not have this fix.
   - |
     if [[ "$TRAVIS_CPU_ARCH" == "ppc64le" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,17 +21,11 @@ jobs:
     python: pypy3
   - arch: ppc64le
     python: 3.5
-  - arch: ppc64le
-    python: 3.6
-  - arch: ppc64le
-    python: 3.7
   - &latest_py3_ppc
     arch: ppc64le
     python: 3.8
   - <<: *latest_py3_ppc
     env: LANG=C
-  - arch: ppc64le
-    python: 3.8-dev
   - arch: ppc64le
     python: 3.9-dev
   - <<: *latest_py3_ppc
@@ -50,6 +44,8 @@ jobs:
 cache: pip
 
 before_install:
+# Fix for PPC64LE, where permissions are fixed to allow tox/pip to install packages. 
+# As travis base images for PPC64LE except for bionic, do not have this fix.
   - |
     if [[ "$TRAVIS_CPU_ARCH" == "ppc64le" ]]; then
       sudo chown -Rfv $USER:$GROUP ~/.cache/pip/wheels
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,20 +16,51 @@ jobs:
...
+      sudo chown -Rfv $USER:$GROUP ~/.cache/pip/wheels
+    fi
+
 install:
 
 # ensure we have recent pip/setuptools/wheel
 - pip install --disable-pip-version-check --upgrade pip setuptools wheel
-# need tox to get started
 - pip install --upgrade tox tox-venv
+# need tox to get started
```
jaraco, line 63: Please restore the location of this comment.

.travis.yml
```diff
@@ -16,20 +16,51 @@ jobs:
...
   - <<: *latest_py3
     env: TOXENV=docs
+  - <<: *latest_py3_ppc
+    env: TOXENV=docs
+
 
 cache: pip
 
+before_install:
+  - |
```
jaraco, line 53: Why is this change necessary only on this platform? Can you add a comment explaining why it's here and if possible what defect requires it?

jaraco (approved):
Thanks for this change. Glad to have better platform test coverage.

Couple of things:

- Travis support might go away at some point. Do you know if the project can get similar platform coverage on Github Actions or Azure Pipelines?
- Looking at the test run, I can't discern from the build matrix in the UI which runs are ppc64le and which are x86. Is there a way to make the architecture visible in the report?
- Probably it's not necessary to test a full matrix of python versions on this platform. Probably just one or two Python versions are suitable to catch platform-specific concerns.

.travis.yml
ezeeyahoo: Travis base image for PPC64LE does not provide user level permission to ~/.cache/pip/wheels, result in build failure during pip install or tox trying to install packages. Therefore this is fix only for ppc64le and for travis build. Going forward, as and when fix will be made in build images, this fix would not create problem.

ezeeyahoo:
Thanks Jaraco for the comment
> 
> Couple of things:
> 
>     * Travis support might go away at some point. Do you know if the project can get similar platform coverage on Github Actions or Azure Pipelines?
> 
This project happens to be a part of the Red Hat distributions, including their ppc64le builds. We are enabling upstream CI for those to help reduce the cost of maintaining/debugging issues and to ensure the port stays current on ppc64le. We have very few that support Power today (jenkins, Travis, and possible github actions). We would like to stay in sync with what you are already doing as much as possible.
>     * Looking at the test run, I can't discern from the build matrix in the UI which runs are ppc64le and which are x86. Is there a way to make the architecture visible in the report?
We can identify x86 by AMD64 and likewise ppc64le in the second coloumn in the travis build matrix.
Another idea could be to create a separate branch for PPC64LE and X86 resp. and then show separate report for each like -
Arch: ppc64le [Build Status](https://travis-ci.com/ezeeyahoo/setuptools)
>     * Probably it's not necessary to test a full matrix of python versions on this platform. Probably just one or two Python versions are suitable to catch platform-specific concerns.
I have reduced the number of jobs to two stable(3.5 and 3.8) and 1 latest dev release, pypy support might come in future, find entry in allow_failures:. 

Thanks

.travis.yml
jaraco: @ezeeyahoo Can you address this one question, then we can merge?

ezeeyahoo: It is needed for PPC64LE only due to the fact that Travis linux base image for this platform lack proper permissions to the directory ~/.cache/pip/wheels that allows user running travis build to install pip packages which is otherwise allowed on linux images for amd64.

Would you like me to add inline comment through a new commit to the same PR?

ezeeyahoo:
Hey Jaraco,

Anything else, that I can do for this PR to go?

.travis.yml
```diff
@@ -16,14 +16,41 @@ jobs:
...
+    python: 3.5
+  - &latest_py3_ppc
+    arch: ppc64le
+    python: 3.8
+  - <<: *latest_py3_ppc
+    env: LANG=C
+  - arch: ppc64le
+    python: 3.9-dev
+  - <<: *latest_py3_ppc
+    env: TOXENV=docs
```
webknjaz, line 32: Is it really necessary to build the docs on PPC?

.travis.yml
```diff
@@ -16,14 +16,41 @@ jobs:
   - python: 3.9-dev
   - <<: *latest_py3
     env: TOXENV=docs
+    #PPC64LE
```
webknjaz, line 19: This comment is unnecessary because it's obvious in what the entries below do.

.travis.yml
```diff
@@ -16,14 +16,41 @@ jobs:
...
+  - arch: ppc64le
```
webknjaz, line 20: It's best to add job names so that it's visible in the UI as @jaraco asked earlier (https://github.com/pypa/setuptools/pull/2379#pullrequestreview-482091500).
Please add `name: ` to all the jobs you've added, similar to:
```suggestion
  - arch: ppc64le
    name: pypy3 @ ppc64le
```

.travis.yml
```diff
@@ -16,14 +16,41 @@ jobs:
...
   allow_failures:
   # suppress failures due to pypa/setuptools#2000
   - python: pypy3
+  - arch: ppc64le
+    python: pypy3
   - <<: *latest_py3
     env: TOXENV=docs
+  - <<: *latest_py3_ppc
+    env: TOXENV=docs
+
```
webknjaz, line 42: I think this should be reverted together with https://github.com/pypa/setuptools/pull/2379/files#r491708329

webknjaz (commented):
In addition to other comments, I think you should extend the `export` in the `after_success:` step to also include the arch.

.travis.yml
webknjaz: Ah, scratch that, I missed a comment about the second column in the UI.

webknjaz:
@ezeeyahoo plz also add a change fragment per https://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request

.travis.yml
jaraco: Yes, in the same PR. New commit or amend an existing commit.

jaraco: Ideally, this comment would also link to a bug filed with Travis that captures the need for this workaround so that someone stumbling on this code could find out if the workaround is still needed.

jaraco: Updated in d78027a25.

.travis.yml
jaraco: Nope. Removed.

.travis.yml
jaraco: Done.

.travis.yml
jaraco: Removed.

PR #2388, opened 2020-09-14
Enable flake8 and cov plugins only when installed.

jaraco:
Avoid installing in PyPy.


jaraco:
Does auto-merge mean nothing to you, Github?

PR #2397, opened 2020-09-23
Apply docs revamp

jaraco:
In #2093, alvyjudy was working on a revamp of the documentation. That (ambitious) effort has stalled out, but this branch remains. There are some latent issues with the documentation (warnings, duplication), but I'm tempted to merge the changes anyway to stop the divergence from master. Assuming the docs are readable in this form, I'm recommending we merge the changes and work out issues in subsequent iterations.


encukou:
This PR breaks existing URLs like https://setuptools.readthedocs.io/en/latest/distutils-legacy.html (mentioned [here](https://mail.python.org/archives/list/python-dev@python.org/message/HDCK3LFOZWDNWLH3NZ2QFNYUPNJAGNP5/) for example).

Would it be possible to add redirects, at least in Read the Docs?

PR #2413, opened 2020-10-04
ignore any exception when loading 2to3

graingert:
## Summary of changes

Related to https://github.com/pypa/setuptools/issues/2086

lib2to3 appears to have a problem loading the pickled pgen files, when running in parallel on pypy: https://github.com/twisted/ldaptor/runs/1205572060#step:12:103 (raises EOFError)

Instead of catching the `EOFError` I catch any `Exception`, as that seems cleaner

I've not added tests yet, as I'd like a concept ACK first, eg, If you'd rather I just remove all use of lib2to3 from the codebase I could do that instead.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -11,7 +11,7 @@ import stat
 
 try:
     from setuptools.lib2to3_ex import Mixin2to3
-except ImportError:
+except Exception:
 
     class Mixin2to3:
         def run_2to3(self, files, doctests=True):
--- a/changelog.d/2413.misc.rst
+++ b/changelog.d/2413.misc.rst
@@ -1,0 +1,1 @@
+Suppress EOF errors (and other exceptions) when importing lib2to3.


setuptools/command/build_py.py
```diff
@@ -11,7 +11,7 @@
 
 try:
     from setuptools.lib2to3_ex import Mixin2to3
-except ImportError:
+except Exception:
```
webknjaz, line 14: I think it's important to avoid catching broad exception classes because it may shadow future bugs that would be hard to debug. Besides, adding non-ImportError exceptions here would require a huge code comment explaining the motivation.

graingert: @webknjaz what about just `EOFError`?

webknjaz: Given that disutils is going to disappear from stdlib, it's probably reasonable to have ImportError too. Although I'm not entirely sure where the fix should be applied to so I'll leave that to @jaraco to decide.

graingert: Sorry, I mean just `ImportError, EOFError`

webknjaz: Yeah, I get what you mean. I just have a feeling that something is off but I can't tell what exactly.

graingert: How about a PR that vendors fissix?

jaraco:
I'm happy to accept this change as a stop-gap measure (bugfix release), but I'd especially like to see a PR removing lib2to3 dependency.

Issue #1698, opened 2019-02-22
Document setuptools.build_meta


pganssle:
There is currently *no documentation for `setuptools.build_meta`*!

We need to cover:

1. What it is - see [this three part series](https://www.bernat.tech/pep-517-and-python-packaging/) for more background on python packaging and PEP 517
2. How to use it - you need to add a `pyproject.toml` to your source root with the following configuration (implementer - please confirm the right versions to pin):
```toml
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build-meta"
```
3. How the semantics of a `setuptools.build_meta` differ from directly invoking `setup.py`; at the moment, the only deliberate difference is that with `setuptools.build_meta`, the source root is not on `sys.path` when you invoke it, meaning that you should either stop importing things from the local path, or add `sys.path.insert(0, os.dirname(__file__))` to the top of your `setup.py` file.
4. The fact that, if you are a `setup.cfg`-only project, when using `pyproject.toml` [*`setup.py` is now optional*](https://github.com/pypa/setuptools/pull/1675)!

We can cover `setuptools.build_meta:__legacy__` in this ticket or in a separate one.

ErikPohl-Lot49-Projects:
@pganssle Hello!  I'd like to look into this one.  Looks like a good challenge to take on!

pganssle:
@ErikPohl-Lot49-Projects Sounds great! This is very important. I will be happy to review as soon as you have a draft.

ErikPohl-Lot49-Projects:
@pganssle 
Thanks!  I'll let you know.

ErikPohl-Lot49-Projects:
@pganssle 
I'm going to find time Friday evening and this weekend to review the specifications.  I'll let you know if I have questions!

brettcannon:
I take it work on this has stalled? (I just tried to find the docs on PEP 517 support in setuptools and came up empty-handed and then found this issue.)

brettcannon:
And a note in the docs that skipping `setup.py` prevents editable installs probably wouldn't go amiss.

alvyjudy:
Hello! Just a trivial note, the build backend in the ``pyroject.toml`` should be ``build_meta`` instead of ``build-meta`` otherwise an error will be thrown

bhrutledge:
[...]

PR #2415, opened 2020-10-07
Link to build_meta documentation

alvyjudy:
## Summary of changes

I added a paragraph of text in the index that briefly mentions how setuptools works under pep517,  and provided a link to the build_meta document as mentioned in (#1698 ). 

At first I thought the quickstart has already covered this document, but judging from the discussion, it seems like it still deserves its own page. It appears that the quickstart should target the end users that are not concerned about this internal module.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/docs/userguide/index.rst
+++ b/docs/userguide/index.rst
@@ -10,6 +10,15 @@ packages.
 Packages built and distributed using ``setuptools`` look to the user like
 ordinary Python packages based on the ``distutils``.
 
+Transition to PEP517
+====================
+
+Since setuptools no longer serves as the default build tool, one must explicitly
+opt in (by providing a ``pyproject.toml`` file) to use this library. The user
+facing part is provided by `PEP517 <https://pypi.org/project/pep517/>`_ and
+backend interface is described :ref:`in this document <../build_meta>`. The
+quickstart provides an overview of the new workflow.
+
 .. toctree::
     :maxdepth: 1
 
--- a/docs/userguide/index.rst
+++ b/docs/userguide/index.rst
@@ -16,7 +16,7 @@ Transition to PEP517
 Since setuptools no longer serves as the default build tool, one must explicitly
 opt in (by providing a :file:`pyproject.toml` file) to use this library. The user
 facing part is provided by tools such as pip and
-backend interface is described :ref:`in this document <../build_meta>`. The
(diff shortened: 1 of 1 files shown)


docs/userguide/index.rst
```diff
@@ -10,6 +10,15 @@ packages.
 Packages built and distributed using ``setuptools`` look to the user like
 ordinary Python packages based on the ``distutils``.
 
+Transition to PEP517
+====================
+
+Since setuptools no longer serves as the default build tool, one must explicitly
+opt in (by providing a ``pyproject.toml`` file) to use this library. The user
```
webknjaz, line 17: Let's use a ```:file:``` role here:
```suggestion
opt in (by providing a :file:`pyproject.toml` file) to use this library. The user
```

docs/userguide/index.rst
```diff
@@ -10,6 +10,15 @@ packages.
...
+facing part is provided by `PEP517 <https://pypi.org/project/pep517/>`_ and
```
webknjaz, line 18: I'm pretty sure that this shouldn't point users to a specific implementation. There's plenty, including pip. I'd rather point to the PEP page using
```rst
:pep:`517`
```

Besides, the CLI present in the pep517 package has always been unstable and experimental, and it is being deprecated in favor of another project called `build` that is about to be accepted under the roof of PyPA: https://mail.python.org/archives/list/pypa-committers@python.org/thread/LNOO7Q2RW55BVYCJK2LP4DEIMHC5D44F/.

alvyjudy:
Thanks for the review. I have made the edits.

docs/userguide/index.rst
```diff
@@ -10,6 +10,15 @@ packages.
 Packages built and distributed using ``setuptools`` look to the user like
 ordinary Python packages based on the ``distutils``.
 
+Transition to PEP517
+====================
+
+Since setuptools no longer serves as the default build tool, one must explicitly
+opt in (by providing a :file:`pyproject.toml` file) to use this library. The user
+facing part is provided by tools such as pip and
+backend interface is described :ref:`in this document <../build_meta>`. The
```
webknjaz, line 19: This ref needs to use the doc role
```suggestion
backend interface is described :doc:`in this document <../build_meta>`. The
```

PR #2414, opened 2020-10-05
Fix ReST syntax error in the PEP 517 link

mgedmin:
I noticed a ReST formatting problem on the quickstart page:

Ekrano nuotrauka iš 2020-10-05 09-54-05

PR #2420, opened 2020-10-12
GHA: Replace 3.9 beta with 3.9 final

hugovk:
## Summary of changes

The CI is failing because it's failing to set up 3.9-beta via deadsnakes.

```
  /usr/bin/python3.9 -mvenv /home/runner/venv-3.9
  Error: Command '['/home/runner/venv-3.9/bin/python3.9', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
Error: The process '/usr/bin/python3.9' failed with exit code 1
```

* https://github.com/pypa/setuptools/runs/1239771433?check_suite_focus=true

Python 3.9.0 final is now available on GitHub Actions as `3.9`.

Let's add that, and we can remove the betas:

* `3.9.0-beta.4 - 3.9.0` from GHA
* `3.9-beta` from deadsnakes

I left in the `3.8-dev` and `3.9-dev` dev versions from deadsnakes. Still useful to keep them both?

I also left in the logic for checking `-beta`, will be useful for when 3.10 reaches beta (or to adapt for alpha). Let me know if you'd prefer it cleaned up.

---

Also we can see from the above that the single failing job caused all the other jobs to be auto-cancelled:

* https://github.com/pypa/setuptools/runs/1239771433?check_suite_focus=true

Instead, we can use `fail-fast: false` which allows the other jobs to run. For example:

* https://github.com/hugovk/setuptools/actions/runs/301588612

### Pull Request Checklist
- [x] Changes have tests [changes sort of _are_ tests]

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -16,6 +16,7 @@ jobs:
       ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       # max-parallel: 5
       matrix:
         python-version:
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -16,7 +16,6 @@ jobs:
       ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
       # max-parallel: 5
       matrix:
         python-version:
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -19,6 +19,7 @@ jobs:
       # max-parallel: 5
       matrix:
         python-version:
+        - 3.9
         - 3.8
         - pypy3
         - 3.7
@@ -31,12 +32,6 @@ jobs:
         # - windows-2019
         # - windows-2016
         include:
-        # Pre-release versions (GH-shipped)
-        - os: ubuntu-20.04
-          python-version: 3.9.0-beta.4 - 3.9.0
-        # Pre-release versions (deadsnakes)
-        - os: ubuntu-20.04
-          python-version: 3.9-beta
         # Dev versions (deadsnakes)
         - os: ubuntu-20.04
           python-version: 3.9-dev
--- a/changelog.d/2420.misc.rst
+++ b/changelog.d/2420.misc.rst
@@ -1,0 +1,1 @@
+Replace Python 3.9.0 beta with 3.9.0 final on GitHub Actions.


.github/workflows/python-tests.yml
```diff
@@ -16,9 +16,11 @@ jobs:
       ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
webknjaz, line 19: I wouldn't do this

webknjaz:
I think it's worth adding a misc change note here

changelog.d/2420.misc.rst
```diff
@@ -0,0 +1 @@
+Replace Python 3.9.0 beta with 3.9.0 final on GitHub Actions.
```
webknjaz, line 1: FWIW I'm trying to make the change notes explanation more illustrative in another PR: https://github.com/pypa/setuptools/pull/2395/files#diff-b0a8c7744649eb7f19a8eb9ccee0eb7bR1.

.github/workflows/python-tests.yml
```diff
@@ -31,12 +33,6 @@ jobs:
         # - windows-2019
         # - windows-2016
         include:
-        # Pre-release versions (GH-shipped)
-        - os: ubuntu-20.04
-          python-version: 3.9.0-beta.4 - 3.9.0
-        # Pre-release versions (deadsnakes)
-        - os: ubuntu-20.04
-          python-version: 3.9-beta
```
webknjaz, line 39: It's funny that I was just about to submit exactly this change right before seeing your PR. So I ended up sending only the metadata update: https://github.com/pypa/setuptools/pull/2421.

.github/workflows/python-tests.yml
hugovk: For much slower CIs and much busier repos, that can be necessary, but it means other PRs (e.g. https://github.com/pypa/setuptools/pull/2414) don't get any results from GHA, and I think's usually it's more useful to get that info:

image

For example, there might be a change that passes on all versions but is incompatible with Python 3.9. It's more useful to see 3.9 fail and the rest pass, to get better feedback. If 3.9 fails and the rest are cancelled, we don't know whether it's only a problem with 3.9 or problem for all.

But happy to remove it if not wanted.

webknjaz: Yeah, I'm often presented with this dilemma too. I guess it's up to the maintainers. I've only wanted to point out that it takes ~10min to complete all jobs (https://github.com/pypa/setuptools/actions/runs/292445192) and you can still get most of that info from other CIs.

webknjaz:
By the way, the original problem was caused by the upstream bug:
https://github.com/deadsnakes/issues/issues/138#issuecomment-707178218

webknjaz (approved):

webknjaz:
@jaraco this PR is clear to be merged w/o any extra reviews. We need to get this done ASAP before anything else in order to fix the CI that makes all of the PR submissions look broken while it's broken because it points to a deadsnakes release that is no longer there.

.github/workflows/python-tests.yml
jaraco: My instinct is to use the default behavior unless there's a compelling reason to deviate.

jaraco:
```suggestion
```

Issue #2424, opened 2020-10-15
Sphinx setup should be stricter


webknjaz:
I noticed that some of the docs pages are unreachable when navigating from the main RTD page. In particular, _I know_ that there's `history.rst` that is only accessible if one knows the URL upfront.

I tracked this to https://github.com/pypa/setuptools/pull/2097 which removes entries from the TOC but doesn't reintroduce them in other places.

Sphinx has a few toggles that make it nitpicky about warnings. I think this should be enabled in the CI to prevent such problems in the future. This should catch implicit orphan pages as well as dead references or typos.

alvyjudy:
I take the blame for a lot of breaking links as I ran out of steam in the middle of the docs revamp. I think jaraco felt that the halfway efforts still made a readable documentation and decided to merge them anyways, and fix the issues subsequently. See (here)[https://github.com/pypa/setuptools/issues/2093#issuecomment-697985806] for the context

I do plan on making more changes here and there to fix those issues and am happy to see the doc is receiving more supports and attention:)

webknjaz:
@alvyjudy please hold on with your PRs. I've got a huge change fixing many refs and enforcing the checks incoming. Let's merge it first so that the CI would verify any further changes automatically.

webknjaz:
Here you go: https://github.com/pypa/setuptools/pull/2427.

PR #2111, opened 2020-05-14
doc overhaul step 3: update userguide

alvyjudy:
#2093  

### Summary of changes

The userguide (originally the setuptools.txt document) is now one step closer to reader-friendliness. It is comprised of three parts:

- an ``index`` that provide an introduction and roadmap
- a ``quickstart`` that provide guidance (minimal knowledge) on the commonly used functionalities (e.g. entry point, automatic package discovery, package dependencies, etc.), each accompanied with a link to a dedicated file for advance use and technical details
- multiple dedicated files for each section introduced in ``quickstart`` and other less commonly used tools

Again, most edits were still cut-and-paste (so the content still assumes the traditional ``setup.py`` usage. It might still take a while before some real edits can take place.

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

### The ``userguide`` directory layout

```
docs/userguide/
    index.txt
    quickstart.txt
    package_discovery.txt
    entry_point.txt
    dependency_management.txt
    datafiles.txt
    development_mode.txt
    distribution.txt
    extension.txt
    declarative_config.txt
    keywords.txt
    commands.txt
```

### Example section in ``quickstart.txt``
image

PR #2426, opened 2020-10-15
Group development related files together

alvyjudy:
## Summary of changes

Built on top of  #2425 , I modified the file position for development related docs and referenced them from the top index:
```
docs/
    index.rst (toc tree now points to development/index.rst)
    development/
        index.rst (originally development.rst)
        developer-guide.rst
        releases.rst
        history.rst
```

Note: I think this means the github PR prompt will also have to be modified as it points toward the developer-guide ( https://setuptools.readthedocs.io/en/latest/developer-guide.html) that is now in a different position

### Pull Request Checklist
- [ ] Changes have tests


docs/index.rst
```diff
@@ -10,5 +10,5 @@ Documentation content:
    :maxdepth: 1
 
    User guide <userguide/index>
-   Development guide <development>
+   Development guide <development/index>
```
webknjaz, line 13: I'm pretty sure that moving the dev guide into a subfolder will break the inclusion of the fragments guide that resides under `changelog.d/`. So this may need to be corrected.

docs/development/index.rst
```diff
@@ -33,3 +33,4 @@ setuptools changes.  You have been warned.
    developer-guide
    formats
    releases
+   history
```
webknjaz, line 36: https://github.com/pypa/setuptools/pull/2425#issuecomment-709554793

webknjaz: Also, there's the same problem as in  https://github.com/pypa/setuptools/pull/2426/files#r505800974

webknjaz:
@alvyjudy

> Note: I think this means the github PR prompt will also have to be modified as it points toward the developer-guide ( setuptools.readthedocs.io/en/latest/developer-guide.html) that is now in a different position

In this case, the change to `.github/pull_request_template.md` should be included in this PR so that it's atomic.
Another question that arises with relation to nesting: wouldn't it be better to switch to `dirhtml` Sphinx builder. It'd produce nicer URLs on RTD and would help with moves like this. cc @jaraco

P.S. This probably deserves a change fragment: https://github.com/pypa/setuptools/tree/master/changelog.d.

alvyjudy:
@webknjaz regarding the changelog, @jaraco [mentioned](https://github.com/pypa/setuptools/pull/2111#issuecomment-629157270) that it might not be necessary to put changelog to different phases of doc  I'm not sure if we should add one here?

webknjaz:
@alvyjudy the changelog is indeed optional for non-essential changes but it's my personal opinion that it's nice to have it most of the time. It's totally up to you, though.

webknjaz:
I've also noticed that `format` got renamed into `deprecated/python_eggs` but is still linked as "format" in TOC in `development.rst`. I'm trying to set up a strict mode for #2424 and discovered this among other missing references. Your move of `development.rst` in this PR is also going to break some of the links so please check the toctree in the new location, you'll probably need to take care of some other entries there too.

webknjaz:
@alvyjudy hey, my PR has been accepted and yours now has conflicts. I'm not planning to submit any more substantial changes in the near future so it's safe for you to solve the conflicts/do rebase. Please also note that you'll probably need to update the PR description to match what's actually in the PR ✌️

PR #2437, opened 2020-10-18
Switch to RTD v2 config

jaraco:


jaraco:
This approach is a trimmed-down version of #2433 that attempts to take on one change at a time (adopting v2 config) while also attempting to keep the config as trim as possible.

jaraco:
> The rtd-bot is activated, but no .github/config.yml found in this repository.
> Make sure that you have it in your default branch.

I tried activating this app for pypa for all repos (as it was for virtualenv only), but that didn't have the intended effect and I don't think it's the right app, so I switched it back to virtualenv only.

webknjaz:
> I tried activating this app for pypa for all repos (as it was for virtualenv only), but that didn't have the intended effect and I don't think it's the right app, so I switched it back to virtualenv only.

It's third-party and outdated. RTD don't have a GitHub App, they use per-repo webhooks to get events about the PRs.

To enable PR builds, toggling just one checkbox (per project) is necessary: https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html.

webknjaz (commented):
@jaraco I see you removed `formats` and it changed the setting on RTD. PDF/htmlzip/epub are no longer present at https://setuptools.readthedocs.io/en/latest/. Was it intentional?

jaraco:
I don't think the other formats were ever intentional. I suspect they were just generated as defaults. Now the defaults are just html, I guess.

PR #2439, opened 2020-10-19
s/517/PEP 517/

mgedmin:
https://setuptools.readthedocs.io/en/latest/userguide/quickstart.html today reads

> The change is driven under 517.

and I think that should be "PEP 517".


webknjaz (approved):

jaraco:
Agreed it should render as `PEP 517`. Shame PEP has to be written twice.

PR #2464, opened 2020-11-28
Add type stub file for setup

### LINKED ISSUE #2345: Add type hints to setuptools
It seems that setuptools doesn't have type hints at the moment:

```
4: error: Skipping analyzing 'setuptools': found module but no type hints or library stubs
4: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

I think that adding type hints at least to the most common entrypoints (e.g. ``setup``) would be great to have, considering that setuptools is quite an important part of the official Python ecosystem.

GergelyKalmar:
## Summary of changes

This PR adds a stub file for the `setup()` function. I've been mostly relying on the [setuptools keywords documentation](https://setuptools.readthedocs.io/en/latest/references/keywords.html) to derive the type hints. I left out the deprecated arguments and the arguments related to Python 2 (the catch-all argument will allow using these anyways). I also marked `version` optional mostly because it can be left out when using tools like [setuptools_scm](https://pypi.org/project/setuptools-scm/).

I'm not quite sure how to write tests for module stub files, so suggestions / additions in this area are welcome.

Closes #2345

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added in `changelog.d`


webknjaz:
Looks like the failing test is unrelated:
```console
stdout = b"/home/runner/work/setuptools/setuptools/.tox/python/bin/python: can't open file 'setup.py': [Errno 2] No such file or directory\n"
```

webknjaz (changes requested):
I like the idea of having the type hints but I think that the implementation should be slightly different: there's no need for separate stub files, just add the typing info right into https://github.com/pypa/setuptools/blob/5cf3865/setuptools/__init__.py#L150 because setuptools requires Python 3.6 (https://github.com/pypa/setuptools/blob/5cf3865/setup.cfg#L48)

GergelyKalmar:
@webknjaz Thanks for the feedback! We could do that, but it would be a bit less elegant in my opinion, mostly because https://github.com/pypa/setuptools/blob/5cf3865/setuptools/__init__.py#L150 forwards its arguments to both `_install_setup_requires` and `distutils.core.setup`, which means that we would need to list the arguments three times (instead of having them just listed once in the stub and using `**attrs` in the actual function). I'm not sure if there's an elegant workaround for this, but if there isn't perhaps maintaining this in the stub file is not the worst idea.

The tests do seem to fail on the plain master branch too, I think they're unrelated. We could add `mypy` and write some test files to be type-checked though for sure, but there's perhaps a more elegant way to do this too.

webknjaz:
I think that keeping a separate stub in sync is harder than having it copy-pasted in code. Also, the copy-paste could be partially eliminated with a decorator if you think that it's that critical.
Anyway, it's up to @jaraco to decide. You should probably wait for his opinion to land before doing anything.

P.S. It's more important to work on making master green now. To eliminate the misleading PR statuses.

GergelyKalmar:
It's fine for me either way – I've chosen this path because that's what was suggested in the original issue discussion. I'm happy to change if @jaraco thinks it's worthwhile.

setuptools/__init__.pyi
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Mapping, Optional, Sequence, Union
+
+import setuptools
+
+
+def setup(
+    name: str,
```
jaraco, line 7: Name is also optional (to supply). Consider [this usage](https://github.com/jaraco/skeleton/blob/main/setup.py), which I recommend. All parameters are specified in setup.cfg.

setuptools/__init__.pyi
```diff
@@ -0,0 +1,44 @@
...
+    version: Optional[str] = None,
```
jaraco, line 8: FYI, version is allowed to be a Number (float). That usage is discouraged, but hasn't yet been deprecated. Definitely don't want to advertise that fact, though, so perhaps this approach is best and violators can use `type:ignore`.

setuptools/__init__.pyi
```diff
@@ -0,0 +1,44 @@
...
+    description: Optional[str] = None,
+    long_description: Optional[str] = None,
+    long_description_content_type: Optional[str] = None,
+    author: Optional[str] = None,
+    author_email: Optional[str] = None,
+    maintainer: Optional[str] = None,
+    maintainer_email: Optional[str] = None,
+    url: Optional[str] = None,
+    download_url: Optional[str] = None,
+    packages: Optional[Sequence[str]] = None,
```
jaraco, line 18: Some of these sequences allow newline- or comma- or space-separated strings. Some may be particular to tuples or lists. Probably specifying a sequence here is best though.

setuptools/__init__.pyi
jaraco, line 8: I'm a little concerned about the "default" value of None. Will this signal to the consumer that passing `None` is the same as not passing the argument? I should think so, but the code expects the parameters to be absent ([example](https://github.com/pypa/setuptools/blob/b6bbe236ed0689f50b5148f1172510b975687e62/setuptools/_distutils/core.py#L100-L103)). Perhaps what's needed is keyword-only parameters with no defaults? e.g.:

```suggestion
    *,
    version: str,
```

jaraco (changes requested):
> We could add `mypy` and write some test files to be type-checked though for sure, but there's perhaps a more elegant way to do this too.

In the feature/skeleton branch, I'm planning to merge jaraco/skeleton into this project, which adds mypy support. Perhaps it would be worth landing that before landing this? Or you could merge that branch with this one and test the two together.

> It's more important to work on making master green now. To eliminate the misleading PR statuses.

Agreed. And done in #2472.

> We could [use inline annotations], but it would be a bit less elegant in my opinion, mostly because https://github.com/pypa/setuptools/blob/5cf3865/setuptools/__init__.py#L150 forwards its arguments to both `_install_setup_requires` and `distutils.core.setup`, which means that we would need to list the arguments three times (instead of having them just listed once in the stub and using `**attrs` in the actual function).

I also agree it would be preferable to use inline annotations. If there are concerns about the underlying implementation (usage), that seems problematic for the `typing` implementation. I'd be okay with having a wrapper for setup, e.g.:

```
def setup(*, name: str, ...):
    attrs = ...
    _setup(**attrs)
```

As I think about it more, I'm not sure it's even possible to declare accurately the types for this function. It might be that one would need to specify a [PEP 589](https://www.python.org/dev/peps/pep-0589/) TypedDict or similar.

That said, I'm okay with a `.pyi` file for now and transition to an inline definition as the type system allows.

setuptools/__init__.pyi
GergelyKalmar: That is an excellent point, I'll change this in a moment.

setuptools/__init__.pyi
GergelyKalmar: Yes, some fields seemed a little convoluted - once we have mypy tests added to the test suite (and assuming there are tests for those invocations) we should see whether these type definitions are good enough.

setuptools/__init__.pyi
GergelyKalmar: Hm, that is a good point. I think making them keyword-only parameters without defaults will be an issue though because `mypy` will warn the user about the missing argument (e.g. `error: Missing named argument "version" for "setup"  [call-arg]`). Note that the stub files are only read by `mypy` therefore this doesn't change the way `setup()` otherwise works and I think from the current documentation it should be clear how `setup()` should be used.

The point about making the arguments keyword-only is valid though, I'll change that. Somehow the ending ellipsis went missing too during the process so I'll add that as well.

GergelyKalmar:
> 
> > We could [use inline annotations], but it would be a bit less elegant in my opinion, mostly because https://github.com/pypa/setuptools/blob/5cf3865/setuptools/__init__.py#L150 forwards its arguments to both `_install_setup_requires` and `distutils.core.setup`, which means that we would need to list the arguments three times (instead of having them just listed once in the stub and using `**attrs` in the actual function).
> 
> I also agree it would be preferable to use inline annotations. If there are concerns about the underlying implementation (usage), that seems problematic for the `typing` implementation. I'd be okay with having a wrapper for setup, e.g.:
> 
> ```
> def setup(*, name: str, ...):
>     attrs = ...
>     _setup(**attrs)
> ```
> 
> As I think about it more, I'm not sure it's even possible to declare accurately the types for this function. It might be that one would need to specify a [PEP 589](https://www.python.org/dev/peps/pep-0589/) TypedDict or similar.
> 

Hm, that would possibly be the most accurate option. Wouldn't that work only with Python 3.8+ though?

GergelyKalmar:
With regards to the `mypy` support, I can surely merge that branch in too. Is it ready to be integrated?

setuptools/__init__.pyi
GergelyKalmar: I've just found out in the [typeshed coding style guide](https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#stub-file-coding-style) that an ellipsis is generally preferred for arguments with a default, therefore I changed `None` to `...` everywhere. This should also solve the issue with the misleading `None` showing up as the default.

jaraco:
> With regards to the `mypy` support, I can surely merge that branch in too. Is it ready to be integrated?

It’s not ready to be integrated with main or this PR. I was suggesting to merge temporarily or to another branch to test.

jaraco:
The mypy support is now integrated, but not enabled. Running the tests with `tox -- --mypy` will run the tests, but the code has many failures. I'm not sure how helpful the plugin will be for this effort until the failures are addressed.

webknjaz:
Time to rebase or close?

PR #2465, opened 2020-11-29
Change documentation theme to Furo

pradyunsg:
I'll add a changelog entry, if we have agreement that this is the direction we want to take. :)


webknjaz (approved):
:+1: I like the idea of having a unified docs theme across the PyPA projects.

@jaraco: The ToCs are removed because the `furo` theme adds its own ToC on each page, in the right sidebar, and is opinionated about not having it duplicated in the corresponding document body.

@pradyunsg: Is https://github.com/pradyunsg/furo/commit/b6c7c76bc1a5aa8f513eafbc5bd87b655a8e801f released? If yes, you could close+reopen this PR to retrigger Netlify and have it build the preview.

jaraco:
> Is pradyunsg/furo@b6c7c76 released? If yes, you could close+reopen this PR to retrigger Netlify and have it build the preview.

I just closed, re-opened, and still had to manually re-deploy, but the error is emitted:

```
12:20:18 PM: Warning, treated as error:
12:20:18 PM: the furo extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
ERROR: InvocationError for command /opt/build/repo/.tox/docs/bin/python -m sphinx -j auto -b html --color -a -n -W -d /opt/build/repo/.tox/.tmp/.doctrees . /opt/build/repo/build/html (exited with code 2)
```

That suggests the fix is not released. Also, that commit is not present in any tags.

pradyunsg:
> @pradyunsg: Is [pradyunsg/furo@b6c7c76](https://github.com/pradyunsg/furo/commit/b6c7c76bc1a5aa8f513eafbc5bd87b655a8e801f) released? If yes, you could close+reopen this PR to retrigger Netlify and have it build the preview.

Not yet, no.

pradyunsg:
@jaraco Try redeploying now?

webknjaz:
@pradyunsg looks like netlify doesn't do rebuilds in reopen. Try rebasing the PR.

pradyunsg:
This removed the tidelift contents from the sidebar.

It'd probably be a good idea to mimic how urllib3 handles the Tidelift callout: https://urllib3.readthedocs.io/en/latest/_sources/index.rst.txt, https://urllib3.readthedocs.io/en/latest/. Let me know if that's desirable and I'll amend this PR for that. :)

docs/index.rst
```diff
@@ -1,13 +1,12 @@
-Documentation
-=============
+setuptools
```
webknjaz, line 1: How about 
```suggestion
|project|
```

jaraco:
Happy to mimic urllib3. I'd like to apply the changes to every tidelift-sponsored project. Let's apply the concept to [jaraco/tidelift](https://github.com/jaraco/tidelift) first, then I can merge the change into setuptools main.

docs/index.rst
jaraco: Have you tried running with that substitution? For me, it causes the build to fail with `/Users/jaraco/code/public/pypa/setuptools/docs/index.rst:4:Undefined substitution referenced: "project".`

jaraco:
I've started working on generalizing the urllib3 technique in jaraco/tidelift#2, but I'm yet unsuccessful in substituting the project name, so could use some help there.

docs/index.rst
pradyunsg: You need to set project in conf.py for that to work.

jaraco: Hmm. That value is [set in config by jaraco.packaging.sphinx](https://github.com/jaraco/jaraco.packaging/blob/38e707641eac9ccfe5f6b8679894dc68a000df20/jaraco/packaging/sphinx.py#L37-L40). My experience has been that setting the property there was equivalent to setting it in conf.py, but perhaps there's another thing that has to happen.

pradyunsg: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#default-substitutions

Aha, project isn't a default substitution in Sphinx. :)

I shouldn't have assumed.

webknjaz: Oh, I usually declare it so I forgot that it's not there by default

webknjaz (approved):

webknjaz:
This PR needs a rebase and here's the demo of solving the Tidelift problem: https://github.com/pypa/setuptools/pull/2708.

webknjaz:
@pradyunsg plz rebase this — it should be unlocked now.

PR #2468, opened 2020-12-04
fix broken link to entry-points details

wimglenn:
See https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points

`For more details, see the [setuptools documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins).`

The link's target is orphaned, and the link itself also does not render correctly (using markdown in a restructured text file)

### DIFF
--- a/docs/pkg_resources.rst
+++ b/docs/pkg_resources.rst
@@ -703,7 +703,7 @@ entry point group and look for entry points named "pre_process" and
 To advertise an entry point, a project needs to use ``setuptools`` and provide
 an ``entry_points`` argument to ``setup()`` in its setup script, so that the
 entry points will be included in the distribution's metadata.  For more
-details, see the [``setuptools`` documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins).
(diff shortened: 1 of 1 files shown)


docs/pkg_resources.rst
```diff
@@ -703,7 +703,7 @@ entry point group and look for entry points named "pre_process" and
 To advertise an entry point, a project needs to use ``setuptools`` and provide
 an ``entry_points`` argument to ``setup()`` in its setup script, so that the
 entry points will be included in the distribution's metadata.  For more
-details, see the [``setuptools`` documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins).
+details, see `Advertising Behavior <https://setuptools.readthedocs.io/en/latest/userguide/entry_point.html#dynamic-discovery-of-services-and-plugins`>_.
```
webknjaz, line 706: To prevent this from happening in the future, it's best to use a reference to the corresponding RST label instead of a hardcoded URL, letting Sphinx figure out the actual link on build:
```suggestion
details, see :ref:`Advertising Behavior<dynamic discovery of services and plugins>`.
```

webknjaz (changes requested):
I like the change but please apply my suggestion to make it more Sphinx-native.

webknjaz (approved):

PR #2483, opened 2020-12-12
Fix repeated command line options handling in distutils.FancyGetopt

### LINKED ISSUE #2482: distutils.FancyGetopt does not hadle repeating options with argument
`setuptools` inherits command line parsing from `distutils` which uses the `distutils.FancyGetopt` class for setting up and performing the parsing. Documentation and comments in the code suggest that it exists a way to specify command line options that can be repeated. However, trying to use this facility for options that take arguments results in only the last specified argument to be retained. It works as expected for boolean options. I believe this is a bug that went so far unnoticed because the code base uses this feature only for the `--verbose` option which does not take arguments.

I know setuptools is moving away from command line invocation and toward a more declarative style. Is there any motivation to fix this? The patch for making `distutils.FancyGetopt.getopt()` do the right thing for repeated options with argumentsis just a few lines and I can submit a PR.

I currently monkey patch the method in my project. Some more monkey patching when it comes to setuptools does not even look that bad ;)

dnicolodi:
Documentation and comments in the code suggest that it should be possible to specify command line options that can be repeated. However, trying to use this facility for options that take arguments results in only the last specified argument to be retained. It works as expected for boolean options. Probably this bug went so far unnoticed because this functionality is used in setuptools or distutils only for the ``--verbose`` option which does not take arguments.

Fixes #2482

### DIFF
--- a/setuptools/_distutils/fancy_getopt.py
+++ b/setuptools/_distutils/fancy_getopt.py
@@ -245,20 +245,30 @@ class FancyGetopt:
             if alias:
                 opt = alias
 
-            if not self.takes_arg[opt]:     # boolean option?
+            if not self.takes_arg[opt]:
+                # boolean option
                 assert val == '', "boolean option can't have value"
                 alias = self.negative_alias.get(opt)
                 if alias:
                     opt = alias
-                    val = 0
+                    val = -1
                 else:
                     val = 1
 
             attr = self.attr_name[opt]
-            # The only repeating option at the moment is 'verbose'.
-            # It has a negative option -q quiet, which should set verbose = 0.
-            if val and self.repeat.get(attr) is not None:
-                val = getattr(object, attr, 0) + 1
+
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for the contrib. Unfortunately, everything under the _distutils name needs to be contributed to [pypa/distutils](/pypa/distutils). Also, I'll probably resist accepting non-critical fixes until setuptools is generally using distutils by default, an effort which is blocked on some complicated ecosystem challenges.

dnicolodi:
Thank you. I didn't see that repo. I thought the `distutils` standard library module was merged into `setuptools`. Should I open a ticket and a pull request there or is `distutils` development effectively halted as you imply?

jaraco:
I don't have a hard and fast rule. I suggest go ahead and submit this PR to that repo. If I don't accept it, I'll probably just leave it to linger until distutils adoption has moved forward.

PR #2486, opened 2020-12-12
Adopt jaraco/skeleton

jaraco:
This change incorporates the best practices [developed and documented at jaraco/skeleton](/jaraco/skeleton). By adopting these changes, Setuptools can inherit the best practices and periodically merge improvements. Highlighted changes include:

- Simple GHA recipe for Linux, Windows, and macOS coverage.
- Drop support for Travis, Appveyor, and Azure Pipelines in favor of GHA.
- Integrated code checks, including flake8, coverage, black, and mypy.

Changes excluded:

- setuptools_scm is not used due to bootstrapping issues.


tox.ini
```diff
@@ -45,20 +34,11 @@ commands=codecov -X gcov --file {toxworkdir}/coverage.xml
...
+	testing
 changedir = docs
 commands =
-        {envpython} -m sphinx \
-          -j auto \
-          -b html \
-          --color \
-          -a \
-          -n \
-          -W \
```
webknjaz, line 58: this change kills strict validation

tox.ini
```diff
@@ -45,20 +34,11 @@ commands=codecov -X gcov --file {toxworkdir}/coverage.xml
...
-          -d "{temp_dir}/.doctrees" \
```
webknjaz, line 59: keeping this cache allows faster rebuilds

tox.ini
```diff
@@ -45,20 +34,11 @@ commands=codecov -X gcov --file {toxworkdir}/coverage.xml
...
 [testenv:docs]
 extras =
-        docs
-        testing
+	docs
+	testing
 changedir = docs
 commands =
-        {envpython} -m sphinx \
-          -j auto \
```
webknjaz, line 53: parallelism makes things faster

tox.ini
```diff
@@ -72,21 +52,22 @@ commands =
 [testenv:release]
 skip_install = True
 deps =
-	wheel
+	pep517>=0.5
```
webknjaz, line 55: Today it's already safe to have the python-build CLI instead, especially, given that pep517 drops its own CLI helpers.

pytest.ini
```diff
@@ -1,8 +1,15 @@
...
 norecursedirs=dist build .tox .eggs
-doctest_optionflags=ELLIPSIS ALLOW_UNICODE
-filterwarnings =
+addopts=
+	--doctest-modules
+	--doctest-glob=pkg_resources/api_tests.txt
+	-r sxX
+doctest_optionflags=ALLOW_UNICODE ELLIPSIS
+# workaround for warning pytest-dev/pytest#6178
+junit_family=xunit2
```
webknjaz, line 9: Do you know that xunit2 drops some of the fields available in xunit1? I've learned that using xunit1 allows representing the test results better.

pytest.ini
```diff
@@ -1,8 +1,15 @@
...
+filterwarnings=
+    # https://github.com/pytest-dev/pytest/issues/6928
+    ignore:direct construction of .*Item has been deprecated:DeprecationWarning
```
webknjaz, line 12: Shouldn't ignores go after the `error` line? I think the order matters.

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+name: Automated Tests
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        python: [3.6, 3.8, 3.9]
+        platform: [ubuntu-latest, macos-latest, windows-latest]
```
webknjaz, line 10: This needs to be revisited as it kills many platforms.

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Install tox
+        run: |
+          python -m pip install tox
+      - name: Run tests
+        run: tox
```
webknjaz, line 22: It's preferable to separate a `--notest` invocation to make the failure reasons easier to distinguish.

.flake8
```diff
@@ -1,12 +1,14 @@
...
-    setuptools/site-patch.py F821
-    setuptools/py*compat.py F811
+	# W503 violates spec https://github.com/PyCQA/pycodestyle/issues/513
+	W503
+	# W504 has issues https://github.com/OCA/maintainer-quality-tools/issues/545
+	W504
+	# Black creates whitespace before colon
+	E203
+	setuptools/site-patch.py F821
+	setuptools/py*compat.py F811
```
webknjaz, line 14: I'm pretty sure `ignore` doesn't work like this. You need to use `per-file-ignores` for this to make sense to flake8

.flake8
```diff
@@ -1,12 +1,14 @@
 [flake8]
-exclude=
-    .tox
-    setuptools/_vendor,
-    pkg_resources/_vendor
+max-line-length = 88
+exclude =
+	setuptools/_vendor
+	pkg_resources/_vendor
 ignore =
```
webknjaz, line 6: It's preferable to use `extend-ignore` so that instead of overriding flake8's default ignores, you'd just append values to that list. This allows us to stop re-adding 503/504 in every project.

webknjaz: https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py#L84

.flake8
domdfcoding: The docs certainly don't mention this usage: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore

webknjaz: I think there was `per-file-ignores =` line at some point in the past above these lines and it probably got removed by accident.

.flake8
jaraco: oh, nice. I was unaware of extend-ignore. That would be a great improvement to jaraco/skeleton. Filed as jaraco/skeleton#28.

.flake8
jaraco: I don't recall, but it falls outside the scope of this effort. I welcome others to investigate and improve in a separate PR.

.github/workflows/main.yml
jaraco: Indeed, it drops support for Python 3.7 and PyPy. It also might drop support beta releases. I wouldn't mind re-introducing pypy support as long as it can be done without overly-complicating the recipe. I'll experiment with that. I omitted Python 3.7 in order to limit the redundant computation required and based on the assumption that if it works on Python 3.6 and 3.8, it almost certainly also works on 3.7.

.github/workflows/main.yml
jaraco: In my experience, the tox output provides a suitable separation of concerns, and the added complexity is rarely worth it. If tox is not providing enough context to easily distinguish the setup phase from the run phase, then I'd say that should be escalated to tox to improve the experience. I'm disinclined to accept the burden of micro-managing the tools' experience across any number of projects. If there's a compelling story here, surely it should be applied at the tool level and not cargo-culted.

pytest.ini
jaraco: I was unaware. Thanks for the reference... and thanks for filing the issue upstream. My instinct is to rely on the degraded behavior of the newer system either until it's improved or until it's deprecated in favor of the prior implementation. Falling back to xunit1 adds technical debt that will likely require eventually switching to xunit2. I'd like to avoid that extra work across these projects.

And more importantly, Setuptools (and probably many projects) already are emitting xunit2 as part of the Pytest 6 release, so this change is a no-op compared to prior runs, so I'm going to defer this as a concern outside this PR.

pytest.ini
jaraco: The "fail on warnings" behavior is project-specific whereas the 'ignore "direct construction of..."' is part of the upstream skeleton. Interleaving these changes would make reconciling changes more difficult.

I tried to understand if or how order matters, so I looked at [the docs](https://docs.python.org/3/library/warnings.html#the-warnings-filter), which say

> any specific warning is matched against each filter specification in the list in turn until a match is found

That suggests that putting `error` at the front would cause all warnings to become errors and no ignores would take effect. Obviously, that's not what's happening, so it seems we can't trust the docs.

So, I'm not going to worry about it too much, especially since that upstream filter can (and will) likely  be removed soon if not today.

tox.ini
jaraco: Perhaps, though it may also be a premature optimization. If it's a generally-useful feature, then I'd rather see it implemented as the default in the builder and otherwise rely on defaults where the impact is small. I just ran the builds locally, and they complete in 11.2s for parallelized and 11.5s otherwise. So, I've already wasted more of my life evaluating the feature than enabling it will ever save :(. YAGNI.

tox.ini
jaraco: This feature is probably worth retaining (done in a283c8776). I'll consider applying this to all projects (jaraco/skeleton#29).

tox.ini
jaraco: I'm not too worried about this. Seems like another premature optimization.

tox.ini
jaraco: Yes, probably. And that should be dealt with outside the scope of this PR. jaraco/skeleton#30

.github/workflows/main.yml
jaraco: Looks like it's possible to simply add pypy3 to the list of python versions, so I'll go with that for now (176622446).

.github/workflows/main.yml
webknjaz: The reason I prefer this is that I can see it separately in the CI steps. And when I open the testing step, it won't start with the `pip install` logs but rather with what actually matters. This largely improves the CI UX.
Also, in the past, there was a nasty tox bug that looked like the test invocation was failing, hiding that the real reason was the install step. Making it separate allowed to actually identify what the problem was.
Besides, this is an idiom that the tox maintainers themselves follow: https://github.com/tox-dev/azure-pipelines-template/blob/30f07f8/run-tox-env.yml#L143.

.flake8
webknjaz: Sure, it's an external effort. I just wanted to point out that these lines most likely don't have any effect at all.

.github/workflows/main.yml
webknjaz: Yep, and they've recently made it possible to separately request pypy3.6 and pypy3.7 which may be useful too.

pytest.ini
webknjaz: It sounds like you're right. Although I'm a bit confused that the docs imply something different.

pytest.ini
webknjaz: it sounds unlikely that junit2 will get that metadata because the underlying standard doesn't have this info

tox.ini
webknjaz: The docs here are quite small but for the skeleton, it'd be a good idea to have it on.

.flake8
```diff
@@ -1,12 +1,14 @@
 [flake8]
-exclude=
-    .tox
```
webknjaz, line 3: Removing this seems to have made the fs scan step unforgivably slow.

Using `extend-exclude` should fix this: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-extend-exclude.

.flake8
```diff
@@ -1,12 +1,14 @@
...
-    setuptools/_vendor,
```
webknjaz, line 4: Removing the trailing comma seems wrong: the official docs show an example with the commas present — https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-exclude.

webknjaz: I did some testing and it seems like commas are indeed unnecessary.

.github/workflows/main.yml
jaraco: Good discussion and good reference. Let's let this approach settle for a bit and revisit in the skeleton in the future. Feel free to file a feature request.

.flake8
jaraco: Let's consider that as part of jaraco/skeleton#28 or possibly as a separate issue.

webknjaz: Yeah, I posted the comments here because this diff has more context in-place.

PR #2494, opened 2020-12-20
Add tests/test*.py to sdist default inclusion list

tomchen:
## Summary of changes

By default, all files matching the pattern **test**/test*.py (not **tests**/test*.py) are included in the source distribution (sdist). This is documented in [Python Packaging User Guide's "Including files in source distributions with MANIFEST.in" page](https://packaging.python.org/guides/using-manifest-in/?highlight=test#how-files-are-included-in-an-sdist).

But, both [pypa/sampleproject](https://github.com/pypa/sampleproject) and [Python Packaging User Guide's "Packaging Python Projects" page](https://packaging.python.org/tutorials/packaging-projects/?highlight=test#creating-the-package-files) suggest the unit test folder be named `tests`.

I opened the issue pypa/sampleproject#133, it looks clear that `tests` is a more common name than `test`, and it may avoid conflicting with the [test](https://docs.python.org/3/library/test.html) standard library module according to some.

Therefore, I request to add "tests/test*.py" to sdist default inclusion list. And "test/test*.py" should be kept for compatibility with old repos.

### Pull Request Checklist
- [x] News fragment added in [`changelog.d/`]

### DIFF
--- a/setuptools/_distutils/command/sdist.py
+++ b/setuptools/_distutils/command/sdist.py
@@ -211,7 +211,7 @@ class sdist(Command):
         """Add all the default files to self.filelist:
           - README or README.txt
           - setup.py
-          - test/test*.py
+          - tests/test*.py and test/test*.py
           - all pure Python modules mentioned in setup script
           - all files pointed by package_data (build_py)
           - all files defined in data_files.
@@ -268,7 +268,7 @@ class sdist(Command):
                     self.warn("standard file '%s' not found" % fn)
 
     def _add_defaults_optional(self):
-        optional = ['test/test*.py', 'setup.cfg']
+        optional = ['tests/test*.py', 'test/test*.py', 'setup.cfg']
         for pattern in optional:
             files = filter(os.path.isfile, glob(pattern))
             self.filelist.extend(files)
--- a/setuptools/command/py36compat.py
+++ b/setuptools/command/py36compat.py
@@ -17,7 +17,7 @@ class sdist_add_defaults:
         """Add all the default files to self.filelist:
           - README or README.txt
           - setup.py
-          - test/test*.py
+          - tests/test*.py and test/test*.py
           - all pure Python modules mentioned in setup script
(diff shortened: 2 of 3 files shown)


setuptools/_distutils/command/sdist.py
```diff
@@ -268,7 +268,7 @@ def _add_defaults_standards(self):
                     self.warn("standard file '%s' not found" % fn)
 
     def _add_defaults_optional(self):
-        optional = ['test/test*.py', 'setup.cfg']
```
jaraco, line 271: Unfortunately, it's not allowed to make changes to distutils here. If you wish to make changes to distutils, they need to happen in the [pypa/distutils](/pypa/distutils) project.

setuptools/command/py36compat.py
```diff
@@ -17,7 +17,7 @@ def add_defaults(self):
         """Add all the default files to self.filelist:
           - README or README.txt
           - setup.py
-          - test/test*.py
```
jaraco, line 20: There's a docstring in this code saying not to change this code.

jaraco:
Probably this issue requires more discussion. It's intertwined with the effort to adopt distutils and the copy of distutils functionality in the compat module. Making this change will add more divergence between distutils implementations. I'm inclined to say it would be best to wait for this change until setuptools owns distutils, at which time fine-tuning such a list would be more straightforward.

I do agree with you, the default of 'test' versus 'tests' violates my expectations, but this shift in expectations happened in spite of 'test' being included as a default for a very long time. Perhaps the better approach here would be to just remove all special files and simply require the user to use setuptools_scm or maintain a manifest of all the sources. After all, fixing this change still doesn't help users who have a `pyproject.toml` or those who choose a different convention for their tests.

The recommended usage here is for the packager to use `setuptools_scm` or provide a MANIFEST.in with the relevant rules.

Since neither of these code bases are meant to be changed here, I'm going to decline the PR.

Thanks anyhow for the effort.

PR #2526, opened 2021-01-08
exclude_source_files but not __init__.py files

amobash2:
When we use exclude_source_files, we can provide another argument include-init-files to ensure if user prefers we keep the __init__.py files.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
Can you explain why someone would want this behavior?

jaraco:
Let's start with filing an issue describing why the current behavior is inadequate or why someone would want this behavior. Feel free to ping me on the issue and reference this PR. We can re-open the PR if the use-case justifies it.

PR #2537, opened 2021-01-17
Remove support for easy_install-based downloads for fetch_build_eggs

jaraco:
This feature has been deprecated since the introduction of pip-based fetch_build_eggs.

### DIFF
--- a/setuptools/installer.py
+++ b/setuptools/installer.py
@@ -7,7 +7,6 @@ from distutils import log
 from distutils.errors import DistutilsError
 
 import pkg_resources
-from setuptools.command.easy_install import easy_install
 from setuptools.wheel import Wheel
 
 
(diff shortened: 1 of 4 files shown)


setuptools/installer.py
```diff
@@ -19,54 +18,11 @@ def _fixup_find_links(find_links):
...
-    return cmd.easy_install(req)
-
-
 def fetch_build_egg(dist, req):  # noqa: C901  # is too complex (16)  # FIXME
     """Fetch an egg needed for building.
 
     Use pip/wheel to fetch/build a wheel."""
-    # Check pip is available.
-    try:
-        pkg_resources.get_distribution('pip')
```
benoit-pierre, line 60: Wouldn't keeping an explicit check with `pkg_resources.get_distribution('pip')` make for a more user-friendly error message if pip is indeed missing?

jaraco: Perhaps, although I found in my experience that the error is "no module named pip", which seems fairly clear. More importantly, as I'd like to at some point remove the dependency on pkg_resources, I'm inclined to omit this usage here. Another advantage of this approach is that it would allow usage when 'pip' is present but its metadata is missing (for whatever reason).

benoit-pierre: Fair enough.

benoit-pierre (approved):

PR #2538, opened 2021-01-17
Repair Descriptions with newlines and emit a warning

### LINKED ISSUE #1390: Newlines in the `description` field produce a malformed PKG-INFO

jaraco:
Closes #1390

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added

### DIFF
--- a/changelog.d/1390.misc.rst
+++ b/changelog.d/1390.misc.rst
@@ -1,0 +1,1 @@
+Validation of Description field now is more lenient, emitting a warning and mangling the value to be valid (replacing newlines with spaces).
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -121,7 +121,9 @@ def read_pkg_file(self, file):
 def single_line(val):
     # quick and dirty validation for description pypa/setuptools#1390
     if '\n' in val:
-        raise ValueError("newlines not allowed")
+        # TODO after 2021-07-31: Replace with `raise ValueError("newlines not allowed")`
+        warnings.UserWarning("newlines not allowed and will break in the future")
+        val = val.replace('\n', ' ')
     return val
 
 


alexshpilkin:
Nit: maybe it’d also make sense to `str.rstrip()` broken descriptions? My guess is that most descriptions with newlines have one at the end. If careless usage of multiline strings (see _e.g._ PyLops/pyproximal@90d37d0eba62bd3de6a74a021de04a2af4473c2d) is frequent enough, it might also make sense to handle newlines and spaces at the beginning as well, and so `str.strip()` the descriptions instead.

jaraco:
Good point, although I do want to be cautious and not overcorrect here. If somehow a space is already present at the beginning or end of the description, I wouldn't want to alter that as part of correcting for newlines, even if that is the correct thing to do. On the other hand, converting a trailing or leading newline to a space seems janky at best.

I think for now I'll stick with this approach, as it still will correct the underlying defect (that metadata gets corrupted) and provide a path forward. I'll defer more sophisticated validation to subsequent efforts (which probably should happen earlier than when writing values out and probably coordinated across multiple forms of input such as setup.py or setup.cfg).

Issue #2541, opened 2021-01-18
SyntaxError: invalid syntax on Python 3.5.2


rsur-dev:
I get the following error:

> File "xxxxxxxxxxxxxxxxxxxxxxxxxx/ve3/lib/python3.5/site-packages/setuptools/command/install_scripts.py", line 17, in run
import setuptools.command.easy_install as ei
File "xxxxxxxxxxxxxxxxxxxxxxxxxx/ve3/lib/python3.5/site-packages/setuptools/command/easy_install.py", line 719
):
^
SyntaxError: invalid syntax

I think is related to the trailing comma on line 718 on setuptools/command/easy_install.py:

def process_distribution( # noqa: C901
  self, requirement, dist, deps=True, *info,
):

I'm using Python 3.5.2.

Thanks,

konradwudkowski:
I've got exactly the same problem and indeed the trailing comma on line 718 is the culprit on python 3.5. Would it be possible to remove it?

philroche:
This appears to have been introduced in 51.3.0 with commit https://github.com/pypa/setuptools/commit/fc891f5cf6d93ad533e2afb5e15a2952408ab358

skaiaa:
+1 same error, python 3.5

czimnick:
+1 same error, python 3.5

rvanlaar:
> +1 same error, python 3.5

Used for Ubuntu 16.04.

di:
cc @webknjaz

webknjaz:
Ah, `setuptools` dropped support for Python 3.5 about 1.5 months ago: https://setuptools.readthedocs.io/en/latest/history.html#v51-0-0.

The metadata clearly says `>= 3.6`: https://github.com/pypa/setuptools/blob/8222d6f/setup.cfg#L30.

So how did you all get a version that pip wasn't even supposed to pull-in? Can this be another instance of Ubuntu/Debian packaging essential Python packages wrong? https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e

If that's the case, I suggest folks complain to their distro packagers because I doubt that the setuptools upstream has the capacity to deal with this, nor can it influence the correctness of packaging debs.

But if you somehow got it via pip, this could be their bug.

I'd say that setuptools-wise, it's a "won't fix" because it's not a bug: testing has been dropped for Python 3.5 from the CIs. And also it's been EOL for almost 5 months already.

cc @jaraco

webknjaz:
[...]

PR #2542, opened 2021-01-19
Remove trailing comma

harlowja:
Hopefully helps fix https://github.com/pypa/setuptools/issues/2541


webknjaz (changes requested):
It's not a bug. Setuptools no longer supports Python 3.5 and attempting to maintain this is just an unnecessary maintenance burden: https://github.com/pypa/setuptools/issues/2541#issuecomment-763235824.

PR #2544, opened 2021-01-21
Remove easy_install script and module.

jaraco:
## Summary of changes

Ref #917

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2544.breaking.rst
+++ b/changelog.d/2544.breaking.rst
@@ -1,0 +1,1 @@
+Removed 'easy_install' top-level model (runpy entry point) and 'easy_install' console script.
--- a/easy_install.py
+++ b/easy_install.py
@@ -1,5 +1,0 @@
-"""Run the EasyInstall command"""
-
-if __name__ == '__main__':
-    from setuptools.command.easy_install import main
-    main()
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,6 @@ project_urls =
 
 [options]
 packages = find_namespace:
-py_modules = easy_install
 # disabled as it causes tests to be included #2505
 # include_package_data = true
 python_requires = >=3.6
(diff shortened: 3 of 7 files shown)


emailweixu:
I am not sure if this is the right location to ask: I used to use easy_install to install eggs. How should I install 3rd-party eggs after easy_install script is removed. And I didn't find "pip install --egg" working.

jaraco:
Your best bet is to stop installing eggs. There are lots of things that are bound to break going forward. Instead, build wheels and install them. If all you have are legacy eggs, consider building a tool to convert those to wheels.

No one is going to stop you from invoking easy_install as a setuptools command, but we won't be supporting that functionality and it could go away at any time (and will go away once setup_requires can be dropped).

emailweixu:
> Your best bet is to stop installing eggs. There are lots of things that are bound to break going forward. Instead, build wheels and install them. If all you have are legacy eggs, consider building a tool to convert those to wheels.

Thanks for the reply. I tried using command "wheel convert" to convert the egg to wheel. But "pip install" failed to install the wheel and reported that the wheel "is not a supported wheel on this platform".  However, the egg can be successfully installed using the hacky method.

learnereveryday:
@emailweixu Thank you so much! This hacky workaround is exactly what I have been looking for.

jaraco:
If you need egg to wheel conversion, I'd bring that up with the wheel project and be prepared to have to provide your own fix. In #2908, I've drafted the removal of easy_install, which will break the hacky workaround. Once that's released, your recourse will be to:

- find a way to convert those eggs to wheels
- use older versions of setuptools to install the egg
- build wheels from source and install those (recommended)

learnereveryday:
@jaraco Got it. Thanks for the info on what to expect. Yes, I think egg to wheel conversion utility tool will be quite helpful, as it is sometimes difficult for the third party enterprise packages in keeping up with the curve.

PR #2564, opened 2021-02-10
cygwinccompiler: Get compilers from CC

longnguyen2004:
## Summary of changes

Modified the `CygwinCCompiler` and `Mingw32CCompiler` to get compilers from the `CC` and `CXX` environment variables, and fallback to generic `gcc` if they are not set. This way, regular users won't be affected by the change, and those who want to use a different compiler can set `CC` and `CXX`.

### Pull Request Checklist

[PR docs]:

### DIFF
--- a/setuptools/_distutils/cygwinccompiler.py
+++ b/setuptools/_distutils/cygwinccompiler.py
@@ -44,6 +44,8 @@ cygwin in no-cygwin mode).
 #   (ld supports -shared)
 # * mingw gcc 3.2/ld 2.13 works
 #   (ld supports -shared)
+# * llvm-mingw with Clang 11 works
+#   (lld supports -shared)
 
 import os
 import sys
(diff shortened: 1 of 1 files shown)


lazka:
I know that's probably not helpful, but I think all those checks for 21(!!) year old software can probably be removed.

longnguyen2004:
Does anyone use a 21-year-old compiler anymore? :)
Personally, I would vote for it to be removed. This will make the implementation much cleaner and simpler. But the setuptools developers might have other opinions.
To further complicate the situation, the current `distutils` module in `cpython` also contains the same code!

lazka:
>  To further complicate the situation, the current `distutils` module in `cpython` also contains the same code!

But that's going away and no longer allows changes -> https://www.python.org/dev/peps/pep-0632/ so that can be ignored.

jaraco:
Changes to distutils should be submitted to [pypa/distutils](/pypa/distutils). Please submit the request there and the changes will be merged into Setuptools.

longnguyen2004:
There's a PR at distutils with this changes: https://github.com/pypa/distutils/pull/26

Issue #2558, opened 2021-02-04
Python 3.10: ModuleNotFoundError: No module named 'distutils.command.bdist_wininst'


hroncok:
Hello. With Python 3.10.0a5+ we have this problem when running setuptools tests in Fedora:

```
==================================== ERRORS ====================================
_____________ ERROR collecting setuptools/command/bdist_wininst.py _____________
setuptools/command/bdist_wininst.py:1: in <module>
    import distutils.command.bdist_wininst as orig
E   ModuleNotFoundError: No module named 'distutils.command.bdist_wininst'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
========================= 2 skipped, 1 error in 1.41s ==========================
```

> The bdist_wininst command deprecated in Python 3.8 has been removed. The bdist_wheel command is now recommended to distribute binary packages on Windows. (Contributed by @vstinner in [bpo-42802](https://bugs.python.org/issue42802).)

https://docs.python.org/3.10/whatsnew/3.10.html#distutils

hroncok:
I suppose the import could be try-excepted and the file might do nothing if nothing is there.

Alternatively, the command could be removed from next setuptools version altogether.

hroncok:
I've tried

```
rm setuptools/command/bdist_wininst.py setuptools/tests/test_bdist_deprecations.py
```

But I've been bit by https://github.com/pytest-dev/pytest/pull/8227

hroncok:
This will be tricky.

The distutils deprecation breaks a lot of tests and integration tests try to use old versions of pip that are not compatible with Python 3.10 due to collections.abc aliases removed from collections.

hroncok:
I'Ve drafted https://github.com/pypa/setuptools/pull/2566 to solve my most immediate issues. That PR allows use to continue testing with Python 3.10.0a5 until a permanent solution is discussed.

vstinner:
Oh, I expected setuptools to be compatible with bdist_wininst removal, I prepared the removal in setuptools with:
* commit https://github.com/pypa/setuptools/commit/5d17586a56077dfa3109a5861cf0ff579095a42e
* issue https://github.com/pypa/setuptools/pull/1986

But it seems like I missed this issue.

jaraco:
> integration tests try to use old versions of pip that are not compatible with Python 3.10 due to collections.abc aliases removed from collections

[...]

PR #2566, opened 2021-02-12
Remove bdist_wininst, Filter out distutils deprecation

hroncok:
## Summary of changes

Python 3.10.0a5 related fix / workaround. See individual commits. 

Closes https://github.com/pypa/setuptools/issues/2558

### Pull Request Checklist
- [ ] Changes have tests -- nope, not sure how
- [x] News fragment added in [`changelog.d/`] -- not yet

[PR docs]:


pytest.ini
```diff
@@ -12,8 +12,6 @@ filterwarnings=
     ignore:direct construction of .*Item has been deprecated:DeprecationWarning
     # Fail on warnings
     error
-    # https://github.com/pypa/setuptools/issues/1823
-    ignore:bdist_wininst command is deprecated
```
hroncok, line 16: This needs to stay for older Pythons.

hroncok:
@jaraco Is this a way to go for now?

jaraco:
Apologies for letting this linger so long. I’ll try to get to it today.

_distutils_hack/__init__.py
```diff
@@ -8,6 +8,11 @@
 is_pypy = '__pypy__' in sys.builtin_module_names
 
 
+warnings.filterwarnings('ignore',
+                        '.+ distutils .+ deprecated',
+                        DeprecationWarning)
```
jaraco, line 13: Can you describe what situation this resolves in the commit message or comment (or link to a description that summarizes the failure mode it addresses)? I'm unable to test on Python 3.10 due to other issues. I'm uneasy adding this change here because it will essentially suppress distutils deprecations anytime Setuptools is imported... and that could mask deprecations that need to be present.

hroncok: My understanding is that setuptools wants to carry their own distutils copy. This switch will happen before distutils is removed from Python, so the deprecation of distutils is well known to setuptools and the users of setuptools don't need to worry about it. Obviously, when users import both distutils and setuptools, they should be warned, but I am not sure how to handle that properly. As for the problems this resolves:

- https://bugzilla.redhat.com/show_bug.cgi?id=1927192
- https://bugzilla.redhat.com/show_bug.cgi?id=1927173

```

==================================== ERRORS ====================================
____________ ERROR collecting tests/unit/test_setuptools_command.py ____________
tests/unit/test_setuptools_command.py:3: in <module>
    from setuptools import dist
/usr/lib/python3.10/site-packages/setuptools/__init__.py:10: in <module>
    import distutils.core
/usr/lib64/python3.10/distutils/__init__.py:16: in <module>
    warnings.warn("The distutils package deprecated and slated for "
E   DeprecationWarning: The distutils package deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.48s ===============================
```

```
==================================== ERRORS ====================================
______________ ERROR collecting test/test_dependency_metadata.py _______________
test/test_dependency_metadata.py:6: in <module>
    from colcon_ros.package_identification.ros import _create_metadata
colcon_ros/package_identification/ros.py:16: in <module>
    from colcon_python_setup_py.package_identification.python_setup_py \
/usr/lib/python3.10/site-packages/colcon_python_setup_py/package_identification/python_setup_py.py:9: in <module>
    import setuptools
/usr/lib/python3.10/site-packages/setuptools/__init__.py:10: in <module>
    import distutils.core
/usr/lib64/python3.10/distutils/__init__.py:16: in <module>
    warnings.warn("The distutils package deprecated and slated for "
E   DeprecationWarning: The distutils package deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
=========================== short test summary info ============================
ERROR test/test_dependency_metadata.py - DeprecationWarning: The distutils pa...
```

jaraco:
At one point, it was proposed that Setuptools might continue to support bdist_wininst after the removal from Python, or at least it might enable support on Python 3.10+. This approach takes a different tack and accelerates the deprecation and imposes it on older Pythons. I'm okay with that, because those users can still use older Setuptools as a stopgap.

_distutils_hack/__init__.py
jaraco: The tests on Setuptools still pass without the commit, so it seems that it's only through other packages that these warnings are emitted. More importantly, these errors are expected and the recommended workaround long-term is to use `SETUPTOOLS_USE_DISTUTILS=local` (which we know doesn't work on Linux yet). So maybe the best bet would be for those environments to suppress the warning rather than asking Setuptools to do so. If Setuptools suppresses the warning now, things may break later. I guess that's acceptable, given that the distutils platform problems are likely to be resolved by then.

jaraco: I still don't understand why Setuptools tests don't encounter this failure mode.

jaraco (approved):
Go ahead and write up a `breaking` changelog and we can get this out.

hroncok:
Amedned. I've also extended the filterwarnings commit message a bit.

hroncok:
@jaraco Could you please merge this PR? It has been approved 3 weeks ago.

jaraco:
Yes. Thanks for your patience.

PR #2571, opened 2021-02-18
Correctly parse cmdclass in setup.cfg. Fixes #2570

### LINKED ISSUE #2570: Support cmdclass in setup.cfg
Having cmdclass defined in setup.cfg causes sdist,  or any other, command to fail because its value is not properly parsed into a dict.

How to replicate

setup.cfg example:
```
[metadata]
name = cmdclass example

[options]
packages = find:
cmdclass =
    sdist = build.CustomSdistCommand
```

setup.py:
```
import setuptools
setuptools.setup()
```

build.py
```
from setuptools.command.sdist import sdist

class CustomSdistCommand (sdist):
    def run(self):
        print('custom steps...')
        super(sdist, self).run()
```

error:
```
❯ python setup.py sdist
  File ".../setup.py", line 5, in <module>
    setuptools.setup()
  File ".../.venv/lib/python3.9/site-packages/setuptools/__init__.py", line 153, in setup
  File ".../lib/python3.9/distutils/core.py", line 134, in setup
  File ".../lib/python3.9/distutils/dist.py", line 483, in parse_command_line
    args = self._parse_command_opts(parser, args)
  File ".../.venv/lib/python3.9/site-packages/setuptools/dist.py", line 916, in _parse_command_opts
    nargs = _Distribution._parse_command_opts(self, parser, args)
  File ".../lib/python3.9/distutils/dist.py", line 539, in _parse_command_opts
    cmd_class = self.get_command_class(command)
  File ".../.venv/lib/python3.9/site-packages/setuptools/dist.py", line 750, in get_command_class
    return self.cmdclass[command]
TypeError: string indices must be integers
```

It happens because self.cmdclass evaluates to the string "sdist = build.CustomSdistCommand" but it should be { "sdist": <class 'build.CustomSdistCommand'> }

Environment details:
```
❯ python
Python 3.9.0 (v3.9.0:9cf6752276, Oct  5 2020, 11:29:23) 
[Clang 6.0 (clang-600.0.57)] on darwin
```

PIP: 21.0.1
setuptools: 53.0.0

Possible solutions:
either change setup.cfg structure introducing options.cmdclass section or fix setuptools/dist.py Distribution._parse_config_files method so that it parses cmdclass value.

Full minimal example:
[example.zip](https://github.com/pypa/setuptools/files/5995374/example.zip)

makeroo:
## Summary of changes

Closes #2570 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -574,6 +574,7 @@ class ConfigOptionsHandler(ConfigHandler):
         parse_list_semicolon = partial(self._parse_list, separator=';')
         parse_bool = self._parse_bool
         parse_dict = self._parse_dict
+        parse_cmdclass = self._parse_cmdclass
 
         return {
             'zip_safe': parse_bool,
@@ -594,6 +595,22 @@ class ConfigOptionsHandler(ConfigHandler):
             'entry_points': self._parse_file,
             'py_modules': parse_list,
             'python_requires': SpecifierSet,
+            'cmdclass': parse_cmdclass,
+        }
+
+    def _parse_cmdclass(self, value):
+        def resolve_class(qualified_class_name):
+            idx = qualified_class_name.rfind('.')
+            class_name = qualified_class_name[idx+1:]
+            pkg_name = qualified_class_name[:idx]
+
+            module = __import__(pkg_name)
+
+            return getattr(module, class_name)
+
+        return {
+            k: resolve_class(v)
+            for k, v in self._parse_dict(value).items()
         }
 
     def _parse_packages(self, value):
--- a/changelog.d/2570.change.rst
+++ b/changelog.d/2570.change.rst
@@ -1,0 +1,1 @@
+Correctly parse cmdclass in setup.cfg.
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -1,3 +1,6 @@
+import types
+import sys
+
 import contextlib
 import configparser
 
@@ -7,6 +10,7 @@ from distutils.errors import DistutilsOptionError, DistutilsFileError
 from mock import patch
 from setuptools.dist import Distribution, _Distribution
 from setuptools.config import ConfigHandler, read_configuration
+from distutils.core import Command
 from .textwrap import DALS
 
 
@@ -852,6 +856,26 @@ class TestOptions:
         with pytest.raises(Exception):
             with get_dist(tmpdir) as dist:
                 dist.parse_config_files()
+
+    def test_cmdclass(self, tmpdir):
+        class CustomCmd(Command):
+            pass
+
+        m = types.ModuleType('custom_build', 'test package')
+
+        m.__dict__['CustomCmd'] = CustomCmd
+
+        sys.modules['custom_build'] = m
+
+        fake_env(
+            tmpdir,
+            '[options]\n'
+            'cmdclass =\n'
+            '    customcmd = custom_build.CustomCmd\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            assert dist.cmdclass == {'customcmd': CustomCmd}
 
 
 saved_dist_init = _Distribution.__init__


makeroo:
Note: I don't know if resolve_class function is a duplicate, don't know setuptools code base in deep. Also, I can't say which approach is the correct one between having cmdclass as property of options section or having a options.cmdclass section.

setuptools/config.py
```diff
@@ -594,6 +595,22 @@ def parsers(self):
...
+            'cmdclass': parse_cmdclass,
+        }
+
+    def _parse_cmdclass(self, value):
+        def resolve_class(qualified_class_name):
+            idx = qualified_class_name.rfind('.')
+            class_name = qualified_class_name[idx+1:]
+            pkg_name = qualified_class_name[:idx]
+
+            module = __import__(pkg_name)
```
jaraco, line 607: Couple of things here. First, `__import__` as used here will always import the top-level module, so if `pkg_name` was `mypkg.mymod`, `pkg_name` would end up being `mypkg` (with `mymod` imported and a property of `mypkg`). Second `importlib.import_module` is preferred over the discouraged `__import__` _and_ it has the nice property of returning the last item, addressing the first concern, so I'd use that.

jaraco (changes requested):
My main concern here is that we're sneaking in a feature that we don't want or that there's a better way to accomplish what we're after.

First, is this option documented in the declarative config? If so, then I'd say that the feature is already supported but broken.

Second, could the underlying use case be satisfied by another technique, namely by having another package that presents a `distutils.commands` entry point? I'm sure one could do that for non-standard commands, but I'm not confident that `distutils.commands` could reliably override setuptools' own `distutils.commands`, and I presume that's the use-case that this bug is intending to resolve. Can you clarify the answer to the question above and describe more about what use cases this feature is intended to solve?

setuptools/config.py
makeroo: Hi Jason,

according to https://setuptools.readthedocs.io/en/latest/references/keywords.html the keyword cmdclass is defined as: "A dictionary providing a mapping of command names to Command subclasses".

but such keyword is not listed in https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html

My interpretation is that declarative config docs are not complete yet. Perhaps I'm wrong, actually I discovered declarative config just a week ago and I'm upgrading my projects to adopt it.

Coming to the use case, my need is to generate some sources before the actual (standard) project build. Eg. embed static js generated files from a node application, thus launching npm build in a subfolder, or compile GRPC proto files as in https://stackoverflow.com/a/54236739.

Before the declarative config the common solution was to derive build_py command. Furthermore, the cmdclass keyword was there to provide custom commands and I thought that declarative config was on the way to support it as well.

I don't understand what you mean by "namely by having another package that presents a distutils.commands entry point".

Finally, I thought setuptools still supported python 2. Later I discovered that p2 support has been dropped with 53.0.0. So yes, resolve_class should be rewritten using importlib.

makeroo: Note: cmdclass is a dict from command name to Commad class, ie. fully qualified class names in setup.cfg.
Example, setup.cfg:

```
cmdclass =
  mycustomcmd  = mycustommodule.CustomCommand
```

where mycustommodule is a file python module in the folder containing setup.cfg too:

```

class CustomCommand(Command):
    ...
```

jaraco: > My interpretation is that declarative config docs are not complete yet.

Another interpretation was that this feature was determined to be unnecessary for declarative config. I don't recall if it was an intentional omission, oversight, or deferred feature.

> Coming to the use case, my need is to generate some sources before the actual (standard) project build. Eg. embed static js generated files from a node application, thus launching npm build in a subfolder, or compile GRPC proto files as in https://stackoverflow.com/a/54236739.

Thanks for sharing the use-case. I can't help but feel like the approach Setuptools presents here is sub-optimal, mainly because the caller is required to override the `build_py` command just to inject a behavior before building.

I'm not sure Setuptools needs to continue to encourage this behavior and should perhaps provide a better interface for projects to inject behavior at various stages, e.g. to allow a project to declare "run 'grpc' before 'build'".

That said, I don't have an obvious solution for this approach, so perhaps it makes sense to honor this setting in setup.cfg.

PR #2573, opened 2021-02-19
Fix error uploading Sphinx doc with upload_docs. Fixes #1060

### LINKED ISSUE #1060: upload_docs raises AttributeError with Sphinx 1.6
In Sphinx 1.6, `builder_target_dir` was replaced with `builder_target_dirs`, see sphinx-doc/sphinx@2afa0b6627f7e5afb188d5a60c8c4767f6250774 and sphinx-doc/sphinx#3476.

This causes an AttributeError in setuptools:

```pytb
  File "setup.py", line 67, in <module>
    setup(**setup_args)
  File "/usr/lib/python3.5/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.5/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.5/distutils/dist.py", line 973, in run_command
    cmd_obj.ensure_finalized()
  File "/usr/lib/python3.5/distutils/cmd.py", line 107, in ensure_finalized
    self.finalize_options()
  File "/home/dmitry/.local/lib/python3.5/site-packages/setuptools/command/upload_docs.py", line 65, in finalize_options
    self.target_dir = build_sphinx.builder_target_dir
  File "/usr/lib/python3.5/distutils/cmd.py", line 103, in __getattr__
    raise AttributeError(attr)
AttributeError: builder_target_dir
```

melissa-kun-li:
## Summary of changes

Closes #1060

Sphinx now uses `builder_target_dirs` in replacement of `builder_target_dir`  https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/setup_command.py#L143-L145. This fix will set the `target_dir` in `upload_docs` to the path of the html builder. 

Note: `Upload_docs` was restored support for sites other than PyPi as noted here: https://github.com/pypa/setuptools/blob/e1ffc2abbae4f2aa78dd09ee9827d754b7702b7b/CHANGES.rst#v3610

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2573.change.rst
+++ b/changelog.d/2573.change.rst
@@ -1,0 +1,2 @@
+Fixed error in uploading a Sphinx doc with the :code:`upload_docs` command. An html builder will be used. 
+Note: :code:`upload_docs` is deprecated for PyPi, but is supported for other sites -- by :user:`melissa-kun-li`
--- a/setuptools/command/upload_docs.py
+++ b/setuptools/command/upload_docs.py
@@ -31,7 +31,7 @@ class upload_docs(upload):
     # supported by Warehouse (and won't be).
     DEFAULT_REPOSITORY = 'https://pypi.python.org/pypi/'
 
-    description = 'Upload documentation to PyPI'
+    description = 'Upload documentation to sites other than PyPi such as devpi'
 
     user_options = [
         ('repository=', 'r',
@@ -59,10 +59,7 @@ class upload_docs(upload):
         if self.upload_dir is None:
             if self.has_sphinx():
                 build_sphinx = self.get_finalized_command('build_sphinx')
-                for (builder, builder_dir) in build_sphinx.builder_target_dirs:
-                    if builder == "html":
-                        self.target_dir = builder_dir
-                        break
+                self.target_dir = dict(build_sphinx.builder_target_dirs)['html']
             else:
                 build = self.get_finalized_command('build')
                 self.target_dir = os.path.join(build.build_base, 'docs')
@@ -70,7 +67,7 @@ class upload_docs(upload):
             self.ensure_dirname('upload_dir')
             self.target_dir = self.upload_dir
         if 'pypi.python.org' in self.repository:
-            log.warn("Upload_docs command is deprecated. Use RTD instead.")
+            log.warn("Upload_docs command is deprecated for PyPi. Use RTD instead.")
         self.announce('Using upload directory %s' % self.target_dir)
 
     def create_zipfile(self, filename):
--- a/setuptools/command/upload_docs.py
+++ b/setuptools/command/upload_docs.py
@@ -2,7 +2,7 @@
 """upload_docs
 
 Implements a Distutils 'upload_docs' subcommand (upload documentation to
-PyPI's pythonhosted.org).
+sites other than PyPi such as devpi).
 """
 
 from base64 import standard_b64encode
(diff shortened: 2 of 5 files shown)


setuptools/command/upload_docs.py
```diff
@@ -2,7 +2,7 @@
 """upload_docs
 
 Implements a Distutils 'upload_docs' subcommand (upload documentation to
-PyPI's pythonhosted.org).
+sites other than PyPi such as devpi).
```
jaraco, line 5: Thanks for taking extra time to clean up docs too!

setuptools/command/upload_docs.py
```diff
@@ -59,7 +59,10 @@ def finalize_options(self):
         if self.upload_dir is None:
             if self.has_sphinx():
                 build_sphinx = self.get_finalized_command('build_sphinx')
-                self.target_dir = build_sphinx.builder_target_dir
+                for (builder, builder_dir) in build_sphinx.builder_target_dirs:
+                    if builder == "html":
+                        self.target_dir = builder_dir
+                        break
```
jaraco, line 65: Looks good. What happens if no builder == 'html'? Should we be concerned about that possibility?

Also, did you consider building a lookup map (dict) to resolve builder -> builder_dir? Were you aware you could construct a [dict from a sequence of pairs](https://docs.python.org/3/tutorial/datastructures.html#dictionaries), something like `dict(build_sphinx.builder_target_dirs)`?

setuptools/tests/test_sphinx_upload_docs.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # A test index.rst for Sphinx
+    with open('build/docs/index.rst', 'w') as f:
+        f.write(".. toctree::\
+                    :maxdepth: 2\
+                    :caption: Contents:")
+
+
+@pytest.mark.usefixtures('sphinx_doc_sample_project')
+@pytest.mark.usefixtures('user_override')
```
jaraco, line 28: Is this override needed? If not, please omit.

jaraco (approved):
This PR is well done. Thanks for putting together the test, which clearly was a lot more effort than the fix itself (not an uncommon scenario). I have no major concerns and just a few comments for your consideration, then we're ready to merge.

setuptools/command/upload_docs.py
melissa-kun-li: I noticed some other places in `upload_docs` I could change the documentation too. I think it would help decrease the confusion about `upload_docs` being deprecated or not

setuptools/command/upload_docs.py
melissa-kun-li: Hmm, that's true. The sphinx documentation says that the html builder is the default so I went with html. But if we want to let people upload any type, I think it would have to be an option in the command when they use `upload_docs` and then we can use whatever they inputted in place of where the `html` code is right now, or and do html by default if they run the command without any type option. Or we could keep as is, what do you think? 

Thanks for the idea, I'll rewrite it with dict!

melissa-kun-li:
@jaraco Thank you for mentoring me through it!

setuptools/command/upload_docs.py
jaraco: Sounds good. I also agree the case where "html" isn't present is an unlikely scenario and surely out of the scope of what to do here. I just wanted to think about that case. Happy to proceed with the current approach. Good reasoning.

jaraco:
Very nice. Thorough and thoughtful work.

changelog.d/2573.change.rst
```diff
@@ -0,0 +1,2 @@
+Fixed error in uploading a Sphinx doc with the :code:`upload_docs` command. An html builder will be used. 
+Note: :code:`upload_docs` is deprecated for PyPi, but is supported for other sites -- by :user:`melissa-kun-li`
```
webknjaz, line 2: FTR PyPI should be spelled with a capital "I"

melissa-kun-li: Ah right! Thanks for pointing that out, will note for future.

PR #2574, opened 2021-02-20
Use file builder from jaraco.path

jaraco:
## Summary of changes

The Setuptools test suite has a nice little tool for creating a small suite of
files (setuptools.tests.files.build_files). I'd like to use that functionalty
elsewhere, so I've extracted it into jaraco.path.

### Pull Request Checklist
- [X] Changes have tests


setup.cfg
```diff
@@ -58,6 +58,7 @@ testing =
     pip>=19.1 # For proper file:// URLs support.
     jaraco.envs
     pytest-xdist
+    jaraco.path
```
benoit-pierre, line 61:
```suggestion
    jaraco.path>=3.2.0
```

jaraco:
The CI reveals that the dependency on `jaraco.path` adds a dependency on pyobjc on macOS, which fails to build on PyPy, so this change may not be viable.

jaraco:
I think I fixed the issue with PyPy (by just excluding the dependency on PyPy), but now there's a new issue where `singledispatch` has deprecation warnings that are errors in Setuptools.

jaraco:
Warnings with singledispatch fixed in the latest patch release, yet still PyPy failing. I must have gotten the environment marker wrong.

Issue #2480, opened 2020-12-10
target directories in data_files are incorrectly changed to lowercase


rhertzog:
Hello, I have this in setup.cfg:
```
[options.data_files]
share/perl5/Debian/Debhelper/Sequence = dh-kaboxer/Sequence/kaboxer.pm
```

But that file ends up installed in `/usr/share/perl5/debian/debhelper/sequence/`. Notice how the target directory is entirely lowercase while the configuration file had uppercase on some characters.

And here the case is meaningful.

This is with python 3.9 and setuptools 50.3.0-1 on a Debian Unstable system.

melissa-kun-li:
Hopefully PR #2580 addresses this :)

PR #2580, opened 2021-02-24
Fix case sensitivity of entry point names and keys in setup.cfg. Fixes #1937

### LINKED ISSUE #1937: setup.cfg: entry_points keys are made lowercase
This breaks when the entry point is actually case-sensitive.

1. `git clone https://github.com/pydoit/doit-plugin-sample`
2. Delete setup.py, use the following setup.cfg:
```text
[metadata]
name = doit-plugin-sample
description = a simple doit command plugin

[options]
py_modules = 
    doit_sample_cmd
install_requires = 
    doit

[options.entry_points]
doit.COMMAND =
    plug_sample = doit_sample_cmd:SampleCmd
```
3. Use this pyproject.toml:
```text
[build-system]
requires = ["setuptools>=30.3.0", "wheel"]
```
4. `pip install .` inside a venv.

Inspecting `venv/*/site-packages/doit_plugin_sample-0.0.0.dist-info/entry_points.txt`:
```text
[doit.command]
plug_sample = doit_sample_cmd:SampleCmd
```

This will not work, as the entry point is `doit.COMMAND`.

melissa-kun-li:
## Summary of changes

Closes #1937

To address the bug of entry point names being made lower-case by Setuptools, this fix preserves case-sensitivity of keys in `setup.cfg` by changing`ConfigParser` to become case-sensitive. Edit: by extension this will also address any issues about keys in `setup.cfg` not retaining case-sensitivity such as in #2480

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/1937.breaking.rst
+++ b/changelog.d/1937.breaking.rst
@@ -1,1 +1,0 @@
-Preserved case-sensitivity of keys in setup.cfg so that entry point names are case-sensitive. Changed sensitivity of configparser -- by :user:`melissa-kun-li`
--- a/changelog.d/1937.breaking.rst
+++ b/changelog.d/1937.breaking.rst
@@ -1,0 +1,1 @@
+Preserved case-sensitivity of keys in setup.cfg so that entry point names are case-sensitive. Changed sensitivity of configparser -- by :user:`melissa-kun-li`
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -802,6 +802,40 @@ class TestOptions:
         with get_dist(tmpdir) as dist:
             assert dist.entry_points == expected
 
+    def test_case_sensitive_entry_points(self, tmpdir):
+        _, config = fake_env(
+            tmpdir,
+            '[options.entry_points]\n'
+            'GROUP1 = point1 = pack.module:func, '
+            '.point2 = pack.module2:func_rest [rest]\n'
+            'group2 = point3 = pack.module:func2\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            assert dist.entry_points == {
+                'GROUP1': [
+                    'point1 = pack.module:func',
+                    '.point2 = pack.module2:func_rest [rest]',
+                ],
+                'group2': ['point3 = pack.module:func2']
+            }
+
+        expected = (
+            '[blogtool.parsers]\n'
+            '.rst = some.nested.module:SomeClass.some_classmethod[reST]\n'
+        )
+
+        tmpdir.join('entry_points').write(expected)
+
+        # From file.
+        config.write(
+            '[options]\n'
+            'entry_points = file: entry_points\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            assert dist.entry_points == expected
+
     def test_data_files(self, tmpdir):
         fake_env(
             tmpdir,
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -820,22 +820,6 @@ class TestOptions:
                 'group2': ['point3 = pack.module:func2']
             }
 
-        expected = (
-            '[blogtool.parsers]\n'
-            '.rst = some.nested.module:SomeClass.some_classmethod[reST]\n'
-        )
-
-        tmpdir.join('entry_points').write(expected)
-
-        # From file.
-        config.write(
-            '[options]\n'
-            'entry_points = file: entry_points\n'
-        )
-
-        with get_dist(tmpdir) as dist:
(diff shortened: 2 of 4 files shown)


setuptools/tests/test_config.py
```diff
@@ -802,6 +802,40 @@ def test_entry_points(self, tmpdir):
...
+        tmpdir.join('entry_points').write(expected)
+
+        # From file.
+        config.write(
+            '[options]\n'
+            'entry_points = file: entry_points\n'
+        )
+
+        with get_dist(tmpdir) as dist:
+            assert dist.entry_points == expected
```
jaraco, line 837: I'm slightly surprised by this behavior. I'm not sure `file:` support was intended for `options.entry_points`, although I agree it would be nice if it were supported. What's more surprising and concerning is that `dist.entry_points` is a different format. I see now that this test is modeled after `test_entry_points`. Since that test already essentially covers this case, I would recommend omitting the second half. WDYT?

jaraco (approved):
This is great. Looks like a simple change, but I know it involved a great deal of work to figure it out. This change will help a lot of people. One small request above then it's good to go.

changelog.d/1937.breaking.rst
```diff
@@ -0,0 +1 @@
+Preserved case-sensitivity of keys in setup.cfg so that entry point names are case-sensitive. Changed sensitivity of configparser -- by :user:`melissa-kun-li`
```
jaraco, line 1: I'm on the fence for declaring this backward-incompatible, though strictly it is. Often when making a backward-incompatible change, it's important to let the users know what action they should expect to take to account for the incompatibility. This message implies what to do and it's good enough. You might consider something like, "Any projects relying on the case-insensitivity will need to adapt to accept the original case as published."

setuptools/tests/test_config.py
melissa-kun-li: I was also unsure about that section of the test. I thought to include it in case the case-sensitive fix broke something that `test_entry_points` was supposed to pass. Although, it does appear to be passing regardless. I'll go ahead and remove it as I'm uncertain about its behaviour.

changelog.d/1937.breaking.rst
melissa-kun-li: Hmmm, I thought about it more and I think I'm going to put it in the category of `change` instead of `breaking`. I think it would be a compatible change with the vast majority of users as it still maintains the convention of lower-case, and in addition is also a bug-fixing change for many other users who require meaningful case-sensitivity. For a smaller subsection of users it may be breaking, but overall I think it could be considered backwards-compatible. I'll also add that message you mentioned so that it's clear!

jaraco: Sounds good, but it looks like this file is still present. I'll remove it.

jaraco:
```
setuptools main $ hub pr checkout 2580
remote: Enumerating objects: 22, done.
remote: Counting objects: 100% (22/22), done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 22 (delta 10), reused 22 (delta 10), pack-reused 0
Unpacking objects: 100% (22/22), 11.48 KiB | 167.00 KiB/s, done.
From https://github.com/pypa/setuptools
 * [new ref]           refs/pull/2580/head -> fix-entry-points-name-case-sensitivity
Switched to branch 'fix-entry-points-name-case-sensitivity'
setuptools fix-entry-points-name-case-sensitivity $ git rm changelog.d/1937.breaking.rst
rm 'changelog.d/1937.breaking.rst'
setuptools fix-entry-points-name-case-sensitivity $ git commit -a -m "Remove 'breaking' changelog, superseded by 'change'."
 
[fix-entry-points-name-case-sensitivity 898a0b59] Remove 'breaking' changelog, superseded by 'change'.
 1 file changed, 1 deletion(-)
 delete mode 100644 changelog.d/1937.breaking.rst
setuptools fix-entry-points-name-case-sensitivity $ git push
Enumerating objects: 5, done.
Counting objects: 100% (5/5), done.
Delta compression using up to 4 threads
Compressing objects: 100% (3/3), done.
Writing objects: 100% (3/3), 322 bytes | 161.00 KiB/s, done.
Total 3 (delta 2), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (2/2), completed with 2 local objects.
To github.com:melissa-kun-li/setuptools.git
   11529db0..898a0b59  fix-entry-points-name-case-sensitivity -> fix-entry-points-name-case-sensitivity
```

jaraco:
Releasing as 53.1.0. Thanks again!

PR #2583, opened 2021-02-28
Handle AttributeError by raising DistutilsSetupError in check_specifier. Fixes #1932

### LINKED ISSUE #1932: check_specifier() raises an exception when not passed as string
`dist.check_specifier()` is used to validate that the given argument is valid. The documentation and the error message it produces clearly states the value should be a string. However, if the value is something other than a string, `DistutilsSetupError` is not raised. This is because `SpecifierSet` class `.split()` assuming that the value is a string. This causes:

```
AttributeError: 'lXXX' object has no attribute 'split'
```
In my case, it happened to be a list containing one valid specifier.

Instead of failing with this error, `check_specifier()` should raise `DistutilsSetupError` to clearly indicate that it was given a value of the wrong type.

This is related to #1869 in that it is hitting the same error. But, I'm suggesting the type be properly checked instead of allowing other types.

melissa-kun-li:
## Summary of changes

Closes  #1932

`Dist.check_specifier` verifies that the specifiers are valid and raises `DistutilsSetupError` if found invalid, however in the situation that the specifier is a non-string (e.g. a list with a string element) it would fail with an `AttributeError` because `SpecifierSet` from _vendor/packaging will assume that the specifier is a string. This fix will change `check_specifier` to handle the `AttributeError` by raising `DistutilsSetupError` when the situation of a non-string specifier occurs. 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/1932.change.rst
+++ b/changelog.d/1932.change.rst
@@ -1,0 +1,1 @@
+Handled :code:`AttributeError` by raising :code:`DistutilsSetupError` in :code:`dist.check_specifier()` when specifier is not a string -- by :user:`melissa-kun-li`
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -292,7 +292,7 @@ def check_specifier(dist, attr, value):
     """Verify that value is a valid version specifier"""
     try:
         packaging.specifiers.SpecifierSet(value)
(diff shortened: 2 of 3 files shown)


jaraco:
This looks perfect. Thanks!

Issue #2469, opened 2020-12-05
[TODO] Enable PR builds in RTD


webknjaz:
@jaraco we've discussed this in comments to other PRs/issues and you seemed to be on board with the idea. Except that the only annoying part is that it cannot be done via a config file but only using the UI.
I thought I'd create a dedicated issue for this task so it wouldn't get lost again.

webknjaz:
@jaraco we've faced an actual problem caused by not having this enabled. It happened 3 weeks ago (https://github.com/pypa/setuptools/pull/2584) but only got noticed 5 days ago (https://github.com/pypa/setuptools/issues/2614).
I think this checkbox should be enabled ASAP.

jaraco:
The checkbox is already checked.

![image](https://user-images.githubusercontent.com/308610/112734772-35d6ea00-8f1e-11eb-9e6a-eae4e09e1967.png)

I did not change it. So there must be more work to be done.

jaraco:
I've unchecked the box, hit save, checked the box, and hit save again. Maybe that will help.

webknjaz:
Sometimes the GH webhooks break. There's a tab to resync them. Or check in the repo settings whether the corresponding hook actually sends PR events to RTD.

webknjaz:
Check https://readthedocs.org/dashboard/setuptools/integrations/ (there's a resync button under the specific hook).

And also check https://github.com/pypa/setuptools/settings/hooks -> `Ctrl+F` for "readthedocs" and see if the `Pull Requests` checkbox is on.

jaraco:
In the hooks settings, I changed "Just the 'push' event" to "send me everything" (even though I want them to send RTD everything). Let's see how that works.

webknjaz:
I retriggered https://github.com/pypa/setuptools/pull/2601 and verified that the RTD status appeared in the list (the failed status is expected because that PR is old and does not have the workaround from `main`. Closing.

PR #2584, opened 2021-03-01
Add mechanism for comparison of setup.py and its equivalent setup.cfg

### LINKED ISSUE #1699: Add mechanism for side-by-side comparison of setup.py and its equivalent setup.cfg
We have many documentation examples that are purely declarative and are either documented as `setup.py` or `setup.cfg`. It would be really awesome if, for each of these, we had the option to have either both versions side-by-side or, even better, in a sort of "tabbed container", like the one in the [code sample at the bottom of this example](https://leetcode.com/articles/median-of-two-sorted-arrays/).

Requirements for this:

1. Cannot *link to* any third-party javascript dependencies. Ideally we wouldn't use any at all, but if you do they must be vendored in the documentation.
2. If javascript is disabled, it has to fall back to something intelligible.

Ideally it would be implemented in pure CSS / HTML if that's at all possible.

amy-lei:
## Summary of changes

Closes #1699

Add the [sphinx-inline-tabs](https://github.com/pradyunsg/sphinx-inline-tabs) extension to allow tabbed containers for comparing `setup.py` and `setup.cfg`. The tabs works even when JS is disabled. This PR also replaces instances of adjacent code-blocks with the new extension. 

However, I wasn't sure if there was an issue with the default styling. It currently looks like this: 
image
The only thing that stood out and possibly needed overriding was the accent color of the tabs.

### Pull Request Checklist
- [ ] ~~Changes have tests~~

[PR docs]:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker']
+extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker', 'sphinx_inline_tabs']
 
 master_doc = "index"
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker', 'sphinx_inline_tabs']
+extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker']
 
 master_doc = "index"
 
@@ -93,3 +93,6 @@ default_role = 'any'
 html_theme = 'alabaster'
 templates_path = ['_templates']
 html_sidebars = {'index': ['tidelift-sidebar.html']}
+
+# Add support for inline tabs
+extensions += ['sphinx_inline_tabs']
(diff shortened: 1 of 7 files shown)


docs/conf.py
```diff
@@ -1,4 +1,4 @@
-extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker']
+extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker', 'sphinx_inline_tabs']
```
jaraco, line 1: This change is causing a nitpicky linter to fail the tests. I'd normally let you fix it, but I probably want to make some concession for other concerns (integration with the [upstream template](https://github.com/jaraco/skeleton)), so I can make the adjustments.

jaraco: I applied a fix in 265fb0eda.

jaraco (approved):
This is beautiful. I honestly did not expect this change to be possible, but the solution you found and implemented is simple and elegant. Thoroughly impressed and grateful.

webknjaz:
Merging this PR has broken the RTD setup 3 weeks ago. Since then, all the builds are failing: https://github.com/pypa/setuptools/issues/2614.
I think we should enable the PR RTD builds toggle (https://github.com/pypa/setuptools/issues/2469) to notice such problems early.

PR #2588, opened 2021-03-03
Provide support for dash-separated keys in setup.cfg, warns for future incompatibility of certain keys

### LINKED ISSUE #1608: extras_require in setup.cfg doesn't allow hyphen in extras name
I'm porting a `setup.py` file to declarative `setup.cfg`. In this package, an extra [contains a hyphen](https://github.com/pypa/twine/blob/6c4d5ecf2596c72b89b969ccc37b82c160645df8/setup.py#L91). This works when defined in `setup.py` but does not when defined in `setup.cfg`. If I change the hyphen to an underscore, it works again, but this would break the extra for users.

To reproduce:

`setup.py`:

```py
from setuptools import setup

setup()
```

`setup.cfg`:

```ini
[options.extras_require]
extra-a =
    pytest
extra_b =
    pytest
```

`tox.ini` (to automate the virtual env creation)

```ini
[testenv:a]
extras = extra-a

[testenv:b]
extras = extra_b
```

Then run:

```
$ tox -e a
$ tox -e b
```

Notice the extras are installed in the "b" environment but not the "a" environment. AFAICT, the only difference is the use of a hyphen instead of an underscore.

melissa-kun-li:
## Summary of changes

Closes #1608

This change will preserve dashes in the keys of `setup.cfg`. This will also implement a compatibility method which will warn users through `UserWarning` if they use an alias involving a dash which should technically be underscore separated (e.g. using `author-email` instead of `author_email`). After warning for these types of aliases, dash to underscore conversion will be performed for compatibility, but in the future Setuptools will no longer support the alias conversion.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


setuptools/dist.py
```diff
@@ -623,6 +623,21 @@ def _parse_config_files(self, filenames=None):  # noqa: C901
             except ValueError as e:
                 raise DistutilsOptionError(e) from e
 
+    def dash_to_underscore_warning(self, opt):
+        if opt in (
+            'home-page', 'download-url', 'author-email',
+            'maintainer-email', 'long-description', 'build-base',
+            'project-urls', 'license-file', 'license-files',
+            'long-description-content-type',
```
jaraco, line 631: How sure are we that this list is exhaustive? If we've missed one, what will happen? Did you consider just checking for `'-' in opt`? I know it was me who suggested the explicit list, so I'm critiquing my own guidance here too.

jaraco (approved):
This is really great. Nice work on creating tests for the warning as well as for the condition that led to this change. Also nice work on the guidance. I'm a little concerned about the explicit list of options. Let's mull over the tradeoffs there and we're good to go.

setuptools/dist.py
jaraco: I like the way you're thinking in 9c2b717.

melissa-kun-li: Hmm that is true. I wrote it with the assumption that we could continue adding to the list if needed, but on second thought that may not be very efficient or compatible. I wrote another fix which could be compatible with more users, which instead of listing the options to warn, it will generally warn if `'-' in opt` and for a certain list it won't do the conversion of hyphen to underscore so that it addresses the issues of PR #1608 for `extras_require` and `data_files`. So if more users report they need the dash, we can add to a smaller list. 

One concern I have with the new fix is that the warning can't distinguish between users who use hyphens meaningfully which may be a confusing error message for them. Although, I think it would also be more compatible for more users. What do you think?

jaraco:
Looks great. Well done!

hugovk:
Hi! I checked https://setuptools.readthedocs.io/en/latest/history.html#v54-1-0 but I wasn't entirely clear: is this meant to warn for all keys using hyphens in `setup.cfg`? For example, I see warnings for these Flake8 options:


```console
$ cat setup.cfg
[flake8]
extend-ignore = E203
max-line-length = 88

[isort]
profile = black

[tool:pytest]
addopts = -ra --color=yes
testpaths = Tests
$ python setup.py --version
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/setuptools/dist.py:633: UserWarning: Usage of dash-separated 'extend-ignore' will not be supported in future versions. Please use the underscore name 'extend_ignore' instead
  warnings.warn(
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/setuptools/dist.py:633: UserWarning: Usage of dash-separated 'max-line-length' will not be supported in future versions. Please use the underscore name 'max_line_length' instead
  warnings.warn(
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/setuptools/dist.py:633: UserWarning: Usage of dash-separated 'extend-ignore' will not be supported in future versions. Please use the underscore name 'extend_ignore' instead
  warnings.warn(
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/setuptools/dist.py:633: UserWarning: Usage of dash-separated 'max-line-length' will not be supported in future versions. Please use the underscore name 'max_line_length' instead
  warnings.warn(
8.2.0.dev0
```

Flake8 themselves say both underscores and hyphens are fine in `setup.cfg`:

https://flake8.pycqa.org/en/latest/user/configuration.html#project-configuration

Thanks!

melissa-kun-li:
@hugovk 
Hi! So prior to v54.1.0, `setup.cfg` read hyphens as underscores internally which is why flake8 said that you can use either hyphens or underscores because it was being converted to underscores in the background anyway. However this caused bugs for users who needed dashes to be preserved in `setup.cfg`. The recent warning is to warn for future deprecation of dashes in keys of `setup.cfg` for those keys which should actually be underscores. So the flake8 documentation was right before, but maybe now we should let them know so they can update it. 

@jaraco

jaraco:
Oh, this was a case I hadn't considered during the review. I had expected it would only apply to sections that setuptools is using. 

The `setup.cfg` file serves as a home for many projects outside the scope of distutils/setuptools. In that case, Setuptools probably shouldn't be as opinionated about the format of the keys. I'd like to avoid other projects needing to mirror this change.

I believe we should come up with a patch to avoid these warnings. I'll convert Hugo's comment to an issue. @melissa-kun-li Would you be interested in working on the patch? In this case, I'd say a test is optional. Use `misc` for the changelog.

melissa-kun-li:
@jaraco Sure! I'll work on a fix, I have an idea in mind

PR #2593, opened 2021-03-06
Warn for uppercase key usage in metadata in setup.cfg, provides compatibility. Fixes #2592

### LINKED ISSUE #2592: "Name" metadata keyword not recognized
Starting with v53.1.0 setuptools seems to no longer recognizes the "Name" key in the [metadata] section of setup.cfg. A "name" key is recognized. 

The metadata spec uses "Name". https://packaging.python.org/specifications/core-metadata/.

Possible related to issue #1937 that was merged in at that release.

melissa-kun-li:
## Summary of changes

Closes #2592

This will warn users if they use an uppercase key in `metadata` in `setup.cfg`. This will also convert any uppercase keys in `metadata` to lowercase to restore some compatibility due to the change by #2580 in which keys in `setup.cfg` became case-sensitive.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


setuptools/dist.py
```diff
@@ -636,6 +637,17 @@ def dash_to_underscore_warning(self, opt, section):
                 % (opt, underscore_opt))
         return underscore_opt
 
+    def uppercase_warning(self, opt, section):
+        if section in ('metadata',) and (any(c.isupper() for c in opt)):
```
webknjaz, line 641: I think it's better to use a guard-expression style similar to the method above. It allows us to dedent the whole block of code and only keep a single line indented:
```suggestion
        if section != 'metadata' or all(not c.isupper() for c in opt):
            return opt

```

webknjaz: Looking closer, I realized that you don't even need that `all()` with a generator looping over vars because you can just call `str.islower()` on the whole opt name to achieve the same:
```suggestion
        if section != 'metadata' or opt.islower():
            return opt

```

setuptools/tests/test_config.py
```diff
@@ -526,6 +526,24 @@ def test_dash_to_underscore_warning(self, tmpdir):
...
+            'description = Some description\n'
+        )
+        msg = ("Usage of uppercase key 'Name' in 'metadata' will be deprecated in "
+               "future versions. "
+               "Please use lowercase 'name' instead")
+        with pytest.warns(UserWarning, match=msg):
+            with get_dist(tmpdir) as dist:
+                metadata = dist.metadata
+                assert metadata.name == 'foo'
+                assert metadata.description == 'Some description'
```
webknjaz, line 545: It's cleaner to keep assertions separate from the patched block. Also `pytest.warns()` is a separate check from the assert statements:
```suggestion

        assert metadata.name == 'foo'
        assert metadata.description == 'Some description'
```

setuptools/dist.py
```diff
@@ -636,6 +637,17 @@ def dash_to_underscore_warning(self, opt, section):
                 % (opt, underscore_opt))
         return underscore_opt
 
+    def uppercase_warning(self, opt, section):
```
webknjaz, line 640: This method name is too broad. It's usually better to (1) be more specific about what it actually does and (2) use a verb because methods/functions do some actions, unlike variables/properties that only hold some values:
```suggestion
    def make_option_lowercase(self, opt, section):
```

setuptools/dist.py
```diff
@@ -636,6 +637,17 @@ def dash_to_underscore_warning(self, opt, section):
...
+        if section in ('metadata',) and (any(c.isupper() for c in opt)):
+            lowercase_opt = opt.lower()
+            warnings.warn(
```
webknjaz, line 643: This should probably have a `DeprecationWarning` argument too.

changelog.d/2592.change.rst
```diff
@@ -0,0 +1 @@
+Warned for usage of uppercase keys in :code:`metadata` in :code:`setup.cfg`, restored some compatiblity by transforming to lowercase if not. Followed the change of keys in :code:`setup.cfg` now being case-sensitive -- by :user:`melissa-kun-li`
```
webknjaz, line 1: FYI RST has double backticks for inline code so there's no need to use an explicit `:code:` role. Also, according to the code, the user-facing change (the impact) is that the option name is normalized, not that the warning is emitted — that warning is additional instrumentation, not the main change:

```suggestion
Made option keys in the ``[metadata]`` section of ``setup.cfg`` case-sensitive. Users having
uppercase option spellings will get a warning suggesting to make them to lowercase
-- by :user:`melissa-kun-li`
```

setuptools/dist.py
```diff
@@ -599,6 +599,7 @@ def _parse_config_files(self, filenames=None):  # noqa: C901
 
                     val = parser.get(section, opt)
                     opt = self.dash_to_underscore_warning(opt, section)
+                    opt = self.uppercase_warning(opt, section)
```
jaraco, line 602:
```suggestion
                    opt = self.make_option_lowercase(opt, section)
```

setuptools/dist.py
jaraco: In my experience, `DeprecationWarnings` are suppressed by default so aren't useful for user-facing warnings. Perhaps a `SetuptoolsDeprecationWarning` would be appropriate, but if I recall correctly, the other warning issued is the default (`UserWarning`). If a different warning class is preferred, let's do that in a separate change.

jaraco:
Thanks Melissa for the follow up and quick response. Thanks Sviatoslav for the edits.

setuptools/dist.py
webknjaz: @jaraco pyca/cryptography does this and it causes problems for people attempting to granularly ignore the thing when the place where it's emitted is in the import chain of the custom exception.

See: https://github.com/pytest-dev/pytest/issues/7524 / https://github.com/pyca/cryptography/issues/5335 / https://github.com/pyca/cryptography/pull/5346 / https://cryptography.io/en/3.4.6/faq.html#i-cannot-suppress-the-deprecation-warning-that-cryptography-emits-on-import

melissa-kun-li:
Thank you @webknjaz and @jaraco for the code review/edits and explanations! :)

PR #2608, opened 2021-03-16
build_meta: produce informative error when a dist is not found

layday:
## Summary of changes

Previously, when `build_sdist` or `build_wheel` were unable
to build a distribution (and were therefore unable to find the
distribution file), they would throw a

    ValueError: not enough values to unpack (expected 1, got 0)

which did not offer any clues as to where the issue might lie.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -101,8 +101,12 @@ def _file_with_extension(directory, extension):
         f for f in os.listdir(directory)
         if f.endswith(extension)
     )
-    file, = matching
-    return file
+    try:
+        return next(matching)
+    except StopIteration:
+        raise ValueError('No distribution was found. The distribution was '
+                         'possibly not built. Ensure that your `setup.py` '
+                         'is not empty and that it calls `setup()`.')
 
 
 def _open_setup_script(setup_script):
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -107,8 +107,7 @@ def _file_with_extension(directory, extension):
         raise ValueError(
             'No distribution was found. Ensure that `setup.py` '
             'is not empty and that it calls `setup()`.')
-    else:
-        return file
+    return file
 
 
 def _open_setup_script(setup_script):
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -104,9 +104,9 @@ def _file_with_extension(directory, extension):
     try:
         return next(matching)
     except StopIteration:
-        raise ValueError('No distribution was found. The distribution was '
-                         'possibly not built. Ensure that your `setup.py` '
-                         'is not empty and that it calls `setup()`.')
+        raise ValueError(
+            'No distribution was found. Ensure that `setup.py` '
+            'is not empty and that it calls `setup()`.')
 
 
 def _open_setup_script(setup_script):
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -102,11 +102,13 @@ def _file_with_extension(directory, extension):
         if f.endswith(extension)
     )
     try:
-        return next(matching)
-    except StopIteration:
+        file, = matching
+    except ValueError:
         raise ValueError(
             'No distribution was found. Ensure that `setup.py` '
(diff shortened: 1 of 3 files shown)


layday:
See https://github.com/pypa/packaging.python.org/issues/838 (and subsequently https://github.com/pypa/packaging-problems/issues/467 and https://github.com/pypa/packaging-problems/issues/468) which motivated this change.

setuptools/build_meta.py
```diff
@@ -101,8 +101,12 @@ def _file_with_extension(directory, extension):
         f for f in os.listdir(directory)
         if f.endswith(extension)
     )
-    file, = matching
-    return file
+    try:
+        return next(matching)
```
jaraco, line 105: Previously, the code would check for exactly one match. Now it only checks for at least one match (and ignores all but the first). Could that silently swallow errors?

setuptools/build_meta.py
```diff
@@ -101,8 +101,12 @@ def _file_with_extension(directory, extension):
...
+    except StopIteration:
+        raise ValueError('No distribution was found. The distribution was '
+                         'possibly not built. Ensure that your `setup.py` '
+                         'is not empty and that it calls `setup()`.')
```
jaraco, line 109: Avoid hanging indents and use of first/second person. Remove 'Distribution was possibly not built', as that's mainly redundant with `no distribution was found`.

```suggestion
        raise ValueError(
            'No distribution was found. Ensure that `setup.py` '
            'is not empty and that it calls `setup()`.')
```

setuptools/tests/test_build_meta.py
```diff
@@ -442,6 +443,19 @@ def test_sys_argv_passthrough(self, tmpdir_cwd):
...
+    def test_build_with_empty_setuppy(self, build_backend, build_hook):
+        files = {'setup.py': ''}
+        path.build(files)
+
+        with pytest.raises(
+                ValueError,
+                match=re.escape(
+                    'No distribution was found. The distribution was '
+                    'possibly not built. Ensure that your `setup.py` '
+                    'is not empty and that it calls `setup()`.')):
```
jaraco, line 456: Simplify match to avoid over-matching the expected text. Just assert on the gist of the message.

```suggestion
                match=re.escape('No distribution was found.')):
```

jaraco (approved):
Looks good. Just a few changes requested.

setuptools/build_meta.py
layday: I suppose, if multiple distributions would be built somehow or if the folder already had a distribution in it but then all bets are off anyway.

setuptools/build_meta.py
```diff
@@ -101,7 +101,12 @@ def _file_with_extension(directory, extension):
         f for f in os.listdir(directory)
         if f.endswith(extension)
     )
-    file, = matching
+    try:
+        file, = matching
+    except ValueError:
+        raise ValueError(
+            'No distribution was found. Ensure that `setup.py` '
+            'is not empty and that it calls `setup()`.')
```
henryiii, line 109: Should this be `raise ValueError(...) from None`, to simplify the output? The original ValueError from unpacking the tuple is not interesting to anyone and adding noise, since it's literally the only thing done here?

layday: That's a good idea.

layday: Then again, somehing else might raise a `ValueError` since the generator isn't consumed before it's unpacked and `ValueError` is so ubiquitous that I'm a little wary of erasing it.

layday: If you'd like to convert the generator to a list comprehension and then raise from `None`, I think that'd be fine.

henryiii: Since the generator is created on the line above, there currently should be no way to for it to be exhausted before it hits this point?

layday: Yes, hence it might raise an unrelated value error in the try block.

henryiii: Ahh, I see what you mean.

PR #2611, opened 2021-03-19
Relax MACOSX_DEPLOYMENT_TARGET check

jmroot:
Only complain if the config target is >= 10.3 and the current target is
< 10.3. The check was originally added to ensure that incompatible
LDSHARED flags are not used, because `-undefined dynamic_lookup` is
used when building for 10.3 and later, and is not supported on older OS
versions. Apart from that, there should be no problem in general with
using an older target.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/setuptools/_distutils/spawn.py
+++ b/setuptools/_distutils/spawn.py
@@ -60,13 +60,17 @@ def spawn(cmd, search_path=1, verbose=0, dry_run=0, env=None):
             if _cfg_target:
                 _cfg_target_split = [int(x) for x in _cfg_target.split('.')]
         if _cfg_target:
-            # ensure that the deployment target of build process is not less
-            # than that used when the interpreter was built. This ensures
-            # extension modules are built with correct compatibility values
+            # Ensure that the deployment target of the build process is not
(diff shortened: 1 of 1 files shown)


jmroot:
Actually, it looks like the only reason this check exists in the first place is that `LDSHARED` is different depending on whether `MACOSX_DEPLOYMENT_TARGET` is less than 10.3 or not. So the only time it actually needs to error is when the config target is >= 10.3 and the current target is < 10.3.

And for that matter, I really doubt if current cpython versions build on Mac OS X 10.2 at all. Maybe it would be best to remove this check entirely and have cpython drop any pretense of supporting linking extensions without `-undefined dynamic_lookup` on macOS.

jmroot:
OK, how about this for now, and we'll see what Python core says about dropping 10.2 support and go from there later.

jaraco:
Any changes to `_distutils` should be filed with pypa/distutils. Thanks.

jmroot:
Ah OK, I thought distutils was part of setuptools now and didn't realise there was a separate repo. Will do.

PR #2620, opened 2021-03-28
License_files improvements

### LINKED ISSUE #2616: [FR] Auto-include license files in source distributions

License files are only included in the source distribution if they are specified in either `license_file`, `license_files` or `MANIFEST.in`. `wheel` automatically adds a few common glob patterns if the `license_files` option isn't specified. https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file

The result is a situation where a license file is included in the wheel, but not in the source distribution.

Match the behavior of `wheel` and include the most common glob patterns automatically if neither `license_file` nor `license_files` is specified.

### Alternative Solutions

 --

### Additional context

PR that originally added support for `license_files`: https://github.com/pypa/setuptools/pull/1767
Commit that added glob patterns in `wheel`: https://github.com/pypa/wheel/commit/59976ab294e1b118f42cab404d95df66ed55f7e4

### Code of Conduct

cdce8p:
## Summary of changes

- Add support for glob patterns in `license_file` and `license_files` options.
- Add default patterns if neither is specified to match behavior of `bdist_wheel`.
https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file
- Added deprecation warning if `license_file` file is used. `license_files` is a full replacement.
- Added additional documentation for the `license_files` option.

The implementation is similar to the one used by `wheel`: [wheel/bdist_wheel.py](https://github.com/pypa/wheel/blob/41b375ef33e3368f6f58c13f9649491d01c2ba5e/src/wheel/bdist_wheel.py#L410-L435)

Closes #2616

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/docs/references/keywords.rst
+++ b/docs/references/keywords.rst
@@ -76,6 +76,18 @@ Keywords
 ``license``
     A string specifying the license of the package.
 
+``license_file``
+
+    .. warning::
+        ``license_file`` is deprecated. Use ``license_files`` instead.
+
+``license_files``
+
+    A list of glob patterns for license related files that should be included.
+    If neither ``license_file`` nor ``license_files`` is specified, this option
+    defaults to ``LICEN[CS]E*``, ``COPYING*``, ``NOTICE*``, and ``AUTHORS*``.
+    Any ``exclude`` specified in ``MANIFEST.in`` will overwrite it.
+
 ``keywords``
     A list of strings or a comma-separated string providing descriptive
     meta-data. See: `PEP 0314`_.
--- a/docs/references/keywords.rst
+++ b/docs/references/keywords.rst
@@ -86,7 +86,6 @@ Keywords
     A list of glob patterns for license related files that should be included.
     If neither ``license_file`` nor ``license_files`` is specified, this option
     defaults to ``LICEN[CS]E*``, ``COPYING*``, ``NOTICE*``, and ``AUTHORS*``.
-    Any ``exclude`` specified in ``MANIFEST.in`` will overwrite it.
 
 ``keywords``
     A list of strings or a comma-separated string providing descriptive
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -213,6 +213,9 @@ class sdist(sdist_add_defaults, orig.sdist):
             patterns.append(opts['license_file'][1])
 
         if 'license_file' not in opts and 'license_files' not in opts:
+            # Default patterns match the ones wheel uses
+            # See https://wheel.readthedocs.io/en/stable/user_guide.html
+            # -> 'Including license files in the generated wheel file'
             patterns = ('LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*')
 
         for pattern in patterns:
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -4,6 +4,8 @@ import os
 import sys
 import io
 import contextlib
+import warnings
+from glob import iglob
 
 from setuptools.extern import ordered_set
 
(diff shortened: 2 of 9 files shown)


docs/references/keywords.rst
```diff
@@ -76,6 +76,18 @@ Keywords
...
+
+    .. warning::
+        ``license_file`` is deprecated. Use ``license_files`` instead.
+
+``license_files``
+
+    A list of glob patterns for license related files that should be included.
+    If neither ``license_file`` nor ``license_files`` is specified, this option
+    defaults to ``LICEN[CS]E*``, ``COPYING*``, ``NOTICE*``, and ``AUTHORS*``.
+    Any ``exclude`` specified in ``MANIFEST.in`` will overwrite it.
```
jaraco, line 89: In my experience, `MANIFEST.in` only affects the sdist. Can you confirm that `exclude` in MANIFEST.in with `LICENSE` present has the same effect as not specifying a license on older Setuptools?

setuptools/command/sdist.py
```diff
@@ -194,29 +195,38 @@ def check_license(self):
...
+            else ordered_set.OrderedSet(license_files)
+
+        if 'license_file' in opts:
+            log.warn(
+                "warning: the 'license_file' option is deprecated, "
+                "use 'license_files' instead")
+            patterns.append(opts['license_file'][1])
+
+        if 'license_file' not in opts and 'license_files' not in opts:
+            patterns = ('LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*')
```
jaraco, line 216: Can you make a comment that indicates these values are shared with wheel?

setuptools/tests/test_egg_info.py
```diff
@@ -644,7 +652,37 @@ def test_setup_cfg_license_file(
             'MANIFEST.in': "exclude LICENSE-XYZ",
             'LICENSE-ABC': "ABC license",
             'LICENSE-XYZ': "XYZ license"
-        }, ['LICENSE-ABC'], ['LICENSE-XYZ'])  # subset is manually excluded
+        }, ['LICENSE-ABC'], ['LICENSE-XYZ']),  # subset is manually excluded
+        pytest.param({
+            'setup.cfg': DALS("""
+                              """),
```
jaraco, line 658: No need for DALS.
```suggestion
            'setup.cfg': "",
```

setuptools/tests/test_virtualenv.py
```diff
@@ -85,7 +85,7 @@ def skip_network(param):
         mark('pip==18.1', issue2599),
         mark('pip==19.3.1', pytest.mark.xfail(reason='pypa/pip#6599')),
         'pip==20.0.2',
-        'https://github.com/pypa/pip/archive/master.zip',
+        'https://github.com/pypa/pip/archive/main.zip',
```
jaraco, line 88: Nice catch.

jaraco (approved):
Nice work. Thorough and clean. Just a couple of nitpicks or requests for additional information, but it's basically ready to go.

docs/references/keywords.rst
cdce8p: I'm not exactly sure what you mean. AFAIK the change itself will only the `sdist` command. `bdist_wheel` will use the `license_file` and `license_files` fields or the default patterns. Therefore it's probably not wise to exclude a license file in the `MANIFEST.in` to begin with. But in case you only care about the source dist, any `exclude` does have president over a matched license file. There are two existing test cases that cover it.
I'm going to remove the last sentence, since it's probably more confusing than helpful especially to beginners.
https://github.com/pypa/setuptools/blob/743298103c05e17a22a74afdf5b2bf47630d2eb4/setuptools/tests/test_egg_info.py#L637-L655

cdce8p:
Thanks for the review! I did a rebase to get rid of the test fix and pushed a small commit to address the comments: https://github.com/pypa/setuptools/pull/2620/commits/608c376e86326c879dd52b56660b2247a3ca854e

As a side note here, I was thinking about if it would be possible to do these steps even earlier. [PEP 639](https://www.python.org/dev/peps/pep-0639/), currently still a draft but quite far along, would add the `License-File` (multiple) field to the package metadata, [Link](https://www.python.org/dev/peps/pep-0639/#license-file-multiple-use). To support that use case, the pattern matching for license files probably needs to happen during the `egg-info` command. What is your opinion regarding this? Could you give my a pointer on how to best approach it?

jaraco:
> What is your opinion regarding this? Could you give my a pointer on how to best approach it?

I'm in favor of exposing the license text(s) in metadata. I don't have any active context on such an effort. I don't even know where license files are presented currently.

I'd start out by filing a bug and describe the problem and start exploring possible solutions. I'd use that forum to examine the code and current behavior and devise a transition to a solution that meets the need. I'd identify projects and people that would be affected by such a change and gather their feedback.

PR #2626, opened 2021-03-29
Fix typo in setup.py example in quickstart guide.

Yhg1s:
Fix a small typo in the 'Basic Use' example of the Quickstart section of the User Guide.


jaraco:
Thanks for the contrib!

PR #2627, opened 2021-03-29
distutils: pass -rpath on macOS when requested

acolinisi:
## Summary of changes

Resurrecting a patch to an old issue that is still relevant: distutils does not set the runtime search path when linking the shared object on macOS, passing a `-L`  argument instead of an `-rpath` argument, like on the other systems. The code in question states that OS X linker doesn't accept `-rpath`, but that code was written 19 years ago, and that comment is no longer true (certainly not on macOS 11.2, but I do not know what is the newest system version for which `-rpath` is not accepted, if any).

The `runtime_library_dirs` configuration option is vital for extensions that need to build against libraries that temporarily exist in a location different from their final installation location (such as during the build of a package for a distro).

Fix -R option of build_ext for OSX

Resolves this old bug against distutils that expired due to PIP 632:
https://bugs.python.org/issue36353

Applies patch originally submitted to CPython:
https://github.com/python/cpython/pull/12418

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2627.change.rst
+++ b/changelog.d/2627.change.rst
@@ -1,0 +1,1 @@
+Honor `runtime_library_dirs` on macOS by passing `-rpath` to the linker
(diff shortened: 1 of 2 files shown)


jaraco:
Changes to distutils should be made at pypa/distutils (and then merged here). Can you contribute the change there? Thanks.

PR #2628, opened 2021-03-30
Write long description in message payload

cdce8p:
## Summary of changes
Write the long description in the message body of the `PKG-INFO` file if the metadata version `>= 2.1`.
https://packaging.python.org/specifications/core-metadata/#description

`wheel` does this conversion already, see: [wheel/metadata.py#L86-L89](https://github.com/pypa/wheel/blob/41b375ef33e3368f6f58c13f9649491d01c2ba5e/src/wheel/metadata.py#L86-L89)

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
Thanks for the contrib.

I'm not sure this change is a safe one. As you've pointed out, the 'wheel' package converts from egg-info to dist-info, and it *assumes* the PKG-INFO file has the description in a field. Won't that break if Setuptools starts putting the long description in metadata 2.1 format? Note also that metadata 2.1 allows the long description (and other multi-line values) to remain in the fields. My instinct is this change is only likely to cause disruption.

Can you write up a motivation about what benefit this change adds or what problem it solves?

setuptools/dist.py
```diff
@@ -67,7 +68,7 @@ def get_metadata_version(self):
     return mv
 
 
-def read_pkg_file(self, file):
+def read_pkg_file(self, file):  # noqa: C901
```
jaraco, line 71: These complexity checks are here for a reason. Can you devise a solution that avoids the complexity?

setuptools/dist.py
```diff
@@ -100,8 +119,9 @@ def _read_list(name):
     else:
         self.download_url = None
 
-    self.long_description = _read_field('description')
-    self.description = _read_field('summary')
+    self.long_description = _read_long_description()
+    if self.long_description is None and self.metadata_version >= StrictVersion('2.1'):
```
jaraco, line 123: Is there a solution that doesn't select on metadata version? Note that long description in Metadata 2.1 could still use the Description field.

setuptools/dist.py
```diff
@@ -83,6 +84,24 @@ def _read_list(name):
             return None
         return values
 
+    def _read_long_description():
+        value = msg['description']
+        if value in ('UNKNOWN', None):
```
jaraco, line 89: I'm uncomfortable with these "UNKNOWN" literals. Does Setuptools add "UNKNOWN" for unsupplied long descriptions?

setuptools/dist.py
```diff
@@ -83,6 +84,24 @@ def _read_list(name):
             return None
         return values
 
+    def _read_long_description():
```
jaraco, line 87: I'd like to avoid copy/pasting this metadata reading code. Let's instead get `importlib_metadata` vendored in here and rely on that to read metadata.

cdce8p:
> I'm not sure this change is a safe one. As you've pointed out, the 'wheel' package converts from egg-info to dist-info, and it _assumes_ the PKG-INFO file has the description in a field. Won't that break if Setuptools starts putting the long description in metadata 2.1 format? Note also that metadata 2.1 allows the long description (and other multi-line values) to remain in the fields. My instinct is this change is only likely to cause disruption.

At least for `wheel`, this isn't an issue. `wheel` uses `email.parser.Parser.parse` to parse `EGG-INFO` and then `msg["description"]` to access the field. If the header field doesn't exist, this will return `None`. If it returns a string, `wheel` will dedent it and write it into the payload of the message, deleting the header field afterwards.
With this PR, the long description will be automatically written into the payload part of `EGG-INFO`, so `wheel` just doesn't need to do the dedent step. It's already present in the payload after `parse`.

> Can you write up a motivation about what benefit this change adds or what problem it solves?

The main point would probably be to improve readability of the `EGG-INFO` file. If the long description is written in the payload instead of the header, all fields are easily readable. Additionally this is also the format that `wheel` uses, so it would lead to a better standard of sorts. Of course it works now too, so there isn't an active need to change it on that front.

setuptools/dist.py
cdce8p: `UNKNOWN` is the `distutils` default for `get_long_description()`. Since this is used when writing the metadata file, either the header field or the payload will contain `UNKNOWN` in case no `description` is given.
https://github.com/pypa/setuptools/blob/03ad13c97fc47b4773c49fa486f7f9b3349b7e56/setuptools/_distutils/dist.py#L1197-L1198

setuptools/dist.py
cdce8p: It will try to read the long description from the header field first. Only if the result is `None` and the `mv >= '2.1'` we try to read it from the payload as well. It will work that way, since in `2.1` an `UNKNOWN` value for the long description is added either in the header field or the payload.

setuptools/dist.py
cdce8p: I'm not sure this would solve it. `imporlib_metadata` doesn't really parse the metadata. It just returns an email message object which can then be inspected. We would still need to do most if not all of these steps.

setuptools/dist.py
cdce8p: Yeah, it would be possible to refactor this function. I'm just not sure it's the right thing to do here. Currently at least, we got all parsing code in one place. Refactor would probably spilt this into multiple functions.

webknjaz: And this is the intended outcome of us having the complexity check enabled: adding noqa effectively ruins the expectation that we keep the code simple at all times. If you don't like doing the refactoring (rightfully) within this PR, please consider submitting a separate one and then rebasing on top of that.

changelog.d/2828.change.rst
```diff
@@ -0,0 +1,2 @@
+Write long description in message payload of PKG-INFO file.
+Changed in metadata version 2.1
```
webknjaz, line 2: @cdce8p please take a look at the explanations in the README here https://github.com/pypa/setuptools/tree/main/changelog.d — it's better to use the past tense consistently (your first sentence is in the present tense, in imperative form). You may optionally sign this fragment with your handle as shown in the examples.

setuptools/dist.py
cdce8p: I can do the refactoring here, that's not the point. What I was trying to say is that I don't believe this to be a good idea. IMO leaving it as it is at the moment is easier to understand than any refactored version would be. That's why I added the `noqa` for only this specific function.

If you still want me to refactor it, please let me know and I'll do it.

changelog.d/2828.change.rst
cdce8p: Using something else then present tense doesn't make sense IMO. Take `Written long description ...`, for example. This would be wrong since it's a continuous process and doesn't stop in the past.

Past tense does make sense when writing about a feature that has been added.

setuptools/dist.py
webknjaz: It's already quite complex. Adding `noqa` would enable people to make it worse over time without alerting them about the problem. So I think that having refactoring can be a good idea if performed properly.

If you look closer, all of these internal helper functions could be pure functions with just one extra argument allowing them to be kept outside of this function body.
OTOH, the problem here is that the abstractions this function operates are not of the same level. Why the data accessors do translation/normalization and also mapping to this object attributes?

It could be easier to move the call to `message_from_file()` into a separate function and have all the data normalization there. Meaning that here you'd just call that helper and get all the data ready to use.

cdce8p:
I've rebased and updated this PR, after #2635 was merged. I would appreciate any additional feedback.

/CC: @jaraco @webknjaz

setuptools/tests/test_egg_info.py
```diff
@@ -855,6 +855,33 @@ def test_long_description_content_type(self, tmpdir_cwd, env):
...
+        # keyword args to the `setup` function results in writing
+        # the description in the message payload of the `PKG-INFO` file
+        # in the `<distribution>.egg-info` directory.
+        self._setup_script_with_requires(
+            "long_description='This is a long description\\nover multiple lines',"
+            "long_description_content_type='text/markdown',"
+        )
+        environ = os.environ.copy().update(
+            HOME=env.paths['home'],
+        )
```
webknjaz, line 869: `dict.update()` does not return anything, it updates the dict in-place. Meaning that `environ=None` always in this test.

I believe what you meant to do is:
```suggestion
        environ = dict(os.environ, HOME=env.paths['home'])
```

setuptools/tests/test_egg_info.py
```diff
@@ -855,6 +855,33 @@ def test_long_description_content_type(self, tmpdir_cwd, env):
...
+            pypath=os.pathsep.join([env.paths['lib'], str(tmpdir_cwd)]),
+            data_stream=1,
+            env=environ,
+        )
+        egg_info_dir = os.path.join('.', 'foo.egg-info')
+        with open(os.path.join(egg_info_dir, 'PKG-INFO')) as pkginfo_file:
+            pkg_info_lines = pkginfo_file.read().split('\n')
+        assert 'Metadata-Version: 2.1' in pkg_info_lines
+        assert '' == pkg_info_lines[-1]
+        long_desc_lines = pkg_info_lines[pkg_info_lines.index(''):]
```
webknjaz, line 881: This looks like too much logic in a test. Isn't the empty line position supposed to be predictable? Just use that number then.

setuptools/dist.py
```diff
@@ -228,6 +238,9 @@ def write_field(key, value):
         for extra in self.provides_extras:
             write_field('Provides-Extra', extra)
 
+    if version >= StrictVersion('2.1'):
```
webknjaz, line 241: You use this comparison twice in this patch. It's time to put it into a var, don't you think? You could have nicer checks like `if at_least_2_1:` and `if not at_least_2_1:`

setuptools/tests/test_egg_info.py
cdce8p: Tbh I just copied that part from the test above without taking a closer look 😅 Turns out it works perfectly even completely without it.

setuptools/tests/test_egg_info.py
cdce8p: True. I was overthinking this one

setuptools/dist.py
cdce8p: I don't really think that it's necessary. At least not just yet. There are also two checks for version `1.2` without an extra variable.

setuptools/tests/test_egg_info.py
webknjaz: Interesting. So this means there's a bug in other tests...

setuptools/dist.py
webknjaz: Fair enough

cdce8p:
@jaraco Would you mind taking a look at this one again? Since both, #2635 and #2641 are now merged, the change is much more reasonable.

mhertz:
Sorry, but is there any way to stop this from happening? Currently I have to downgrade python-setuptools to 0.56 every time building an egg, and I see no switches plus no matter how messing with setup.py then does this i.e adds description or Unknown into payload which result in error in program using the egg(yeah issue with program I know and did report bug there, but in mean-time). The metadata spec states Can place in payload, but always happens.

Thanks in advance.

jaraco:
@mhertz Sorry to hear this caused disruption. Now that I think about it, forcing projects to jump from metadata 1.x to 2.x probably should have been a backward-incompatible change.

What you're proposing is that Setuptools provide a flag and two different ways of rendering the metadata. Such a change would almost certainly complicate the implementation and require additional tests - a lot of work for one use-case. It sounds like you have already identified a fix for the root cause and also a (albeit clumsy) workaround (downgrade Setuptools).

Why not work to expedite the fix in the offending program?

This project may consider an option to fall back to the prior behavior, maybe even though a third-party plugin. But before it does that, it would really help to have an issue to track the issue. In particular:

- What are the circumstances that produce the unwanted behavior (what program or programs fail to work with the newer metadata format)?
- In what environments does this occur and why is downgrading Setuptools problemmatic?
- What is the impact of the unwanted behavior? How much is the disruption?
- How can one replicate the undisireable behavior (minimal repro)?

Finally, would you be willing to help contribute the solution?

mhertz:
Thanks alot for your nice detailed reply, much appreciated :)

You're right of-course and I agree fully. I'm not good enough to make such changes honestly, and don't see the need on second thought, well more after reading your post, but regardless :) 

It's of-course better to fix offending program than to complicate this project and wrong way to go about it obviously. Anyway I just though as when the metadata spec states that it _can_ do this, then maybe was a way to control this, or atleast explain the circumstances of this i.e. when happens - I checked if was contollable in some way already e.g. from long_description_content_type or alike etc, but didn't find any, but as said, agree with this isn't an issue of this project, and also downgrading for me is an easy workaround for the offending program in question, obviously, and sorry for exagerating the issue :)

The offending program is the deluge torrent client, which accepts plugins in egg format, and when enabling such plugin then also reads various fields from PKG-INFO e.g. name, version and description etc, and then shows error as description now returns none. By that token, I apologise as also just realized that I actually entirely misunderstood the issue altogether, as instead was that the offending program relies on description alias which also is warned upon in docs, unless i'm mistaken again, and instead should use summary merely, which also btw was my original posted solution in my bug-ticket to project, though instead should have made PR in retrospect, but just thought the issue was about description now being in payload, but which is long-description and unrelated here, sorry. I'll update my deluge bug-ticket shortly, as currently have erronimous info in it about root-cause, but again, nothing to do with you guys. https://dev.deluge-torrent.org/ticket/3476 

So, thanks again for your response, and i'm sorry for the noise and agree with your points entirly, thank you :)

jaraco:
>  I believe description = sumarry and long-description = description etc, and not even sure I'll 100% correct here still lol :)

Yes. I believe you got this right this time. Sorry it's confusing, but glad you're getting it figured out.

PR #2629, opened 2021-04-01
fix: more focused output if setup.py is empty

henryiii:
## Summary of changes

Followup to #2608. Makes sure an unrelated ValueError is not caught by the try/except, then removes the extra exception printout. Also moves the msg to the line above to avoid a double message printout in the traceback.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -97,16 +97,16 @@ def _get_immediate_subdirectories(a_dir):
 
 
 def _file_with_extension(directory, extension):
-    matching = (
+    matching = [
         f for f in os.listdir(directory)
         if f.endswith(extension)
-    )
+    ]
     try:
         file, = matching
     except ValueError:
-        raise ValueError(
-            'No distribution was found. Ensure that `setup.py` '
-            'is not empty and that it calls `setup()`.')
+        msg = ('No distribution was found. Ensure that `setup.py` '
+               'is not empty and that it calls `setup()`.')
+        raise ValueError(msg) from None
     return file
 
 


henryiii:
This should already be tested, can add more if needed, also don't know if it's large enough to warrant a changelog entry, happy to add one if it is.

henryiii:
> ERROR: HTTP error 404 while getting https://github.com/pypa/pip/archive/master.zip

Don't think I caused this... :) I'll wait to see what the feedback / interest is before retriggring.

setuptools/build_meta.py
```diff
@@ -97,16 +97,16 @@ def _get_immediate_subdirectories(a_dir):
 
 
 def _file_with_extension(directory, extension):
-    matching = (
+    matching = [
```
jaraco, line 100: I saw the conversation that inspired this change, but if I hadn't, I wouldn't understand why this was here and I'd be reluctant to accept the change. Moreover, without that context or seeing this diff, I'd be inclined to change this back to a generator expression to build on lazy evaluation.

I'm slightly inclined to think this code is becoming overly-protective and it may not be worth the effort to try to protect against unexpected value errors or to hide them from the output.

jaraco (commented):
The more I think about it, I think I prefer the existing implementation to the change proposed here. Let's just stick with what we have. If it turns out to cause actual consternation, we can revisit, but for now, I think the existing implementation is satisfactory.

Still, thanks for the contribution and consideration.

PR #2641, opened 2021-04-16
Always use latest metadata version for PKG-INFO

cdce8p:
## Summary of changes

`setuptools` should always try to use the latest supported metadata version when writing the `PKG-INFO` file.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2641.change.rst
+++ b/changelog.d/2641.change.rst
@@ -1,0 +1,2 @@
+Setuptools will now always try to use the latest supported
+metadata version for ``PKG-INFO``. - by :user:`cdce8p`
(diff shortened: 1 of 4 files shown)


jaraco:
Looks like by merging this after some of the other recent merges, the tests now fail.

jaraco:
I hadn't noticed it before, but `pkg_info_lines[7]` (or any use of a specific index) is a red flag for brittle behavior.

cdce8p:
> I hadn't noticed it before, but `pkg_info_lines[7]` (or any use of a specific index) is a red flag for brittle behavior.

Makes sense, will remember that. I originally switched to it after I got the feedback to do so: https://github.com/pypa/setuptools/pull/2628#discussion_r614237882

PR #2645, opened 2021-04-17
Add ``License-File`` field to package metadata

### LINKED ISSUE #2631: [FR] Include ``License-File`` field in package metadata

With `importlib.metadata` it's now easily possible to read the metadata and all files of a specific package. A topic that remains difficult though is the identification of license files. `setuptools` and `wheel` use the `license_file` / `license_files` files to identify those, but these fields aren't included in the final metadata. That means that external project can only guess which once are the correct license files.

PEP 639 plans to add the [`License-File`](https://www.python.org/dev/peps/pep-0639/#license-file-multiple-use) (multiple) field to the package metadata that should contain the path to a license file relative to `.egg-info` / `.dist-info`.

The PEP is currently still in draft mode, but pretty far along. The `License-File` field isn't part of the recent discussions anymore.

As a first step, I propose adding the `License-File` field to the `PKG-INFO` file as additional metadata. A second part could be to include all license files in the `.egg-info` directory.

### Alternative Solutions

_No response_

### Additional context

Discussion for PEP 639: https://discuss.python.org/t/pep-639-improving-license-clarity-with-better-package-metadata/2154

### Code of Conduct

I agree to follow the PSF Code of Conduct

cdce8p:
## Summary of changes

This PR adds the ``License-File`` field to the package metadata. Although not yet part of the official metadata spec, including this field will enable third party tools easier access to all included license files through `importlib.metadata`. This is especially useful if a file doesn't have an easily recognizable name or is in an unusual location.

`setuptools` already parses the `license_file` and `license_files` options and cross-references them with all project files to see which ones should be included. With this change, the result of that would be made available for others to use as well.

Closes #2631

--
For reference: [PEP 639](https://www.python.org/dev/peps/pep-0639/), currently still in a draft, plans to add the `License-File` field to the metadata spec, [link](https://www.python.org/dev/peps/pep-0639/#license-file-multiple-use)

--
Example use case
```py
from importlib import metadata
meta = metadata.metadata('my-project')
project_files = metadata.files('my-project')

license_file_paths = [
    path for path in project_files
    if str(path) in meta.get_all('License-File')
]
```

## Open questions
**1. What should the value of `License-Path` be for files not in main project directory?**
From a technical standpoint it would probably be best to use the actual relative path and mirror that in the `.dist-info` directory. However, at the moment `wheel` flattens the folder structure and just puts all license files into `.dist-info`, AFAIK. This could lead to name conflicts.

**2. Should license files be copied into the `.egg-info` directory if the legacy install is used?**

**3. How should `license_files` and `MANIFEST.in` continue to work together?**
At the moment, the `MANIFEST.in` is parse at the end and thus overwrites all previous file includes. However, `license_files_computed` will only look at the `license_file` and `license_files` options (unless we parse the manifest there too). That means that the `License-File` field might contain files that are later excluded by the manifest.
-> I would recommend to change the behavior so that the `license_file` and `license_files` options take precedent over the manifest. A license file is most likely excluded by accident anyway. Additionally, specifying `license_files =` (without any values) won't add any default patterns and thus not match any license files.

## Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -4,9 +4,6 @@ import os
 import sys
 import io
 import contextlib
-from glob import iglob
-
-from setuptools.extern import ordered_set
 
 from .py36compat import sdist_add_defaults
 
@@ -190,46 +187,3 @@ class sdist(sdist_add_defaults, orig.sdist):
                 continue
             self.filelist.append(line)
         manifest.close()
-
-    def check_license(self):
-        """Checks if license_file' or 'license_files' is configured and adds any
-        valid paths to 'self.filelist'.
-        """
-        opts = self.distribution.get_option_dict('metadata')
-
-        files = ordered_set.OrderedSet()
-        try:
-            license_files = self.distribution.metadata.license_files
-        except TypeError:
-            log.warn("warning: 'license_files' option is malformed")
-            license_files = ordered_set.OrderedSet()
-        patterns = license_files if isinstance(license_files, ordered_set.OrderedSet) \
-            else ordered_set.OrderedSet(license_files)
-
-        if 'license_file' in opts:
-            log.warn(
-                "warning: the 'license_file' option is deprecated, "
-                "use 'license_files' instead")
-            patterns.append(opts['license_file'][1])
-
-        if 'license_file' not in opts and 'license_files' not in opts:
-            # Default patterns match the ones wheel uses
-            # See https://wheel.readthedocs.io/en/stable/user_guide.html
-            # -> 'Including license files in the generated wheel file'
-            patterns = ('LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*')
-
-        for pattern in patterns:
-            for path in iglob(pattern):
-                if path.endswith('~'):
-                    log.debug(
-                        "ignoring license file '%s' as it looks like a backup",
-                        path)
-                    continue
-
-                if path not in files and os.path.isfile(path):
-                    log.info(
-                        "adding license file '%s' (matched pattern '%s')",
-                        path, pattern)
-                    files.add(path)
-
-        self.filelist.extend(sorted(files))
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -142,7 +142,7 @@ def read_pkg_file(self, file):
         self.provides = None
         self.obsoletes = None
 
-    self.license_files_computed = _read_list_from_msg(msg, 'license-file')
+    self.license_files = _read_list_from_msg(msg, 'license-file')
 
 
 def single_line(val):
@@ -216,7 +216,7 @@ def write_pkg_file(self, file):  # noqa: C901  # is too complex (14)  # FIXME
         for extra in self.provides_extras:
             write_field('Provides-Extra', extra)
 
-    self._write_list(file, 'License-File', self.license_files_computed)
+    self._write_list(file, 'License-File', self.license_files or [])
 
     file.write("\n%s\n\n" % self.get_long_description())
 
@@ -421,7 +421,6 @@ class Distribution(_Distribution):
         'provides_extras': ordered_set.OrderedSet,
         'license_file': None,
         'license_files': None,
-        'license_files_computed': list,
     }
 
     _patched_dist = None
@@ -603,7 +602,7 @@ class Distribution(_Distribution):
                 if path not in files and os.path.isfile(path):
                     files.add(path)
 
-        self.metadata.license_files_computed = sorted(files)
+        self.metadata.license_files = sorted(files)
 
     # FIXME: 'Distribution._parse_config_files' is too complex (14)
     def _parse_config_files(self, filenames=None):  # noqa: C901
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -419,7 +419,6 @@ class Distribution(_Distribution):
         'long_description_content_type': lambda: None,
         'project_urls': dict,
         'provides_extras': ordered_set.OrderedSet,
-        'license_file': lambda: None,
         'license_files': lambda: None,
     }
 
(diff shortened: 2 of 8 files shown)


setuptools/command/sdist.py
```diff
@@ -190,46 +187,3 @@ def read_manifest(self):
...
-            else ordered_set.OrderedSet(license_files)
-
-        if 'license_file' in opts:
-            log.warn(
-                "warning: the 'license_file' option is deprecated, "
-                "use 'license_files' instead")
-            patterns.append(opts['license_file'][1])
-
-        if 'license_file' not in opts and 'license_files' not in opts:
-            # Default patterns match the ones wheel uses
```
jaraco, line 216: What happens to these default patterns?

setuptools/dist.py
```diff
@@ -406,7 +411,9 @@ class Distribution(_Distribution):
         'long_description_content_type': None,
         'project_urls': dict,
         'provides_extras': ordered_set.OrderedSet,
-        'license_files': ordered_set.OrderedSet,
+        'license_file': None,
+        'license_files': None,
+        'license_files_computed': list,
```
jaraco, line 416: Could it be possible to avoid this diff by using `license_files` where `license_files_computed` is found (and thus not introduce a new name)?

setuptools/command/sdist.py
jaraco: Nevermind - I see it now in _finalize_license_files.

setuptools/dist.py
```diff
@@ -565,6 +572,31 @@ def _clean_req(self, req):
...
+            patterns = ('LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*')
+
+        for pattern in patterns:
+            for path in iglob(pattern):
+                if path.endswith('~'):
+                    continue
+                if path not in files and os.path.isfile(path):
+                    files.add(path)
+
+        self.metadata.license_files_computed = sorted(files)
```
jaraco, line 598: Can we retain order of license files rather than enforcing order? For the manifest, that's less important, but for metadata, it may prove useful to honor the user's supplied order.

jaraco (approved):
I like where this is going. Just a couple of concerns.

setuptools/dist.py
cdce8p: The user only provides patterns which are used for `glob`. As such, I don't think retaining any order would make much sense.

cdce8p:
I've updated the PR. Additionally, I also added a commit to change when license files are added to the source list. As explained in the open questions section (`3. How should license_files and MANIFEST.in continue to work together?`), to make sure a license file which is listed in `License-File` is included, the `license_files` option needs to be able to overwrite the `MANIFEST.in`

setuptools/dist.py
jaraco: Alright. I'll probably re-write it to avoid sorting or at least avoid sorting user-supplied values.

setuptools/dist.py
```diff
@@ -406,7 +411,8 @@ class Distribution(_Distribution):
         'long_description_content_type': None,
         'project_urls': dict,
         'provides_extras': ordered_set.OrderedSet,
-        'license_files': ordered_set.OrderedSet,
+        'license_file': None,
```
jaraco, line 414: I had kind-of hoped this would go away, especially since the value is deprecated. I tried removing it and got an error, but I think it can be eliminated by being more defensive at the one point where it's queried.

setuptools/dist.py
```diff
@@ -406,7 +411,8 @@ class Distribution(_Distribution):
...
+        'license_files': None,
```
jaraco, line 415: Here it probably makes sense to just declare the default here rather than declare a null value and then set a default if the value is a null value.

jaraco:
I added a couple more comments, but I'll take the first stab at adjusting the behavior.

setuptools/dist.py
cdce8p: That was my initial idea as well. However, with that we can't determined if the user didn't specify the option (thus the defaults would be loaded) or left it empty (in which case on license files would be included). Both would equal an empty list.

setuptools/dist.py
cdce8p: I'm happy to remove it if you want. Personally, I would also be open to remove the option `license_file` entirely and return an error. Should be easy enough for the user to fix then.

setuptools/dist.py
jaraco: Oh, that's very interesting. And the test cases capture this aspect. Consider me impressed.

cdce8p:
@jaraco Pushed new change to address some of your points. First, I modified the logic to match license files so that the user sorting is kept as best as possible. Second, I removed the default `license_file` metadata attribute. Let me know if you what me to change anything else.

cdce8p:
**Small update**: I needed to revert the removal of the `license_file` attribute. Without it the deprecation warning wouldn't be printed (even with `-Wd`).

jaraco:
I spent some time trying to simplify the logic by having the defaults for "license_files" be defined in the defaults for license files (rather than having a null value as the default and defining defaults in the finalizer), but when doing so, I encountered two problems:

- The config parser [won't override a default with an explicit value if `bool(default) == True`](https://github.com/pypa/setuptools/blob/c6a761fd4dd86283f25d828c7b9b56d764dcf84b/setuptools/config.py#L212-L214).
- The logic for the default/fallback values is contingent both on `license_file` and `license_files` not being specified, so until `license_file` can be removed, the two settings are inherently entangled.

At some point soon, I'd like to remove support for license_file and remove that check for overriding a non-False default. For now this should do.

I've added one additional commit that removes the sorting entirely and uses a generator to build the list. I see I don't have push access to the PR, so I'll try sending a PR to your branch. If that doesn't work, I'll probably just merge it all together manually.

cdce8p:
> I've added one additional commit that removes the sorting entirely and uses a generator to build the list. I see I don't have push access to the PR, so I'll try sending a PR to your branch. If that doesn't work, I'll probably just merge it all together manually.

Strange. I've allowed edits / commits by maintainers 🤔 Anyway, I merged your changes and, while I was add it, I rebased the PR to resolve the merge conflict.

jaraco:
> Strange. I've allowed edits / commits by maintainers 🤔 Anyway, I merged your changes and, while I was add it, rebased the PR to resolve the merge conflict.

Seems the mistake is on my side, an artifact of having switched from `hub` to `gh` and `gh` not handling pr checkouts as elegantly (https://github.com/cli/cli/issues/2189).

hroncok:
I think I might have found a problem with this, see https://github.com/pypa/setuptools/issues/2739

PR #2653, opened 2021-04-24
Distutils refresh

jaraco:
Sync with pypa/distutils@7423f0767a.


jaraco:
I'm confused. I only expected this PR to include a handful of these commits. I need to figure out why the whole history of distutils is appearing in the PR.

jaraco:
After merging #2657, the commit log is much shorter, just including the recent changes.

PR #2658, opened 2021-04-25
Add ability to wrap project_url labels with quotation marks

cdce8p:
## Summary of changes
To specify `project_urls` inside the `setup.cfg` file, a developer has to use a dictionary style.
```ini
project_urls =
    Bug tracker = https://github.com/...
```
This works and produces the desired output. However, many syntax highlight don't recognize this correctly (including Github). A reasonable next step would be to use quotation marks for the project_url label, like so
```ini
project_urls =
    "Bug tracker" = https://github.com/...
```
The issue this PR addresses is, that in the last case the quotation mark would be included in the final metadata and subsequently also be present on pypi.
```
Project-URL: "Bug tracker", https://github.com/...
```

**Links**

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2658.change.rst
+++ b/changelog.d/2658.change.rst
@@ -1,0 +1,1 @@
+Added ability wrap project_url labels with quotation marks. - by :user:`cdce8p`
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -269,7 +269,7 @@ class ConfigHandler:
             if sep != separator:
                 raise DistutilsOptionError(
                     'Unable to parse option value to dict: %s' % value)
-            result[key.strip()] = val.strip()
+            result[key.strip().strip('\"\'')] = val.strip()
 
         return result
 
--- a/setuptools/tests/test_egg_info.py
+++ b/setuptools/tests/test_egg_info.py
@@ -855,7 +855,43 @@ class TestEggInfo:
         assert expected_line in pkg_info_lines
         assert 'Metadata-Version: 2.1' in pkg_info_lines
 
-    def test_project_urls(self, tmpdir_cwd, env):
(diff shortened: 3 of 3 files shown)


jaraco:
I'm slightly averse to merging this change for these reasons:

- It reinforces an anti-pattern of wrapping space-separated phrases or names with quotes when those quotes are in fact unnecessary.
- Other keys in setup.cfg aren't quoted. Users may be tempted to use `"name" = project` or `"name" = "project"` or to put quotes around entry points as well.
- It provides a crutch for syntax formatters to continue not to support simple, unquoted names.
- The format itself supports the simple, unquoted names. Other consumers of this file would not be expected to parse these select values in this special way.
- The solution allows quotes to be applied improperly, such as in `"'Bug tracker'''"' = https://github.com/...`.

Since the file syntax itself supports spaces in the names, it seems to me that the formatters should fix the bug that they're not formatting unquoted keys properly. Have you explored that route?

webknjaz (changes requested):
I don't think that adding crutches to please the syntax highlighters that parse files improperly is a valid motivation for such a change. If anything, that software should fix their lexers.

Also, I don't believe setuptools is the right place for implementing a special setuptools-flavored `.ini` syntax. If this is a truly desirable change, it should be proposed to the configparser itself, in stdlib.

cdce8p:
Ok, you convinced me. I didn't know that spaces are valid inside keywords.
As for the highlighting issue, I've opened https://github.com/textmate/ini.tmbundle/pull/8 to improve the regex. The repo is used by at least VS Code and Github for syntax highlighting of configuration files. Let's see if it get accepted.

PR #2663, opened 2021-05-03
Added Visual Studio Express 2017 support

dofuuz:
## Summary of changes

Added Visual Studio Express 2017 detection to `setuptools/msvc.py`.

### Some details

Setuptools cannot detect Visual Studio Express 2017 automatically for now.

It's detecting Visual Studio using this command:
`vswhere -latest -prerelease -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath -products *`

This prints nothing because component ID `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` is missing in VS Express 2017
https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-express?view=vs-2017

So, it should include VS Express explicitly using this command:
`vswhere -latest -prerelease -requiresAny -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -requires Microsoft.VisualStudio.Workload.WDExpress -property installationPath -products *`

Or, detect ARM compiler using component ID `Microsoft.VisualStudio.Component.VC.Tools.ARM`, and `Microsoft.VisualStudio.Component.VC.Tools.ARM64`.

Detecting ARM compiler(eb27045) might be more good solution because setuptools supports ARM. But there are some corner case to deal with. (It seems like ARM compiler can be installed without buildtools.)

So, I just included VS Express workload ID.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


abravalheri:
Hi @dofuuz, we recently received a bug report (https://github.com/pypa/setuptools/issues/3946), which seems to indicate that the `-requiresAny` argument in this PR is not compatible with older versions of Visual Studio 2017.

After having a look on the release notes, it seems that `-requiresAny` was added in [v2.3.2](https://github.com/microsoft/vswhere/releases/tag/2.3.2), released on 23 Jan 2018.

By correlating the date with the [Visual Studio 2017 releases](https://learn.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes-history), we can say that the earliest possible Visual Studio to include this (or a later) version of `vswhere` (supporting `-requiresAny`) is version 15.5.5. But it is more likely that the update was made in version 15.6.

This means that setuptools currently is not compatible with Visual Studio 2017 with versions prior to 15.6.

Do you agree with this assessment? (or maybe you have a different explanation?)

Do you have any suggestion on how to tackle this incompatibility?

dofuuz:
~~Solution: Just update VS2017 to recent version~~ (JK)

We can use workaround something like this:
```python
...
    if not root:
        return None, None

    for require in ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "Microsoft.VisualStudio.Workload.WDExpress"):
        try:
            path = subprocess.check_output([
                join(root, "Microsoft Visual Studio", "Installer", "vswhere.exe"),
                "-latest",
                "-prerelease",
                "-requires", require,
                "-property", "installationPath",
                "-products", "*",
            ]).decode(encoding="mbcs", errors="strict").strip()
        except (subprocess.CalledProcessError, OSError, UnicodeDecodeError):
            return None, None

        path = join(path, "VC", "Auxiliary", "Build")
        if isdir(path):
            return 15, path

    return None, None
```

dofuuz:
@abravalheri 

~~Solution: Just update VS2017 to recent version~~ (JK)

We can use workaround something like this:
```python
...
    if not root:
        return None, None

    for require in ("Microsoft.VisualStudio.Component.VC.Tools.x86.x64", "Microsoft.VisualStudio.Workload.WDExpress"):
        try:
            path = subprocess.check_output([
                join(root, "Microsoft Visual Studio", "Installer", "vswhere.exe"),
                "-latest",
                "-prerelease",
                "-requires", require,
                "-property", "installationPath",
                "-products", "*",
            ]).decode(encoding="mbcs", errors="strict").strip()
        except (subprocess.CalledProcessError, OSError, UnicodeDecodeError):
            return None, None

        path = join(path, "VC", "Auxiliary", "Build")
        if isdir(path):
            return 15, path

    return None, None
```

PR #2665, opened 2021-05-05
CHANGES.rst: Fix version

hugovk:
## Summary of changes

Fix double version in changelog.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
The change that introduced this error was 529419e71f496fcf6274ebdb4af96905314b5e7f, which was introduced by running `tox -e finalize`. It seems towncrier is no longer able to properly finalize the release, so while this change will address the symptom, the root cause, probably introduced in #2654, will still need to be addressed.

PR #2667, opened 2021-05-08
Fixed a couple of cases where generated shebangs are invalid

### LINKED ISSUE #1744: Editable/develop install with spaces in path results in unusable #!
As I dig deeper I get a bit more confused but it seems that `pip install --editable` installs are not creating sensible `#!`s for POSIX when there are spaces in the path.

`--editable`
```
#!"/home/altendky/very_sneaky/venv with spaces/bin/python3"
```

regular
```
#!/bin/sh
'''exec' "/home/altendky/very_sneaky/venv with spaces/bin/python3" "$0" "$@"
' '''
```

The undesired results seem to come from:
https://github.com/pypa/setuptools/blob/7be2d14dcd669d6010621a8e5db23763a63cf55e/setuptools/command/easy_install.py#L2047

From what I can tell the regular result is built by `distlib` (vendored in `pip`).

https://bitbucket.org/pypa/distlib/src/3b0fd333c8fb15bc04e570a23ee4836caefb7951/distlib/scripts.py#lines-139:168

```python
    def _build_shebang(self, executable, post_interp):
        """
        Build a shebang line. In the simple case (on Windows, or a shebang line
        which is not too long or contains spaces) use a simple formulation for
        the shebang. Otherwise, use /bin/sh as the executable, with a contrived
        shebang which allows the script to run either under Python or sh, using
        suitable quoting. Thanks to Harald Nordgren for his input.

        See also: http://www.in-ulm.de/~mascheck/various/shebang/#length
                  https://hg.mozilla.org/mozilla-central/file/tip/mach
        """
        if os.name != 'posix':
            simple_shebang = True
        else:
            # Add 3 for '#!' prefix and newline suffix.
            shebang_length = len(executable) + len(post_interp) + 3
            if sys.platform == 'darwin':
                max_shebang_length = 512
            else:
                max_shebang_length = 127
            simple_shebang = ((b' ' not in executable) and
                              (shebang_length <= max_shebang_length))

        if simple_shebang:
            result = b'#!' + executable + post_interp + b'\n'
        else:
            result = b'#!/bin/sh\n'
            result += b"'''exec' " + executable + post_interp + b' "$0" "$@"\n'
            result += b"' '''"
        return result
```

The details are provided from Linux.  The original issue was reported to me from macOS.

<details><summary>Recreation Details</summary>

```
 ~/very_sneaky   master ●  virtualenv -p python3 venv\ with\ spaces
Running virtualenv with interpreter /home/altendky/.pyenv/shims/python3
Using base prefix '/home/altendky/.pyenv/versions/3.7.2'
New python executable in /home/altendky/very_sneaky/venv with spaces/bin/python3
Also creating executable in /home/altendky/very_sneaky/venv with spaces/bin/python
Installing setuptools, pip, wheel...
done.
```

```
 ✘  ~/very_sneaky   master ●  git clone https://github.com/ambv/black
Cloning into 'black'...
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 2494 (delta 0), reused 1 (delta 0), pack-reused 2489
Receiving objects: 100% (2494/2494), 3.00 MiB | 3.93 MiB/s, done.
Resolving deltas: 100% (1690/1690), done.
```

```
 ~/very_sneaky   master ●  venv\ with\ spaces/bin/pip install -e black
Obtaining file:///home/altendky/very_sneaky/black
    Preparing wheel metadata ... done
Collecting appdirs (from black==19.3b0)
  Using cached https://files.pythonhosted.org/packages/56/eb/810e700ed1349edde4cbdc1b2a21e28cdf115f9faf263f6bbf8447c1abf3/appdirs-1.4.3-py2.py3-none-any.whl
Collecting toml>=0.9.4 (from black==19.3b0)
  Using cached https://files.pythonhosted.org/packages/a2/12/ced7105d2de62fa7c8fb5fce92cc4ce66b57c95fb875e9318dba7f8c5db0/toml-0.10.0-py2.py3-none-any.whl
Collecting attrs>=18.1.0 (from black==19.3b0)
  Using cached https://files.pythonhosted.org/packages/23/96/d828354fa2dbdf216eaa7b7de0db692f12c234f7ef888cc14980ef40d1d2/attrs-19.1.0-py2.py3-none-any.whl
Collecting click>=6.5 (from black==19.3b0)
  Using cached https://files.pythonhosted.org/packages/fa/37/45185cb5abbc30d7257104c434fe0b07e5a195a6847506c074527aa599ec/Click-7.0-py2.py3-none-any.whl
Installing collected packages: appdirs, toml, attrs, click, black
  Running setup.py develop for black
Successfully installed appdirs-1.4.3 attrs-19.1.0 black click-7.0 toml-0.10.0
```

```
 ~/very_sneaky   master ●  venv\ with\ spaces/bin/black --help
Failed to execute process 'venv with spaces/bin/black'. Reason:
The file 'venv with spaces/bin/black' does not exist or could not be executed.
```

```
 ✘  ~/very_sneaky   master ●  head venv\ with\ spaces/bin/black
#!"/home/altendky/very_sneaky/venv with spaces/bin/python3"
# EASY-INSTALL-ENTRY-SCRIPT: 'black','console_scripts','black'
__requires__ = 'black'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
```

```

 ✘  ~/very_sneaky   master ●  venv\ with\ spaces/bin/pip freeze --all
appdirs==1.4.3
attrs==19.1.0
-e git+https://github.com/ambv/black@cea13f498418784e22f8fbd78db3f9240a2bad11#egg=black
Click==7.0
pip==19.0.3
setuptools==41.0.0
toml==0.10.0
wheel==0.33.1
```

```
 ~/very_sneaky   master ●  cat venv\ with\ spaces/bin/pip
#!/bin/sh
'''exec' "/home/altendky/very_sneaky/venv with spaces/bin/python3" "$0" "$@"
' '''
# -*- coding: utf-8 -*-
import re
import sys

from pip._internal import main

if __name__ == '__main__':
    sys.exit(main())
```

</details>

AWhetter:
## Summary of changes

This change fixes two different cases where the shebang line is invalid:
* The length of the path to the Python executable is above the system's limit for the length of the shebang.
* The path to the Python executable contains spaces. Currently these produce a shebang with a quoted path, which is invalid.

In both of these cases, a script is generated that will execute using /bin/sh but is still valid Python. This is similar behaviour to distlib and therefore also to pip when a regular install is being performed.

Closes #1744

### Pull Request Checklist
- [x] Changes have tests


setuptools/tests/test_easy_install.py
```diff
@@ -1011,7 +1011,22 @@ def test_get_script_header_exe_with_spaces(self):
         actual = ei.ScriptWriter.get_header(
             '#!/usr/bin/python',
             executable='"' + self.exe_with_spaces + '"')
-        expected = '#!"%s"\n' % self.exe_with_spaces
```
AWhetter, line 1014: This test has been changed because the shebang line that is generated is invalid.

setuptools/tests/test_easy_install.py
```diff
@@ -1038,7 +1053,9 @@ def test_from_environment_with_spaces_in_executable(self):
         os.environ.pop('__PYVENV_LAUNCHER__', None)
         cmd = ei.CommandSpec.from_environment()
         assert len(cmd) == 1
-        assert cmd.as_header().startswith('#!"')
```
AWhetter, line 1041: Similarly here, this test has been changed because the shebang line that is generated is invalid.

jaraco:
> This is similar behaviour to distlib.

Would it be possible to just adopt distlib?

AWhetter:
I did consider it but the bit we need is a method on a class that's used to write out the files itself (https://bitbucket.org/pypa/distlib/src/4f54a3117db58f6bb00c69ab9a892170b91acd3b/distlib/scripts.py#lines-127:156). The method could be static so we could hack around it and do something like `distlib.scripts.ScriptMaker._build_shebang(None, cmdline.encode("utf-8"), b"")`, or we could construct a `ScriptMaker` even though we don't need it? Alternatively we could adopt distlib more formally and make it responsible for writing out scripts, but I think that'll be tough to do whilst maintaining the current API of CommandSpec.

jaraco:
> Alternatively we could adopt distlib more formally and make it responsible for writing out scripts, but I think that'll be tough to do whilst maintaining the current API of CommandSpec.

Adopting distlib would fix lots of broken behaviors in this project. I realize it's a substantially larger scope than patching over this one undesirable behavior, but I'm loathe to duplicate even more behavior. I haven't looked at CommandSpec in a while, but IIRC, that is an internal interface, so it may be relatively straightforward to simply replace all of the script writing with the distlib implementation. Is that something you'd be willing to explore?

AWhetter:
Absolutely! I'll give it a go and see what I can come up with.

AWhetter:
Is there a formal definition of what's public on the easy_install class and what's not?

I'm currently looking at changing (https://github.com/pypa/setuptools/blob/317ad2c4c9edf793e65372e80ec2d64f4edf147f/setuptools/command/easy_install.py#L790) but it would no longer call `self.write-script()`. I also need to change `self.install_script()` (https://github.com/pypa/setuptools/blob/317ad2c4c9edf793e65372e80ec2d64f4edf147f/setuptools/command/easy_install.py#L796) and it would also not call `self.write_script()`. Ideally `install_script()` wouldn't get called at all because distlib cannot take arbitrary script text. It only accepts a file path or entry point definition, which means that we have to write out the script text to a temporary file if the method must continue to exist.
If it's easier I can submit what I have as a draft so that you have a more concrete view on what I'm describing.

AWhetter:
I've gone ahead and pushed what I have so far as a work in progress

jaraco:
> Is there a formal definition of what's public on the easy_install class and what's not?

I don't believe there is, but in general, the public API for distutils command classes (such as `easy_install`) is just its command interface. I'd say it's safe to assume that most of the interfaces are private to Setuptools.

> It only accepts a file path or entry point definition, which means that we have to write out the script text to a temporary file if the method must continue to exist.

That's unfortunate. What I'd recommend in that case is two-fold:

- First, do as you've planned and write the text out to a temporary file to work with distlib as it is.
- Second, file a feature request issue with distlib to expose the necessary functionality to support this use-case.

PR #2677, opened 2021-05-19
Publish draft change notes

jaraco:
A reprise of #2650 with a fix for publishing clean release notes when no changes are pending.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,3 +1,6 @@
+from pathlib import Path
+
+
 extensions = ['sphinx.ext.autodoc', 'jaraco.packaging.sphinx', 'rst.linker']
 
 master_doc = "index"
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -143,3 +143,10 @@
     python=('https://docs.python.org/3', None),
     python2=('https://docs.python.org/2', None),
 )
+
+# Add support for the unreleased "next-version" change notes
+extensions += ['sphinxcontrib.towncrier']
+# Extension needs a path from here to the towncrier config.
```
webknjaz, line 149: This comment is slightly inaccurate. The path is not "from here" but from the current working directory that is set on the process level. In the case of this project, it's set in `tox.ini`.
At the moment, this value is passed through to the subprocess invocation IIRC but I'll probably try to make it more intuitive in the future by calculating the path relative to `conf.py` (if that's even possible in Sphinx's runtime — I haven't checked).

jaraco: Indeed. I'd probably make it relative to some stable path in the sphinx config.

Issue #2691, opened 2021-05-31
[BUG] Generated PKG_INFO file is not reproducible


rpurdie:
### setuptools version

57.0.0

### Python version

3.9

### OS

Linux

### Additional environment information

We're seeing this in Yocto Project builds (cross compiling for many different architectures/targets on multiple host distros)

### Description

We're seeing an issue where the PKG-INFO file generated depends on the order of the files on the disk as returned by os.listdir(). This probably isn't intended. This involves the License-File lines, for example for python-packaging, one build shows:

License-File: LICENSE 
License-File: LICENSE.APACHE
License-File: LICENSE.BSD

and the other shows:

License-File: LICENSE 
License-File: LICENSE.BSD
License-File: LICENSE.APACHE

where the last two entries are reversed. The output from our CI showing the issue is here:

https://autobuilder.yocto.io/pub/repro-fail/oe-reproducible-20210531-niv755r2/packages/diff-html/

### Expected behavior

The files would have deterministic output regardless of the order of the files on disk.

### How to Reproduce

Changing _expand_patterns() in setuptools/dist.py:

Index: setuptools-57.0.0/setuptools/dist.py
===================================================================
--- setuptools-57.0.0.orig/setuptools/dist.py
+++ setuptools-57.0.0/setuptools/dist.py
@@ -15,7 +15,7 @@ import distutils.command
 from distutils.util import strtobool
 from distutils.debug import DEBUG
 from distutils.fancy_getopt import translate_longopt
-from glob import iglob
+from glob import glob
 import itertools
 import textwrap
 from typing import List, Optional, TYPE_CHECKING
@@ -603,7 +603,7 @@ class Distribution(_Distribution):
         return (
             path
             for pattern in patterns
-            for path in iglob(pattern)
+            for path in sorted(glob(pattern), reverse=True)
             if not path.endswith('~')
             and os.path.isfile(path)
         )

and looking at PKG-INFO in python-packaging, then removing the reverse=True will show the output depends on the order of the files on disk (from glob which uses listdir).

### Output

See above, output of the License-File headers changes

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

CAM-Gerlach:
FYI, looks like this issue was solved by PR #2692 , but that PR wasn't closed-tagged so this issue didn't close.

Also, thanks for bringing this up; I'll mention this in the PEP 639 spec.

PR #2692, opened 2021-05-31
setuptools/dist: Fix reproducibility issue by sorting globbing

rpurdie:
globbing uses os.listdir() which returns order on disk which is unsorted.
Ultimately this leads to random ordering of License-File entries in PKG-INFO
and means the resulting output isn't deterministic.

Adding a sort fixes that. Switch to glob instead ot iglob since sorted()
removes the benefit of an iterator.

https://github.com/pypa/setuptools/issues/2691

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>

## Summary of changes

Closes 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -15,7 +15,7 @@ import distutils.command
 from distutils.util import strtobool
 from distutils.debug import DEBUG
 from distutils.fancy_getopt import translate_longopt
-from glob import iglob
+from glob import glob
 import itertools
 import textwrap
 from typing import List, Optional, TYPE_CHECKING
@@ -603,7 +603,7 @@ class Distribution(_Distribution):
         return (
             path
             for pattern in patterns
-            for path in iglob(pattern)
+            for path in sorted(glob(pattern))
             if not path.endswith('~')
             and os.path.isfile(path)
         )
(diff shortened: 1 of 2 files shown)


jaraco:
> Switch to glob instead ot iglob since sorted()
> removes the benefit of an iterator.

I think I'd still prefer an iterator.

Can you create a test to capture this expectation?

webknjaz (commented):
Could you add a change not file following this instruction? https://setuptools.readthedocs.io/en/latest/development/developer-guide.html#adding-change-notes-with-your-prs

setuptools/dist.py
```diff
@@ -603,7 +603,7 @@ def _expand_patterns(patterns):
         return (
```
webknjaz, line 603: If you want reproducibility, why not sort the filtered result? Of course, it'll turn a generator expression into a list or something...

jaraco: I prefer this minimal approach that retains the order of each pattern. If you were to sort the entire result, you would lose fidelity about the order in which the patterns were supplied. Thus if the user specifies LICENSE_B, LICENSE_A as separate globs, they'll get them in that order.

PR #2693, opened 2021-06-01
Adding codecov configuration

### LINKED ISSUE #2642: Resurrect reporting to codecov
### setuptools version

`main`

### Python version

N/A

### OS

N/A

_No response_

### Description

I've been doing some security audit after the recent codecov supply chain attack and discovered that there's been no coverage updates on the service since Dec 20.

I tracked it down to this change: https://github.com/pypa/setuptools/pull/2486.

Funny enough, I've left a number of comments reviewing that but still missed this issue ;)

It used to be integrated in Travis CI and AppVeyor
- https://github.com/pypa/setuptools/pull/2486/files?file-filters%5B%5D=.ini&file-filters%5B%5D=.toml&file-filters%5B%5D=.yml#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L60
- https://github.com/pypa/setuptools/pull/2486/files?file-filters%5B%5D=.ini&file-filters%5B%5D=.toml&file-filters%5B%5D=.yml#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dcL43

and the toxenv is still in place but its invocation never survives the migration to GHA and so this needs to be corrected.

Fixing this should be pretty straightforward — just integrate https://github.com/codecov/codecov-action.

### Expected behavior

Coverage to be actually reported to codecov.

### How to Reproduce

Look at https://app.codecov.io/gh/pypa/setuptools/commits — the last known report was 4 month ago.

### Output

N/A

### Code of Conduct

tanvimoharir:
## Summary of changes

Added minimal configuration for codecov.

Closes #2642 

### Pull Request Checklist
- [x] Changes have tests (N/A)
- [x] News fragment added in [`changelog.d/`]. (N/A?)

- [x]  I have hereby reviewed ,and agree to follow, the PSF Code of Conduct

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,11 @@ jobs:
       - name: Run tests
         run: tox
       - name: Upload coverage to Codecov
-        run: tox -e coverage,codecov
+        uses: codecov/codecov-action@v1
+        with:
+          functionalities: gcov
+          files: ./coverage.xml
+          verbose: true # optional (default = false)
 
   release:
     needs: test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,8 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: Install coveragepy tool
+        run: pip install coverage
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,10 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
+          flags: >-
+            GHA,
+            ${{ runner.os }},
+            ${{ matrix.python }}
           functionalities: gcov
           verbose: true # optional (default = false)
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,6 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           flags: >-  # Mark which lines are covered by which envs
-            GHA,
             ${{ runner.os }},
             ${{ matrix.python }}
           functionalities: gcov  # disable C coverage autodiscovery
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,10 +27,6 @@ jobs:
             GHA,
             ${{ runner.os }},
             ${{ matrix.python }}
-          flags: >-
-            GHA,
-            ${{ runner.os }},
-            ${{ matrix.python }}
           functionalities: gcov
           verbose: true # optional (default = false)
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,13 +22,6 @@ jobs:
         run: tox
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
-        with:
-          flags: >-
-            GHA,
-            ${{ runner.os }},
-            ${{ matrix.python }}
-          functionalities: gcov
-          verbose: true # optional (default = false)
 
   release:
     needs: test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,6 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           functionalities: gcov
-          files: ./coverage.xml
           verbose: true # optional (default = false)
 
   release:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,10 +20,11 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: After tests
+        run: tox -e coverage,codecov
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
 
   release:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,11 @@ jobs:
         run: pip install coverage
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
+        with:
+          flags: >-
+            GHA,
+            ${{ runner.os }},
+            ${{ matrix.python }}
 
   release:
     needs: test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,11 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: true
 
   release:
     needs: test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,12 +20,8 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
-      - name: After tests
-        run: tox -e coverage,codecov
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
-        with:
-          fail_ci_if_error: true
+        run: tox -e coverage,codecov
 
   release:
     needs: test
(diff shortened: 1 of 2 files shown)


Ely68 (approved):

Ely68 (approved):

.github/workflows/main.yml
```diff
@@ -20,6 +20,8 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: Upload coverage to Codecov
+        run: tox -e coverage,codecov
```
webknjaz, line 24: Could you use https://github.com/codecov/codecov-action instead?

tanvimoharir: Sure, thanks.

.github/workflows/main.yml
```diff
@@ -20,6 +20,11 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          functionalities: gcov
```
webknjaz, line 26: How about adding some markers?
```suggestion
          flags: >-
            GHA,
            ${{ runner.os }},
            ${{ matrix.python }}
          functionalities: gcov
```

webknjaz:
We'll need to wait for @jaraco to allow the CI run, I don't have access to that button. But looks good otherwise.

.github/workflows/main.yml
tanvimoharir: Sure,I will add these as well.

webknjaz: FYI it's best to accept changes via GH UI because this way it properly marks discussion threads as resolved and also credits the suggester in the commit message as a co-author.

tanvimoharir: Thanks, I wasn't aware of this.

jaraco: I see `functionalities: gcov` disables gcov. What is gcov and why does it need to be disabled?

.github/workflows/main.yml
```diff
@@ -20,6 +20,15 @@ jobs:
...
         run: tox
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          flags: >-
+            GHA,
+            ${{ runner.os }},
+            ${{ matrix.python }}
+          functionalities: gcov
+          verbose: true # optional (default = false)
```
jaraco, line 31: My preference would be to start without any options unless the project specifically benefits from the options. Why enable verbose?

jaraco (approved):
This change looks acceptable. I don't yet see any results in the runners. I'll probably remove all of the options that aren't necessary and trust that the defaults are good enough.

jaraco:
Looks like codecov isn't finding any coverage reports. See if you can find a way to retain those artifacts for upload, preferably without altering any existing lines.

jaraco (changes requested):
Looks like codecov isn't finding any coverage reports. See if you can find a way to retain those artifacts for upload, preferably without altering any existing lines.

.github/workflows/main.yml
webknjaz: My understanding is that gcov is a coverage format for code written in C. By default, codecov attempts scanning the disk for various formats because it doesn't know what technologies are present, I suppose this disables probing some formats.

https://en.wikipedia.org/wiki/Gcov

webknjaz:
GHA logs have this:
> ```console
> ==> Python coveragepy not found
> ```

:thinking: Does it expect that `pip install coverage` is executed beforehand? It shouldn't need this.

webknjaz:
> I'll probably remove all of the options that aren't necessary

Flags are necessary because these are useful in the report view showing what envs hit certain lines. So plz recover that option. And disabling gcov sounds right too. You can also specify `files:` to point exactly to the file produced by pytest-cov.

webknjaz:
>  Does it expect that `pip install coverage` is executed beforehand? It shouldn't need this.

Looks like it's exactly what it expects: the README example shows installing+running pytest-cov in a bare env (w/o a venv created by tox).

.github/workflows/main.yml
```diff
@@ -20,6 +20,8 @@ jobs:
           python -m pip install tox
       - name: Run tests
         run: tox
+      - name: Upload coverage to Codecov
```
webknjaz, line 23: If my guess is correct, this will fix the issue.
```suggestion
      - name: Install coveragepy tool
        run: pip install coverage
      - name: Upload coverage to Codecov
```

tanvimoharir:
> > I'll probably remove all of the options that aren't necessary
> 
> Flags are necessary because these are useful in the report view showing what envs hit certain lines. So plz recover that option. And disabling gcov sounds right too. You can also specify `files:` to point exactly to the file produced by pytest-cov.

Sure, I will add it.

webknjaz:
@tanvimoharir could you recover the action options under `with:` per my earlier comments?

@jaraco you'll need to re-approve the CI run for us to see if the patch works.

tanvimoharir:
> @tanvimoharir could you recover the action options under `with:` per my earlier comments?
> 
> @jaraco you'll need to re-approve the CI run for us to see if the patch works.

Thanks, I have added back the actions options per your earlier comment.

.github/workflows/main.yml
```diff
@@ -20,6 +20,15 @@ jobs:
...
         run: tox
+      - name: Install coveragepy tool
+        run: pip install coverage
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          flags: >-
+            GHA,
+            ${{ runner.os }},
+            ${{ matrix.python }}
```
webknjaz, line 31: Looks like I was wrong about the root cause of the missing report. In the earlier config (pre https://github.com/pypa/setuptools/pull/2486), the CIs created a junit-compatible XML file and now they don't. Here's how to recover this:
```suggestion
        run: tox -- --junit-xml=.test-results/pytest/cov.xml
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v1
        with:
          file: .test-results/pytest/cov.xml  # Explicitly point at the correct coverage report
          flags: >-  # Mark which lines are covered by which envs
            GHA,
            ${{ runner.os }},
            ${{ matrix.python }}
          functionalities: gcov  # disable C coverage autodiscovery
```

.github/workflows/main.yml
```diff
@@ -19,7 +19,16 @@ jobs:
         run: |
           python -m pip install tox
       - name: Run tests
-        run: tox
+        run: tox -- --junit-xml=.test-results/pytest/cov.xml
```
jaraco, line 22: Does codecov use junitxml? I'm pretty sure junitxml is for test results and not coverage results. I suspect this setting is unnecessary and instead we need a setting to cause coverage results to be exported to an xml file.

.github/workflows/main.yml
```diff
@@ -19,7 +19,16 @@ jobs:
...
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          file: .test-results/pytest/cov.xml  # Explicitly point at the correct coverage report
```
jaraco, line 26: This setting is documented as optional. Let's find out what the default setting is for the codecov uploader (I think it's `coverage.xml`) and make sure to produce the file using that filename and eliminate this optional setting.

.github/workflows/main.yml
```diff
@@ -19,7 +19,16 @@ jobs:
...
+          flags: >-  # Mark which lines are covered by which envs
+            GHA,
```
jaraco, line 28: Since there are no environments that don't use Github actions, it's probably premature to mark these tests as GHA. Probably better to eliminate it until it's needed.

.github/workflows/main.yml
jaraco: Fixed in 725be0142.

.github/workflows/main.yml
jaraco: Fixed in 725be0142.

.github/workflows/main.yml
jaraco: Fixed in a77319fff.

jaraco (approved):

jaraco:
Thanks for the diligent work on putting this together.

.github/workflows/main.yml
webknjaz: Isn't it the same?

webknjaz: Oh, I confused test results with coverage

tanvimoharir:
Thanks @webknjaz and @jaraco 
I'm still new to this and this was a good learning for me. I will continue to learn more and find some other (beginner friendly) issues.

PR #2697, opened 2021-06-16
Fix problems when sys.executable is slightly different to argv[0]

### LINKED ISSUE #216: Generated launchers won't launch if there is a space in the path
Originally reported by: **stevedower (Bitbucket: [stevedower](http://bitbucket.org/stevedower), GitHub: Unknown)**

---

(This is related to https://bitbucket.org/pypa/setuptools/issue/188/nt-escaping-in-shebang-in-easy_install, but that one is now being treated as a feature request and I believe the two changes would be unrelated anyway.)

The latest setuptools generates .exe launchers that include a shebang line pointing to the python.exe. If there is a space in this path, the path is quoted.

However, when the launcher runs, it will add a second set of quotes to the path leading to an error like this:

``` text
Fatal error in launcher: Unable to create process using '""D:\...\Test VEnv\Scripts\python.exe"" "D:\...\Test VEnv\Scripts\easy_install.exe" '
```

I believe the issue is in the launcher rather than the shebang line, since quoting that path is the only way you could support arguments as well as spaces on Windows. IMO, the launcher should not add more quotes to the shebang line when launching.

My repro steps (run at the command prompt):

``` text
C:\Python27\python.exe -m virtualenv "Test VEnv"
cd Test VEnv
Scripts\python.exe -m pip install -U setuptools==4.0.1

rem The current easy_install.exe was generated with setuptools 3.something and it works fine

rem The next line generates a new easy_install.exe with setuptools 4.0.1 and it doesn't work
Scripts\python.exe -m pip install -U setuptools==4.0.0
Scripts\easy_install.exe
```

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/216

set-soft:
## Summary of changes

This patch fixes problems when Python is installed on a directory with spaces on Windows and the `sys.executable` variable doesn't exactly match `argv[0]`.

Closes #216   At least on my system this was the root of the problem.

### DIFF
--- a/setuptools/command/install_scripts.py
+++ b/setuptools/command/install_scripts.py
@@ -42,7 +42,7 @@ class install_scripts(orig.install_scripts):
         if is_wininst:
             exec_param = "python.exe"
             writer = ei.WindowsScriptWriter
-        if exec_param == sys.executable:
+        if os.path.abspath(exec_param) == os.path.abspath(sys.executable):
             # In case the path to the Python executable contains a space, wrap
             # it so it's not split up.
             exec_param = [exec_param]


setuptools/command/install_scripts.py
```diff
@@ -42,7 +42,7 @@ def run(self):
         if is_wininst:
             exec_param = "python.exe"
             writer = ei.WindowsScriptWriter
-        if exec_param == sys.executable:
+        if os.path.abspath(exec_param) == os.path.abspath(sys.executable):
```
webknjaz, line 45: Could you add a test for this?

webknjaz (changes requested):
Also, besides the test, this change requires a changelog entry.
Here's how to do this: https://setuptools.readthedocs.io/en/latest/development/developer-guide.html#adding-change-notes-with-your-prs.

set-soft:
Sorry, but I can't add a test. The error is quite obvious, comparing files by its raw name is a wrong thing to do, no matters the OS.
I work on Linux and I was trying to solve a users problem on Windows.

jaraco:
After some inspection, I believe this change doesn't achieve the intended goal. Looking at the code, `exec_param` will be `"python.exe"` on Windows. That means that depending on what the current working directory is, `abspath(exec_param)` will be essentially `os.path.join(os.getcwd(), "python.exe")`. For example:

```
>>> os.path.abspath('python.exe') 
'C:\\python.exe' 
```

I'm struggling to imagine a scenario where that's the right thing to compare against `sys.executable`, which is always absolute to begin with.

Coupled with the fact that I was unable to replicate the referenced issue, I suggest to go back to the drawing board. Please at the very least replicate the issue in your environment, make the patch, and then show that the patch corrects the issue. If possible, trace the code at the point before the patch and show what the values are that are implicated in the failure.

Thanks again for the contrib, but I'm declining this one.

set-soft:
> After some inspection, I believe this change doesn't achieve the intended goal. Looking at the code, `exec_param` will be `"python.exe"` on Windows. That means that depending on what the current working directory is, `abspath(exec_param)` will be essentially `os.path.join(os.getcwd(), "python.exe")`. For example:
> 
> ```
> >>> os.path.abspath('python.exe') 
> 'C:\\python.exe' 
> ```

Not in my case. The value is a full path. And the patch could also be:

```
-        if exec_param == sys.executable:
+       if exec_param == sys.executable or os.path.abspath(exec_param) == os.path.abspath(sys.executable):
```

set-soft:
> Please at the very least replicate the issue in your environment, make the patch, and then show that the patch corrects the issue. If possible, trace the code at the point before the patch and show what the values are that are implicated in the failure.

The patch was verified to work in my environment! Is not a theoretical patch.

webknjaz:
Then, a test is needed to prove that it doesn't break envs that aren't yours + it'd be a regression protection. For the maintainers to accept this, there must be at least something they could verify.

Issue #2699, opened 2021-06-17
[FEATURE] Allow paths outiside of the source code for long_description


airvzxf:
### setuptools version

setuptools>=57.0.0

### Python version

Python 3.9

### OS

Arch Linux

### Additional environment information

It happens when I try to use a README.md file outside the source folder.

### Description

In the file `setup.cfg`, when I try to use the read me file `../README.md` it is not working.

Error: 
```text
distutils.errors.DistutilsOptionError: `file:` directive can not access /home/wolf/workspace/projects/sniparinject/README.md

ERROR Backend subproccess exited when trying to invoke get_requires_for_build_sdist
```

The file exists, looks the parsing of the `setup.cfg` is not recognized in the paths outside the source code, for example `../README.md` is not working but `./other_folder/README.md` works.

I found this issue https://github.com/pypa/setuptools/issues/1635, which was related to my problem, but I am not agreed. 

### Expected behavior

Same as the `setup.py`, which recognize the paths outside the source code.

```python
import setuptools

with open('../README.md', 'r', encoding="utf-8") as fh:
    long_description = fh.read()

setuptools.setup(
    long_description=long_description,
)
```

### How to Reproduce

```bash
git clone --no-checkout https://github.com/airvzxf/sniparinject.git
cd sniparinject || exit 123
git checkout af2215e3661e06c2d2569949d9b19f6f549a90e8

./script/setup-dev.bash

# Display read me file.
ls -lha README.md

cd ./src || exit

# Display read me file.
ls -lha ../README.md

# Check if the setup cfg has the parameter: long_description = file: ../README.md
cat setup.cfg

rm -fR build dist ./*.egg-info
source ../venv/bin/activate
python3 -m pip install --upgrade pip build twine
python3 -m build
```

### Output

```console
/home/wolf/workspace/projects/sniparinject/script/pypi-build.bash

CURRENT_PATH=$(dirname "${0}")
cd "${CURRENT_PATH}" || exit

cd ../src || exit
rm -fR build dist ./*.egg-info
source ../venv/bin/activate
# This file must be used with "source bin/activate" *from bash*
# you cannot run it directly

deactivate () {
    # reset old environment variables
    if [ -n "${_OLD_VIRTUAL_PATH:-}" ] ; then
        PATH="${_OLD_VIRTUAL_PATH:-}"
        export PATH
        unset _OLD_VIRTUAL_PATH
    fi
[...]

PR #2701, opened 2021-06-18
Allows to use files outside of the main directory in setup.cfg file

airvzxf:
## Summary of changes

This change allows to use the file outside the main directory. For example, in the file `setup.cfg`, section `long_description` the user can use an outside file `file: ../README.md`.

Closes https://github.com/pypa/setuptools/issues/2699

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -308,11 +308,8 @@ class ConfigHandler:
         """Represents value as a string, allowing including text
         from nearest files using `file:` directive.
 
-        Directive is sandboxed and won't reach anything outside
-        directory with setup.py.
-
         Examples:
-            file: README.rst, CHANGELOG.md, src/file.txt
+            file: README.rst, CHANGELOG.md, src/file.txt ../file.txt
 
         :param str value:
         :rtype: str
@@ -330,15 +327,8 @@ class ConfigHandler:
         return '\n'.join(
             cls._read_file(path)
             for path in filepaths
-            if (cls._assert_local(path) or True)
-            and os.path.isfile(path)
+            if os.path.isfile(path)
         )
-
-    @staticmethod
-    def _assert_local(filepath):
-        if not filepath.startswith(os.getcwd()):
-            raise DistutilsOptionError(
-                '`file:` directive can not access %s' % filepath)
 
     @staticmethod
     def _read_file(filepath):
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -309,7 +309,7 @@ class ConfigHandler:
         from nearest files using `file:` directive.
 
         Examples:
-            file: README.rst, CHANGELOG.md, src/file.txt ../file.txt
+            file: README.rst, CHANGELOG.md, src/file.txt, ../file.txt
 
         :param str value:
         :rtype: str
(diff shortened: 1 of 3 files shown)


webknjaz (changes requested):
I'm pretty sure that it's not a bug but a designed behavior. Even the test name says "sandbox". Referencing files outside the project is prone to errors and security bugs. I think, as a workaround, you could symlink it. Also, how do you include it in an sdist if it's outside?

airvzxf:
All the information and my arguments are in the Issue https://github.com/pypa/setuptools/issues/2699, but I'll try to resume in this comment.

> Even the test name says "sandbox". Referencing files outside the project is prone to errors and security bugs.

In my point of view, say that take file outside the project is a security issue is relative. Because if I have the permissions to access to a root file via `setuptools`, means I can copy manually this file in the project directory. What is the preventing this security code changes? Is the person who build the application the only one who can modify anything before upload to the `PyPi`?


- If you think that my PR is not relevant, and you didn't merge my code. At least add in these pieces of code a warning message saying: `For security reasons, you could't use file outside of this directory: /home/my-user/projects/hulala/src/`
> Also, how do you include it in a `sdist` if it's outside?

This is the interesting part, if you review my code changes and description in the issue. It isn't copying the file into the project directory, it is copy the content and put in some place, in this case in the `PKG-INFO` file. Means I am able to have a README.md file in `/root/hack-me/README.md` and put this content in my project `/home/my-user/projects/hulala/src/hulala.egg/PKG-INFO`. Do I expose that is essential in this PR?

> I think, as a workaround, you could symlink it.

As I mentioned in the created issue. I am able to fix using the `setup.py` without problems `with open('../README.md', 'r', encoding="utf-8") as fh:`, also create a symlink looks ugly and not make sense to have a file. Let me put an example: Imagine that my documentation in `./hulala/src/README.md` is referring to the application, this documentation not necessary needs to have the documentation for the `PyPi`. In fact, it shouldn't have the `PyPi` documentation. Let say that this documentation is in `./hulala/docs/pypi/README.md`. Then, when I create the package, I want to specify in the file `setup.cfg`:
- `long_description = file: ../docs/pypi/README.md`
- `long_description = file: /home/my-user/hulala/docs/pypi/README.md`

Do  you get my point?

setuptools/config.py
```diff
@@ -308,11 +308,8 @@ def _parse_file(cls, value):
         """Represents value as a string, allowing including text
         from nearest files using `file:` directive.
 
-        Directive is sandboxed and won't reach anything outside
-        directory with setup.py.
```
jaraco, line 312: This change removes the intentional constraint to disallow loading from outside the project directory. This change will affect any config option that uses `_parse_file`. If we wish to change the behavior here to be more lenient, we should at least enumerate the directives that might use this behavior. Is it only `long_description` or are there others?

jaraco (commented):
> > Also, how do you include it in a `sdist` if it's outside?
> 
> This is the interesting part, if you review my code changes and description in the issue. It isn't copying the file into the project directory, it is copy the content and put in some place, in this case in the `PKG-INFO` file.


I think you may have missed the point. If your project loads its long_description from `../README`, then you build an sdist and ship that off to PyPI, and then I download the sdist and try to build it, it will fail to build because ../README doesn't exist.

What's worse, though, is if you use `long_description = file:/etc/passwd`, and then you build an sdist and ship that off to your colleague who builds a wheel on their build server and installs that to their applications, now `importlib.metadata.metadata('your-package')['Description']` will have the contents of `/etc/passwd` for the build server.

I'm sorry, but it's not about security, but it's about having hermetic packages whose sources are transportable.

I think you may wish to update the error message for clarity, but I recommend to do that in a separate PR.

PR #2712, opened 2021-06-30
Implement #1681 (`glob:` directive for `[options.data_files]`)

### LINKED ISSUE #1681: glob patterns for data_files
Would be great to be able to use glob patterns for `data_files`, so that one can write something like this:

```
[options.data_files]
share/myapp/doc = doc/*.md
```

This would be consistent with the way `package_data` works.

darkvertex:
## Summary of changes

Hello! 👋 First time contributor here.

This change (which closes #1681) introduces an optional new `glob:` directive just for values of `[options.data_files]` which will expand the given patterns to the corresponding relative paths.

It is also possible to do a mix of some regular strings and some using the new directive in the same dict. (See new unittest for an example.)

I took the liberty of updating `docs/userguide/declarative_config.rst` to mention the new directive. Please advise if it should be noted elsewhere in the documentation.

I tried to stay within the style of the file and added a new unittest for this new feature. Let me know what you think or if you wish to reword or adjust anything.

### Pull Request Checklist
- [✅] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/config.py
+++ b/setuptools/config.py
@@ -9,6 +9,7 @@ import importlib
 from collections import defaultdict
 from functools import partial
 from functools import wraps
+from glob import iglob
 import contextlib
 
 from distutils.errors import DistutilsOptionError, DistutilsFileError
@@ -254,6 +255,33 @@ class ConfigHandler:
             value = value.split(separator)
 
         return [chunk.strip() for chunk in value if chunk.strip()]
+
+    @classmethod
+    def _parse_list_glob(cls, value, separator=','):
+        """Equivalent to _parse_list() but expands 'glob:' directives using glob().
+
+        However, unlike glob(), the resolved results stay relative paths.
+
+        :param value:
+        :param separator: List items separator character.
+        :rtype: list
+        """
+        directive = 'glob:'
+        values = cls._parse_list(value, separator=separator)
+        expanded_values = []
+        for value in values:
+            trimmed_value = value.strip()
+            if trimmed_value.startswith(directive):
+                # take what is after "glob:"
+                value = trimmed_value.split(directive, 1)[-1].strip()
+                value = os.path.abspath(value)
+                # and expand it while keeping as a relative path:
+                value = sorted(os.path.relpath(path, os.getcwd()) for path in iglob(value))
+                expanded_values.extend(value)
+            else:
+                expanded_values.append(value)
+
+        return expanded_values
 
     @classmethod
     def _parse_dict(cls, value):
@@ -711,5 +739,5 @@ class ConfigOptionsHandler(ConfigHandler):
 
         :param dict section_options:
         """
-        parsed = self._parse_section_to_dict(section_options, self._parse_list)
+        parsed = self._parse_section_to_dict(section_options, self._parse_list_glob)
         self['data_files'] = [(k, v) for k, v in parsed.items()]
--- a/docs/userguide/declarative_config.rst
+++ b/docs/userguide/declarative_config.rst
@@ -68,6 +68,7 @@ boilerplate code in some cases.
         site.d/00_default.conf
         host.d/00_default.conf
     data = data/img/logo.png, data/svg/icon.svg
+    fonts = glob: data/fonts/*.ttf, glob: data/fonts/*.otf
 
 Metadata and options are set in the config sections of the same name.
 
@@ -156,6 +157,8 @@ Special directives:
       The ``file:`` directive is sandboxed and won't reach anything outside
       the directory containing ``setup.py``.
 
+* ``glob:`` - Value will be treated as a glob() pattern and expanded accordingly.
+
 
 Metadata
 --------
@@ -225,7 +228,7 @@ package_data             section                                              [#
 exclude_package_data     section
 namespace_packages       list-comma
 py_modules               list-comma
-data_files               dict                                 40.6.0
+data_files               glob:, dict                          40.6.0
 =======================  ===================================  =============== =========
 
 **Notes**:
--- a/setuptools/tests/test_config.py
+++ b/setuptools/tests/test_config.py
@@ -893,6 +893,41 @@ class TestOptions:
             ]
             assert sorted(dist.data_files) == sorted(expected)
 
+    def test_data_files_globby(self, tmpdir):
+        fake_env(
+            tmpdir,
+            '[options.data_files]\n'
+            'cfg =\n'
+            '      a/b.conf\n'
+            '      c/d.conf\n'
+            'data = glob: *.dat\n'
+            'icons = \n'
+            '      glob: *.ico\n'
+            'audio = \n'
+            '      glob:*.wav\n'
+            '      sounds.db\n'
+        )
+
+        # Create dummy files for glob()'s sake:
+        tmpdir.join('a.dat').write('')
+        tmpdir.join('b.dat').write('')
+        tmpdir.join('c.dat').write('')
+        tmpdir.join('a.ico').write('')
+        tmpdir.join('b.ico').write('')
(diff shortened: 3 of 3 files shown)


changelog.d/2712.change.rst
```diff
@@ -0,0 +1,2 @@
+Added `glob:` directive for use inside `[options.data_files]` values, with accompanying unittests.
+Updated declarative_config.rst docs to explain new `glob:` directive.
```
eli-schwartz, line 2: A changelog entry can probably skip mentioning that:
- the change comes with unittests
- the change includes documentation updates

Maybe this info would fit better in the commit message?

jaraco (commented):
This work is really good... good enough to merge after fixing a couple of nitpicks (done).

My only hesitation is that `license_files`, which was introduced recently, also accepts globs, but using a different syntax (no `glob:` prefix) and relies on [postprocessing](https://github.com/pypa/setuptools/blob/0aa3576ae23fd88d450e41931a6703e8eb96e9c4/setuptools/dist.py#L583-L599) to expand the globs.

At some point, it would probably be worthwhile to harmonize those two mechanisms (if that makes sense). And of course, it might prove easiest to harmonize the two before this setting adopts the glob feature and users come to depend on it.

@darkvertex Would you be willing to analyze these two settings and make a recommendation on how you'd like to proceed?

darkvertex:
@jaraco Hi! Thanks for the review and doing the little adjustments.

> This work is really good... good enough to merge after fixing a couple of nitpicks (done).

Thanks! :)

---

> My only hesitation is that `license_files`, which was introduced recently, also accepts globs, but using a different syntax (no `glob:` prefix) and relies on [postprocessing](https://github.com/pypa/setuptools/blob/0aa3576ae23fd88d450e41931a6703e8eb96e9c4/setuptools/dist.py#L583-L599) to expand the globs.
> 
> At some point, it would probably be worthwhile to harmonize those two mechanisms (if that makes sense). And of course, it might prove easiest to harmonize the two before this setting adopts the glob feature and users come to depend on it.
> 
> @darkvertex Would you be willing to analyze these two settings and make a recommendation on how you'd like to proceed?

@jaraco If I understand correctly, the `license_files` with glob support is already released to the public, right? Doesn't that make it too late to add the notion of the `glob:` prefix/directive to it and telling people to explicitly use that when they want globbing?

Or did you mean to harmonize as in make both functions reuse the same glob expansion logic?

I see two avenues of going about this:

---

A) If we want to make `license_files` support the `glob:` directive but not expand patterns by default -- which it does currently expand, as I understand it, so it would be a *breaking change* but I think it's the most logical/harmonious way to bring both together -- then it'd be super easy to just change line 557 in config.py:
https://github.com/darkvertex/setuptools/blob/data_files_glob_directive/setuptools/config.py#L557
from `'license_files': parse_list,` to `'license_files': parse_list_glob,` and do away with the `_expand_patterns(` stuff in line 583 of `dist.py`:
https://github.com/pypa/setuptools/blob/0aa3576ae23fd88d450e41931a6703e8eb96e9c4/setuptools/dist.py#L583
but keeping the logic about searching for some default patterns (`patterns = ('LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*')`) when no values are declared whatsoever.

---

B) If you wanted to simply bring some of the core glob logic together to reduce repetition / duplication, not changing any behaviour of `license_files`, I could move `_expand_patterns()`:
https://github.com/pypa/setuptools/blob/0aa3576ae23fd88d450e41931a6703e8eb96e9c4/setuptools/dist.py#L602
into `config.py` and introduce the os.path.relpath() bit I had in my expansion -- which does not seem to break any related tests; I tried it:
```
    def _expand_patterns(patterns):
        """
        >>> list(Distribution._expand_patterns(['LICENSE']))
        ['LICENSE']
        >>> list(Distribution._expand_patterns(['setup.cfg', 'LIC*']))
        ['setup.cfg', 'LICENSE']
        """
        return (
            os.path.relpath(path, os.getcwd())
            for pattern in patterns
            for path in sorted(iglob(pattern))
            if not path.endswith('~')
            and os.path.isfile(path)
        )
```
then I could use it in line 280:
https://github.com/darkvertex/setuptools/blob/data_files_glob_directive/setuptools/config.py#L280
where it says:
```
                value = sorted(
                    os.path.relpath(path, os.getcwd()) for path in iglob(value))
```
and edit it to reuse the function:
```
                value = sorted(cls._expand_patterns([value]))
```
and then make sure it still gets used in `dist.py`'s `_finalize_license_files()` func (after importing it at the top of the file, of course.)

There's already a few things being imported from config.py in dist.py, so it would not be wise to move that logic completely to dist.py as that could cause a circular import situation.

---

How would you guys like to proceed?

_Personally_ I prefer option A: make `license_files` use the new `glob:` directive and break its globbing behaviour from happening by default when not empty, but I totally understand if we want to avoid pissing people off since it seems to be a released feature (though then again, it is very recent so maybe it wouldn't piss off that many! ha)

eli-schwartz:
We could also slightly reverse tracks and decide to expand globs by default in data_files. I discussed the tradeoffs in https://github.com/pypa/setuptools/issues/1681#issuecomment-869209762

jaraco:
> If you wanted to simply bring some of the core glob logic together to reduce repetition / duplication

I wasn't thinking about avoiding code duplication, so thanks for raising that possibility. I do, however, want to focus primarily on the user interface first and then address implementation details subsequently.

I'm liking option A, and I do think that's the best approach.

In that case, it doesn't particularly affect this PR, so I'll likely merge this soon.

darkvertex:
As requested by @jaraco [in his comment](https://github.com/pypa/setuptools/issues/2720#issuecomment-882877355) I've adapted the code such that globbing is implicit.

Looks like someone needs to approve the Github workflows again for me since I'm a first-time contributor. Can someone help?

setuptools/config.py
```diff
@@ -255,6 +256,43 @@ def _parse_list(cls, value, separator=','):
...
+        values = cls._parse_list(value, separator=separator)
+        expanded_values = []
+        for value in values:
+            trimmed_value = value.strip()
+
+            # Has globby characters?
+            if any(char in value for char in glob_characters):
+                value = os.path.abspath(value)
+
+                # check if this path has globby characters but does in fact exist:
```
jaraco, line 280: Thanks for thinking of this case. I also agree it's an edge case and unlikely to be depended on, especially since `*` matches `*` and `?` matches `?`. The only situation where this would be a problem is if there were _also_ adjacent data files that were not meant to be matched by the glob but were after this change. That seems remote enough a possibility that I'd prefer to defer this special handling until such a time that it's needed.

darkvertex: Fair enough. I'll take it out.

darkvertex:
I fixed the line-too-long flake8 warning. Sorry about that.

Can someone re-run the tests please?

darkvertex:
No rush, but just sharing that I'll be AFK on vacation for three weeks starting next week. So if anyone wants any further little edits from me, now is the time to ask for them (or else see ya late August.)

darkvertex:
@jaraco I see you tried to create a release but Github Actions workflow failed due to a test failure.

Thought you'd like to know that it would appear that it's an issue with `sphinx` being incompatible with Python 3.10 and apparently not an issue with the code of setuptools:
https://github.com/sphinx-doc/sphinx/issues/9505#issuecomment-891650831

I also found this fix for the line causing the type error seen in this repo's [failed build](https://github.com/pypa/setuptools/runs/3443509191#step:5:113):
https://github.com/sphinx-doc/sphinx/commit/8b2031c747e7c7e6b845ee2e3db47de617d33cc6
...but it doesn't seem to be in the latest stable sphinx release yet.

PR #2716, opened 2021-07-04
Remove ssl_support.

### LINKED ISSUE #2715: [BUG] ssl.match_hostname is deprecated
### setuptools version

main

### Python version

3.10b something

### OS

CI

_No response_

### Description

After merging jaraco/skeleton, which includes running tests on Python 3.10, some deprecation warnings manifest as errors: https://github.com/pypa/setuptools/runs/2984838195

### Expected behavior

Setuptools should be deprecation-free.

### How to Reproduce

Follow the CI steps.

### Output

n/a

### Code of Conduct

jaraco:
Fixes #2715.

It appears that `ssl.match_hostname` has been deprecated since Python 3.7 and without any replacement. Perhaps that means that hostname verification is built into the default URL openers, or maybe it means the functionality isn't supported any longer. In either case, since the Setuptools package index code is deprecated, there's no point in trying to drag this functionality into the future, so just rely on what on the default implementation for HTTPS connections.


amotl:
### Thoughts about replacement for `find_ca_bundle`

Dear Jason,

at https://github.com/crate/crate-python/pull/523#issuecomment-1433170310, we discovered setuptools dropped the `setuptools.ssl_support` module.

We have been using `setuptools.ssl_support.find_ca_bundle` within our test suite, which croaks like [^1] when using a more recent version of `setuptools`.

Maybe you have any sort of suggestion how to resolve that, other than vendoring the previous version of the `setuptools.ssl_support` module into the failing testsuite?

With kind regards,
Andreas.

[^1]: https://github.com/crate/crate-python/actions/runs/5480142715/job/14840796866?pr=558#step:4:226

PR #2723, opened 2021-07-09
Fixes #2722 Adds an environment variable SETUPTOOLS_EXT_SUFFIX

### LINKED ISSUE #2722: [FR] Allow environment variable override of EXT_SUFFIX

I'm trying to cross-compile other people's packages, but the value of `EXT_SUFFIX` is always taken from the host Python.

When I set `SETUPTOOLS_EXT_SUFFIX` environment variable, the value is used instead of `sysconfig.get_config_var("EXT_SUFFIX")`. This would allow me to set the suffix precisely for all packages in a build.

As I've already set environment variables to control the builds, another environment variable is the best solution for matching these. (Command-line and per-environment settings may be lost during the build process.)

### Alternative Solutions

I have previously tried building with other tools (and even written an entire backend that can handle this), but doing it for arbitrary projects means porting them to build backends other than setuptools.

It would be possible to extract the wheels and rename individual files, however, this leaves traces and interferes with reproducibility.

Rather than a completely open-ended environment variable, setuptools could try and automatically determine what the suffix should be. (But the point of overriding the automatic detection is to avoid this kind of logic and let things be set in the environment, so I'd probably _still_ want the option of a complete override.)

### Additional context

_No response_

### Code of Conduct

zooba:
Adds an environment variable SETUPTOOLS_EXT_SUFFIX to override the suffix inferred from the host Python runtime.

Closes #2722

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -104,14 +104,20 @@ class build_ext(_build_ext):
                 self.write_stub(package_dir or os.curdir, ext, True)
 
     def get_ext_filename(self, fullname):
-        filename = _build_ext.get_ext_filename(self, fullname)
+        so_ext = os.getenv('SETUPTOOLS_EXT_SUFFIX')
+        if so_ext:
+            filename = os.path.join(*fullname.split('.')) + so_ext
+        else:
+            filename = _build_ext.get_ext_filename(self, fullname)
+            so_ext = get_config_var('EXT_SUFFIX')
+
         if fullname in self.ext_map:
             ext = self.ext_map[fullname]
             use_abi3 = getattr(ext, 'py_limited_api') and get_abi3_suffix()
             if use_abi3:
-                so_ext = get_config_var('EXT_SUFFIX')
                 filename = filename[:-len(so_ext)]
-                filename = filename + get_abi3_suffix()
+                so_ext = get_abi3_suffix()
+                filename = filename + so_ext
             if isinstance(ext, Library):
                 fn, ext = os.path.splitext(filename)
                 return self.shlib_compiler.library_filename(fn, libtype)
--- a/changelog.d/2722.change.rst
+++ b/changelog.d/2722.change.rst
@@ -1,0 +1,1 @@
+Added support for ``SETUPTOOLS_EXT_SUFFIX`` environment variable to override the suffix normally detected from the ``sysconfig`` module.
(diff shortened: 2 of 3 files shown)


zooba:
I went looking for somewhere good to add this to the docs, but nothing stood out. Any suggestions?

zooba:
Not sure whether you normally squash merge these, but please do. Otherwise, I'll merge all the changes together - there's no interesting progression in the commits, just me making mistakes ;)

zooba:
Did my own rebase (to cover up my shame...)

setuptools/command/build_ext.py
```diff
@@ -104,14 +104,20 @@ def copy_extensions_to_source(self):
                 self.write_stub(package_dir or os.curdir, ext, True)
 
     def get_ext_filename(self, fullname):
-        filename = _build_ext.get_ext_filename(self, fullname)
+        so_ext = os.getenv('SETUPTOOLS_EXT_SUFFIX')
+        if so_ext:
+            filename = os.path.join(*fullname.split('.')) + so_ext
+        else:
+            filename = _build_ext.get_ext_filename(self, fullname)
+            so_ext = get_config_var('EXT_SUFFIX')
```
jaraco, line 112: Have you considered adding a hook inside `get_config_var` or `get_ext_filename` that would address the need more fundamentally?

zooba: Yeah, that's where I started. Obviously that only works when you also opt in to using the `distutils` that comes with `setuptools`, but that's fine.

The bigger issue is that the code later in this function tries to strip off the suffix later, which is not possible to do if `get_ext_filename` uses anything other than `get_condig_var('EXT_SUFFIX')` (at least without the same logic as what I added here anyway).

Adding the hook in `setuptools._distutils.sysconfig.get_config_var` would have made some sense. However, I assume at some point in the future that the patches for `build_ext` will eventually merge _somewhere_, and at some point `distutils.sysconfig` will be replaced with the standard library's `sysconfig` module. If the hook were in `_distutils`, there's a pretty good chance it would be forgotten later and the functionality would be lost. This way, there's no chance of it being forgotten in any refactorings.

zooba: (And yes, the tests would detect the functionality being lost in a refactoring. But tracking down the root cause and handling it - probably just like this - would take up someone's day and I don't want to inflict that on them!)

ax3l:
Shouldn't this also add the same logic to `get_ext_filename`?

https://github.com/pypa/setuptools/blob/v60.9.0/setuptools/_distutils/command/build_ext.py#L675-L683

PR #2725, opened 2021-07-09
Fixes #2724 Adds detection of Windows ARM64 build environments

### LINKED ISSUE #2724: [BUG] win-arm64 platform is not detected
### setuptools version

latest

### Python version

any

### OS

Windows

Only relevant when `%VSCMD_ARG_TGT_ARCH%` is set to `arm64` (for example, in a configured build environment).

### Description

For other values (`x86`, `x64`, `arm`), the `VSCMD_ARG_TGT_ARCH` variable is used to select the platform specifier (see setuptools/_distutils/util.py line 100ish). There is no value for `arm64`, and so it builds for the current platform rather than cross-compiling.

### Expected behavior

When the value is `arm64`, it should select platform `win-arm64`. This allows later steps to select the correct build tools and cross-compile.

Note that you also need to [obtain and] ensure that setuptools will find the ARM64 compatible .lib files for Python. Putting them into a venv's `libs` directory works fine (provided you make sure the build happens in that venv, which is not hard).

### How to Reproduce

1. Open regular command prompt
2. `set VSCMD_ARG_TGT_ARCH=arm64`
3. `pip wheel --no-binary :all: --no-deps --no-cache --no-build-isolation markupsafe` (or any other simple module with native code)

The generated wheel will be for the same platform as the Python it was run with, rather than respecting the target architecture.

If you `set VSCMD_ARG_TGT_ARCH=arm` without getting the right libs, the compilation fails and it generates a `none-any` wheel because markupsafe knows how to handle this. Other packages will just hard fail.

### Output

```console
Collecting markupsafe                                                                                                   
  Downloading MarkupSafe-2.0.1.tar.gz (18 kB)                                                                           
Building wheels for collected packages: markupsafe                                                                      
  Building wheel for markupsafe (setup.py) ... done                                                                     
  Created wheel for markupsafe: filename=MarkupSafe-2.0.1-cp39-cp39-win_amd64.whl size=17214 sha256=cd682d318cfd3a5805c032fa6b1d9e40def490f62be7b4057fcd67cf4f6cc1dc                                                                            
  Stored in directory: C:\...\Temp\pip-ephem-wheel-cache-lgs8awq4\wheels\9f\6d\c8\1f59b07cf85ae842908006ec28f4477f7e4578df72c3eb0e46                                                                                     
Successfully built markupsafe                                                                                           
```

Note that the wheel platform is `win32` or `win_amd64`, and not `win_arm64`.

### Code of Conduct

zooba:
Adds detection of Windows ARM64 build environments for cross-compilation.

Closes #2724

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 57.0.0
+current_version = 57.1.0
 commit = True
 tag = True
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,23 @@
+v57.1.0
+-------
+
+
+Changes
+^^^^^^^
+* #2692: Globs are now sorted in 'license_files' restoring reproducibility by eliminating variance from disk order.
+* #2714: Update to distutils at pypa/distutils@e2627b7.
+* #2715: Removed reliance on deprecated ssl.match_hostname by removing the ssl support. Now any index operations rely on the native SSL implementation.
+
+Documentation changes
+^^^^^^^^^^^^^^^^^^^^^
+* #2604: Revamped the backward/cross tool compatibility section to remove
+  some confusion.
+  Add some examples and the version since when ``entry_points`` are
+  supported in declarative configuration.
+  Tried to make the reading flow a bit leaner, gather some informations
+  that were a bit dispersed.
+
+
 v57.0.0
 -------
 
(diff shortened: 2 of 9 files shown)


zooba:
Personally I think this is trivial enough to get by without a test. Setting up the build environment that needs this is a bit of a mess, but without this fix upstream it's worse because you need to force a patched setuptools into pip's isolated environments.

jaraco:
This change looks fine, except changes to `_distutils` should be applied to pypa/distutils. Let me see if I can grab the commit and merge it in here.

jaraco:
Pushed as pypa/distutils@f15f814e9e.

jaraco:
Released as 57.2.0. Thanks!

PR #2731, opened 2021-07-18
remove lib2to3 usage

### LINKED ISSUE #2086: lib2to3 will be removed in Python 3.10

graingert:
## Summary of changes

remove any mention of 2to3

Closes #2086

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
Thanks for working on this. I see this change includes changes to `_distutils`. Those changes should be applied first to [pypa/distutils](/pypa/distutils). Will it be possible to submit the same changes to that repo in a way such that they can land there first, and such that when applied at the PR here, there's no diff for `setuptools._distutils`?

graingert:
> Thanks for working on this. I see this change includes changes to `_distutils`. Those changes should be applied first to [pypa/distutils](/pypa/distutils). Will it be possible to submit the same changes to that repo in a way such that they can land there first, and such that when applied at the PR here, there's no diff for `setuptools._distutils`?

hello, does this mean I need to get the change into github.com/python/cpython ?

jaraco:
No. Distutils in cpython is frozen, deprecated, and slated for removal. pupa/distutils is the official maintenance line for that work until Setuptools adopts it completely.

graingert:
> No. Distutils in cpython is frozen, deprecated, and slated for removal. pupa/distutils is the official maintenance line for that work until Setuptools adopts it completely.

PR here for the distutils readme: https://github.com/pypa/distutils/pull/48/files

graingert:
> Thanks for working on this. I see this change includes changes to `_distutils`. Those changes should be applied first to [pypa/distutils](/pypa/distutils). Will it be possible to submit the same changes to that repo in a way such that they can land there first, and such that when applied at the PR here, there's no diff for `setuptools._distutils`?

those changes are now here: https://github.com/pypa/distutils/pull/49

jaraco:
I've merged the changes into distutils, then merged those changes into this branch and then rebased the non-distutils changes on top of those. I then fixed the linter warnings and updated the changelog entry. I believe it's now ready to go.

KOLANICH:
This change should be reverted, since it forces people to do either of things:

* drop support of python 2
* drop support of modern versions of setuptools as a build backend (and probably fork setuptools, causing fragmentation of infrastructure and bloat)

Whether people must drop python 2 in their code - it is their decision, and `setuptools` should not dictate it.

As a workaround, support of `use_lib2to3` **can** be moved into a setuptools plugin (it should be there from the beginning, but it is likely that there were no suppkrt if plugins in setuptools that time), but it would require breaking changes in many projects.

graingert:
> This change should be reverted, since it forces people to do either of things:
> 
>     * drop support of python 2
> 
>     * just drop support of modern versions of setuptools as a build backend (and probably fork setuptools, causing fragmentation of infrastructure and bloat)
> 
> 
> Whether people must drop python 2 in their code - it is their decision, and `setuptools` should not dictate it.
> 
> As a workaround, support of `use_2to3` **can** be moved into a setuptools plugin (it should be there from the beginning, but it is likely that there were no suppkrt if plugins in setuptools that time), but it would require changes in many projects (specifying that such a package must be installed in `pyproject.toml` and `setup.{py,cfg}`).

The other option is for package authors to apply `modernize` and commit the changes, producing a universal wheel and removing the need for a deprecated transpile step

jaraco:
There are other options, such as publishing python 2 and python 3 wheels (and owning the transpile process locally), possibly using 3to2 to backport the changes to Python 2.

> it is their decision, and setuptools should not dictate it.

Setuptools dropped support for Python 2 a while ago. Those still relying on Python 2 code should expect to run up against compatibility issues.

> changes in many projects

I'm interested in what projects are affected. I'm under the impression there are a small number of rarely-used projects relying on this behavior. Please do share any projects that are broken by this change. I'd like to help push those projects forward. If the number is substantial, we might consider rolling back this change.

> moved into a setuptools plugin

I welcome others to tackle this challenge. I believe the effort to be high and the value proposition as small.

KOLANICH:
>I'm under the impression there are a small number of rarely-used projects relying on this behavior. Please do share any projects that are broken by this change.

Not so many:

* `setup.py` : https://github.com/search?q=use_2to3+AND+path%3A%2F+AND+name+AND+install_requires+AND+setuptools+AND+setup+AND+extension%3Apy&type=Code (1,331 code results)

* `setup.cfg`: https://github.com/search?q=use_2to3+AND+path%3A%2F+AND+name+AND+install_requires+AND+metadata+AND+options+AND+extension%3Acfg&type=Code (15 code results)

>Setuptools dropped support for Python 2 a while ago. Those still relying on Python 2 code should expect to run up against compatibility issues.

Surely. But here is a difference. If someone wanted to support both versions, they used `use_2to3`. Then when a package was built from sdists on python 3, it was transpiled into it. Probably it was a big mistake to introduce this option at all.

But it was done and there are  some packages that were trying to support both versions now depend on this option. This means that when setuptools is upgraded on **machines of python __3__ users**, who use the packages trying to support both versions, they will run into problems. &lt; ffffrustration and anger here &gt;

It means that the affected packages need to be **urgently** fixed, because all the software dependent on these packages got broken in a widespread use case of installing it from source from git. The easiest solution is to just use `2to3` (since it is already installed and known and is guaranteed to produce the workable code, because the code produced that way was already battle-tested) and to remove that option from the packages' `setup.cfg` and `setup.py` and just drop support of python 2. (Disclaimer: I have never supported running upon python 2 in my own packages, I came into python after 3 was already fully usable).

The alternative solution suggested by @graingert is a much better one.

>I welcome others to tackle this challenge. I believe the effort to be high and the value proposition as small.

Why do you think so? I already have [a setuptools plugin](https://github.com/kaitaiStructCompile/kaitaiStructCompile.setuptools/blob/master/kaitaiStructCompile/setuptools.py#L285), though it does another task, it should be fairly easy to utilise its code for writing a plugin calling 2to3. Though I am not going to implement such a plugin, since I personally don't use python 2.

erwinlem:
"Errors should never pass silently."

Enabling use_2to3 should be a failing error at least, it's never going to work. Took me an hour to figure out why [nose](https://github.com/nose-devs/nose/) would not compile.

jaraco:
And that commit was released in 58.0.2.

PR #2732, opened 2021-07-18
Remove `pass` after `log.info` and `DistutilsPlatformError`

donno2048:
## Summary of changes

Simply some unnecessary code removed


jaraco:
Any contributions to `_distutils` should be submitted to [pypa/distutils](/pypa/distutils). Thanks.

PR #2735, opened 2021-07-19
Remove repeated information in same sentence

lmmx:
Minor edit to setuptools docs on PEP 517

## Summary of changes

Removed a repeated part of the sentence in the setuptools docs. Does not touch any code.


jaraco:
Thanks for noticing and caring enough to send a PR.

Issue #2704, opened 2021-06-19
[BUG] "Unknown MS Compiler version" when compiling with mingw32


hickford:
From https://bugs.python.org/issue25251#msg251758

> Hi distutils. I previously used compiler=mingw32 with success. Today I installed Visual Studio 2015. Now compiler=mingw32 gives me an error
>
> File "c:\python35\lib\distutils\cygwinccompiler.py", line 86, in get_msvcr     raise ValueError("Unknown MS Compiler version %s " % msc_ver) ValueError: Unknown MS Compiler version 1900
>
> Any ideas? I don't understand why what msvc I have installed affects build with mingw32

This code still exists in https://github.com/pypa/setuptools/blob/main/setuptools/_distutils/cygwinccompiler.py

hickford:
https://bugs.python.org/issue25251#msg251758 includes a fix. I'll share a pull request

imba-tjd:
See my PR https://github.com/pypa/distutils/pull/41 and its discussion.

Though I don't have the experience to ensure it's OK

imba-tjd:
This can be closed.

kerneltravel:
may caused by 32bit / 64bit python and lib mis-match

PR #2742, opened 2021-07-26
Improve MSVC support in cygwin compiler.

hickford:
Patch from https://bugs.python.org/issue25251

Fixes https://github.com/pypa/setuptools/issues/2704


jaraco:
Changes to `_distutils` need to be contributed to [pypa/distutils](/pypa/distutils) and they will be merged into here.

PR #2762, opened 2021-08-30
Update codecov.yml

### LINKED ISSUE #2727: Adding codecov causes spurious PR check failures
In #2693, this project added codecov back in. Following that addition, I created a pull request (#2726) that didn't change any line of code but still somehow caused a check failure (indicating that code coverage was [reduced by .04%](https://github.com/pypa/setuptools/runs/3040300168) and causing the checks to fail.

I'm hoping @tanvimoharir or @webknjaz can figure out how to either:

- fix the root cause of these spurious check failures, or
- devise a project-agnostic way to make the checks less sensitive (allowed to pass unless there are significant regressions in coverage).

I guess I'd also entertain simply disabling the ability for the checks to fail (report coverage, but never fail), which would be comparable to the prior behavior.

The reason I want the solution to be project-agnostic is because I'd like to be able to apply this technique across a suite of projects and I don't want each project to have to custom-tune the coverage in order to reach a reasonably quiet check. It's okay if a project wants to tune to be more sensitive, but the default should be to produce far less noise than signal.

tanvimoharir:
## Summary of changes

This is a change to in codecov configurations to prevent failures as those seen post #2693 

Closes #2727 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
This change appears to change three different settings in the configuration. Can you explain the purpose of each of those settings?

tanvimoharir:
> This change appears to change three different settings in the configuration. Can you explain the purpose of each of those settings?

The changes are currently adapted from the configurations initially recommended at #2727 (the main change being reducing the threshold )
These are not final and I'm still looking into details of initial failures.

PR #2764, opened 2021-09-05
Declare dependencies in metadata

jaraco:
I'm attempting once again to allow Setuptools to declare dependencies. Due to known bootstrapping issues, it's not possible to declare simply the dependencies (especially any dependencies that require setuptools to install) without considering the bootstrapping issue.

This change offers a fallback to add a directory of dependencies to be added to sys.path when Setuptools/pkg_resources is imported/invoked.


webknjaz (commented):
I think I'd prefer the vendored blobs not to be checked into Git but rather gitignored. There could be a script that would populate them separately. Thinking along the lines of what I attempted in https://github.com/python/cpython/pull/12791.

jaraco:
I'm still unsure why the test failures are occurring. They don't occur for me locally, even on Python 3.6. The failures are all in the `test_pip_upgrade_from_source` tests. I have Python 3.6.6 while the Github actions has 3.6.14.

jaraco:
The failures have at least two modes. On pip 9.0.1, it fails early checking for setuptools being installed. On 18.3, the tests fail when attempting to install the dependencies:

```
Processing ./setuptools-58.0.0.post20211020-py3-none-any.whl
Collecting pyparsing==2.2.1 (from setuptools==58.0.0.post20211020)
  Using cached https://files.pythonhosted.org/packages/42/47/e6d51aef3d0393f7d343592d63a73beee2a8d3d69c22b053e252c6cfacd5/pyparsing-2.2.1-py2.py3-none-any.whl
Collecting ordered-set==3.1.1 (from setuptools==58.0.0.post20211020)
  Using cached https://files.pythonhosted.org/packages/a3/b7/d4d69641cbe707a45c23b190f2d717466ba5accc4c70b5f7a8a450387895/ordered-set-3.1.1.tar.gz
    Complete output from command python setup.py egg_info:
    ModuleNotFoundError: No module named 'setuptools'
```

I believe the issue is that prior to pip 20, there wasn't adequate support for building source packages without setuptools being implicitly installed.

I think I'm prepared to declare non-support for installation from source with these older pips.

jaraco:
I believe the reason why the tests don't fail for me locally is because I have a local pip cache with things like 'ordered-set' already built into a wheel.

jaraco:
> I think I'd prefer the vendored blobs not to be checked into Git but rather gitignored. There could be a script that would populate them separately. Thinking along the lines of what I attempted in [python/cpython#12791](https://github.com/python/cpython/pull/12791).

I'd consider something along these lines. I think we can consider that change separately, no?

jaraco:
FFS. Now tests are failing with some error in flake8.

```
________________________________ FLAKE8-check _________________________________
[gw0] linux -- Python 3.6.15 /home/runner/work/setuptools/setuptools/.tox/python/bin/python
.tox/python/lib/python3.6/site-packages/pluggy/_hooks.py:265: in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
.tox/python/lib/python3.6/site-packages/pluggy/_manager.py:80: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
.tox/python/lib/python3.6/site-packages/_pytest/runner.py:170: in pytest_runtest_call
    raise e
.tox/python/lib/python3.6/site-packages/_pytest/runner.py:162: in pytest_runtest_call
    item.runtest()
.tox/python/lib/python3.6/site-packages/pytest_flake8.py:127: in runtest
    self.statistics)
.tox/python/lib/python3.6/site-packages/py/_io/capture.py:150: in call
    res = func(*args, **kwargs)
.tox/python/lib/python3.6/site-packages/pytest_flake8.py:222: in check_file
    app.report_errors()
.tox/python/lib/python3.6/site-packages/flake8/main/application.py:309: in report_errors
    results = self.file_checker_manager.report()
.tox/python/lib/python3.6/site-packages/flake8/checker.py:249: in report
    results_reported += self._handle_results(filename, results)
.tox/python/lib/python3.6/site-packages/flake8/checker.py:161: in _handle_results
    physical_line=physical_line,
.tox/python/lib/python3.6/site-packages/flake8/style_guide.py:430: in handle_error
    code, filename, line_number, column_number, text, physical_line
.tox/python/lib/python3.6/site-packages/flake8/style_guide.py:581: in handle_error
    self.formatter.handle(error)
.tox/python/lib/python3.6/site-packages/flake8/formatting/base.py:100: in handle
    self.write(line, source)
.tox/python/lib/python3.6/site-packages/flake8/formatting/base.py:203: in write
    self._write(line)
.tox/python/lib/python3.6/site-packages/flake8/formatting/base.py:187: in _write
    sys.stdout.buffer.write(output.encode() + self.newline.encode())
E   AttributeError: '_io.StringIO' object has no attribute 'buffer'
------------------------------ Captured log call -------------------------------
WARNING  flake8.options.manager:manager.py:221 option --max-complexity: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
```

jaraco:
Aaah. It's because there was a lint error (unused import) and the error handling between pytest and flake8 is broken.

jaraco:
Strangely, when I create a virtualenv locally, I don't end up with pyparsing in it. Is it `pytest-virtualenv` that's adding it?

jaraco:
Okay, I see it now. pyparsing is present in `c:\hostedtoolcache\windows\python\3.9.7\x64\lib\site-packages`. But virtualenv isn't supposed to give access to system-site-packages by default.

```
~ $ virtualenv --help | grep 'site-packages dir'
  --system-site-packages        give the virtual environment access to the system site-packages dir (default: False)
```

So why is a package from the system site packages leaking into a (bare) virtualenv?

jaraco:
Ugh. The problem is that `bare_virtualenv.run` doesn't honor the virtualenv on Windows:

```
> c:\code\public\pypa\setuptools\setuptools\tests\test_virtualenv.py(48)test_clean_env_install() 
-> bare_virtualenv.run(['python', 'setup.py', 'install'], cd=tmp_src)
(Pdb) bv = bare_virtualenv 
(Pdb) bv.run(['python', '-c', 'import sys; print(sys.executable)']) 
C:\Python39\python.exe 
```

jaraco:
Okay, so in pytest-virtualenv, it's crucial to pass `.python` for the executable or use the full string syntax. Otherwise, the subprocess may find "python" on the path and not in the virtualenv.

webknjaz:
> Point taken, but what happens if the technique for bundling dependencies has negative downstream
> effects? Then the only option is to back out the whole change and not just the bundling behavior.

That's an interesting question. Note that downstream distro packages absolutely love unbundling the vendored deps and depend on packages that are already packaged for their ecosystem. So I'd say they already have to deal with stuff like this.
Of course, it may somehow be problematic as those packages may have an undesired version.
There's a solution for this — ansible/ansible has a special location with a `.gitignore` in its tree — a `_vendor/` folder. The assumption is that if the packagers have problems with getting the desired deps packaged separately, they could just drop them into that folder and those deps would become available on the import path.
Essentially, the point is that downstreams usually have their own ways to populate this sort of stuff (or vice versa — purge the vendored content).

> I agree that it's undesirable to have this bulk content in the repo, so I would like to avoid that.
> 
> I'd also like to be able to support pip install from Github, e.g.:
> 
> ```
> $ pip-run git+https://github.com/pypa/setuptools@feature/unvendor -- -c "import setuptools, appdirs; print(appdirs.__file__)"
> ...
> Successfully installed appdirs-1.4.3 more-itertools-8.8.0 ordered-set-3.1.1 packaging-20.4 pyparsing-2.2.1 setuptools-58.2.0.post20211022 six-1.16.0
> /var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pip-run-bcrng0_h/appdirs.py
> ```
> 
> It's not obvious to me how that could work if the content isn't already bundled in the repo.

May I suggest implementing the download logic + caching in the PEP 517 in-tree build backend? If the dirs are already populated, it'd skip the download (after checking the checksums, I guess).

P.S. Apparently pip's vendoring script is now available on PyPI: https://github.com/pypa/pip/tree/main/src/pip/_vendor#automatic-vendoring / https://pypi.org/project/vendoring/. Maybe it's worth checking it out.

> Plus, I observe that the content is currently _already_ bundled in both `pkg_resources/_vendor` and `setuptools/_vendor`, so this change limits that effect.
> 
> I can rewrite the history to remove .whl files for now, as they're unusable.

Yeah, I'm mostly unhappy with things like whl/exe at this point. It'd be great to catch these with linters. Maybe there's some pre-commit hook repo disallowing blobs?

setuptools/__init__.py
```diff
@@ -13,6 +13,7 @@
 
 from ._deprecation_warning import SetuptoolsDeprecationWarning
 
+import _setuptools_vendored  # noqa: F401
```
webknjaz, line 16: I recommend adding a code comment with an explanation of what this does and why it's important to keep it in this specific position.

changelog.d/2764.breaking.rst
```diff
@@ -0,0 +1 @@
+Setuptools now declares its dependencies in metadata but also vendors libraries satisfying those dependencies if they're not present. This change means that Setuptools cannot install from source under pip earlier than 20.
```
webknjaz, line 1: @hroncok FYI

jaraco:
I'm abandoning this effort for now.

PR #2770, opened 2021-09-06
Fail fast when use_2to3 is supplied.

### LINKED ISSUE #2769: [BUG] Should fail hard when installing packages that want 2to3
### setuptools version

setuptools==58.0.0

### Python version

Python 3.9

### OS

Linux

_No response_

### Description

Tried to install a package that uses 2to3 (specifically https://pypi.org/project/demjson/), got a message from pip like:
```
Successfully installed demjson-2.2.3
```
But really it wasn't successful, and then get confusing `SyntaxError` exceptions.

### Expected behavior

I have now worked out that this is because of the change to remove `2to3` from setuptools, which I think is a perfectly reasonable change, and because my setuptools version wasn't pinned. But setuptools is not doing something the package is asking for, and the result is going to be something that does not run (though I guess there are caveats here).

I would have expected to get an error like "This module wants to use 2to3, but this is not supported [link to github issue about removing 2to3 support]". At least, that would certainly have saved me time.

### How to Reproduce

Install `demjson==2.2.3` with `setuptools==58.0.0`. Attempt to import and get a syntax error.

### Output

```console
$ pip install -U setuptools==58.0.0 && pip install demjson==2.2.3 && python -c 'import demjson'
Collecting setuptools==58.0.0
  Using cached setuptools-58.0.0-py3-none-any.whl (816 kB)
    Found existing installation: setuptools 56.0.0
    Uninstalling setuptools-56.0.0:
      Successfully uninstalled setuptools-56.0.0
Successfully installed setuptools-58.0.0
Collecting demjson==2.2.3
  Using cached demjson-2.2.3.tar.gz (131 kB)
Using legacy 'setup.py install' for demjson, since package 'wheel' is not installed.
Installing collected packages: demjson
    Running setup.py install for demjson ... done
Successfully installed demjson-2.2.3
  File "/home/user/example/venv/lib/python3.9/site-packages/demjson.py", line 645
    class json_int( (1L).__class__ ):    # Have to specify base this way to satisfy 2to3
                      ^
SyntaxError: invalid syntax
```

### Code of Conduct

jaraco:
Fixes #2769.

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -289,6 +289,10 @@ def assert_bool(dist, attr, value):
         raise DistutilsSetupError(tmpl.format(attr=attr, value=value))
 
 
+def invalid(dist, attr, value):
+    raise DistutilsSetupError(f"{attr} is invalid.")
+
+
 def check_requirements(dist, attr, value):
     """Verify that install_requires is a valid requirements list"""
     try:
--- a/changelog.d/2769.misc.rst
+++ b/changelog.d/2769.misc.rst
@@ -1,0 +1,1 @@
+Build now fails fast when ``use_2to3`` is supplied.
--- a/setup.cfg
+++ b/setup.cfg
@@ -124,6 +124,7 @@ distutils.setup_keywords =
 	dependency_links = setuptools.dist:assert_string_list
 	test_loader = setuptools.dist:check_importable
 	test_runner = setuptools.dist:check_importable
+	use_2to3 = setuptools.dist:invalid
 egg_info.writers =
 	PKG-INFO = setuptools.command.egg_info:write_pkg_info
 	requires.txt = setuptools.command.egg_info:write_requirements


setuptools/dist.py
```diff
@@ -289,6 +289,10 @@ def assert_bool(dist, attr, value):
         raise DistutilsSetupError(tmpl.format(attr=attr, value=value))
 
 
+def invalid(dist, attr, value):
+    raise DistutilsSetupError(f"{attr} is invalid.")
```
webknjaz, line 293: @jaraco in the light of https://github.com/pypa/setuptools/issues/2775, maybe this could at least be turned into
```suggestion
    if not value:  # it's okay, if it's False
        warnings.warn(f"Remove the {attr} setting from your packaging.")
        return
    raise DistutilsSetupError(f"{attr} is invalid.")
```

jaraco: I originally was writing the function with this check, but then I wanted to make the "check" function suitable for any invalid value. I'll review the referenced bug and consider this proposal.

webknjaz: Yeah, it turned out more impactful on the scale than you've probably expected. Seems rather urgent.

samdoran:
@jaraco Thank you for the fast and excellent fix. Much appreciated.

Issue #2776, opened 2021-09-07
[BUG] PIP3 Install hangs indefinitely


mgorsk1:
### setuptools version

setuptools==58.0.2

### Python version

Python 3.7

### OS

CentOS Linux 7 (in Docker)

### Additional environment information

Running code in gitlab ci/cd docker runner. Gitlab version 14.0.1.

### Description

We run pip3 install --upgrade setuptools pip and after that pip3 install -r requirements-dev.txt with following content:

```
mock
coverage
flake8
mypy
pytest
apache-airflow==1.10.11
types-PyYAML
types-requests
```

Until yesterday this step finished in under 2 minutes with setuptools==58.0.0. Today newer version of setuptools has been pulled which made step timeout after 1hr. The job is stuck on pip3 install step with croniter package being consistently the last one (stuck for ~55 minutes) before timeout happens.

### Expected behavior

pip3 install -r requirements-dev.txt finishes ok

### How to Reproduce

run pip3 install --upgrade setuptools pip and after that pip3 install -r requirements-dev.txt with following content:

```
mock
coverage
flake8
mypy
pytest
apache-airflow==1.10.11
types-PyYAML
types-requests
```

### Output

```console
  Downloading https://xxx/artifactory/api/pypi/ISP-PyPI/packages/packages/28/27/f2c883373aa58fe77ecfc09a984f1ae52bc05790c0eee66b87f65d14cbfe/croniter-0.3.23.tar.gz (18 kB)
INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. If you want to abort this run, you can press Ctrl + C to do so. To improve how pip performs, tell us what happened here: https://pip.pypa.io/surveys/backtracking
  Downloading https://xxx/artifactory/api/pypi/ISP-PyPI/packages/packages/79/f1/3d9dac6f43a1f789ad79c1dc2589e7389fc4f0c4753a451485ecb326c357/croniter-0.3.22.tar.gz (18 kB)
  Downloading https://xxx/artifactory/api/pypi/ISP-PyPI/packages/packages/69/84/1a839fdb454929b2f5f34dfc0f6ae9c393193e4c088fa62b37ec318c0df4/croniter-0.3.20.tar.gz (18 kB)
ERROR: Job failed: execution took longer than 1h0m0s seconds
```

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

webknjaz:
[...]

Issue #2775, opened 2021-09-07
[BUG] error in XXXX setup command: use_2to3 is invalid when false.


miguelfito:
### setuptools version

setuptools==58.0.2

### Python version

Python 3.6.9, Python 3.9.2

### OS

Ubuntu 18.04.5 LTS, Debian 11 

### Additional environment information

This happens installing any Python2 package when trying to use `use_2to3` on setuptool execution

### Description

Error installing a package that requires to use `use_2to3` on its installation

### Expected behavior

The package should be installed as previous versions do (like `setuptools==58.0.1`)

### How to Reproduce

1. Generate a `virtualenv` and activate it.
2. Make sure you're environment tools are updated.
3. Install a Python2 package that requires to use `use_2to3` on its installation.

### Output

1. Generate a `virtualenv` like:
```console
virtualenv .env -p /usr/bin/python3 && source .env/bin/activate
created virtual environment CPython3.9.2.final.0-64 in 121ms
  creator CPython3Posix(dest=/home/miguelfito/tmp/pruebas/devops_housekeeping_test/.env, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/miguelfito/.local/share/virtualenv)
    added seed packages: pip==20.3.4, pkg_resources==0.0.0, setuptools==44.1.1, wheel==0.34.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
```
2. Make sure you're environment tools are updated:
```console
python3 -m pip install --upgrade pip setuptools wheel
Looking in indexes: ****
Requirement already satisfied: pip in ./.env/lib/python3.9/site-packages (20.3.4)
Collecting pip
  Downloading pip-21.2.4-py3-none-any.whl (1.6 MB)
     |████████████████████████████████| 1.6 MB 5.4 MB/s 
Requirement already satisfied: setuptools in ./.env/lib/python3.9/site-packages (44.1.1)
Collecting setuptools
  Downloading setuptools-58.0.2-py3-none-any.whl (816 kB)
     |████████████████████████████████| 816 kB 20.9 MB/s 
Requirement already satisfied: wheel in ./.env/lib/python3.9/site-packages (0.34.2)
Collecting wheel
  Downloading wheel-0.37.0-py2.py3-none-any.whl (35 kB)
[...]

PR #2777, opened 2021-09-07
Do not error if use_2to3 is set to a false value

plumdog:
## Summary of changes

I think this slight change fixes an issue where setting `use_2to3=False` would fail fast when really it ought to be ignored.

I think this results in behaviour closer to what I had in mind with #2769.

### Pull Request Checklist
- [ ] Changes have tests - no because I can't see where one would go


haoyuanli:
should 58.0.2 be yanked? I assume this will go into a 58.0.3

setuptools/dist.py
```diff
@@ -289,8 +289,11 @@ def assert_bool(dist, attr, value):
         raise DistutilsSetupError(tmpl.format(attr=attr, value=value))
 
 
-def invalid(dist, attr, value):
-    raise DistutilsSetupError(f"{attr} is invalid.")
+def invalid_ignored_if_false(dist, attr, value):
+    if not value:
+        warnings.warn("{attr} is ignored")
```
webknjaz, line 294: I think this should be
```suggestion
        warnings.warn("{attr} is ignored", DistDeprecationWarning)
```
(following the style in this module).

changelog.d/2777.misc.rst
```diff
@@ -0,0 +1 @@
+Build does not fail fast when ``use_2to3`` is supplied but set to a false value.
```
webknjaz, line 1: I think, this should be a bugfix, not a misc change.

webknjaz (commented):
In general, I think it's a move in the right direction but could you also add a regression test for this scenario? I'd like to see both positive and negative test cases added.

changelog.d/2777.misc.rst
jaraco: Bugfix is misc. Naming is weird.

setup.cfg
```diff
@@ -124,7 +124,7 @@ distutils.setup_keywords =
 	dependency_links = setuptools.dist:assert_string_list
 	test_loader = setuptools.dist:check_importable
 	test_runner = setuptools.dist:check_importable
-	use_2to3 = setuptools.dist:invalid
+	use_2to3 = setuptools.dist:invalid_ignored_if_false
```
jaraco, line 127:
```suggestion
	use_2to3 = setuptools.dist:invalid_if_true
```

changelog.d/2777.misc.rst
webknjaz: Ah, I see. The towncrier config in this repo doesn't declare a separate bugfix type. Many other projects do so. This is what confused me :)

webknjaz:
FTR somebody needs to evaluate this as a possible fix for #2776.

jaraco:
> should 58.0.2 be yanked? I assume this will go into a 58.0.3

What value would there be in yanking 58.0.2?

jaraco:
> In general, I think it's a move in the right direction but could you also add a regression test for this scenario? I'd like to see both positive and negative test cases added.

I'm happy to skip regression tests for these tweaks.

haoyuanli:
> > should 58.0.2 be yanked? I assume this will go into a 58.0.3
> 
> What value would there be in yanking 58.0.2?

No, there wouldn't be, if a 58.0.3 is released.

potiuk:
We have some packages as dependencies, which have 2_to_3 set to true (specifically FlaskOpenId  which has not been updated since 2016, so this is likely it won't be updated. https://github.com/mitsuhiko/flask-openid/ Any chance setuptools can simply ignore 2_to_3 setting, rather than ignore when false ?

plumdog:
@potiuk See https://github.com/pypa/setuptools/issues/2775#issuecomment-914606508 and https://github.com/pypa/setuptools/issues/2769 for rationale behind not ignoring the flag - the installed package almost certainly won't be importable.

potiuk:
Yeah. Absolutely. Actually we had an interesting case because the Flask-openid actually was working with just removing 2_to_3 - because it was really small, the changes were harmless (like removing u` from unicod strings) and the only non-compatible change was under "if python < 2" anyway. Thankfully we got response from the maintainers and we became maintainers ourselves and we released a new version of package (after manually converting it with 2_to_3 and making it python3 - only). We also released a .wh;l package which remove the need of using setuptools at all.

I agree it's not possible and tha approach you come up with with "false" is good.

PR #2780, opened 2021-09-09
Deprecate bdist_rpm

hugovk:
## Summary of changes

Similar to https://github.com/pypa/setuptools/pull/2040, deprecate `bdist_rpm`.

For #1988.

---

Also some test/CI improvements (please let me know if these should be in another PR):

* Ignore a generated file from running tox locally
* CI: test slower PyPy3 first, to make full use of parallel jobs to finish faster
* CI: turn on pytest colours for GitHub Actions logging, makes it more readable

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/1988-change.rst
+++ b/changelog.d/1988-change.rst
@@ -1,2 +1,2 @@
-Deprecated the ``bdist_rom`` command. Binary packages should be built as wheels instead.
+Deprecated the ``bdist_rpm`` command. Binary packages should be built as wheels instead.
 -- by :user:`hugovk`
--- a/changelog.d/1988-change.rst
+++ b/changelog.d/1988-change.rst
@@ -1,0 +1,2 @@
+Deprecated the ``bdist_rom`` command. Binary packages should be built as wheels instead.
+-- by :user:`hugovk`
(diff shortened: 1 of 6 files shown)


changelog.d/1988-change.rst
```diff
@@ -0,0 +1,2 @@
+Deprecated the ``bdist_rom`` command. Binary packages should be built as wheels instead.
```
jmarshall, line 1:
```suggestion
Deprecated the ``bdist_rpm`` command. Binary packages should be built as wheels instead.
```

hugovk: Thanks! Mistyped that nearly every time but missed one :)

pytest.ini
```diff
@@ -4,6 +4,7 @@ addopts=
 	--doctest-modules
 	--doctest-glob=pkg_resources/api_tests.txt
 	-r sxX
+	--color=yes
```
jaraco, line 7: This project attempts to avoid proliferation of non-default options or to apply "preferred settings" across projects in jaraco/skeleton. In any case, this change is seemingly unrelated to the described change, so I'm going to omit it for now.

```suggestion
```

thisch:
There are users (like us) that are still using bdist_rpm to create rpms. I don't really understand the rational behind removing `bdist_rpm`. It would be nice if you could take a look at this comment https://github.com/pypa/setuptools/issues/1988#issuecomment-613093759 from April and answer the questions from @Cimbali.

PR #2800, opened 2021-09-22
sdist: Accept -u/--owner and -g/--group options

### LINKED ISSUE #1893: Please restore --owner= and --group= options to setuptools' sdist command
Tar archives record the Unix ownership (user and group) of each file they contain.  This can cause undesirable variation between source tarballs generated from the same VCS checkout but in different build environments, so `distutils.command.sdist.sdist` has options `--owner=[username]` and `--group=[groupname]` to override the recorded ownership.  However, `setuptools.command.sdist.sdist` supersedes the `user_options` array of its parent class and makes most of the options inaccessible, including `--owner` and `--group`.

Please restore these options.  The simplest change to achieve this would be to copy the `--owner` and `--group` array entries from `distutils.command.sdist.sdist.user_options` to `setuptools.command.sdist.sdist.user_options`.  I have written a monkeypatch in my `setup.py` that does exactly this, so I can confirm that no other changes to setuptools' code are required.  However, a more future-proof change would be to have setuptools.command.sdist.sdist copy its parent class's `user_options` array and *remove* only those entries that are definitely not wanted.

jmarshall:
## Summary of changes

Controlling the file ownership recorded in tar archives is useful for those striving towards reproducible builds, as discussed in #1893. These options are already understood by `distutils.command.sdist.sdist`, so just need to be accepted by `setuptools.command.sdist.sdist` to be propagated.

I didn't see any tests of the adjacent `sdist` options such as `--keep-temp` or `--dist-dir`, so as the change is to data only and effectively just forwards to already existing distutils code, I haven't added any additional tests. If there is somewhere appropriate to add a new test, please recommend where and I will add something appropriate. A similar comment applies to documentation.

Closes #1893.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/2800.change.rst
+++ b/changelog.d/2800.change.rst
@@ -1,0 +1,3 @@
+Added ``--owner`` and ``--group`` options to the ``sdist`` command,
+for specifying file ownership within the produced tarball (similarly
+to the corresponding distutils ``sdist`` options).
--- a/setuptools/command/sdist.py
+++ b/setuptools/command/sdist.py
@@ -31,6 +31,10 @@ class sdist(sdist_add_defaults, orig.sdist):
         ('dist-dir=', 'd',
          "directory to put the source distribution archive(s) in "
          "[default: dist]"),
+        ('owner=', 'u',
+         "Owner name used when creating a tar file [default: current user]"),
+        ('group=', 'g',
+         "Group name used when creating a tar file [default: current group]"),
     ]
 
     negative_opt = {}


jaraco:
Doing some more investigation, I see that this support was added to distutils in Python 2.7 and slated for Python 3.2, but rolled back (see 5f1980bbc9 and [bpo-6516](https://bugs.python.org/issue6516) and [bpo-19544](https://bugs.python.org/issue19544).

I wish I understood what factors led to the omission of other options from distutils. It looks like the masking of distutils options was intentional in f3694a038.

Ideally, reproducible builds should be possible without supplying non-default options, but for now, this approach is the least invasive and should ease the merge with distutils.

PR #2803, opened 2021-09-26
fix: cannot set root package properly

songmeixu:
According to https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages, the root package can be set through `package_dir` by two ways:
1. `'': <root-path>`
2. `'<root-pkg>': <the-dirpath-of-its-init-file>`

However, the case is it only get changed if the key `''` in `package_dir`. Otherwise, it just use a default value: the distribution root (`os.cwd()`) .

This make errors for pkg without a standard source directory layout. 
E.g. when a pkg installed on develop mode, and with `package_dir` set not with a key `:`. Then the installed pkg wouldn't find itself with right root path.

### DIFF
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -218,8 +218,13 @@ class egg_info(InfoCommon, Command):
             ) from e
 
         if self.egg_base is None:
-            dirs = self.distribution.package_dir
-            self.egg_base = (dirs or {}).get('', os.curdir)
+            dirs = self.distribution.package_dir or {}
+            if '' in dirs:
+                self.egg_base = dirs['']
+            elif self.egg_name in dirs:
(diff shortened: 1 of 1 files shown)


jaraco:
I'd like to see a couple of things here:

- More explanation about the flaw that's encountered. Do you have a sample project that fails under the current behavior? How does it fail?
- Tests. Since this change doesn't introduce any tests to capture the desired behavior, the behavior that's introduced here could be removed and tests would still pass. Also, the tests could provide links back to the justification for the behavior (the explanation).

I recommend to start with the explanation, because it may be that there is a better recommendation that avoids this undesirable behavior altogether, and maybe the solution is to disallow the usage that led to the undesirable behavior.

jaraco:
Feel free to revisit when you have time to address the concerns. We can either re-open this PR or create a new one.

Issue #2329, opened 2020-08-14
builds proceed with zero configuration


hroncok:
Hello, my understanding is that the `build_meta:__legacy__` backend should behave like `python setup.py ...` calls.

However, I've realized that it proceeds fine without any `setup.py` / `setup.cfg` files:

```
$ mkdir empty
$ cd empty/
$ pip list
Package    Version
---------- -------
packaging  20.4
pip        20.2.2
pyparsing  2.4.7
setuptools 49.6.0   # also happens with 41.6.0 and probably anything in between
six        1.15.0
wheel      0.35.0
$ python
>>> from setuptools.build_meta import __legacy__
>>> __legacy__.build_wheel('.')
running bdist_wheel
running build
installing to build/bdist.linux-x86_64/wheel
running install
running install_egg_info
running egg_info
creating UNKNOWN.egg-info
writing UNKNOWN.egg-info/PKG-INFO
writing dependency_links to UNKNOWN.egg-info/dependency_links.txt
writing top-level names to UNKNOWN.egg-info/top_level.txt
writing manifest file 'UNKNOWN.egg-info/SOURCES.txt'
reading manifest file 'UNKNOWN.egg-info/SOURCES.txt'
writing manifest file 'UNKNOWN.egg-info/SOURCES.txt'
Copying UNKNOWN.egg-info to build/bdist.linux-x86_64/wheel/UNKNOWN-0.0.0-py3.8.egg-info
running install_scripts
creating build/bdist.linux-x86_64/wheel/UNKNOWN-0.0.0.dist-info/WHEEL
creating '/home/churchyard/tmp/empty/tmpy86e5sx4/UNKNOWN-0.0.0-py3-none-any.whl' and adding 'build/bdist.linux-x86_64/wheel' to it
adding 'UNKNOWN-0.0.0.dist-info/METADATA'
adding 'UNKNOWN-0.0.0.dist-info/WHEEL'
adding 'UNKNOWN-0.0.0.dist-info/top_level.txt'
adding 'UNKNOWN-0.0.0.dist-info/RECORD'
removing build/bdist.linux-x86_64/wheel
'UNKNOWN-0.0.0-py3-none-any.whl'
>>> ^D
[empty]$ ls
build  UNKNOWN-0.0.0-py3-none-any.whl  UNKNOWN.egg-info
```

This might be a problem in scripted environments. I would have expect an error here. Is this behavior deliberate or accidental?

pganssle:
[...]

PR #2817, opened 2021-10-15
Implement metadata validation (not used)

### LINKED ISSUE #2799: setuptools.build_meta proceeds with no setup.py or setup.cfg [BUG]
### setuptools version

setuptools==58.1.0

### Python version

Python 3.9

### OS

Linux

_No response_

### Description

Running `pip install .` in a directory that contains nothing but a `pyproject.toml` with these contents:
```toml
[build-system]
requires = ["setuptools", "wheel"]
```
successfully installs a distribution containing no modules or packages and no useful metadata:
```console
$ cat /tmp/38/lib/python3.8/site-packages/UNKNOWN-0.0.0.dist-info/METADATA
Metadata-Version: 2.1
Name: UNKNOWN
Version: 0.0.0
Summary: UNKNOWN
Home-page: UNKNOWN
License: UNKNOWN
Platform: UNKNOWN

UNKNOWN
```

### Expected behavior

I would expect the build backend to fail, due to it lacking any useful configuration, rather than succeeding in building a package that is empty except for placeholder metadata.

This may be a duplicate of #2329, but that one was specifically scoped to `build_meta:__legacy__`, and all of the discussion was about the legacy backend. In this case, you can see that the same issue occurs even with the non-legacy backend.

### How to Reproduce

See description

### Output

See description

### Code of Conduct

pganssle:
## Summary of changes

Change to the `Distribution` class to require that some metadata exists before actually doing anything.

This is just a draft. It still needs test, presumably linting and formatting, and I'm not even confident that this is conceptually the right place for this logic, but I thought I'd throw this approach out there as an option for discussion.

Closes #2799, #2329 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


webknjaz:
> This is just a draft.

Mind converting it to a draft, then? So that it's visually distinctive in the PR list..

setuptools/dist.py
```diff
@@ -466,6 +466,29 @@ def __init__(self, attrs=None):
...
+
+        if missing:
+            if len(missing) == 1:
+                message = "%s attribute" % missing[0]
+            else:
+                message = "%s and %s attributes" % (", ".join(missing[:-1]),
+                                                    missing[-1])
+            raise DistutilsSetupError(
+                "Required package metadata is missing: please supply the %s." % message
+            )
```
webknjaz, line 485: Nit: use a guard expression to dedent this.

```suggestion
        if not missing:
            return

        if len(missing) == 1:
            message = "%s attribute" % missing[0]
        else:
            message = "%s and %s attributes" % (", ".join(missing[:-1]),
                                                missing[-1])
        raise DistutilsSetupError(
            "Required package metadata is missing: please supply the %s." % message
        )
```

jaraco (commented):
I have some ideas I'll add to the issue.

jaraco:
I spent some more time looking at this today with the expectation that I'd merge it without the version requirement, but as I looked into it more, I'm far from confident that overriding `run_commands` is the right hook point. I suspect it should happen late in `Distribution.finalize_options` or perhaps explicitly between construction and running the commands, something like:

```
diff --git a/setuptools/_distutils/core.py b/setuptools/_distutils/core.py
index d603d4a45..5ab4bbddf 100644
--- a/setuptools/_distutils/core.py
+++ b/setuptools/_distutils/core.py
@@ -142,6 +142,8 @@ def setup (**attrs):
     if _setup_stop_after == "commandline":
         return dist
 
+    getattr(dist, '_validate_metadata', lambda: None)()
+
     # And finally, run all the commands found on the command line.
     if ok:
         try:
```

But I'm not even sure that's early enough. In theory, additional commands could alter the metadata (e.g. `setup.py name_from_folder egg_info install`). Of course, such a usage is deprecated, but that doesn't mean it isn't a supported configuration now, and it would be invalidated be the proposed change.

I'd like to avoid adding more interactions between distutils and Setuptools classes (Distribution in this case).

So for now, I'd like to keep the _validate_metadata method, and defer the method of integration.

PR #2824, opened 2021-10-22
Deprecate setup_requires, setup.py install, and easy_install command.

jaraco:
### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/917.change.rst
+++ b/changelog.d/917.change.rst
@@ -1,0 +1,1 @@
+``setup.py install`` and ``easy_install`` commands are now officially deprecated. Use other standards-based installers (like pip) and builders (like build). Workloads reliant on this behavior should pin to this major version of Setuptools.
--- a/pytest.ini
+++ b/pytest.ini
@@ -29,3 +29,7 @@ filterwarnings=
 
 	# https://github.com/pypa/setuptools/issues/2823
 	ignore:setup_requires is deprecated.
+
+	# https://github.com/pypa/setuptools/issues/917
+	ignore:setup.py install is deprecated.
+	ignore:easy_install command is deprecated.
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -153,6 +153,12 @@ class easy_install(Command):
     create_index = PackageIndex
 
     def initialize_options(self):
+        warnings.warn(
+            "easy_install command is deprecated. "
+            "Use build and pip and other standards-based tools.",
+            EasyInstallDeprecationWarning,
+        )
+
         # the --user option seems to be an opt-in one,
         # so the default should be False.
         self.user = 0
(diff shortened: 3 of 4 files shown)


changelog.d/917.change.rst
```diff
@@ -0,0 +1 @@
+``setup.py install`` and ``easy_install`` commands are now officially deprecated. Use other standards-based installers (like pip) and builders (like build). Workloads reliant on this behavior should pin to this major version of Setuptools.
```
webknjaz, line 1: I wonder if it'd be useful to link https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html from the changelog.

jaraco: Done in 7b5158b10.

orbea:
You're going to break entire workflows and distros with this, please reconsider.

jaraco:
> You're going to break entire workflows and distros with this, please reconsider.

I'm tempted to un-deprecate `setup_requires` as a parameter, but to deprecate/remove only the `fetch_build_eggs` part. In other words, continue to allow `setup_requires` to supply requirements for PEP 517 hooks (`get_requires_for_build_sdist` and `get_requires_for_build_wheel`).

The problem is [the current implementation](https://github.com/pypa/setuptools/blob/8ab7799032fd7811c55388ce04df90e486370106/setuptools/build_meta.py#L136-L147) is a hack around `fetch_build_eggs`. I guess it's a manageable approach.

To be sure, the intention is to break workflows away from this unsafe and clumsy usage, but to give projects time to adapt. I'll probably want to wait until `--use-pep517` is the default in pip before removing the behavior.

jaraco:
Follow up in #2877.

orbea:
Thanks for the attention. For what its worth not everyone uses pip.  Distros and some other users just need a reliable way to install projects into potentially arbitrary locations. I agree the setup.py is not ideal, but I strongly request any enhancements or replacements take care to consider this workflow.

The GNU standards explain well what is often expected.

https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

Additionally some build environments may not have working networking so being able to continue to provide all dependencies as system packages would be helpful.

jaraco:
> Distros and some other users just need a reliable way to install projects into potentially arbitrary locations.

Backing out the `setup_requires` deprecation won't help here. Distros still need to use pip or some other PEP 517 compliant installer. Keeping `setup_requires` around for PEP 517 usage does mean that packagers will be able to define build-time requirements at build time in Python code (setup.py) or declare them in setup.cfg in addition to being able to declare them in pyproject.toml. While at first blush, this seems like a redundancy, it does have a nice separation of responsibility. One can relegate the build-requires in pyproject.toml to defining the build backend requirements, whereas `setup_requires` can indicate the project's specific build-time requirements (and even calculate them based on environment-specific factors).

`setup.py install` remains deprecated and it's up to downstream packagers to get on board with the PEP 517 protocol for building/installing Python packages.

PR #2835, opened 2021-10-27
docs: clarify option usage

layday:
## Summary of changes

This (a) attempts to frame setuptools  data inclusion keywords in terms
of wheel and source distributions and (b) to warn users about common
pitfalls.

Closes —

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
I would recommend not to think about `sdist` and `wheel` as comparable formats, but instead to think about them as two sides of the build/install step.

`sdist` is simply a collection of sources as a tarball, comparable to getting the sources from github. The MANIFEST is one way of specifying which sources to include in the source dist. Others include simply relying on the defaults, using a file-finder plugin (like setuptools_scm), or possibly relying on files implied by other settings.

When building a wheel, on the other hand, you're specifying the files that, given the sources (either from a checkout or sdist or simple files), will be built and installed.

There's a loose relationship between the two formats, but they're largely unrelated. The main thing that's important is that any files that are needed to build the wheel should also be made present in the sdist (otherwise when building from sdist, the package will produce a different install than when building from repo source).

I hope that helps explain why these seemingly similar formats actually have very different semantics about how to specify the inputs.

layday:
It doesn't really; the semantics are muddied.  The options do not behave in the way they do because the formats are not 'comparable'.  They've simply evolved over the course of some fifteen odd years along with the packaging ecosystem as a whole.  A lot of what might've seemed good in principle did not pan out in practice.  As a result data file inclusion in setuptools is a major pain point for newcomers just as much as it is for seasoned developers.  The nomenclature is problematic and the interaction between the options both surprising and surprisingly complex.  The least we can do is spell it all out in the documentation, so when things don't work in the way people might expect, they might have some indication as to why.

abravalheri:
Thanks @jaraco for the information. Indeed, if we consider that in order to be included to the wheel the data file should first be "added to the sdist", the same logical expressions for the inclusion of files in the distribution can be derived from the [experiment](https://github.com/abravalheri/experiment-setuptools-package-data).

So I understand that the desired behaviour is:

1. default setuptools algorithm/file-finder plugins/``MANIFEST.in`` list files to be included in the `sdist`
2. ``include_package_data=True`` allow all the files inside the package directories in the `sdist` to be added to the `wheel`.
3. ``package_data`` _SHOULD_ add extra files (e.g. compiled artifacts/code generated on the fly) to both ``sdist`` and ``wheel``.
4. ``exclude_package_data`` refines the behaviour of ``include_package_data`` and ``package_data``.

However item 4. is not working correctly due to [the bug pointed out by @jaraco in  #1461](https://github.com/pypa/setuptools/issues/1461#issuecomment-749092366), and is the source of all the confusion.

layday:
Closing as https://github.com/pypa/setuptools/pull/2844 will fix some of this behaviour and there's no agreement on framing data packaging in terms of sdists and wheels.

PR #2836, opened 2021-10-27
Duplicate all of setuptools' entry points to simplify bootstrapping

### LINKED ISSUE #2828: [FR] Allow bootstraping setuptools by running dependencies from source

`setuptools` requires `wheel` to be installed to bootstrap itself, but `wheel` requires `setuptools` to build, preventing bootstrapping.

Even though `setuptools` now bootstraps itself via a PEP 517 builder, not everyone is able to do the same. Linux distributions, for example, need to bootstrap it from scratch.

I would like an option, perhaps an environment variable, to skip the dependency checks, and to make the `bdist_wheel` always available. The easiest way to do this is to run everything from source, which trips up `setuptools`' dependency checks, and also as a consequence makes the `bdist_wheel` command unavailable.

### Alternative Solutions

I have written https://github.com/FFY00/python-bootstrap to help with this task, but currently, I need to run `setup.py egg_info` on setuptools, copy that to the package directory, in order to trick setuptools to think it is installed, and then do the same for wheel, only after that I can actually build wheels for `setuptools` and `wheel`.
https://github.com/FFY00/python-bootstrap/blob/4004fda86bc82470b0ee1b564764f8ba9960b493/bootstrap/build.py#L43
https://github.com/FFY00/python-bootstrap/blob/bf68a5ff74706e348de1f3dddeaf51bc5f37a449/bootstrap/__init__.py#L109
See https://github.com/pypa/setuptools/issues/2088#issuecomment-950233014 for more details.

Ideally, `setuptools` would actually be able to bootstrap itself without any dependencies.
See https://github.com/takluyver/flit/tree/master/flit_core for example.

### Additional context

_No response_

### Code of Conduct

layday:
## Summary of changes

Also adds `bdist_wheel`.

This is to obviate the intermediate `egg_info` step in both setuptools and wheel.

Closes #2828.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/bootstrap.egg-info/entry_points.txt
+++ b/bootstrap.egg-info/entry_points.txt
@@ -1,14 +1,60 @@
+# Keep this in sync with setup.cfg.
+
 [distutils.commands]
+alias = setuptools.command.alias:alias
+bdist_egg = setuptools.command.bdist_egg:bdist_egg
+bdist_rpm = setuptools.command.bdist_rpm:bdist_rpm
+build_clib = setuptools.command.build_clib:build_clib
+build_ext = setuptools.command.build_ext:build_ext
+build_py = setuptools.command.build_py:build_py
+develop = setuptools.command.develop:develop
+dist_info = setuptools.command.dist_info:dist_info
+easy_install = setuptools.command.easy_install:easy_install
 egg_info = setuptools.command.egg_info:egg_info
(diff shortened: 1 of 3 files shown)


layday:
Leaving this to someone more knowledgeable.

PR #2839, opened 2021-10-30
Maintain `requires` order in METADATA.

msuozzo:
## Summary of changes

Remove a source of instability in the ordering of requires entries. The order is generally specified by the package author and maintained through the build process. From a brief attempt to trace through all the related interfaces, it seems this is the only code that changes the order of the requirements and so would be the only source of the observed inconsistency of Requires-Dist entries.

For example, workflows that build -> install -> build would have the order of requires changed by these sorted() calls.

If this sorting is desirable, the other setuptools and wheel interfaces that interact with the requirements should also sort them. Otherwise, it's valuable to retain the order across all parts of the system.

### Pull Request Checklist
- [X] Changes have tests
- [X] News fragment added in [`changelog.d/`].

[PR docs]:


msuozzo:
@di this seems to impact wheel reproducibility.

jaraco:
I like this approach. I prefer not retain order rather than enforce order.

I'm not sure who added those sorted lines, but if the tests pass, then the sorting behavior is not tested, so it's fine to change it. Would you consider adding a test to capture the expectation this change introduces (so that someone else doesn't come along and suggest to sort the values again)?

di (approved):
This makes sense to me, I see no reason why these would need to be sorted.

jaraco:
The tests seem to be indicating that the sorting is needed. Would you be willing to investigate and determine what expectation is missed?

msuozzo:
Yep I haven't had the time to check out the failures but I'm planning on circling back to this in a bit.

msuozzo:
The sorting looks like it was only required because of a separate order-stability issue with pkg_resources. `frozenset` isn't stable while `dict` is, at least for recent Python versions, so the type+logic change should ensure the requires order remains constant.

msuozzo:
Is this good to merge now? Anything more I should test?

jaraco:
> Would you consider adding a test to capture the expectation this change introduces (so that someone else doesn't come along and suggest to sort the values again)?

msuozzo:
Ah didn't see that comment before. Added a test to more reliably fail if the order changes.

msuozzo:
Friendly ping

abravalheri:
Sorry @msuozzo, we might need to get the CI back into shape before running your tests (just to make sure everything is OK).

pkg_resources/__init__.py
```diff
@@ -3047,12 +3047,12 @@ def reqs_for_extra(extra):
                 if not req.marker or req.marker.evaluate({'extra': extra}):
                     yield req
 
-        common = frozenset(reqs_for_extra(None))
+        common = dict.fromkeys(reqs_for_extra(None))
```
abravalheri, line 3050: There is a difference here: `frozenset` would yield an immutable data structure, while `dict` is mutable...

But I suppose it does not make a difference in this particular implementation because `common` is not being directly modified/assigned to another variable/data-structure until the end of the function, right? Or is there any chance other piece of code would accidentally mutate it?

abravalheri (approved):

pkg_resources/__init__.py
msuozzo: Yeah it's a bit less expressive a type but given that it's only used locally, I don't see that risk extending beyond the function. Were it to be 10s of lines of code, it might make sense to take precautions but since it's just a handful of statements, I think it's tolerable.

abravalheri: Does it make sense to use something like `types.MappingProxy` to retain the original intention (of being read-only) and be future-proof?

msuozzo: Done

PR #2865, opened 2021-11-09
Test SETUPTOOLS_USE_DISTUTILS = stdlib, local on GH Actions

### LINKED ISSUE #2861: SETUPTOOLS_USE_DISTUTILS=local is not tested by CI infrastructure
### setuptools version

any

### Python version

any

### OS

any

_No response_

### Description

There is no testing of the vendored distutils that are activated by `SETUPTOOLS_USE_DISTUTILS=local`.

### Expected behavior

The vendored distutils, currently again only used when `SETUPTOOLS_USE_DISTUTILS=local` is set, should be checked by `tox` and GH Actions workflows.

### How to Reproduce

`$ git grep -l SETUPTOOLS_USE_DISTUTILS`

### Output

```
$ git grep -l SETUPTOOLS_USE_DISTUTILS
CHANGES.rst
_distutils_hack/__init__.py
docs/deprecated/distutils-legacy.rst
setup.py
setuptools/tests/test_distutils_adoption.py
```

### Code of Conduct

mkoeppe:
## Summary of changes

This also makes the integration workflow added in https://github.com/pypa/distutils/pull/67 actually work.

Closes #2861 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,6 @@ jobs:
         - ubuntu-latest
         - macos-latest
         - windows-latest
-      fail-fast: false
     runs-on: ${{ matrix.platform }}
     env:
       SETUPTOOLS_USE_DISTUTILS: ${{ matrix.distutils }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,9 @@ jobs:
   test:
     strategy:
       matrix:
+        distutils:
+        - stdlib
+        - local
         python:
         - pypy3
         - 3.6
@@ -16,6 +19,8 @@ jobs:
         - macos-latest
         - windows-latest
     runs-on: ${{ matrix.platform }}
+    env:
+      SETUPTOOLS_USE_DISTUTILS: ${{ matrix.distutils }}
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,7 @@ jobs:
         - ubuntu-latest
         - macos-latest
         - windows-latest
+      fail-fast: false
     runs-on: ${{ matrix.platform }}
     env:
       SETUPTOOLS_USE_DISTUTILS: ${{ matrix.distutils }}
--- a/tox.ini
+++ b/tox.ini
@@ -12,11 +12,7 @@ deps =
 	# TODO: remove after Sphinx>4.1.2 is available.
 	sphinx@git+https://github.com/sphinx-doc/sphinx; python_version>="3.10"
 commands =
-	# We invoke pytest like this to avoid the warning:
-	#   UserWarning: Distutils was imported before Setuptools,
-	#   but importing Setuptools also replaces the `distutils`
-	#   module in `sys.modules`.
-	pytest -c 'import setuptools; import pytest; pytest.main("{posargs}".split())'
+	pytest {posargs}
 usedevelop = True
 extras = testing
 passenv =
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ toxworkdir={env:TOX_WORK_DIR:.tox}
 
 [testenv]
 deps =
+	pytest
 	# workaround for sphinx-doc/sphinx#9562
 	# TODO: remove after Sphinx>4.1.2 is available.
 	sphinx@git+https://github.com/sphinx-doc/sphinx; python_version>="3.10"
--- a/tox.ini
+++ b/tox.ini
@@ -15,6 +15,7 @@ commands =
 usedevelop = True
 extras = testing
 passenv =
+	SETUPTOOLS_USE_DISTUTILS
 	windir  # required for test_pkg_resources
 
 [testenv:docs]
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,11 @@ deps =
 	# TODO: remove after Sphinx>4.1.2 is available.
 	sphinx@git+https://github.com/sphinx-doc/sphinx; python_version>="3.10"
 commands =
-	pytest {posargs}
+	# We invoke pytest like this to avoid the warning:
+	#   UserWarning: Distutils was imported before Setuptools,
+	#   but importing Setuptools also replaces the `distutils`
+	#   module in `sys.modules`.
+	pytest -c 'import setuptools; import pytest; pytest.main("{posargs}".split())'
 usedevelop = True
 extras = testing
 passenv =
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,6 @@ toxworkdir={env:TOX_WORK_DIR:.tox}
 
 [testenv]
 deps =
-	pytest
 	# workaround for sphinx-doc/sphinx#9562
 	# TODO: remove after Sphinx>4.1.2 is available.
 	sphinx@git+https://github.com/sphinx-doc/sphinx; python_version>="3.10"
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,5 @@
 import sys
+import warnings
 
 
 pytest_plugins = 'setuptools.tests.fixtures'
@@ -27,3 +28,11 @@ collect_ignore = [
 if sys.version_info < (3, 6):
     collect_ignore.append('docs/conf.py')  # uses f-strings
     collect_ignore.append('pavement.py')
+
+
+if sys.version_info > (3, 10):
+    # https://github.com/pypa/setuptools/pull/2865#issuecomment-965700112
+    warnings.filterwarnings(
+        'ignore',
+        'The distutils.sysconfig module is deprecated, use sysconfig instead',
+    )
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,4 @@
 import sys
-import warnings
 
 
 pytest_plugins = 'setuptools.tests.fixtures'
@@ -28,24 +27,3 @@ collect_ignore = [
 if sys.version_info < (3, 6):
     collect_ignore.append('docs/conf.py')  # uses f-strings
     collect_ignore.append('pavement.py')
-
-
-if sys.version_info > (3, 10):
-    # https://github.com/pypa/setuptools/pull/2865#issuecomment-965700112
-    warnings.filterwarnings(
-        'ignore',
-        'The distutils.sysconfig module is deprecated, use sysconfig instead',
-    )
-
-
-is_pypy = '__pypy__' in sys.builtin_module_names
-if is_pypy:
-    # Workaround for pypa/setuptools#2868
-    warnings.filterwarnings(
-        'ignore',
-        'Distutils was imported before setuptools',
-    )
-    warnings.filterwarnings(
-        'ignore',
-        'Setuptools is replacing distutils',
-    )
--- a/conftest.py
+++ b/conftest.py
@@ -36,3 +36,16 @@ if sys.version_info > (3, 10):
         'ignore',
         'The distutils.sysconfig module is deprecated, use sysconfig instead',
     )
+
+
+is_pypy = '__pypy__' in sys.builtin_module_names
+if is_pypy:
+    # Workaround for pypa/setuptools#2868
+    warnings.filterwarnings(
+        'ignore',
+        'Distutils was imported before setuptools',
+    )
+    warnings.filterwarnings(
+        'ignore',
+        'Setuptools is replacing distutils',
+    )
--- a/pytest.ini
+++ b/pytest.ini
@@ -34,3 +34,15 @@ filterwarnings=
 
 	# https://github.com/pypa/setuptools/issues/2497
 	ignore:.* is an invalid version and will not be supported::pkg_resources
+
+	# https://github.com/pypa/setuptools/pull/2865#issuecomment-965700112
+	# ideally would apply to Python 3.10+ when
+	# SETUPTOOLS_USE_DISTUTILS=stdlib but for
+	# https://github.com/pytest-dev/pytest/discussions/9296
+	ignore:The distutils.sysconfig module is deprecated, use sysconfig instead
+
+	# Workaround for pypa/setuptools#2868
+	# ideally would apply to PyPy only but for
+	# https://github.com/pytest-dev/pytest/discussions/9296
+	ignore:Distutils was imported before setuptools
+	ignore:Setuptools is replacing distutils
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,7 +55,7 @@ testing =
 	mock
 	flake8-2020
 	virtualenv>=13.0.0
-	pytest-virtualenv>=1.2.7
+	pytest-virtualenv @ git+https://github.com/jaraco/pytest-plugins@distutils-deprecated#subdirectory=pytest-virtualenv
 	wheel
 	paver
 	pip>=19.1 # For proper file:// URLs support.
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ testing =
 	wheel
 	paver
 	pip>=19.1 # For proper file:// URLs support.
-	jaraco.envs
+	jaraco.envs>=2.2
 	pytest-xdist
 	sphinx
 	jaraco.path>=3.2.0
--- a/setuptools/tests/test_distutils_adoption.py
+++ b/setuptools/tests/test_distutils_adoption.py
@@ -9,6 +9,9 @@ import jaraco.envs
 import path
 
 
+IS_PYPY = '__pypy__' in sys.builtin_module_names
+
+
 class VirtualEnv(jaraco.envs.VirtualEnv):
     name = '.env'
     # Some version of PyPy will import distutils on startup, implicitly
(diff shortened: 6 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -15,7 +18,10 @@ jobs:
         - ubuntu-latest
         - macos-latest
         - windows-latest
+      fail-fast: false
```
jaraco, line 21: I'm guessing this change is an artifact of working through the changes? Is there a reason to keep this change going forward?

tox.ini
```diff
@@ -7,14 +7,20 @@ toxworkdir={env:TOX_WORK_DIR:.tox}
...
 	# workaround for sphinx-doc/sphinx#9562
 	# TODO: remove after Sphinx>4.1.2 is available.
 	sphinx@git+https://github.com/sphinx-doc/sphinx; python_version>="3.10"
 commands =
-	pytest {posargs}
+	# We invoke pytest like this to avoid the warning:
+	#   UserWarning: Distutils was imported before Setuptools,
+	#   but importing Setuptools also replaces the `distutils`
+	#   module in `sys.modules`.
+	pytest -c 'import setuptools; import pytest; pytest.main("{posargs}".split())'
```
jaraco, line 19: I'm uneasy about this change. I'd like to know more about why distutils is imported in pytest and if it's something that can be fixed.

jaraco (approved):
Looks great. Just a couple of questions.

.github/workflows/main.yml
mkoeppe: In my test, some of the `pypy` tests failed early for unrelated reasons and immediately canceled the runs that I was interested in.

I would suggest to use `fail-fast: false` - all test runs are pretty short, so it's unlikely that this is problematic

tox.ini
mkoeppe: +1 if you can find a better solution

jaraco: Here's the error I see:

```
setuptools github_workflows_SETUPTOOLS_USE_DISTUTILS $ $SETUPTOOLS_USE_DISTUTILS='local' ./.tox/python/bin/pytest
  File "/Users/jaraco/code/public/pypa/setuptools/./.tox/python/bin/pytest", line 8, in <module>
    sys.exit(console_main())
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 185, in console_main
    code = main()
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 143, in main
    config = _prepareconfig(args, plugins)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 318, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_callers.py", line 55, in _multicall
    gen.send(outcome)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/helpconfig.py", line 100, in pytest_cmdline_parse
    config: Config = outcome.get_result()
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1003, in pytest_cmdline_parse
    self.parse(args)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1283, in parse
    self._preparse(args, addopts=addopts)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1191, in _preparse
    self.hook.pytest_load_initial_conftests(
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
    return outcome.get_result()
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
    res = hook_impl.function(*args)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 1070, in pytest_load_initial_conftests
    self.pluginmanager._set_initial_conftests(early_config.known_args_namespace)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 503, in _set_initial_conftests
    self._try_load_conftest(current, namespace.importmode)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 508, in _try_load_conftest
    self._getconftestmodules(anchor, importmode)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 536, in _getconftestmodules
    mod = self._importconftest(conftestpath, importmode)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 587, in _importconftest
    self.consider_conftest(mod)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 663, in consider_conftest
    self.register(conftestmodule, name=conftestmodule.__file__)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 443, in register
    self.consider_module(plugin)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 669, in consider_module
    self._import_plugin_specs(getattr(mod, "pytest_plugins", []))
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 676, in _import_plugin_specs
    self.import_plugin(import_spec)
  File "/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/_pytest/config/__init__.py", line 703, in import_plugin
    __import__(importspec)
  File "/Users/jaraco/code/public/pypa/setuptools/setuptools/__init__.py", line 9, in <module>
    import _distutils_hack.override  # noqa: F401
  File "/Users/jaraco/code/public/pypa/setuptools/_distutils_hack/override.py", line 1, in <module>
    __import__('_distutils_hack').do_override()
  File "/Users/jaraco/code/public/pypa/setuptools/_distutils_hack/__init__.py", line 68, in do_override
    warn_distutils_present()
  File "/Users/jaraco/code/public/pypa/setuptools/_distutils_hack/__init__.py", line 23, in warn_distutils_present
    warnings.warn(
UserWarning: Distutils was imported before Setuptools, but importing Setuptools also replaces the `distutils` module in `sys.modules`. This may lead to undesirable behaviors or errors. To avoid these issues, avoid using distutils directly, ensure that setuptools is installed in the traditional way (e.g. not an editable install), and/or make sure that setuptools is always imported before distutils.
```

mkoeppe: Yes, it looked like this without the workaround

jaraco: It doesn't appear that simply importing pytest is the culprit:

```
setuptools github_workflows_SETUPTOOLS_USE_DISTUTILS $ cat .tox/python/bin/pytest
#!/Users/jaraco/code/public/pypa/setuptools/.tox/python/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from pytest import console_main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(console_main())
setuptools github_workflows_SETUPTOOLS_USE_DISTUTILS $ .tox/python/bin/python -c "import pytest, sys; print('distutils' in sys.modules)"
False
```

jaraco: I don't see anything in the pytest code that imports distutils, so I'm guessing there's another pytest plugin that's getting loaded that imports distutils before importing setuptools.

jaraco: Looks like the culprit is virtualenv:

```
>>> import importlib.metadata as md
>>> for ep in md.entry_points(group='pytest11'):
...   ep.load()
...   print({'distutils'} & set(sys.modules))
... 
<module 'pytest_flake8' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_flake8.py'>
set()
<module 'pytest_enabler' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_enabler/__init__.py'>
set()
<module 'pytest_checkdocs' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_checkdocs/__init__.py'>
set()
<module 'pytest_black' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_black.py'>
set()
<module 'pytest_mypy' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_mypy.py'>
set()
<module 'pytest_shutil.workspace' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_shutil/workspace.py'>
set()
<module 'xdist.plugin' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/xdist/plugin.py'>
set()
<module 'xdist.looponfail' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/xdist/looponfail.py'>
set()
<module 'pytest_virtualenv' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_virtualenv.py'>
{'distutils'}
<module 'pytest_forked' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_forked/__init__.py'>
{'distutils'}
<module 'pytest_cov.plugin' from '/Users/jaraco/code/public/pypa/setuptools/.tox/python/lib/python3.10/site-packages/pytest_cov/plugin.py'>
{'distutils'}
```

jaraco: [confirmed](https://github.com/man-group/pytest-plugins/blob/2929621566d6510849878d913da6509635e666e1/pytest-virtualenv/pytest_virtualenv.py#L5-L8)

jaraco: Confirmed that with https://github.com/man-group/pytest-plugins/pull/189, the problem goes away.

.github/workflows/main.yml
jaraco: I'd prefer to avoid introducing variance from best practices where possible, and I presume fail-fast is a best practice as it's the default and the behavior deployed in all my other repos. So I'll omit it for now. If there's a good reason to introduce it, let's do that separately.

jaraco: Fixed in ec1c3f34c.

jaraco:
Well that's interesting. Tests are now failing on Python 3.10 for `stdlib` because Setuptools itself imports `distutils.dist`, which imports `distutils.command.build_scripts` which imports `distutils.sysconfig`, which emits a deprecation warning. It seems that deprecation warning had been masked by the pre-test invocation of it in pytest-virtualenv. Interestingly, the failure isn't triggered on macOS and Python 3.10. I don't know why. I believe the best thing to do here is acknowledge and ignore that deprecation warning.

jaraco:
It does look like PyPy is also, separately, importing `distutils` early, breaking the `local` tests.

jaraco:
It looks like those last two commits aren't having any effect, probably because they take effect before the warnings are applied in the pytest.ini. I need to find a hook that applies them after the warnings filters are applied in pytest.ini.

jaraco:
I see now that pytest configures the warnings for each item, loading from the ini file settings each time. So it seems it may not be possible to add contextual warnings (only for Python 3.10 or PyPy) (https://github.com/pytest-dev/pytest/blob/5d87a27434dada56ba69ef7b5fa014c321dc46f4/src/_pytest/warnings.py#L52).

jaraco:
There's still a lingering error on DISTUTILS=local for PyPy. Pip is getting a BackendInvalid exception from pep517. Unfortunately, it masks all the details, so it's difficult to determine anything about why it's invalid.

jaraco:
I've found another use-case that fails on PyPy. If you set TOX_WORK_DIR to `/tox` (or anywhere outside the source tree), simply running `tox` will fail if `SETUPTOOLS_USE_DISTUTILS=local`, because a BackendInvalid will be raised when tox attempts to replace setuptools in the virtualenv, but PyPy has already imported it from the old install. It apparently doesn't happen when `TOX_WORK_DIR` isn't set, presumably because the virtualenv exists somewhere in the source tree (`./.tox`).

jaraco:
I realize now, the main reason why I didn't implement the tests this way in the first place is because it multiplies the test suite by 2x when most of the tests are irrelevant to the difference. That is, the intention was for all tests to run under `SETUPTOOLS_USE_DISTUTILS=local` by default and for `SETUPTOOLS_USE_DISTUTILS=stdlib` to be stable but deprecated. And then there is the `distutils_adoption` tests that check the main assumptions about the adoption differences. Unfortunately, it turned out that due to monkeypatching in Fedora and Debian and others, it's not been possible to make `SETUPTOOLS_USE_DISTUTILS=local` the default, so it's worthwhile having these tests around duplicated for now.

jaraco:
It's so close to passing. It's failing now on PyPy on macOS, one test failure that I've not seen before.

jaraco:
The remaining failure is in `test_distutils_local`, which previously was xfailed on PyPy. I saw it passing on Unix, so thought it could be allowed again, but alas, it seems not. My guess is that distutils gets imported before .pth files are processed, so the precedence doesn't take place.

mkoeppe:
Looking great, thanks for all your work!

webknjaz:
@jaraco did this break the CI?

jaraco:
I don't think it did. I'm pretty sure it was https://github.com/pypa/pip/pull/10641, presumably fixed in f2de34767 / #2874.

abravalheri:
I think there might be a problem with using:

```diff
[options.extras_require]
testing =
        ...
-	pytest-virtualenv>=1.2.7
+	pytest-virtualenv @ git+https://github.com/jaraco/pytest-plugins@distutils-deprecated#subdirectory=pytest-virtualenv
```

in `setup.cfg`.

I am almost sure PyPI forbid uploading packages with dependencies outside of PyPI.
`pypa/build` may also have problems when parsing the dependencies for PEP517 style builds...

jaraco:
Confirmed. Good catch. I can see in https://github.com/pypa/setuptools/runs/4196605393?check_suite_focus=true this output:

> Error during upload. Retry with the --verbose option for more details.
> HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
> Invalid value for requires_dist. Error: Can't have direct dependency: "pytest-virtualenv @ git+https://github.com/jaraco/pytest-plugins@distutils-deprecated#subdirectory=pytest-virtualenv ; extra == 'testing'"

So the v59 release was rejected.

I'm not sure what to do with this other than to fork the project or port its functionality to another project.

I've noticed that this project (Setuptools) already is using jaraco.envs for some of the tests requiring isolated environments for testing. Perhaps the best thing would be to migrate the tests to jaraco.envs.

webknjaz:
PyPI does not validate the deps. They may be satisfied from third-party sources.

abravalheri:
> PyPI does not validate the deps. They may be satisfied from third-party sources.

@webknjaz I had a problem like this around a month ago with another project. I don't know when PyPI started forbidding distributions that list dependencies via URLs, but they do it now for sure.

You can also check the [CI logs](https://github.com/pypa/setuptools/runs/4196605393?check_suite_focus=true):

image

PR #2866, opened 2021-11-10
Incorporate pypa/distutils@f1b0a2b.

jaraco:
- Fix PyPy3.8 install locations
- distutils/ccompiler.py (CCompiler.has_function): Do not fail if self.outputdir is set
- distutils/tests/test_unixccompiler.py (UnixCCompilerTestCase.test_has_function): New
- .github/workflows/main.yml: Add integration testing with setuptools
- Use simple expression for resolution of outputdir.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,40 @@ jobs:
       - name: Run tests
         run: tox
 
+  ci_setuptools:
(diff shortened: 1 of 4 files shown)


jaraco:
See 77114aede, where I reverted the integration tests from distutils.

PR #2876, opened 2021-11-13
Fixed string list concatenation error

gentlegiantJGC:
I am using [build](https://github.com/pypa/build) to build my project and want to pass custom settings into the setup.py file.

The [suggested way](https://github.com/pypa/build/issues/328) is to use the `config-setting` flag chained with your `global-option` flag however this errors if only one global option is defined because they pass it in as a string not a list of strings.

Your code tries concatenating a list of strings with a single string which causes an error.

## Summary of changes

If `config_settings["--global-option"]` is a string wrap it in a list


jaraco:
Can you provide an example of what usage results in a list and what usage results in a string, and what commands you run to get that behavior? Does setuptools need to accept both forms (list and string)? I believe the `global-option` is poorly-defined, so this might be a good time to evaluate a better way for Setuptools to solicit configuration.

In any case, it would also be nice to have a test for the desired behavior to capture that it's required and why.

gentlegiantJGC:
This command does not work
`python -m build -C--global-option=--find-libs=True`

This command does work
`python -m build -C--global-option=--find-libs=True -C--global-option=--find-libs=True`

I am using the `--config-setting` argument in build (`-C` for short) which is parsing the command into a dictionary.

In the first case that is
```py
{
    "--global-option": "--find-libs=True"
}
```
and in the second case is 
```py
{
    "--global-option": [
        "--find-libs=True",
        "--find-libs=True"
    ]
}
```
If it finds the same argument more than once it will convert it to a list and append subsequent values.

That dictionary gets passed onto setuptools where it pulls out the `--global-option` value which assumes it is a list. That is where the error happens.
https://github.com/pypa/setuptools/blob/main/setuptools/build_meta.py#L213

jaraco:
My god, that syntax is atrocious (`-C--global-option=--find-libs=True`). Do we know where `--global-option` is defined outside of the setuptools implementation?

gentlegiantJGC:
Yeh it isn't great.
Build just takes all the config settings and feeds them to the build system (in this case setuptools). They don't have a flag just for global option because I think that is a thing implemented by setuptools and they work with other build systems afaik.

Build parses the `--config-setting` here https://github.com/pypa/build/blob/main/src/build/__main__.py#L346
Setuptools parses the `--global-option` here https://github.com/pypa/setuptools/blob/main/setuptools/build_meta.py#L213
I don't know if it is used anywhere else or if this was just implemented there.

PR #2923, opened 2021-12-11
Add integration testing with python from Debian, Ubuntu, Fedora, Gentoo, Arch Linux, OpenSUSE, Slackware

mkoeppe:
## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -95,7 +95,7 @@ jobs:
       fail-fast: false
       max-parallel: 32
       matrix:
-        tox_system_factor: [ubuntu-trusty, ubuntu-xenial, ubuntu-bionic, ubuntu-focal, ubuntu-hirsute, ubuntu-impish, ubuntu-jammy, debian-stretch, debian-buster, debian-bullseye, debian-bookworm, debian-sid, linuxmint-17, linuxmint-18, linuxmint-19, linuxmint-19.3, linuxmint-20.1, linuxmint-20.2, linuxmint-20.3, fedora-26, fedora-27, fedora-28, fedora-29, fedora-30, fedora-31, fedora-32, fedora-33, fedora-34, fedora-35, fedora-36, centos-7, centos-8, gentoo-python3.9, archlinux-latest, opensuse-15, opensuse-15.3, opensuse-tumbleweed, slackware-14.2, ubuntu-bionic-i386, manylinux-2_24-i686, debian-buster-i386, centos-7-i386]
+        tox_system_factor: [ubuntu-trusty, ubuntu-xenial, ubuntu-bionic, ubuntu-focal, ubuntu-hirsute, ubuntu-impish, ubuntu-jammy, debian-stretch, debian-buster, debian-bullseye, debian-bookworm, debian-sid, linuxmint-17, linuxmint-18, linuxmint-19, linuxmint-19.3, linuxmint-20.1, linuxmint-20.2, linuxmint-20.3, fedora-26, fedora-27, fedora-28, fedora-29, fedora-30, fedora-31, fedora-32, fedora-33, fedora-34, fedora-35, centos-7, centos-8, gentoo-python3.9, archlinux-latest, opensuse-15, opensuse-15.3, opensuse-tumbleweed, slackware-14.2, ubuntu-bionic-i386, manylinux-2_24-i686, debian-buster-i386, centos-7-i386]
         tox_packages_factor: [minimal, standard]
     env:
       TOX_ENV: docker-${{ matrix.tox_system_factor }}-${{ matrix.tox_packages_factor }}
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -1,0 +1,155 @@
+name: Run Sage CI for Linux
+
+## This GitHub Actions workflow provides:
+##
+##  - portability testing, by building and testing this project on many platforms
+##
+##  - continuous integration, by building and testing other software
+##    that depends on this project.
+##
+## It runs on every pull request and push of a tag to the GitHub repository.
+##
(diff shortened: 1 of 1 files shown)


jaraco:
It looks like some tests are unsuccessful. Are these tests going to be flaky? Will you be available to help maintain these checks if included?

mkoeppe:
The failure in `opensuse-tumbleweed` (https://github.com/pypa/setuptools/runs/4494440976?check_suite_focus=true) looks like a legitimate incompatibility of setuptools with their Python to me.

I'll revise the list of platforms, adding some newer ones and removing flaky ones.

I maintain these things in the Sage project. Occasionally the list of platforms needs to be updated, and I'll be happy to send a PR, but it should otherwise not require much maintenance.

.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,155 @@
...
+## The testing can be monitored in the "Actions" tab of the GitHub repository.
+##
+## After all jobs have finished (or are canceled) and a short delay,
+## tar files of all logs are made available as "build artifacts".
+##
+## This GitHub Actions workflow uses the portability testing framework
+## of SageMath (https://www.sagemath.org/).  For more information, see
+## https://doc.sagemath.org/html/en/developer/portability_testing.html
+
+## The workflow consists of two jobs:
```
webknjaz, line 21: I think it may be better to add these jobs to the existing workflow making it possible to integrate all the jobs in a single branch protection check job depending on the required jobs.

mkoeppe: That could also be done via https://docs.github.com/en/actions/using-workflows/reusing-workflows

mkoeppe: I have a preference for `ci-sage.yml` to be a separate file because I have provided versions of it for various projects, so it's easier for me to keep it up to date.

webknjaz: Yep, reusable workflows are a good idea except they don't work well within the same repository (because you cannot refer to the same commit when including them).

mkoeppe: I didn't know that, thanks

webknjaz: I don't have access to this repo like Jason does :) But I still occasionally review things to help out.

As a recommendation to you personally — you could create a repository dedicated to this reusable workflow and parameterize it. This way you could contribute it as reusable to many projects but only the list of the matrix targets would need to be maintained on this side. Then, you'd occasionally update the SHA-1 or the tag in repos where contribute, instead of having to synchronize the whole file.

mkoeppe: Thanks, that's a good idea. Noted at https://trac.sagemath.org/ticket/29060

jaraco: I slightly prefer the separate workflow because it provides separation based on custody/origin. Good suggestions on limiting the amount of copy/synchronization. Since this project doesn't use protections, let's go with this proposal for now and consider merging separately.

abravalheri:
Hi @mkoeppe, thank you very much for working in improving the tests. It is an extremely important (and difficult) thing to do!

I noticed that the introduced workflow is [failing](https://github.com/pypa/setuptools/runs/5066567338?check_suite_focus=true) recently with some weird error messages:

> [The short ebuild name "tox" is ambiguous. Please specify](https://github.com/pypa/setuptools/runs/5066567539?check_suite_focus=true#step:8:82)

> [/sage/build/bin/sage-bootstrap-python: error: none of python python3 python3.10 python3.9 python3.8 python3.7 python2.7 python3.6 python2 is a suitable Python](https://github.com/pypa/setuptools/runs/5066567338?check_suite_focus=true#step:8:92)

> [Bootstrap failed. Either upgrade autotools and gettext; or run bootstrap with the -d option to download the auto-generated files instead.](https://github.com/pypa/setuptools/runs/5066567338?check_suite_focus=true#step:8:117)

The intriguing part is that this happens for a PR that has no code changes, it just modifies the configuration for pytest: #3080 (i.e., it expands the amount of warnings being ignored, it is not a big configuration change either...)

Is there any chance you have ever seen these error messages before or have any information about these failures? I am not very familiar with the `sage` tools and the CI steps you have set up, so I am kind of lost 😅 on what can be done to fix these problems. Any help would be very appreciated.

mkoeppe:
The Sage portability CI provisions installations of system packages based on its database that maps packages to their names in various distributions. The failure for `gentoo-python3.9-standard` comes from a flaw in this mapping, for which we already have a fix in https://trac.sagemath.org/ticket/33269; once this is merged into our `develop` branch, this will no longer fail in this GH Actions workflow.

It appears that `centos-8` can no longer be tested (it is past EOL and it looks from the log like the packages are now no longer available). 
This distribution name can just be removed from `ci-sage.yml`. I can send a PR if nobody beats me to it.

As replacements we have `centos-stream-8`, `centos-stream-9` (https://trac.sagemath.org/ticket/33196), which can be added as soon as this is merged into our `develop` branch.

mkoeppe:
OK, I have opened #3081 for the update

abravalheri:
Thank you very much @mkoeppe for having a look on this and providing a PR!

di:
@jaraco I think the large # of jobs this introduces to setuptools' GitHub workflows is slowing down builds for other PyPA projects. There is a [limit of 20 concurrent jobs](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) org-wide for our plan, and setuptools has gone from 36 jobs in #3078 to 119 jobs here. I've noticed a significant backlog over at https://github.com/pypa/warehouse/actions over the last few days and suspect this might be why.

Any thoughts on how we could get this down to a more reasonable number of concurrent jobs per workflow run?

(cc @ewdurbin)

PR #2934, opened 2021-12-19
Render `distutils` commit links properly in documentation

### LINKED ISSUE #2884: [Docs] GitHub reference in changelog rendered as email
### Summary

In [the changelog for v59.0.0](https://setuptools.pypa.io/en/latest/history.html#v59-0-0), a `distutils` commit is referenced using `pypa/distutils@f1b0a2b` in the RST source. Although it should link to pypa/distutils@f1b0a2b (`https://github.com/pypa/distutils/commit/f1b0a2b`) as GitHub automatically does, it instead renders the URL as `mailto:pypa/distutils@f1b0a2b`, which is incorrect.

### OS / Environment

N/a. This is in the source code.

### Additional Information

I would solve this myself, but I don't know the best solution. I can see that some things like GitHub issues are automatically linked, but I'm not sure if it should be added to the linking code in [`docs/conf.py`](https://github.com/pypa/setuptools/blob/4b980ef4072a817aae0da3643d0fa70c30fcb6cf/docs/conf.py) or just manually made a link.

### Code of Conduct

aphedges:
## Summary of changes

There are seven references to `distutils` commits that Sphinx renders as email addresses. I have added a rule to convert these commits into the proper GitHub URLs. I opted to explicitly encode `distutils` into the pattern because it is the only repository for which the documentation references commits. I felt that trying something more general would be more error prone, and this implementation at least solves the current problem. It might be worth revisiting if this documentation links to commits from other repositories.

Closes #2884

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,6 +69,10 @@ link_files = {
                 url='{GH}/pypa/distutils/issues/{distutils}',
             ),
             dict(
+                pattern=r'pypa/distutils@(?P<distutils_commit>[\da-f]+)',
+                url='{GH}/pypa/distutils/commit/{distutils_commit}',
+            ),
+            dict(
                 pattern=r'^(?m)((?P<scm_version>v?\d+(\.\d+){1,2}))\n[-=]+\n',
                 with_scm='{text}\n{rev[timestamp]:%d %b %Y}\n',
             ),


jaraco:
Love it. Thanks!

PR #2953, opened 2021-12-22
Fixed a bug that easy install incorrectly parsed Python 3.10 version string

liuzhe-lz:
## Summary of changes

Fixed a bug that easy install incorrectly parsed Python 3.10 version string.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
Looks like this bug was reported in #3001.

Can you add a test?

liuzhe-lz:
Sorry I don't have time for now.
Maybe you can help add one. Or I can do it sometime next week.

> Looks like this bug was reported in #3001.
> 
> Can you add a test?

Sorry I don't have time for now.
Maybe you can help add one. Or I can do it sometime next week.

abravalheri:
As the existing tests are not running properly, I think the best here would be an update, So here it goes...

@Mergifyio update

abravalheri:
Hi @liuzhe-lz, @jaraco I added some tests to the proposed fix. Please feel free to remove it, change or add any other tests.

(I also reordered a bit the commits, because I wanted to see the test failing before the fix is applied and study how to better represent the situation in #3001).

---

Notes:

I noticed the `stdlib` version of distutils sometimes create a path that contains a literal string `'None'`, which seems to be a bug already solved in the vendored version.

Another thing that I noticed is that for Windows 3.10, there might be a `ValueError: invalid variable 'py_version_nodot_plat'`, which I guess is already solved by https://github.com/pypa/distutils/commit/1560a1fe4e56e00392d3658a59e6fc7238e26a8a and will be eventually pulled into `setuptools`.

jaraco:
Thanks everyone!

PR #2962, opened 2021-12-23
distutils shim should ignore setuptools on another path

### LINKED ISSUE #2957: [BUG] DistutilsMetaFinder breaks distutils imports when old setuptools is higher on path
### setuptools version

setuptools==60.0.4

### Python version

python3.8

### OS

RHEL8, FreeBSD12/13, Ubuntu 20.04

Any environment where an old setuptools (eg OS package) is higher in `sys.path`, but a newer setuptools is installed anywhere on `sys.path` (with its distutils shimming `.pth` file active)

### Description

Any import of `distutils` will bomb when a new setuptools is present, but an older setuptools is higher on the path (since the new setuptools' path shim blindly tries to import from `setuptools._distutils`, which doesn't exist in old versions of setuptools). This has probably been the case ever since the distutils shim was created, but now that it's active by default, it's a bigger problem.

The place we hit this was running `pip install` from a venv with `--system-site-packages`- pip's build isolation installs latest `setuptools`, but then shuffles `sys.path` around such that the ancient OS-packaged `setuptools` is higher on the path. The first thing that tries to import `distutils` invokes the metapath finder, which has no validation or recovery and just lets the `ModuleNotFoundError` fly.

At first blush, it seems like the metapath shim's `spec_for_distutils` should try to at least verify the co-location of `setuptools` with the shim impl (ie, that they're under the same path prefix) and/or that the `_distutils` subpackage is available via `find_spec` before it actually imports it. Mixing and matching the shim version with random setuptools versions seems like a recipe for trouble down the road...

I could probably throw together a PR for this if there's a consensus that this approach is sane.

### Expected behavior

The presence of a new-but-not-loadable setuptools on sys.path doesn't break old setuptools/distutils.

### How to Reproduce

(NB: this is a much simpler repro than the OS installed setuptools that's actually failing, but illustrates the problem more easily on any OS)
```
python3 -m venv /tmp/bang && source /tmp/bang/bin/activate  # give us a clean venv to work in
python -m pip install 'setuptools>60'  # install new setuptools to venv site-packages
python -m pip install 'setuptools<45' -t .  # install old setuptools to cwd- higher than site-packages on path for most systems
python -c 'import distutils'
```

### Output

```console
...
  File "/tmp/bang/lib64/python3.9/site-packages/_distutils_hack/__init__.py", line 92, in create_module
    return importlib.import_module('setuptools._distutils')
  File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/home/mdavis/setuptools/__init__.py", line 6, in <module>
    import distutils.core
  File "/tmp/bang/lib64/python3.9/site-packages/_distutils_hack/__init__.py", line 92, in create_module
    return importlib.import_module('setuptools._distutils')
  File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named 'setuptools._distutils'

```

### Code of Conduct

nitzmahone:
## Summary of changes
Prevents `_distutils_hack` finder shim from interacting with a possibly incompatible setuptools higher on sys.path

Closes #2957

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -58,7 +58,9 @@ def ensure_local_distutils():
 
     # check that submodules load as expected
     core = importlib.import_module('distutils.core')
-    assert '_distutils' in core.__file__, core.__file__
+    # FIXME: this assertion blows up if the MetaFinder below has no-opped on a
+    #  setuptools from another path
+    # assert '_distutils' in core.__file__, core.__file__
 
 
 def do_override():
@@ -85,6 +87,15 @@ class DistutilsMetaFinder:
     def spec_for_distutils(self):
         import importlib.abc
         import importlib.util
+        from pathlib import Path
+
+        st_mod = importlib.import_module('setuptools')
+
+        # prevent this import redirection shim from interacting with a possibly
+        # incompatible setuptools in another location on sys.path
+        # see pypa/setuptools#2957
+        if not Path(__file__).parents[1].samefile(Path(st_mod.__file__).parents[1]):
+            return None
 
         class DistutilsLoader(importlib.abc.Loader):
 
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -85,20 +85,18 @@ class DistutilsMetaFinder:
     def spec_for_distutils(self):
         import importlib.abc
         import importlib.util
-        from pathlib import Path
 
-        st_mod = importlib.import_module('setuptools')
-
-        # prevent this import redirection shim from interacting with a possibly
-        # incompatible setuptools in another location on sys.path
-        # see pypa/setuptools#2957
-        if not Path(__file__).parents[1].samefile(Path(st_mod.__file__).parents[1]):
+        try:
+            mod = importlib.import_module('setuptools._distutils')
+        except Exception:
+            # an older Setuptools without a local distutils is taking
+            # precedence, so fall back to stdlib. Ref #2957.
             return None
 
         class DistutilsLoader(importlib.abc.Loader):
 
             def create_module(self, spec):
-                return importlib.import_module('setuptools._distutils')
+                return mod
 
             def exec_module(self, module):
                 pass
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -58,9 +58,7 @@ def ensure_local_distutils():
 
     # check that submodules load as expected
     core = importlib.import_module('distutils.core')
-    # FIXME: this assertion blows up if the MetaFinder below has no-opped on a
-    #  setuptools from another path
-    # assert '_distutils' in core.__file__, core.__file__
+    assert '_distutils' in core.__file__, core.__file__
 
 
 def do_override():
--- a/changelog.d/2962.misc.rst
+++ b/changelog.d/2962.misc.rst
@@ -1,0 +1,1 @@
+Avoid attempting to use local distutils when the presiding version of Setuptools on the path doesn't have one.


nitzmahone:
I actually found a bunch of other potential problems while trying to get this working. Bigger picture- a lot of the problem cases are due to the `.pth` shim hack being in a separate top-level package than setuptools itself. Is there a reason the `.pth` can't be under the main `setuptools` top-level package? (I got nailed with similar problems on pyyaml when we moved the optional `_yaml` native extension top-level package into a subpackage, making the old top-level package visible from lower entries on sys.path)

jaraco:
> Is there a reason the .pth can't be under the main setuptools top-level package?

Yes. `.pth` files are only processed if they appear directly in site directories (as registered by `site.addsitedir`), usually a small subset of paths on sys.path.

jaraco:
Maybe a better approach would be to update the DistutilsLoader to always ensure it imports `setuptools._distutils` from the intended path (same as the .pth file).

nitzmahone:
> Yes. .pth files are only processed if they appear directly in site directories (as registered by site.addsitedir), usually a small subset of paths on sys.path.

Ah of course- I forgot how the actual mechanism functioned under the covers... 

> update the DistutilsLoader to always ensure it imports setuptools._distutils from the intended path (same as the .pth file).

I'm not sure that'd cover all the problem cases- unless I'm misunderstanding your approach, that'd open up some strange mismatch possibilities where you're directly importing content from packages that might end up being shadowed on sys.path (since IIRC, each .pth from a registered site dir is going to run as sys.path is being built up, but we don't know until the user actually imports setuptools or distutils which site dir is going to "win").

I hate to even suggest this, but it seems like the only way to account for not knowing which setuptools will "win" in the face of early startup sys.path shenanigans by users/tools is to have each .pth install a dynamically-typed shim- ie, each .pth directly defines and installs its own shim type that will only service its associated setuptools instance. So if there were 10 different installs of setuptools visible (and thus 10 .pth's that ran at startup), there'd actually be 10 shims of different types (that didn't import anything from setuptools/_distutils_hack packages to avoid chicken/egg shadowing problems), and only one should ever actually "work". The one shim that actually does its work on the first user import of setuptools/distutils should theoretically be able to clean the others up off the metapath. I think that model would also get rid of the need for the separate `_distutils_hack` package altogether (and problems from its cross-path visibility), at the expense of potentially needing to duplicate some of the shim code that the setuptools package init uses directly. IIUC the reason for the separate `_distutils_hack` package is to avoid the .pth invoking any of the setuptools package init too early- this would preserve that as well.

I can probably throw together a prototype of this if you want a strawman to poke at- especially with the shimming being on by default now (which, side note/other bug: it looks like it isn't for the .pth part right now- still relies on the envvar alone), it seems like the scope of this problem and/or potential future incompatibilities with mismatched shim versions + setuptools versions is only going to get bigger.

_distutils_hack/__init__.py
```diff
@@ -58,7 +58,9 @@ def ensure_local_distutils():
 
     # check that submodules load as expected
     core = importlib.import_module('distutils.core')
-    assert '_distutils' in core.__file__, core.__file__
+    # FIXME: this assertion blows up if the MetaFinder below has no-opped on a
+    #  setuptools from another path
+    # assert '_distutils' in core.__file__, core.__file__
```
jaraco, line 63: I'm trying to understand why this check is a problem. `ensure_local_distutils` is only called by `do_override`, which is only called by importing setuptools. But if an old version of setuptools is earlier on sys.path, it will not invoke the override.

jaraco: In 41f61f3cc, I've removed the commented code and restored the expectation to see what breaks.

jaraco: Aha. I see now (in the failed RTD build). Because the check for "matching setuptools" was very particular, and because another copy of Setuptools was taking precedence, _and_ because local checkouts don't get the `.pth` treatment, the shim wasn't getting loaded for any invocation (I think).

In da8c68c49, I'm exploring allowing any setuptools to satisfy the distutils requirement.

jaraco:
> since IIRC, each .pth from a registered site dir is going to run as sys.path is being built up, but we don't know until the user actually imports setuptools or distutils which site dir is going to "win"

In [my investigation](https://github.com/pypa/setuptools/issues/2957#issuecomment-1002653724), I found that site paths are always appended, so the first .pth file to run should take precedence... and any site dirs added later should be shadowed by the first. The only concern then is if an older setuptools without the shim appears earlier on sys.path, and in that case, it should be safe to disable the shim, since that version of setuptools doesn't expect a local distutils.

jaraco:
I'm liking this implementation better, for its simplicity and reliance on import semantics. Let's see if it solves the primary issue(s) and we can evaluate if the other [identified issues](https://github.com/pypa/setuptools/pull/2962#issuecomment-1000497527) justify further accommodation.

PR #2964, opened 2021-12-24
WIP direct module load shims

### LINKED ISSUE #2957: [BUG] DistutilsMetaFinder breaks distutils imports when old setuptools is higher on path

nitzmahone:
## Summary of changes
* .pth uses direct shim module load from sibling setuptools dir instead of import
* import shims attempt to remove themselves/others when it's clear they're not needed (eg, setuptools path mismatch from .pth)
* removes _distutils_hack top-level package

Closes #2957

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -5,7 +5,8 @@ import functools
 import os
 import re
 
-import _distutils_hack.override  # noqa: F401
+from ._distutils_shim import do_override  # noqa: F401
+do_override()
 
 import distutils.core
 from distutils.errors import DistutilsOptionError
--- a/setuptools/_distutils_shim.py
+++ b/setuptools/_distutils_shim.py
@@ -74,6 +74,9 @@ def do_override():
 
 
 class DistutilsMetaFinder:
+    # for forensic purposes, since we may not be able to get at the module object easily
+    _fromfile = __file__
+
     def find_spec(self, fullname, path, target=None):
         if path is not None:
             return
@@ -86,8 +89,16 @@ class DistutilsMetaFinder:
         import importlib.abc
         import importlib.util
 
-        class DistutilsLoader(importlib.abc.Loader):
+        st = importlib.import_module('setuptools')
+        if os.path.dirname(st.__file__) != os.path.dirname(__file__):
+            # no-op; the setuptools we imported was from a foreign site-packages dir, remove self from meta_path
+            remove_shim()
+            return None
+        else:
+            # we're a sibling of the setuptools that actually got imported; kill any other shims
+            remove_foreign_shims()
 
+        class DistutilsLoader(importlib.abc.Loader):
             def create_module(self, spec):
                 return importlib.import_module('setuptools._distutils')
 
@@ -130,7 +141,8 @@ DISTUTILS_FINDER = DistutilsMetaFinder()
 
 
 def ensure_shim():
-    DISTUTILS_FINDER in sys.meta_path or add_shim()
+    if not (any(s for s in sys.meta_path if type(s).__name__ == 'DistutilsMetaFinder' and getattr(s, '_fromfile', None) == DistutilsMetaFinder._fromfile)):
+        add_shim()
 
 
 @contextlib.contextmanager
@@ -150,4 +162,16 @@ def remove_shim():
     try:
         sys.meta_path.remove(DISTUTILS_FINDER)
     except ValueError:
+        pass
+
+
+def remove_foreign_shims():
+    """
+    Remove setuptools metapath shims that might've been installed by other installations on sys.path
+    """
+    try:
+        foreign_shims = [l for l in sys.meta_path if type(l).__name__ == 'DistutilsMetaFinder' and l is not DISTUTILS_FINDER]
+        for shim in foreign_shims:
+            sys.meta_path.remove(shim)
+    except Exception:
         pass
--- a/_distutils_hack/override.py
+++ b/_distutils_hack/override.py
@@ -1,1 +1,0 @@
-__import__('_distutils_hack').do_override()
--- a/setup.py
+++ b/setup.py
@@ -51,11 +51,15 @@ class install_with_pth(install):
     """
 
     _pth_name = 'distutils-precedence'
+    # NB: this abuses the fact that site.py directly execs this with a `sitedir` local that contains the current sitedir it's working on; __file__ is no good here
     _pth_contents = textwrap.dedent("""
         import os
+        from importlib.util import spec_from_file_location, module_from_spec
         var = 'SETUPTOOLS_USE_DISTUTILS'
         enabled = os.environ.get(var, 'local') == 'local'
-        enabled and __import__('_distutils_hack').ensure_shim()
+        mod = module_from_spec(spec_from_file_location('_shimmod', os.path.join(sitedir, 'setuptools', '_distutils_shim.py'))) if enabled else None
+        mod and mod.__spec__.loader.exec_module(mod)
+        mod and mod.ensure_shim()
         """).lstrip().replace('\n', '; ')
 
     def initialize_options(self):


setuptools/__init__.py
```diff
@@ -5,7 +5,8 @@
 import os
 import re
 
-import _distutils_hack.override  # noqa: F401
+from ._distutils_shim import do_override  # noqa: F401
+do_override()
```
jaraco, line 9: Introducing imperative code ahead of imports is going to cause flake8 errors on all the subsequent imports. I'd suggest to just keep the previous technique where the shim is a package and a module of the package can be imported to invoke the behavior. i.e. `mv _distutils_hack setuptools/_distutils_shim`

setuptools/_distutils_shim.py
```diff
@@ -74,6 +74,9 @@ def do_override():
 
 
 class DistutilsMetaFinder:
+    # for forensic purposes, since we may not be able to get at the module object easily
+    _fromfile = __file__
```
jaraco, line 78: I believe accessing `__file__` may cause problems under a zipimport scenario. That's something we'd want to verify before accepting.

jaraco (changes requested):
My biggest concern about this change is moving the hack into the setuptools namespace. I would have liked to have this package in some shared namespace, but the fact of the matter is:

- importing setuptools invokes a lot of behavior, including [monkeypatching distutils](https://github.com/pypa/setuptools/blob/db37ab088bc826bdd407b141d68e4aab13771dda/setuptools/__init__.py#L242).
- importing setuptools imports pkg_resources, which invokes a lot more behavior and is known to be slow.

It's simply not acceptable to force every environment to import setuptools on startup. Therefore, the shim needs not to import pkg_resources or setuptools in order to install the shim. The import of either should be deferred until distutils is imported.

nitzmahone:
> My biggest concern about this change is moving the hack into the setuptools namespace
It's very explicitly *not* importing setuptools or pkg_resources- merely direct loading the module code for the shim from the sibling setuptools. Try it: 
```
>>> import sys
>>> sys.meta_path
[<_shimmod.DistutilsMetaFinder object at 0x7f053265e790>, <_virtualenv._Finder object at 0x7f05327dcf70>, <class '_frozen_importlib.BuiltinImporter'>, <class '_frozen_importlib.FrozenImporter'>, <class '_frozen_importlib_external.PathFinder'>]
>>> any(m for m in sys.modules.keys() if m in ['setuptools', 'distutils', 'pkg_resources'])
False
```

It's using the assumption that the `setuptools` package that's a sibling to the .pth in the filesystem is the only one that it should be allowed to service. I haven't played with the zip side of things, but assuming it works the way I expect, it could probably be located with a `find_spec` without tripping the package init, though I'm not sure what assumptions could be made about co-location in that case...

jaraco:
In 9c9c91c0a3951a7c521bd6eddc3aa5f84f0cfd9f, I've applied a similar patch to address a separate concern (race between .pth handling and import invocation).  We'll need to resolve the conflict.

> It's very explicitly _not_ importing setuptools or pkg_resources- merely direct loading the module code for the shim from the sibling setuptools structure.

Oh, nice. I'd missed that nuance. I like that behavior.

I do regret the added complexity to the implementation. 

> import shims attempt to remove themselves/others when it's clear they're not needed

I wonder if this is a premature optimization... Can we consider this change separately if it's not needed to fix the reported issue?

setuptools/_distutils_shim.py
```diff
@@ -130,7 +141,8 @@ def frame_file_is_setup(frame):
 
 
 def ensure_shim():
-    DISTUTILS_FINDER in sys.meta_path or add_shim()
+    if not (any(s for s in sys.meta_path if type(s).__name__ == 'DistutilsMetaFinder' and getattr(s, '_fromfile', None) == DistutilsMetaFinder._fromfile)):
+        add_shim()
```
jaraco, line 145: Oh, this is unfortunate. Because the shim might be invoked independently several times, it has to detect different instances of the same name and origin. This complexity is one argument for keeping _distutils_hack as a top-level module in order to rely on import semantics to resolve exactly one instance of the module/finder per origin.

I'd like to explore an approach that doesn't require touching some many different aspects of the implementation.

jaraco:
Let's try #2962 first and see how that goes.

setuptools/_distutils_shim.py
nitzmahone: That was the point though- since the requirements for the shim are certainly a moving target, the only sure way to get the right shim behavior for any given version of setuptools is to let each setuptools load its own locally, then clean up the others when one actually "fires". It's definitely "icky", but given the restrictions of what you can actually do in a `.pth` file, I can't think of another way that will also accommodate evolving shims.

setuptools/__init__.py
nitzmahone: This one's easy to solve any number of ways since it's a real import (I was just lazy with the PoC), but I think you're already starting to see the problems that the extra top-level package can cause (eg, https://github.com/pypa/setuptools/commit/d0da0142609318e534d5874bfa0c44d1254f9e38)

PR #2974, opened 2021-12-26
Enable logging through Python logging module

### LINKED ISSUE #2023: Recommended practice for logging in plugins?
I'm currently adding a `setuptools.finalize_distribution_options` hook to one of my packages, and I felt it would be appropriate if it emitted some informational messages when invoked through `egg_info`, `sdist`, or `bdist_wheel`.  I thus added several calls to `disutils.log.info()` to the code, but whenever I run `egg_info` etc., the messages are not emitted, even though the rest of the hook function does its thing and even though the output that is emitted appears to also be at `INFO` level.  The messages show up if I output them with `print()`, but that's clearly the wrong thing to do, as then they also show up when the `--quiet` option is given and when running things like `setup.py --version`.

Just how should setuptools plugins perform logging?  Should they even do it at all?

### LINKED ISSUE #2973: Replacement needed for distutils.log
As distutils is deprecated and uses its own logging infrastructure that predates the Python logging module, Setuptools should expose something better for plugins to use. See #2023 and pypa/wheel#428 for background.

jaraco:
- Add setuptools.log to supersede distutils.log. Ref #2973.

## Summary of changes

Closes #2973. Closes #2023.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -18,6 +18,7 @@ from setuptools.extension import Extension
 from setuptools.dist import Distribution
 from setuptools.depends import Require
 from . import monkey
+from . import logging
 
 
 __all__ = [
@@ -149,6 +150,7 @@ def _install_setup_requires(attrs):
 
 def setup(**attrs):
     # Make sure we have any requirements needed to interpret 'attrs'.
+    logging.configure()
     _install_setup_requires(attrs)
     return distutils.core.setup(**attrs)
 
--- a/setuptools/logging.py
+++ b/setuptools/logging.py
@@ -1,5 +1,7 @@
 import sys
 import logging
+import distutils.log
+from . import monkey
 
 
 def _not_warning(record):
@@ -20,3 +22,9 @@ def configure():
     handlers = err_handler, out_handler
     logging.basicConfig(
         format="{message}", style='{', handlers=handlers, level=logging.DEBUG)
+    monkey.patch_func(set_threshold, distutils.log, 'set_threshold')
+
+
+def set_threshold(level):
+    logging.root.setLevel(level*10)
+    return set_threshold.unpatched(level)
--- a/setuptools/logging.py
+++ b/setuptools/logging.py
@@ -1,0 +1,22 @@
+import sys
+import logging
+
+
+def _not_warning(record):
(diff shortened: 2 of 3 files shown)


setuptools/__init__.py
```diff
@@ -149,6 +150,7 @@ def finalize_options(self):
 
 def setup(**attrs):
     # Make sure we have any requirements needed to interpret 'attrs'.
+    logging.configure()
```
abravalheri, line 153: I wonder if this function should also be called from the 'build_meta'... (in theory for PEP517-style buils that is the first piece of setuptools coda that run, right?)

abravalheri (commented):
That looks like a very nice move!

The CI errors seem to be just a hiccup, and pass on the [second run](https://github.com/pypa/setuptools/actions/runs/1624930237p)

setuptools/__init__.py
jaraco: Good tip. I looked briefly and wasn't sure where to invoke that behavior. But moreover, I was concerned that adding it in build_meta would cause it to be called twice (once when build_meta is setup and again when setup.py is executed by build_meta). My instinct was that this hook would happen exactly once, and if a future refactor went to remove setup.py, this call could be moved then.  I also note that for distutils, logging didn't get setup until somewhere around Distribution.finalize_options, so this configuration will happen even earlier than for distutils. It might be nice for it to happen even earlier, so it could affect code in setup.py prior to invoking setup(), but I'm okay with this for now.

setuptools/logging.py
```diff
@@ -0,0 +1,30 @@
...
+    for compatibilty with distutils.log but may change in
+    the future.
+    """
+    err_handler = logging.StreamHandler()
+    err_handler.setLevel(logging.WARNING)
+    out_handler = logging.StreamHandler(sys.stdout)
+    out_handler.addFilter(_not_warning)
+    handlers = err_handler, out_handler
+    logging.basicConfig(
+        format="{message}", style='{', handlers=handlers, level=logging.DEBUG)
```
dolfinus, line 24: Why default level is DEBUG instead of INFO or WARNING?

PR #2990, opened 2021-12-31
set origin= for distutils.__spec__

asottile:
set origin so spec finding reports the correct location -- https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec

## Summary of changes

Closes https://github.com/asottile/reorder_python_imports/issues/210

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -118,7 +118,9 @@ class DistutilsMetaFinder:
             def exec_module(self, module):
                 pass
 
-        return importlib.util.spec_from_loader('distutils', DistutilsLoader())
+        return importlib.util.spec_from_loader(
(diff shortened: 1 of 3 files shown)


jaraco:
Inferring from the bug report, the issue is that the DistutilsMetaFinder is creating a spec without a valid origin, creating a spec that is inferred to be a namespace package.

I might re-arrange the test a bit, but this is great. Thanks for the contrib.

jaraco:
I confirmed that 91ce77e captures the failed expectation when run with SETUPTOOLS_USE_DISTUTILS=local, which is run in CI, so I believe it to be a more selective and less entangled test.

Issue #3016, opened 2022-01-08
Cygwin CI has started failing


jaraco:
CI is failing during the install step with this output:

```
Run choco install git gcc-core python38-devel python38-pip --source cygwin
Chocolatey v0.11.3
[NuGet] Could not connect to the feed specified at 'https://community.chocolatey.org/api/v2/'. Please verify that the package source (located in the Package Manager Settings) is valid and ensure your network connectivity.
cygwin not installed. The package was not found with the source(s) listed.
 Source(s): 'https://community.chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://docs.chocolatey.org/en-us/troubleshooting for more
 assistance.
Installing the following packages:
git;gcc-core;python38-devel;python38-pip
By installing, you accept licenses for the packages.

Chocolatey installed 0/0 packages. 
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
The system cannot find the file specified
C:\\tools\\cygwin\\bin\\bash: D:\a\_temp\9a3f62ea-c178-4826-b186-e16cec4eb03b.ps1:3
Line |
   3 |  C:\\tools\\cygwin\\bin\\bash -l -x -c 'python3.8 -m pip install tox'
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The term 'C:\\tools\\cygwin\\bin\\bash' is not recognized as a name of a cmdlet, function, script
     | file, or executable program. Check the spelling of the name, or if a path was included, verify that
     | the path is correct and try again.

Error: Process completed with exit code 1.

```

jaraco:
I manually released 60.4 and 60.5, which were blocked by this issue.

jaraco:
@mkoeppe Would you be willing to look into the issue?

lazka:
We could switch to the official action like in https://github.com/pypa/distutils/blob/788cc159e4d734b972e22ccf06dbcd8ed8f94885/.github/workflows/main.yml#L35 to avoid depending on choco

mkoeppe:
I would think this was just an intermittent problem. Cygwin tests based on choco continue to work normally for us (https://github.com/sagemath/sage/runs/4752774122)

lazka:
I've created #3033 either way. Feel free to merge or close.

PR #3033, opened 2022-01-17
CI/cygwin: use cygwin/cygwin-install-action instead of choco

lazka:
Use the official GH action for installing cygwin instead of
going through choco.

See #3016

-----

The job is still disabled. I can re-enable it if wanted.


abravalheri:
Can we change line 45 `if: ${{ false }}  # failing #3016` to see if this change solve the problems with cygwin?

abravalheri:
It seems this solves the problem! Thank you very much @lazka!

abravalheri (approved):

PR #3048, opened 2022-01-24
doc: Fix trailing spaces, tabs, and missing newlines at end of file.

JulienPalard:
I spotted those while working on cpython's rstlint, which I test on other repos to see if it can be usefull. I now that's not much :D


abravalheri:
Thank you very much @JulienPalard! Every contribution is welcome 😄

PR #3060, opened 2022-01-29
Rely on session context when building sdist and wheel

### LINKED ISSUE #3059: Some setuptools tests run against stale artifacts
In #2968, the team introduced the `setuptools_wheel` fixture, but due to a possible deficiency in pytest, that fixture gets used across test runs, even with a clean checkout.

I discovered this issue while working on adding some new vendored dependencies that relied on their package_data being present. During early testing, the package data was missing, but even after correcting the issue in the Setuptools build, for some reason tests relying on setuptools_wheel were still failing. During debugging, I created a breakpoint to test:

```diff
diff --git a/setuptools/tests/fixtures.py b/setuptools/tests/fixtures.py
index 9b91d7d71..d6d2be06a 100644
--- a/setuptools/tests/fixtures.py
+++ b/setuptools/tests/fixtures.py
@@ -79,6 +79,7 @@ def setuptools_sdist(tmp_path_factory, request):
 @pytest.fixture(scope="session")
 def setuptools_wheel(tmp_path_factory, request):
     with contexts.session_locked_tmp_dir(tmp_path_factory, "wheel_build") as tmp:
+        breakpoint()
         dist = next(tmp.glob("*.whl"), None)
         if dist:
             return dist
```

And during debugging, confirmed that the artifact existed already on the first run and had yesterday's date on it:

```
setuptools/tests/test_distutils_adoption.py 
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/jaraco/code/public/pypa/setuptools/setuptools/tests/fixtures.py(83)setuptools_wheel()
-> dist = next(tmp.glob("*.whl"), None)
(Pdb) n
> /Users/jaraco/code/public/pypa/setuptools/setuptools/tests/fixtures.py(84)setuptools_wheel()
-> if dist:
(Pdb) dist
PosixPath('/private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pytest-of-jaraco/wheel_build/setuptools-60.5.4.post20220128-py3-none-any.whl')
```

And after exiting the tests, confirmed that the artifact is still present in my temp directory:

```
setuptools feature/vendor-jaraco-text $ ls /private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/pytest-of-jaraco/wheel_build
setuptools-60.5.4.post20220128-py3-none-any.whl
```

Looking into the implementation, I see a few potential problems:

- I don't see how [tmp_path_factory](https://docs.pytest.org/en/6.2.x/_modules/_pytest/tmpdir.html#TempPathFactory) does any cleanup. It seems like a bad design if `pytest-of-jaraco` is going to get reused across runs and possible across projects.
- session_locked_temp_dir [invokes mkdir directly](https://github.com/pypa/setuptools/blob/aa3d9b93f4b46cad03d3eeb174430668d565bb06/setuptools/tests/contexts.py#L110), bypassing `mktemp`, the primary mechanism for getting a directory managed by the factory.
- session_locked_temp_dir operates on the [parent of the temp dir](https://github.com/pypa/setuptools/blob/aa3d9b93f4b46cad03d3eeb174430668d565bb06/setuptools/tests/contexts.py#L105), which is presumably also unmanaged.

As I think about it more, I think I understand the issue. `getbasetemp()` does by default create a new, numbered temporary directory for the session. Because of (3), that session is bypassed and a directory is created outside of the management of pytest, so doesn't get cleaned up and can get re-used across sessions.

jaraco:
## Summary of changes

Closes #3059

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/tests/contexts.py
+++ b/setuptools/tests/contexts.py
@@ -99,12 +99,26 @@ def suppress_exceptions(*excs):
         pass
 
 
+def multiproc(request):
+    """
+    Return True if running under xdist and multiple
+    workers are used.
+    """
+    try:
+        worker_id = request.getfixturevalue('worker_id')
+    except Exception:
+        return False
+    return worker_id != 'master'
+
+
 @contextlib.contextmanager
-def session_locked_tmp_dir(tmp_path_factory, name):
+def session_locked_tmp_dir(request, tmp_path_factory, name):
     """Uses a file lock to guarantee only one worker can access a temp dir"""
-    root_tmp_dir = tmp_path_factory.getbasetemp().parent
-    # ^-- get the temp directory shared by all workers
-    locked_dir = root_tmp_dir / name
+    # get the temp directory shared by all workers
+    base = tmp_path_factory.getbasetemp()
+    shared_dir = base.parent if multiproc(request) else base
+
+    locked_dir = shared_dir / name
     with FileLock(locked_dir.with_suffix(".lock")):
         # ^-- prevent multiple workers to access the directory at once
         locked_dir.mkdir(exist_ok=True, parents=True)
--- a/setuptools/tests/fixtures.py
+++ b/setuptools/tests/fixtures.py
@@ -64,7 +64,8 @@ def sample_project(tmp_path):
 
 @pytest.fixture(scope="session")
 def setuptools_sdist(tmp_path_factory, request):
-    with contexts.session_locked_tmp_dir(tmp_path_factory, "sdist_build") as tmp:
+    with contexts.session_locked_tmp_dir(
+            request, tmp_path_factory, "sdist_build") as tmp:
         dist = next(tmp.glob("*.tar.gz"), None)
         if dist:
             return dist
@@ -78,7 +79,8 @@ def setuptools_sdist(tmp_path_factory, request):
 
 @pytest.fixture(scope="session")
 def setuptools_wheel(tmp_path_factory, request):
-    with contexts.session_locked_tmp_dir(tmp_path_factory, "wheel_build") as tmp:
+    with contexts.session_locked_tmp_dir(
+            request, tmp_path_factory, "wheel_build") as tmp:
         dist = next(tmp.glob("*.whl"), None)
         if dist:
             return dist


jaraco:
Github has messed up this PR. The diff is wrong and there are too many commits. If I look at the branch, it shows correctly that this branch is one commit ahead of main:

image

That's not pertinent to the failures, but it surely doesn't help in the analysis.

jaraco:
Aha. I did some searching and found [this article](https://medium.com/@kozyreva.hanna/pytest-xdist-with-session-scoped-fixture-b9884bd7bb29), which looks suspiciously similar, so I suspect its where the inspiration was drawn from, and confirms my misunderstanding about session fixtures. They're not automatically synchronized across workers in xdist, so each worker attempts to generate their own.

jaraco:
I distilled the [two issues](https://medium.com/@jaraco/there-are-issues-with-this-technique-87018a6ae391) I found with that article.

abravalheri:
Hi @jaraco, thank you very much for the deep investigation.
Indeed, I did not realise the shortcoming of this technique...

I understand that this PR will solve the problem if `xdist` is not used.

Does it make sense / is it worthy to manually remove the created directory after `yield locked_dir` and transform `setuptools_wheel` and `setuptools_sdist` in "yield fixtures"? Would that solve the problem?

abravalheri (approved):

jaraco:
> I understand that this PR will solve the problem if `xdist` is not used, but when `xdist` is being used, things remain more or less the same. Is that correct?

Yes. That's right. As I was investigating, I learned that `tmp_path_factory.getbasetemp()` changes depending on whether xdist is enabled, and it was only in the case that xdist was disabled or using only one process that the state was getting stored outside the session dir.

> Does it make sense / is it worthy to manually remove the created directory after the `yield` inside the context?

Originally, I was concerned, but once I figured out the basedir mismatch and also learned that pytest normally leaves files behind to clean them up in the future, so I'm okay with it now that it gets cleaned up with or without xdist.

PR #3064, opened 2022-02-01
[`pyproject.toml`: part 1] Update `build_meta` to use dist objects instead of simply running `setup.py`

abravalheri:
> *This is the first part of #2970, split into several stacked PRs as requested in https://github.com/pypa/setuptools/pull/2970#pullrequestreview-868151658*

We can rely on `distutils.core.run_setup` to get access to the distribution object while also postponing the execution of the build commands.

This allows directly manipulation of the distribution object (which is for example useful to add configurations from different files such as `pyproject.toml`) as well as introspecting it for metadata (such as `setup_requires`).

## Summary of changes

- Use `distutils.core.run_setup` to obtain a dist object in `build_meta` - 8773aa266ee09ac408b2b295488d638ebfcca321
- Use the distribution object to run commands in `build_meta` - 3d7e7df6acfa397fecb759922523e16536ae2fcb
- Replace `_get_build_requires` by using the dist object - ec5b43b900908beadb82619b91e64ea2e648ec3e
- Add some entry-points to the bootstrap - 6bbc7547cc5932f0a470d8d6fc176f3fc96c675d
- Make sure changes in `build_meta` are compatible with tests - 2f991555571b0e7300962e3f183b3f2413c8590c
- Make `script_name` on `build_meta` less error prone - 90f216c9cbc910ed48d0748dd9064b7036e86bb0

Please see the commit message of each individual commit for more information about the changes.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/bootstrap.egg-info/entry_points.txt
+++ b/bootstrap.egg-info/entry_points.txt
@@ -2,6 +2,11 @@
 egg_info = setuptools.command.egg_info:egg_info
 build_py = setuptools.command.build_py:build_py
 sdist = setuptools.command.sdist:sdist
+dist_info = setuptools.command.dist_info:dist_info
+develop = setuptools.command.develop:develop
+install_scripts = setuptools.command.install_scripts:install_scripts
+bdist_egg = setuptools.command.bdist_egg:bdist_egg
+install_egg_info = setuptools.command.install_egg_info:install_egg_info
 
 [distutils.setup_keywords]
 include_package_data = setuptools.dist:assert_bool
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -47,6 +47,17 @@ __all__ = ['get_requires_for_build_sdist',
            'build_sdist',
            '__legacy__']
 
+SETUP_SCRIPT = "setup.py"
+CONFIG_FILE = "pyproject.toml"
+LEGACY_CONFIG_FILE = "setup.cfg"
+SAFE_SCRIPT_NAME = "%%setuptools.build_meta%%"
+# ^-- distutils.sdist:sdist._add_defaults_standards will try to add `script_name`
+#     to the list of files to include in the distribution.
+#     If the SETUP_SCRIPT does not exist, the best is to use some made-up name
+#     that is unlikely to exist as a file.
+#     If we use `sys.argv[0]` setuptools might try to add things like
+#     `.tox/python/bin/pytest` to the tar.gz (and this will result in an error)
+
 
 @contextlib.contextmanager
 def no_install_setup_requires():
@@ -132,14 +143,17 @@ class _BuildMetaBackend(object):
         config_settings.setdefault('--global-option', [])
         return config_settings
 
-    def _get_dist(self, setup_script="setup.py"):
+    def _get_dist(self):
         """Retrieve a distribution object already configured."""
 
-        if os.path.exists(setup_script) and os.stat(setup_script).st_size > 0:
+        if os.path.exists(SETUP_SCRIPT) and os.stat(SETUP_SCRIPT).st_size > 0:
             with no_install_setup_requires(), _patch_distutils_core():
-                dist = distutils.core.run_setup(setup_script, stop_after="init")
+                dist = distutils.core.run_setup(SETUP_SCRIPT, stop_after="init")
+                dist.script_name = SETUP_SCRIPT
+                # ^-- preserve _add_defaults_standards behaviour
         else:
             dist = setuptools.dist.Distribution()
+            dist.script_name = SAFE_SCRIPT_NAME
 
         dist.parse_config_files()
         dist.finalize_options()
@@ -156,7 +170,6 @@ class _BuildMetaBackend(object):
         # Note that we can reuse our build directory between calls
         # Correctness comes first, then optimization later
         dist = self._get_dist()
-        dist.script_name = sys.argv[0]
         dist.script_args = args
         dist.parse_command_line()
 
@@ -256,13 +269,12 @@ class _BuildMetaLegacyBackend(_BuildMetaBackend):
         # '' into sys.path. (pypa/setuptools#1642)
         sys_path = list(sys.path)           # Save the original path
 
-        setup_script = "setup.py"
-        if not os.path.exists(setup_script) or os.stat(setup_script).st_size == 0:
-            msg = f"__legacy__ backend conflicts with empty/missing {setup_script!r}"
+        if not os.path.exists(SETUP_SCRIPT) or os.stat(SETUP_SCRIPT).st_size == 0:
+            msg = f"__legacy__ backend conflicts with empty/missing {SETUP_SCRIPT!r}"
             warnings.warn(msg, setuptools.SetuptoolsDeprecationWarning)
             return super().run_command(*args)
 
-        script_dir = os.path.dirname(os.path.abspath(setup_script))
+        script_dir = os.path.dirname(os.path.abspath(SETUP_SCRIPT))
         if script_dir not in sys.path:
             sys.path.insert(0, script_dir)
 
@@ -271,7 +283,7 @@ class _BuildMetaLegacyBackend(_BuildMetaBackend):
         # setup.py script. ==> This is already handled in distutils.core
         try:
             with no_install_setup_requires(), _patch_distutils_core():
-                distutils.core.run_setup(setup_script, args)
+                distutils.core.run_setup(SETUP_SCRIPT, args)
         finally:
             # While PEP 517 frontends should be calling each hook in a fresh
             # subprocess according to the standard (and thus it should not be
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -39,6 +39,7 @@ from uuid import uuid4
 import setuptools
 import distutils
 from ._reqs import parse_strings
+from setuptools.extern.packaging.requirements import Requirement
 
 __all__ = ['get_requires_for_build_sdist',
            'get_requires_for_build_wheel',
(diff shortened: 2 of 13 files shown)


bootstrap.egg-info/entry_points.txt
```diff
@@ -2,6 +2,11 @@
 egg_info = setuptools.command.egg_info:egg_info
 build_py = setuptools.command.build_py:build_py
 sdist = setuptools.command.sdist:sdist
+dist_info = setuptools.command.dist_info:dist_info
+develop = setuptools.command.develop:develop
+install_scripts = setuptools.command.install_scripts:install_scripts
+bdist_egg = setuptools.command.bdist_egg:bdist_egg
+install_egg_info = setuptools.command.install_egg_info:install_egg_info
```
abravalheri, line 9: Pending review comment: https://github.com/pypa/setuptools/pull/2970#discussion_r795206791

setuptools/tests/test_build_meta.py
```diff
@@ -440,18 +442,28 @@ def run():
...
-        with pytest.raises(AssertionError):
-            build_backend.build_sdist("temp")
+
+        # TODO: Clarify the use case motivating this test
+        # Why is it necessary that the assertion error is raised?
+        # Or is it just a side effect that was found to happen,
+        # but it is not necessary for setuptools to work properly?
+        # >>> with pytest.raises(AssertionError):
+
+        build_backend.build_sdist("temp")
```
abravalheri, line 452: Pending review comment: https://github.com/pypa/setuptools/pull/2970#discussion_r795206170

setuptools/tests/test_build_meta.py
```diff
@@ -440,18 +442,28 @@ def run():
...
-                match=re.escape('No distribution was found.')):
-            getattr(build_backend, build_hook)("temp")
+        # TODO: Clarify why is it necessary that the error is raised?
+        # Or is it just a side effect that was found to happen,
+        # but it is not necessary for setuptools to work properly?
+        # >>> with pytest.raises(
+        # >>>         ValueError,
+        # >>>         match=re.escape('No distribution was found.')):
+
+        getattr(build_backend, build_hook)("temp")
```
abravalheri, line 466: Pending review comment: https://github.com/pypa/setuptools/pull/2970#discussion_r795206170

abravalheri:
For some reason the diff (and the list of commits) are not updating according to the latest version of `support-pyproject-aggregate`...

[This diff](https://github.com/pypa/setuptools/pull/3064/files/da71fab6210e14e81047909c1b66444062842cfe..a956b0b62d58cf4a7c1f5e97cec49a4446392fb0) is more appropriate... Ideally the list of commits should have only 8 commits...

setuptools/build_meta.py
```diff
@@ -26,53 +26,38 @@
...
-        orig = distutils.core.Distribution
-        distutils.core.Distribution = cls
-        try:
-            yield
-        finally:
-            distutils.core.Distribution = orig
+SETUP_SCRIPT = "setup.py"
+CONFIG_FILE = "pyproject.toml"
+LEGACY_CONFIG_FILE = "setup.cfg"
+SAFE_SCRIPT_NAME = "%%setuptools.build_meta%%"
```
webknjaz, line 54: If this needs to be non-existing, why not add a random factor to it, like `uuid.uuid4()`?

setuptools/build_meta.py
```diff
@@ -110,51 +95,88 @@ def _file_with_extension(directory, extension):
...
+        yield  # do nothing, already using the improved version of distutils
+        return
+
+    def _exec(code, global_vars):
+        try:
+            fid, tmp = tempfile.mkstemp(suffix=f"{uuid4()}-setup.py", text=False)
+            os.close(fid)  # Ignore the low level API
+            with open(tmp, "wb") as f:
+                f.write(code)
+            with tokenize.open(tmp) as f:
```
webknjaz, line 119: Can this be done in memory?

setuptools/build_meta.py
```diff
@@ -110,51 +95,88 @@ def _file_with_extension(directory, extension):
...
+                code = f.read().replace(r'\r\n', r'\n')
+        finally:
+            os.remove(tmp)
```
webknjaz, line 122: The implementation is quite straightforward here https://github.com/python/cpython/blob/81c7204/Lib/tokenize.py#L392-L405. and could be emulated a bit more efficiently with smth like
```suggestion
        from io import BytesIO, TextIOWrapper

        with BytesIO(code) as bytes_code_buffer:
            encoding, _lines = detect_encoding(bytes_code_buffer.readline)
            bytes_code_buffer.seek(0)
            with TextIOWrapper(buffer, encoding, line_buffering=True) as text:
                code = text.read().replace(r'\r\n', r'\n')
```

setuptools/build_meta.py
```diff
@@ -110,51 +95,88 @@ def _file_with_extension(directory, extension):
...
+        orig_exec(code, {**global_vars, "__name__": "__main__"})
+
+    def _run_commands(dist):
+        try:
+            dist.run_commands()
+        except Exception as ex:
+            raise SystemExit("error:" + str(ex))
```
webknjaz, line 129: How about
```suggestion
            raise SystemExit(f"error:{ex!s}") from ex
```
?

(Also, why there's no space after the colon?)

setuptools/build_meta.py
```diff
@@ -110,51 +95,88 @@ def _file_with_extension(directory, extension):
...
+        """Retrieve a distribution object already configured."""
 
-        sys.argv = sys.argv[:1] + ['egg_info'] + \
-            config_settings["--global-option"]
-        try:
-            with Distribution.patch():
-                self.run_setup()
-        except SetupRequirementsError as e:
-            requirements += e.specifiers
+        if os.path.exists(SETUP_SCRIPT) and os.stat(SETUP_SCRIPT).st_size > 0:
```
webknjaz, line 150: Could you put this condition into a label-variable so that it's self-explanatory?
```suggestion
        non_empty_setup_py_present = os.path.exists(SETUP_SCRIPT) and os.stat(SETUP_SCRIPT).st_size > 0
        if non_empty_setup_py_present:
```

setuptools/build_meta.py
```diff
@@ -246,33 +264,35 @@ class _BuildMetaLegacyBackend(_BuildMetaBackend):
...
-    def run_setup(self, setup_script='setup.py'):
+
+    def run_command(self, *args):
         # In order to maintain compatibility with scripts assuming that
         # the setup.py script is in a directory on the PYTHONPATH, inject
         # '' into sys.path. (pypa/setuptools#1642)
         sys_path = list(sys.path)           # Save the original path
 
-        script_dir = os.path.dirname(os.path.abspath(setup_script))
+        if not os.path.exists(SETUP_SCRIPT) or os.stat(SETUP_SCRIPT).st_size == 0:
```
webknjaz, line 274: Same as above, it's best to have an explanatory var
```suggestion
        setup_py_is_empty = not os.path.exists(SETUP_SCRIPT) or os.stat(SETUP_SCRIPT).st_size == 0
        if setup_py_is_empty:
```

setuptools/tests/test_build_meta.py
```diff
@@ -132,7 +132,9 @@ def get_build_backend(self):
     @pytest.fixture(params=defns)
     def build_backend(self, tmpdir, request):
         path.build(request.param, prefix=str(tmpdir))
-        with tmpdir.as_cwd():
+        with tmpdir.as_cwd(), warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            warnings.filterwarnings("ignore", "__legacy__ backend conflicts")
```
webknjaz, line 137: Have you considered using a `with pytest.warns()` check instead of just suppressing them?

setuptools/tests/test_build_meta.py
```diff
@@ -440,18 +442,28 @@ def run():
...
     @pytest.mark.parametrize('build_hook', ('build_sdist', 'build_wheel'))
     def test_build_with_empty_setuppy(self, build_backend, build_hook):
         files = {'setup.py': ''}
         path.build(files)
 
-        with pytest.raises(
-                ValueError,
-                match=re.escape('No distribution was found.')):
-            getattr(build_backend, build_hook)("temp")
+        # TODO: Clarify why is it necessary that the error is raised?
```
webknjaz, line 459: Maybe it's related to an empty `setup.py` with no metadata (not even a package name)?

webknjaz:
@abravalheri FYI my comments are mostly readability/maintainability related. I haven't looked too deeply into the side effects of the patch.

setuptools/build_meta.py
abravalheri: Thank you very much for the improvement!

setuptools/tests/test_build_meta.py
abravalheri: I tried to remove all filter warnings I managed, and raise exceptions to avoid accidental behaviour.

Relevant commits: https://github.com/pypa/setuptools/pull/3064/commits/660a273f67dc0aa3e9d2818079051a4b927d91bf and https://github.com/pypa/setuptools/pull/3064/commits/f5fe86f2ac6162537699557067e1cbb2620b9226

This is related to https://github.com/pypa/setuptools/pull/3064#discussion_r803011890.

---

Rewriting this particular test in terms of `pytest.warns` is very complicated. When I was adding tests, I tried to keep the structure very similar to the existing code in the file to avoid being to disruptive and rewriting everything. Unfortunately the existing test uses some tricks, e.g. `TestBuildMetaLegacyBackend` inherits from `TestBuildMetaBackend` to reuse test cases, and fixtures are given via `pytest.fixture(params=...)` instead of `pytest.mark.parametrize`, etc...

This made it very difficult to not rely on warning filters on #3068... (although I tried, I did not manage to remove all the filters)

setuptools/tests/test_build_meta.py
abravalheri: The solution for this is presented in  https://github.com/pypa/setuptools/commit/660a273f67dc0aa3e9d2818079051a4b927d91bf (https://github.com/pypa/setuptools/pull/3064#discussion_r803011890.)

abravalheri:
Thank you very much @webknjaz for the very kind review and all the improvements suggested. I tried to make my very best to address your comments. In some cases I took a different approach, but I left some comments explaining my reasoning.

abravalheri:
Due to the problems reported in https://github.com/pypa/setuptools/pull/2970#issuecomment-1036078047, I decided to drop this part of the change and rework the following PRs to work without it.

The attempt of "modernising" the `build_meta` code is recoded in https://github.com/abravalheri/setuptools/tree/modernise-build-meta, so we can try it in a future attempt.

PR #3065, opened 2022-02-01
[`pyproject.toml`: part 1] Separate `setup.cfg` parsing and extract common post-processing functions

abravalheri:
> *This is part of #2970, split into several stacked PRs as requested in https://github.com/pypa/setuptools/pull/2970#pullrequestreview-868151658*
> *Follows #3064*

> **Update:** *#3064 was dropped because of the reasons discussed in https://github.com/pypa/setuptools/pull/2970#issuecomment-1036078047. The following PRs will be changed to work without it.*

We can split the process of interpreting configuration files into 2 steps:

1. The parsing the file contents from strings to value objects that can be understand by Python (for example a string with a comma separated list of keywords into an actual Python list of strings).
2. The expansion (or post-processing) of these values according to the semantics ``setuptools`` assign to them (for example a configuration field with the ``file:`` directive should be expanded from a list of file paths to a single string with the contents of those files concatenated)

The idea of this change is to extract the functions responsible for (2.) into a new module, so they can be reused between different config file formats.

## Summary of changes

1. Rename existing `setuptools.config` to `setuptools.config.setupcfg` - 300ebba7fe200ba32c9f35d9f1e71aecde573a84
2. Extract post-processing functions from `setuptools.config.setupcfg` into `setuptools.config.expand` - b0346a8e5164e5fca718bc28ff0a2595618db72b
3. Allow `root_dir` to be explicit in `setuptools.config.expand` functions - 9c596074d2f3f2afe7169a8a2bcfbfbc7b257613
4. Allow single strings in `setuptools.config.expand.read_files` - 8b0a15a5462c36e2f535ccd533573c6fd75d4b50
5. Adequate `test_setupcfg` to latest changes in `setuptools.config.setupcfg` - 09499a078d31025e562937f6206453a9d7e63442

Please see the commit message of each individual commit for more information about the changes.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


setuptools/config/expand.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def cmdclass(values, package_dir=None, root_dir=None):
+    """Given a dictionary mapping command names to strings for qualified class
+    names, apply :func:`resolve_class` to the dict values.
+    """
+    return {k: resolve_class(v, package_dir, root_dir) for k, v in values.items()}
+
+
+def find_packages(*, namespaces=False, root_dir=None, **kwargs):
```
abravalheri, line 224: Pending review comment: https://github.com/pypa/setuptools/pull/2970#discussion_r778432831

changelog.d/3065.change.rst
```diff
@@ -0,0 +1,4 @@
+Refactored ``setuptools.config`` by separating configuration parsing (specific
```
webknjaz, line 1: Is this change user-facing? Maybe change the fragment type to `misc` if it's internal?

setuptools/config/expand.py
```diff
@@ -0,0 +1,288 @@
...
+from itertools import chain
+
+from distutils.errors import DistutilsOptionError
+
+chain_iter = chain.from_iterable
+
+
+class StaticModule:
+    """
+    Attempt to load the module by the name
```
webknjaz, line 33: This feels wrong. This verb at the beginning suggests that this thing the docstring describes is callable (function/method or has `__call__()`). But it isn't. Maybe consider describing what this object is? And move the side-effect description to where it happens, like the initializer docstring.

setuptools/config/expand.py
```diff
@@ -0,0 +1,288 @@
...
+            return next(
+                ast.literal_eval(statement.value)
+                for statement in self.module.body
+                if isinstance(statement, ast.Assign)
+                for target in statement.targets
+                if isinstance(target, ast.Name) and target.id == attr
+            )
+        except Exception as e:
+            raise AttributeError(
+                "{self.name} has no attribute {attr}".format(**locals())
```
webknjaz, line 54: `str.fomat_map()` could be seen as slightly easier to read.
```suggestion
                "{self.name} has no attribute {attr}".format_map(locals())
```

webknjaz: OTOH I don't see why don't you just use an f-string
```suggestion
                f"{self.name} has no attribute {attr}"
```

setuptools/config/expand.py
```diff
@@ -0,0 +1,288 @@
...
+        del self.self
+
+    def __getattr__(self, attr):
+        try:
+            return next(
+                ast.literal_eval(statement.value)
+                for statement in self.module.body
+                if isinstance(statement, ast.Assign)
+                for target in statement.targets
+                if isinstance(target, ast.Name) and target.id == attr
```
webknjaz, line 50: I can't parse the precedence here. Could you rewrite it splitting it into digestible chunks? Maybe construct a series of generator expressions if that helps.

setuptools/config/expand.py
```diff
@@ -0,0 +1,288 @@
...
+
+def glob_relative(patterns, root_dir=None):
+    """Expand the list of glob patterns, but preserving relative paths.
+
+    :param list[str] patterns: List of glob patterns
+    :param str root_dir: Path to which globs should be relative
+                         (current directory by default)
+    :rtype: list
+    """
+    glob_characters = ('*', '?', '[', ']', '{', '}')
```
webknjaz, line 66: It seems like a set could be semantically more correct here
```suggestion
    glob_characters = {'*', '?', '[', ']', '{', '}'}
```

setuptools/config/expand.py
webknjaz: I think that turning the switch on the namespaces default could be useful in a separate PR but would be out of the scope of this effort.

webknjaz:
Is `setuptools/tests/config/__init__.py` really necessary?

setuptools/tests/config/test_expand.py
```diff
@@ -0,0 +1,106 @@
...
+    values = expand.read_attr('lib.mod.VALUES', {'lib': 'pkg/sub'}, tmp_path)
+    assert values['c'] == (0, 1, 1)
+
+
+def test_resolve_class():
+    from setuptools.command.sdist import sdist
+    assert expand.resolve_class("setuptools.command.sdist.sdist") == sdist
+
+
+def test_find_packages(tmp_path):
```
webknjaz, line 83: It seems like this test is actually several parametrized variations. Would it be possible to use `@pytest.mark.parametrize` here?

setuptools/tests/config/test_expand.py
```diff
@@ -0,0 +1,106 @@
...
+    assert values['b'] == {42}
+
+    # Make sure the same APIs work outside cwd
+    assert expand.read_attr('pkg.sub.VERSION', root_dir=tmp_path) == '0.1.1'
+    values = expand.read_attr('lib.mod.VALUES', {'lib': 'pkg/sub'}, tmp_path)
+    assert values['c'] == (0, 1, 1)
+
+
+def test_resolve_class():
+    from setuptools.command.sdist import sdist
```
webknjaz, line 79: Any reason this import is not at the top? Can it poison the imported module cache for subsequent tests?

setuptools/tests/config/test_expand.py
```diff
@@ -0,0 +1,106 @@
...
+        "dir1/file3.txt",
+        "a.ini",
+        "b.ini",
+        "dir1/c.ini",
+        "dir1/dir2/a.ini",
+    }
+
+    write_files({k: "" for k in files}, tmp_path)
+    patterns = ["**/*.txt", "[ab].*", "**/[ac].ini"]
+    with pushd(tmp_path):
```
webknjaz, line 30: Is `pushd()` significant for the tests here? If not, I suggest switching to the built-in fixture and doing a variation of `monkeypatch.chdir()`.

setuptools/tests/config/test_expand.py
```diff
@@ -0,0 +1,106 @@
...
+    files = {
+        "a.txt": "a",
+        "dir1/b.txt": "b",
+        "dir1/dir2/c.txt": "c"
+    }
+    write_files(files, tmp_path)
+    with pushd(tmp_path):
+        assert expand.read_files(list(files)) == "a\nb\nc"
+
+    with pushd(tmp_path / "dir1"), pytest.raises(DistutilsOptionError):
```
webknjaz, line 46: It is highly recommended to always use a `match` argument together with `pytest.raises`/`pytest.warns`/`pytest.deprecated` so that the test is precise and doesn't pass based on side-effects.

setuptools/config/expand.py
abravalheri: Thank you very much @webknjaz. This comes from the [original implementation](https://github.com/pypa/setuptools/blob/fe7975d091483e8d6d4c9befea38ccfeb29e8e49/setuptools/config.py#L22) that was moved/extracted.

I will try to think what would be more appropriate.

setuptools/tests/config/test_expand.py
abravalheri: I tried to move it to the top and the tests still pass. I don't remember exactly my initial motivations so let's just move it up!

setuptools/tests/config/test_expand.py
abravalheri: Thank you very much for the suggestion, I tried to add a match for all checks in this test file.

abravalheri:
Thank you very much for the kind very @webknjaz and valuable suggestions!

PR #3070, opened 2022-02-01
Skip non-string values from sysconfig.get_config_vars()

### LINKED ISSUE #3063: easy_install doesn't handle sysconfig vars being integers.
In the `create_home_path` function of `setuptools/command/easy_install.py` the code fails with an AttributeError if `path` is not a string. Since #3027 `self.config_vars` has been created from `sysconfig.get_config_vars()`, where the values can be strings or integers, such as `'WITH_PYMALLOC': 1`.

https://github.com/pypa/setuptools/blob/b2ba0e39918d58bd6c6b15093ddcdc42605040ce/setuptools/command/easy_install.py#L241

https://github.com/pypa/setuptools/blob/b2ba0e39918d58bd6c6b15093ddcdc42605040ce/setuptools/command/easy_install.py#L1337-L1338

Because integers do not have a `startswith` method an AttributeError is raised. Checking whether `path` is a string should fix this.

I tried to reproduce this locally with `easy_install`, but couldn't. However, it can be observed in pip's testsuite: https://github.com/pypa/pip/runs/5002678226

domdfcoding:
## Summary of changes

`sysconfig.get_config_vars()` can return values which are not strings (including integers), but the current code assumes all values are strings. This results in an `AttributeError`. Adding a check for non-string values and skipping them avoids the error.

Closes #3063

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3070.misc.rst
+++ b/changelog.d/3070.misc.rst
@@ -1,0 +1,1 @@
+Avoid AttributeError in easy_install.create_home_path when sysconfig.get_config_vars values are not strings.
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -1334,9 +1334,7 @@ class easy_install(Command):
         if not self.user:
             return
         home = convert_path(os.path.expanduser("~"))
-        for path in self.config_vars.values():
-            if not isinstance(path, str):
-                continue
+        for path in only_strs(self.config_vars.values()):
             if path.startswith(home) and not os.path.isdir(path):
                 self.debug_print("os.makedirs('%s', 0o700)" % path)
                 os.makedirs(path, 0o700)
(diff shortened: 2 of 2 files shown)


datumbox:
@domdfcoding Thanks for issuing a patch.

@jaraco Are there plans for releasing a hotfix or this is unlikely? The latest version of setuptools brake a couple of jobs downstream on PyTorch and I would like to know if a new version is coming or pin to the previous one. Thanks!

abravalheri:
@datumbox there was some hick-up with the CI (sometimes PyPy on Windows fails to allocate enough memory), but there should be some release out: https://github.com/pypa/setuptools/releases/tag/v60.7.0.

Please let us know how it goes :)

datumbox:
@abravalheri Yes, that worked! Our CI just "saw" the new version and the failure is now resolved. Thanks a lot for the help!

Issue #3085, opened 2022-02-05
Setuptools should not rely on pkg_resources


jaraco:
This issue basically tracks the root goal of obviating pkg_resources, in particular that Setuptools depends on it.

Over the past few years, myself and others have worked hard to create `importlib.resources` and `importlib.metadata` in the goal of superseding pkg_resources. These libraries are now very stable and replacing much of the functionality provided by pkg_resources.

In https://github.com/python/importlib_metadata/issues/38, I examined the setuptools package to determine its level of dependency on pkg_resources. Unsurprisingly, there is a diverse set of interfaces on which setuptools relies.

Nevertheless, this project should work toward Setuptools adopting the importlib features in place of pkg_resources.

romuald:
This somehow breaks projects entry points to pyramid/paste it seems, which aren't able to find their entry point anymore (I have no idea what's the cause)

Error for Google:
> LookupError: Entry point 'main' not found in egg 'my.package' (dir: xxx/.pyenv-python3.7/lib/python3.7/site-packages; protocols: paste.app_factory, paste.composite_factory, paste.composit_factory; entry_points: )

armanckeser:
The same pr causes this issue I think #3102

jaraco:
Now that the setuptools code base no longer depends on pkg_resources, I'm considering once again republishing pkg_resources as a separate package.

I've reactivated the pypa/pkg_resources repo and re-run a filter operation on this repo to re-capture the history of that code there:

```
git filter-repo --path pkg_resources/ --path api_tests.txt --path pkg_resources.py --path pkg_resources.txt --path setuptools/tests/test_resources.py tests/api_tests.txt tests/test_pkg_resources.py
```

I'm now pondering publishing that code to see if the setuptools tests can continue to run by requiring that package at test time.

jaraco:
Attempting to remove pkg_resources, the pbr tests fail with this traceback:

```
  File "/Users/jaraco/draft/.venv/lib/python3.13/site-packages/pbr/core.py", line 53, in <module>
    from pbr import util
  File "/Users/jaraco/draft/.venv/lib/python3.13/site-packages/pbr/util.py", line 75, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

It seems that pbr is reliant on the interface.

jaraco:
[...]

PR #3087, opened 2022-02-06
Remove reliance on pkg_resources for entry points

jaraco:
Ref #3085


jaraco:
That's weird. Tests are timing out on PyPy (only). The same thing happens on my local machine.

Curiously, if I disable the pytest-xdist plugin, the tests run. So it seems xdist is implicated.

jaraco:
I do see that at least one test passed on pypy for SETUPTOOLS_USE_DISTUTILS=local. I don't get it - there's next to nothing in this change that's related to distutils or pypy.

jaraco:
An earlier test is passing or nearly passing. Maybe the issue is transient.

jaraco:
Issue was apparently transient. Seems tests are getting flakier and flakier. Not good considering it’s now taking hours to work through jobs.

abravalheri:
> Issue was apparently transient. Seems tests are getting flakier and flakier. Not good considering it’s now taking hours to work through jobs.

Yes, I noticed in previous PRs that the combination Windows + PyPy 3.7 is specially flaky. In most of the cases I observed, the error happens on `test_build_meta.py` with `self.pool.submit`. For some reason the combination seems to lead to memory allocation errors (lack of available memory in the worker, maybe? That would make sense with the behaviour you observed with `xdist`). Other time it just takes for ever and ends up cancelled.

These are just a few of the examples I found recently:
- https://github.com/pypa/setuptools/runs/5056604070?check_suite_focus=true#step:5:1583
- https://github.com/pypa/setuptools/runs/5021063326?check_suite_focus=true#step:5:1690

changelog.d/3085.change.rst
```diff
@@ -0,0 +1 @@
+Setuptools no longer relies on pkg_resources for entry point handling.
```
webknjaz, line 1: (nit)
```suggestion
Setuptools no longer relies on ``pkg_resources`` for entry point handling.
```

setuptools/_entry_points.py
```diff
@@ -0,0 +1,79 @@
...
+    return ep.group, ep.name
+
+
+def validate(eps: metadata.EntryPoints):
+    """
+    Ensure entry points are unique by group and name and validate the pattern.
+    """
+    for ep in ensure_unique(eps, key=by_group_and_name):
+        # exercise one of the dynamic properties to trigger validation
+        ep.extras
```
webknjaz, line 41: Maybe consider reusing the helper?
```suggestion
        ensure_valid(ep)
```

jaraco: Thanks for flagging this, as I'd lost track of it.

You're right. I was noticing that `ensure_valid` wasn't used anywhere. Some consolidation is called for. If I was going to invoke `ensure_valid`, I'd also want to remove the preceding comment, which becomes redundant, so I'll figure something out separately.

jaraco: Fixed in f4b0e3ac5.

Issue #3093, opened 2022-02-11
Setuptools consuming excessive CI resources


jaraco:
@jaraco I think the large # of jobs this introduces to setuptools' GitHub workflows is slowing down builds for other PyPA projects. There is a [limit of 20 concurrent jobs](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) org-wide for our plan, and setuptools has gone from 36 jobs in #3078 to 119 jobs here. I've noticed a significant backlog over at https://github.com/pypa/warehouse/actions over the last few days and suspect this might be why.

Any thoughts on how we could get this down to a more reasonable number of concurrent jobs per workflow run?

(cc @ewdurbin)

_Originally posted by @di in https://github.com/pypa/setuptools/issues/2923#issuecomment-1035840098_

jaraco:
I too have noticed that Setuptools tests now take forever to run, and turning down concurrency, if that would help the other projects would only throttle Setuptools.

Maybe the new platform tests could be set up to be run only for tagged commits (releases) or for PRs with a particular tag to drastically limit the scope of these platform tests.

jaraco:
I believe that push should limit the execution of these tests. Let's see if that improves the situation.

di:
Thanks!

blink1073:
I'm happy to open a PR to add [Using concurrency to cancel any in-progress job or run](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run) as well

jaraco:
Sure. Let’s give it a try!

jaraco:
Something else I'd like to consider - the matrix of tests is doubled by the SETUPTOOLS_USE_DISTUTILS=local/stdlib. Soon I'd like to remove that matrix, maybe to replace it with a single job to capture regressions in that space.

blink1073:
You can leave it out of the matrix and add an explicit include entry with that option.

blink1073:
(I'm happy to do that as well)

jaraco:
It seems the release process is broken due to concurrency cancellation:

![image](https://user-images.githubusercontent.com/308610/154278836-aa912e33-a5a1-470c-957c-1844a8e1b0af.png)

Because a tagged commit is also a mainline commit with the same SHA hash, the second one without the tag takes precedence, so the tagged commit gets cancelled and no release happens.

I'm not sure what needs to be done to address the issue.

[...]

PR #3097, opened 2022-02-11
Using concurrency to cancel any in-progress job or run

blink1073:
## Summary of changes
Uses [concurrency to cancel any in-progress job or run](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run) to help address #3093

Closes 

### Pull Request Checklist
- [ ] Changes have tests (N/A)
- [ ] News fragment added in [`changelog.d/`].  (N/A)

[PR docs]:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,10 @@ name: tests
 
 on: [push, pull_request, workflow_dispatch]
 
+concurrency: 
+  group: tests-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   test:
     strategy:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,8 +2,8 @@ name: tests
 
 on: [push, pull_request, workflow_dispatch]
 
-concurrency: 
-  group: tests-${{ github.head_ref || github.run_id }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 jobs:


.github/workflows/main.yml
```diff
@@ -2,6 +2,10 @@ name: tests
 
 on: [push, pull_request, workflow_dispatch]
 
+concurrency: 
+  group: tests-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
di, line 7: My recommendation would be to do [what `pip` does in it's CI suite](https://github.com/pypa/pip/blob/ec8edbf5df977bb88e1c777dd44e26664d81e216/.github/workflows/ci.yml#L19-L21):

```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
  cancel-in-progress: true
```

Since `run_id` is a unique number for each job run, this wouldn't have much of an effect e.g. with multiple pushes to a single branch.

And I think grouping on `head_ref` would have the effect of cancelling in-progress runs on commits pushed to the default branch, which IMO is undesirable.

blink1073: Makes sense, thanks!

abravalheri:
I don't know exactly but I have the impression that this change accidentally is preventing the `release` job to run on tags...

For example if we look at the [workflow for the tag v60.9.1](https://github.com/pypa/setuptools/actions/runs/1844730255) (Fig. 1), the message seems to indicate that it got cancelled by GitHub because jobs for the commit being pushed to the main branch got enqueued just after the ones for the tag (Fig. 2):

> image
> Fig 1: Workflow for tag is getting cancelled because of a higher priority request


> image
> Fig 2: The higher priority request seems to be the normal push to `main` of the same commit

This might be related to the lack of determinism in the order workflows are triggered...

I don't know exactly how to solve this, but maybe it is a good idea to add the name of the event triggering the workflow to the `group` string? (if that is even possible)

@blink1073, @di what do you think?

PR #3114, opened 2022-02-16
Add separate release workflow

blink1073:
## Summary of changes
Adds an extra discriminator for `${{ github.ref }}` in the concurrency group.
This avoids a race condition in workflow runs which caused the release job to be canceled for either a tag or a commit with the same sha.

Part of #3093 

### Pull Request Checklist
- [x] Changes have tests
- [ ] News fragment added in [`changelog.d/`].  N/A

[PR docs]:


abravalheri:
Hi @blink1073 thank you very much for providing this PR!

I was trying to go through GitHub's docs trying to understand this config with `release: types: [published]`.

Does this solve the problem with tests never running discussed in https://github.com/pypa/setuptools/issues/3093#issuecomment-1041535777?

blink1073:
The latest commit effectively restores the original behavior before concurrency limiting as much as possible.   The only difference is that we don't run the unit tests explicitly before release.

abravalheri:
I was not expecting how **more** annoying PyPy+Windows tests can be...
It is like they are just waiting to fail in different ways...

blink1073:
Yeah, it is pain squared

blink1073:
Okay, with the latest change I believe we have what we want.  We avoid canceling subsequent pushes *and* pushes that are to different git references, `refs/head/main` vs `refs/tags/foo`.

blink1073:
Closing in favor of #3116

PR #3133, opened 2022-02-24
Tests using virtual environments: pass the copy of existing env to the subprocesses

befeleme:
## Summary of changes

In specifically set environment like the build environment in Fedora, we noticed some of the tests running subprocesses were failing. Typically they didn't see the correct paths, leading to multiple `ModuleNotFoundError: No module named setuptools` and failed meta-tests in [`test_virtualenv`](https://github.com/pypa/setuptools/blob/main/setuptools/tests/test_virtualenv.py#L95) (`DID NOT RAISE...`).
The changes: passing the copy of the environment minus PYTHONPATH for the subprocesses made the test suite pass during our builds.

### Pull Request Checklist
- [ ] Changes have tests (I'd say N/A - those are test fixes)

[PR docs]:

### DIFF
--- a/setuptools/tests/environment.py
+++ b/setuptools/tests/environment.py
@@ -22,7 +22,7 @@ class VirtualEnv(jaraco.envs.VirtualEnv):
         # - tox isn't used to run tests and
         # - PYTHONPATH is set to point to a specific setuptools codebase and
         # - no custom env is explicitly set by a test
-        # that PYTHONPATH leaks to the spawned processes.
+        # PYTHONPATH will leak into the spawned processes.
         # In that case tests look for module in the wrong place (on PYTHONPATH).
         # Unless the test sets its own special env, pass a copy of the existing
         # environment with removed PYTHONPATH to the subprocesses.
--- a/setuptools/tests/environment.py
+++ b/setuptools/tests/environment.py
@@ -18,6 +18,19 @@ class VirtualEnv(jaraco.envs.VirtualEnv):
     def run(self, cmd, *args, **kwargs):
         cmd = [self.exe(cmd[0])] + cmd[1:]
         kwargs = {"cwd": self.root, **kwargs}  # Allow overriding
+        # In some environments (eg. downstream distro packaging), where:
+        # - tox isn't used to run tests and
+        # - PYTHONPATH is set to point to a specific setuptools codebase and
+        # - no custom env is explicitly set by a test
+        # that PYTHONPATH leaks to the spawned processes.
+        # In that case tests look for module in the wrong place (on PYTHONPATH).
+        # Unless the test sets its own special env, pass a copy of the existing
+        # environment with removed PYTHONPATH to the subprocesses.
+        if "env" not in kwargs:
+            env = dict(os.environ)
+            if "PYTHONPATH" in env:
+                del env["PYTHONPATH"]
+            kwargs["env"] = env
         return subprocess.check_output(cmd, *args, **kwargs)
 
 
--- a/setuptools/tests/fixtures.py
+++ b/setuptools/tests/fixtures.py
@@ -98,7 +98,18 @@ def venv(tmp_path, setuptools_wheel):
     env = environment.VirtualEnv()
     env.root = path.Path(tmp_path / 'venv')
     env.req = str(setuptools_wheel)
-    return env.create()
+    # In some environments (eg. downstream distro packaging),
+    # where tox isn't used to run tests and PYTHONPATH is set to point to
+    # a specific setuptools codebase, that PYTHONPATH leaks to the spawned
+    # processes.
+    # env.create() should install the just created setuptools
+    # wheel, but it doesn't if it finds another existing matching setuptools
+    # installation present on PYTHONPATH:
+    # `setuptools is already installed with the same version as the provided
+    # wheel. Use --force-reinstall to force an installation of the wheel.`
+    # This prevents leaking PYTHONPATH to the created environment.
+    with contexts.environment(PYTHONPATH=None):
+        return env.create()
 
 
 @pytest.fixture
--- a/setuptools/tests/fixtures.py
+++ b/setuptools/tests/fixtures.py
@@ -100,7 +100,7 @@ def venv(tmp_path, setuptools_wheel):
     env.req = str(setuptools_wheel)
     # In some environments (eg. downstream distro packaging),
     # where tox isn't used to run tests and PYTHONPATH is set to point to
-    # a specific setuptools codebase, that PYTHONPATH leaks to the spawned
+    # a specific setuptools codebase, PYTHONPATH will leak into the spawned
     # processes.
     # env.create() should install the just created setuptools
     # wheel, but it doesn't if it finds another existing matching setuptools
--- a/changelog.d/3133.misc.rst
+++ b/changelog.d/3133.misc.rst
@@ -1,0 +1,1 @@
+Enhanced isolation of tests using virtual environments - PYTHONPATH is not leaking to spawned subprocesses  -- by :user:`befeleme`


abravalheri:
@befeleme thank you very much for working on this.

There are somethings I don't understand: in the tests the venvs are supposed to have their own installation of setuptools, and it is part of the test the fact that they don't share the same setuptools installation...

When looking here:

https://github.com/pypa/setuptools/blob/6193a69d8608b6ca7e0e5ac1223d63abb55eee3b/setuptools/tests/environment.py#L19

and in [jaraco.envs](https://github.com/jaraco/jaraco.envs/blob/2d3b27254c97c792fc3f6cd352f6cb7001fdbdf5/jaraco/envs.py#L38-L40), the environment object seems to do the right thing to find the correct executable inside the virtual env...

When looking at:

https://github.com/pypa/setuptools/blob/6193a69d8608b6ca7e0e5ac1223d63abb55eee3b/setuptools/tests/fixtures.py#L100

https://github.com/pypa/setuptools/blob/6193a69d8608b6ca7e0e5ac1223d63abb55eee3b/setuptools/tests/fixtures.py#L109

and [here](https://github.com/jaraco/jaraco.envs/blob/2d3b27254c97c792fc3f6cd352f6cb7001fdbdf5/jaraco/envs.py#L33), and [here](https://github.com/jaraco/jaraco.envs/blob/2d3b27254c97c792fc3f6cd352f6cb7001fdbdf5/jaraco/envs.py#L29), I cannot understand how `setuptools` is not getting installed when it should, or why `setuptools` is getting installed when it should not not...

Can you elaborate a bit more on how have you reached the conclusion that missing env vars are the root cause of this problem?

setuptools/tests/test_distutils_adoption.py
```diff
@@ -53,7 +53,7 @@ def test_distutils_stdlib(venv):
     """
     Ensure stdlib distutils is used when appropriate.
     """
-    env = dict(SETUPTOOLS_USE_DISTUTILS='stdlib')
+    env = {**os.environ, **dict(SETUPTOOLS_USE_DISTUTILS='stdlib')}
```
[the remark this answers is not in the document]
abravalheri: Sorry, just a curiosity question here: do the `sysconfig` installation schemes affect `virtualenv` installation schemes? (Previously my thought was that they only affect system-wide installations)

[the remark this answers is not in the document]
abravalheri: If we are allowing other variables, I think we should be more specific here. For example it would not be nice to have things like `PYTHONHOME`, `PYTHONSTARTUP`, etc leaking...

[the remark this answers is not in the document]
hroncok: It shouldn't affect virtualenv, will recheck later. This part of the patch was written first, so maybe it is no longer necessary with the other part.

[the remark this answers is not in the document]
abravalheri: Thank you @hroncok!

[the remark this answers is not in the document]
befeleme: As @hroncok said, I'm checking the patch without changes in `test_distutils_adoption` in Fedora's build environment and it still works fine. It's redundant here, I will remove it in an amended commit.

abravalheri:
Thank you very much @befeleme for the explanation. Nice to see that we are starting to get to the bottom of the problems :)

Please correct me here if I am wrong, If I understand it correctly this problem will always happen if someone, instead of running the tests via `tox`, uses `PYTHONPATH` to point to a specific setuptools codebase, right? One `PYTHONPATH` is set, it will leak to the spawned processed when `env` is not explicitly passed.

I always think that making tests more isolated is a good thing, so removing items from the environment variables that are not explicitly set should be fine!

However, I would avoid removing items from the `env` dictionary when it is explicitly set. For whatever reason, tests might need to set `PYTHONPATH`, so I would like to avoid surprises for the developers.

setuptools/tests/environment.py
```diff
@@ -18,6 +18,13 @@ class VirtualEnv(jaraco.envs.VirtualEnv):
     def run(self, cmd, *args, **kwargs):
         cmd = [self.exe(cmd[0])] + cmd[1:]
         kwargs = {"cwd": self.root, **kwargs}  # Allow overriding
+        # If no environment was passed as an argument
+        # create a copy of the current env to use when spawning a subprocess
+        # Unset PYTHONPATH - tests in venv need only that venv's paths in sys.path
+        env = kwargs.get("env") or dict(os.environ)
```
abravalheri, line 24: Is it possible to not touch an explicitly given `env` here? In some scenarios tests might explicitly set `PYTHONPATH`.

setuptools/tests/environment.py
```diff
@@ -18,6 +18,13 @@ class VirtualEnv(jaraco.envs.VirtualEnv):
     def run(self, cmd, *args, **kwargs):
         cmd = [self.exe(cmd[0])] + cmd[1:]
         kwargs = {"cwd": self.root, **kwargs}  # Allow overriding
+        # If no environment was passed as an argument
+        # create a copy of the current env to use when spawning a subprocess
+        # Unset PYTHONPATH - tests in venv need only that venv's paths in sys.path
```
abravalheri, line 23: I think it would be nice to summarise the PR discussion in this comment and mention for which reason PYTHONPATH would be set in the first place, and how it affects the tests...

setuptools/tests/fixtures.py
```diff
@@ -98,7 +98,10 @@ def venv(tmp_path, setuptools_wheel):
     env = environment.VirtualEnv()
     env.root = path.Path(tmp_path / 'venv')
     env.req = str(setuptools_wheel)
-    return env.create()
+    # Don't pass PYTHONPATH from the current environment to the created venv
+    # It only needs its own paths in sys.path
```
abravalheri, line 102: I think it would be nice to also explain the reasoning here (summarise the PR discussion).

It is not obvious why `PYTHONPATH` cannot be in the environment when doing `env.create()`

abravalheri:
@befeleme there are somethings that are still not obvious to me, could you try to help me understand?

> test_distutils_local doesn't find .env in the path
> test_distutils_local_with_setuptools can't import setuptools, test_test_command_install_requirements can't import pkg_resources.

If `PYTHONPATH` is not an environment variable in the subprocess, why would these tests fail? The venv is supposed to have an installation of setuptools from the wheel/sdist right? And `SETUPTOOLS_USE_DISTUTILS="local"` should prevent `distutils` from being loaded from the standard library...

setuptools/tests/fixtures.py
hroncok: In this case, when PYTHONPATH is set during `create()` (which actually installs packages, despite the name that does not clearly indicate that), setuptools is not installed (pip says something along the lines of "the same version fo setuptools is already installed, skipping" and does not fail)

hroncok: I believe this also answers the second question from https://github.com/pypa/setuptools/pull/3133#issuecomment-1050081209

hroncok: Here is what happens exactly:

```
(__venv__) [tmp]$ pip list
Package    Version
---------- -------
pip        21.2.3
setuptools 60.9.3

(__venv__) $ pip install .../setuptools-60.9.3-py3-none-any.whl 
Processing .../setuptools-60.9.3-py3-none-any.whl
setuptools is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.

(__venv__) $ echo $?
0

```

befeleme:
> Please correct me here if I am wrong, If I understand it correctly this problem will always happen if someone, instead of running the tests via `tox`, uses `PYTHONPATH` to point to a specific setuptools codebase, right? Once `PYTHONPATH` is set, it will leak to the spawned processed when `env` is not explicitly passed.

Yes, this is what happens.

> However, I would avoid removing items from the `env` dictionary when it is explicitly set. For whatever reason, tests might need to set `PYTHONPATH`, so I would like to avoid surprises for the developers.

Yes, this is possible. I've run the test suite on a modified patch, when I only touch environment which hasn't been explicitly set by test and the tests pass. Which puzzled me at first, but then given the venv is created and installs package in a context unsetting `PYTHONPATH`, this makes sense, as it's already removed from the way.

Regarding the second problem, I think @hroncok's clarification better explains the point I've missed. 

I'll send the updated (and better explained) patch soon. Thank you for the well-pointed questions!

setuptools/tests/fixtures.py
abravalheri: Amazing! Thank you very much for the clarification @hroncok.
Now everything makes sense 😄

abravalheri:
Thank you very much @befeleme! Sorry for disturbing you with a lot of questions.

> I'll send the updated (and better explained) patch soon. Thank you for the well-pointed questions!

Does it make sense to install `setuptools` in the virtual environment with `-I` (to force the installation)? Will it make the solution more generic?

setuptools/tests/fixtures.py
```diff
@@ -98,7 +98,17 @@ def venv(tmp_path, setuptools_wheel):
     env = environment.VirtualEnv()
     env.root = path.Path(tmp_path / 'venv')
     env.req = str(setuptools_wheel)
-    return env.create()
+    # In some environments (eg. downstream distro packaging),
+    # where tox isn't used to run tests and PYTHONPATH is set to point to
+    # a specific setuptools codebase, that PYTHONPATH leaks to the spawned
+    # processes.
+    # env.create() should install the just created setuptools
+    # wheel, but it doesn't if it finds another wheel present on PYTHONPATH:
```
hroncok, line 106:
```suggestion
    # wheel, but it doesn't if it finds existing matching setuptools installation present on PYTHONPATH:
```

befeleme:
I have just sent the updated patch. The comments are quite verbose but I hope for a good clarity of the explanation (and I've never managed to keep it short, eh).

> Sorry for disturbing you with a lot of questions.

No, it's great! This discussion helped me understand the issue much deeper. :)

befeleme:
The last force-push invalidated the approved workflow. @abravalheri, could you please rerun it? I'm ready with the updates.

setuptools/tests/fixtures.py
```diff
@@ -98,7 +98,18 @@ def venv(tmp_path, setuptools_wheel):
     env = environment.VirtualEnv()
     env.root = path.Path(tmp_path / 'venv')
     env.req = str(setuptools_wheel)
-    return env.create()
+    # In some environments (eg. downstream distro packaging),
+    # where tox isn't used to run tests and PYTHONPATH is set to point to
+    # a specific setuptools codebase, that PYTHONPATH leaks to the spawned
```
abravalheri, line 103: Thank you very much @befeleme, that is a very great summary. 

```suggestion
    # a specific setuptools codebase, PYTHONPATH will leak into the spawned
```

I think it read better like this.

setuptools/tests/environment.py
```diff
@@ -18,6 +18,19 @@ class VirtualEnv(jaraco.envs.VirtualEnv):
     def run(self, cmd, *args, **kwargs):
         cmd = [self.exe(cmd[0])] + cmd[1:]
         kwargs = {"cwd": self.root, **kwargs}  # Allow overriding
+        # In some environments (eg. downstream distro packaging), where:
+        # - tox isn't used to run tests and
+        # - PYTHONPATH is set to point to a specific setuptools codebase and
+        # - no custom env is explicitly set by a test
+        # that PYTHONPATH leaks to the spawned processes.
```
abravalheri, line 25:
```suggestion
        # PYTHONPATH will leak into the spawned processes.
```

abravalheri (approved):
Thank you very much @befeleme, that is a very great work, and followed by an in depth-investigation!

setuptools/tests/fixtures.py
abravalheri: I will just go ahead and commit the suggestions, this way the CI can run all at once.

abravalheri:
@Mergefyio rebase

abravalheri:
@Mergifyio rebase

abravalheri:
I am sorry @befeleme, I was hopping that 5ae9aa41369b8b0c8e1710475988ac0e9e3cf431 would have fixed the problems with the CI but apparently we still have other problems to solve in the main branch.

Maybe this PR will take a bit longer to land (until we fix the problems in the main).

jaraco:
@mergifyio rebase

PR #3151, opened 2022-03-05
Fix editable --user installs with build isolation

hexagonrecursion:
## Summary of changes

Make `setup.py develop --user` install to the user site packages directory even if it is disabled in the current interpreter.

Closes https://github.com/pypa/setuptools/issues/3019

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3151.breaking.rst
+++ b/changelog.d/3151.breaking.rst
@@ -1,0 +1,1 @@
+Made ``setup.py develop --user`` install to the user site packages directory even if it is disabled in the current interpreter.
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -169,12 +169,8 @@ class easy_install(Command):
         self.install_data = None
         self.install_base = None
         self.install_platbase = None
-        if site.ENABLE_USER_SITE:
-            self.install_userbase = site.USER_BASE
-            self.install_usersite = site.USER_SITE
-        else:
-            self.install_userbase = None
-            self.install_usersite = None
+        self.install_userbase = site.USER_BASE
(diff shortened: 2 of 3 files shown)


abravalheri:
Thank you very much for investigating this @hexagonrecursion and providing a fix!

I changed a little bit your test code to avoid using directly the `py` library (according to their warning on https://pypi.org/project/py/, this library should not be used in new code). I hope you don't mind the changes (I just found it easy to do it directly instead of going back and forth requesting changes in the review).

abravalheri:
Thank you very much for investigating this @hexagonrecursion and providing a fix!

I changed a little bit your test code to avoid using directly the `py` library (according to their warning on https://pypi.org/project/py/, this library should not be used in new code). I hope you don't mind the changes (I just found it easy to do it directly instead of going back and forth requesting changes in the review).

abravalheri:
This approach seems reasonable to me, if the user is explicitly asking for installing in the home, we should honour that.

Question: if I understood it correctly we are still keeping the isolation in terms of not allowing libraries installed in the user home to be used during the build, right? (The only thing that changes is that we "break the isolation" just when it is time to install the `.pth` file)

hexagonrecursion:
> 

correct

abravalheri (approved):
Thank you very much @hexagonrecursion, this looks good!

hexagonrecursion:
@jaraco ping

jaraco (approved):
LGTM

kynan:
FTR: the first setuptools release to include this fix is 62.0.0 from 2022-04-04

PR #3161, opened 2022-03-09
Add pre-commit and use pyupgrade to upgrade to Python 3.7+ syntax

hhartzer:
## Summary of changes

Adds pre-commit and uses pyupgrade to upgrade to Python 3.7+ syntax.

Happy to remove pre-commit if this would be better one-off.


abravalheri:
Hi @hhartzer thank you very much for the PR.

Although I really like pre-commit in repositories, I think this PR will break all the PRs that are currently open... rebasing would be super painful. I am also not sure 

Is there any way to do it very, very gradually?

hhartzer:
No problem! Thanks for replying so quickly!

Unfortunately, there's no easy option to put in changes like this without impacting existing pull requests. The easiest way might be for someone with an open pull request to run `pyupgrade` on their own before rebasing.

Easiest way would be to copy the .`pre-commit-config.yaml` file manually and run `pre-commit run --all-files`, commit that, and then try to rebase.

There may be some other options and ideally pre-commit would run in CI to help maintain this. I will say that `pyupgrade` makes it really easy. When 3.7 gets EOLed, you can bump it to 3.8 and upgrade all of the syntax at once.

Another possibility might be to target specific folders at once that don't have open MRs. I guess the good thing though is that it's deterministic and once ran, should have no more conflicts (in theory!).

I'd appreciate any ideas anyone else has, as introducing code quality/formatting does have some unfortunate drawbacks.

agronholm:
Yeah, changes such as this will impact existing PRs but it's unavoidable unless all PRs are merged first, and I don't think that's very realistic. Contributors just have to deal with it.

akx:
Hmm... 3.7 didn't add any new syntax, did it? f-strings are a Python 3.6 feature.

You could also run https://github.com/ikamensh/flynt if you're f-stringing things; it tends to catch some instances pyupgrade doesn't. (I also tend to do a manual pass looking for `.format()` and `["'] %`; PyCharm's "convert to f-string" intention is very handy there.)

jaraco:
This PR makes changes to _distutils, which is maintained at [pypa/distutils](/pypa/distutils). Please add that also as an exclude.

setuptools/_distutils/command/bdist_dumb.py
```diff
@@ -90,7 +90,7 @@ def run(self):
 
         # And make an archive relative to the root of the
         # pseudo-installation tree.
-        archive_basename = "%s.%s" % (self.distribution.get_fullname(),
+        archive_basename = "{}.{}".format(self.distribution.get_fullname(),
```
jaraco, line 93: It looks like the tool breaks hanging indents. It's probably acceptable for codebases that use something like black to autoformat, but this project hasn't been able to adopt black yet, so it'll require manual reformatting of affected files.

setuptools/tests/test_windows_wrappers.py
```diff
@@ -108,7 +108,7 @@ def test_basic(self, tmpdir):
         ]
         proc = subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stdin=subprocess.PIPE)
-        stdout, stderr = proc.communicate('hello\nworld\n'.encode('ascii'))
+        stdout, stderr = proc.communicate(b'hello\nworld\n')
```
jaraco, line 111: I'm not sure I agree with this change. Perhaps I'm being nitpicky, but I chose to use explicit encoding instead of bytes literals for clarity and symmetry with the output handling. Perhaps better, though, would be to use `text=True` or `universal_newlines=True` to enable text mode in the subprocess.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+# Update with: pre-commit autoupdate --freeze
+
+exclude: '_vendor'
+
+repos:
+-   repo: https://github.com/asottile/pyupgrade
+    rev: e695ecd365119ab4e5463f6e49bea5f4b7ca786b  # frozen: v2.31.0
+    hooks:
+    -   id: pyupgrade
+        args: [--py37-plus]
```
jaraco, line 13: Normally, I like to manage changes like these in [jaraco/skeleton](/jaraco/skeleton). I'm not confident, however, that this change could be accepted there. Is this change a best practice or a setuptools-specific change? How about let's treat this concern (a pre-commit hook) separately from the code changes?

setuptools/tests/test_windows_wrappers.py
jaraco: See a56be0e5, where I'm experimenting with just using `text=True`.

.pre-commit-config.yaml
agronholm: I'm not sure why the author pinned this tool to an old version, but I at least don't consider that to be a best practice. Or do you mean using pyupgrade (in pre-commit) in general? That I do in all of my new repositories.

jaraco: I mean pyupgrade and pre-commit in general. In the skeleton, I've added 'black' as a pre-commit hook, but I removed blacken docs (https://github.com/jaraco/skeleton/commit/719a7ced8a1713b7fe94d842a8f6fec7425b8a0a). Last night, I explored the possibility of using pyupgrade, but found it [doesn't support compatible changes](https://github.com/asottile/pyupgrade/issues/659). Hence, I'm reluctant to apply it as a best practice or even in this project. I don't want to enforce a behavior that's known unsafe and hostile.

jaraco: In other projects, I've started applying pyupgrade at strategic times and not used any pre-commit hook. Let's abandon that concern for now and focus on simply running the upgrade.

jaraco:
Given this concern is stale, I'm going to close it, but I'd like to see it applied at some time that's good for the project. It may be worthwhile to apply black first.

Issue #3166, opened 2022-03-10
[BUG] wheel install_as_egg does not honor file mode


delijati:
### setuptools version

 setuptools.__version__ '60.2.0'

### Python version

>= 3.8

### OS

Manjaro

### Description

Files installed via `Wheel.install_as_egg` have the same file mode as installed via `pip install`

### Expected behavior

Files installed via `Wheel.install_as_egg` have the same file mode as installed via `pip install`

### How to Reproduce

Install via pip:
```
virtuelenv env
env/bin/pip install kaleido
```
Check mode:
```console
❯ stat -c "%a" env/lib/python3.8/site-packages/kaleido/executable/kaleido                                         
755
```

Install via `install_as_egg`:
```
import setuptools.wheel

if __name__ == "__main__":
    whl = setuptools.wheel.Wheel("kaleido-0.2.1-py2.py3-none-manylinux1_x86_64.whl")
    whl.install_as_egg("./foo")
```

Check mode:
```console
❯ stat -c "%a" foo/kaleido/executable/kaleido                 
644
```

### Output

```console
# ABOVE
```

delijati:
Seams to be a general python issue in `zipfile` -> https://stackoverflow.com/questions/39296101/python-zipfile-removes-execute-permissions-from-binaries

PR #3167, opened 2022-03-11
 [BUG] wheel install_as_egg does not honor file mode

delijati:
### Pull Request Checklist
- [x] Changes have tests
- [?] News fragment added in [`changelog.d/`].

[PR docs]:

See https://github.com/pypa/setuptools/issues/3166 for more info.

### DIFF
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -6,6 +6,8 @@
 from distutils.sysconfig import get_config_var
 from distutils.util import get_platform
 import contextlib
+import pathlib
+import stat
 import glob
 import inspect
 import os
@@ -614,3 +616,85 @@ def test_wheel_is_compatible(monkeypatch):
     monkeypatch.setattr('setuptools.wheel.sys_tags', sys_tags)
     assert Wheel(
         'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
+
+
+def test_wheel_mode():
+    @contextlib.contextmanager
+    def build_wheel(extra_file_defs=None, **kwargs):
+        file_defs = {
+            'setup.py': (DALS(
+                '''
+                # -*- coding: utf-8 -*-
+                from setuptools import setup
+                import setuptools
+                setup(**%r)
+                '''
+            ) % kwargs).encode('utf-8'),
+        }
+        if extra_file_defs:
+            file_defs.update(extra_file_defs)
+        with tempdir() as source_dir:
+            path.build(file_defs, source_dir)
+            runsh = pathlib.Path(source_dir) / "script.sh"
+            os.chmod(runsh, 0o777)
+            subprocess.check_call((sys.executable, 'setup.py',
+                                   '-q', 'bdist_wheel'), cwd=source_dir)
+            yield glob.glob(os.path.join(source_dir, 'dist', '*.whl'))[0]
+
+    params = dict(
+        id='script',
+        file_defs={
+            'script.py': DALS(
+                '''
+                #/usr/bin/python
+                print('hello world!')
+                '''
+            ),
+            'script.sh': DALS(
+                '''
+                #/bin/sh
+                echo 'hello world!'
+                '''
+            ),
+        },
+        setup_kwargs=dict(
+            scripts=['script.py', 'script.sh'],
+        ),
+        install_tree=flatten_tree({
+            'foo-1.0-py{py_version}.egg': {
+                'EGG-INFO': [
+                    'PKG-INFO',
+                    'RECORD',
+                    'WHEEL',
+                    'top_level.txt',
+                    {'scripts': [
+                        'script.py',
+                        'script.sh'
+                    ]}
+
+                ]
+            }
+        })
+    )
+
+    project_name = params.get('name', 'foo')
+    version = params.get('version', '1.0')
+    install_tree = params.get('install_tree')
+    file_defs = params.get('file_defs', {})
+    setup_kwargs = params.get('setup_kwargs', {})
+
+    with build_wheel(
+        name=project_name,
+        version=version,
+        install_requires=[],
+        extras_require={},
+        extra_file_defs=file_defs,
+        **setup_kwargs
+    ) as filename, tempdir() as install_dir:
+        _check_wheel_install(filename, install_dir,
+                             install_tree, project_name,
+                             version, None)
+        w = Wheel(filename)
+        script_sh = pathlib.Path(install_dir) / w.egg_name() / "EGG-INFO" / "scripts" / "script.sh"
+        assert script_sh.exists()
+        assert oct(stat.S_IMODE(script_sh.stat().st_mode)) == "0o777"
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -618,7 +618,6 @@ def test_wheel_is_compatible(monkeypatch):
         'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
 
 
-@pytest.mark.skipif(sys.platform == 'win32', reason='non-Windows only')
 def test_wheel_mode():
     @contextlib.contextmanager
     def build_wheel(extra_file_defs=None, **kwargs):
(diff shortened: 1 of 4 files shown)


delijati:
@abravalheri can you take a look :)

abravalheri:
@Mergifyio rebase

abravalheri:
Hi @delijati thank you very much for creating the PR.
I can try to have a look, but this is one of the areas of `setuptools` that I am not that familiar.

setuptools/tests/test_wheel.py
```diff
@@ -614,3 +616,85 @@ def sys_tags():
...
+        install_requires=[],
+        extras_require={},
+        extra_file_defs=file_defs,
+        **setup_kwargs
+    ) as filename, tempdir() as install_dir:
+        _check_wheel_install(filename, install_dir,
+                             install_tree, project_name,
+                             version, None)
+        w = Wheel(filename)
+        script_sh = pathlib.Path(install_dir) / w.egg_name() / "EGG-INFO" / "scripts" / "script.sh"
```
abravalheri, line 698:
```suggestion
        base = script_sh = pathlib.Path(install_dir) / w.egg_name()
        script_sh = base / "EGG-INFO" / "scripts" / "script.sh"
```

setuptools/tests/test_wheel.py
```diff
@@ -614,3 +616,86 @@ def sys_tags():
...
+        install_requires=[],
+        extras_require={},
+        extra_file_defs=file_defs,
+        **setup_kwargs
+    ) as filename, tempdir() as install_dir:
+        _check_wheel_install(filename, install_dir,
+                             install_tree, project_name,
+                             version, None)
+        w = Wheel(filename)
+        base = script_sh = pathlib.Path(install_dir) / w.egg_name()
```
abravalheri, line 698: Sorry, I introduced this by accident

```suggestion
        base = pathlib.Path(install_dir) / w.egg_name()
```

abravalheri:
Hi @delijati, I did change the PR a little bit for the following reasons:

- I split a line that was too long to prevent flake8 from failing
- I reordered the commits so I could have a look on the tests failing prior to the fix being implemented.

I hope you don't mind these changes. Please feel free to revert if that works better for you.

abravalheri (commented):
Thank you very much for the PR. I had a look on the code and it looks good. I just left a few notes bellow:

#### General remarks about the reported issue/proposed fix:

- The `install_as_egg` interface is a non-public interface used internally by setuptools.
  This function, as the name suggests, is very likely to be deprecated, but is currently being used by `easy_install`.
- `easy_install` despite being deprecated, is still an internal mechanism used by setuptools for editable installs.
- In this context, this PR makes sense, since it can fix potential issues for users using editable installs.

#### Review comments:

- The CI output seems to indicate that this implementation is OS-specific. Is this approach safe to use on Windows environments? What would be the impact for the users?
- The implementation overrides a non-public API of `ZipFile`.
  I understand that the method is not completely private, but this also means that we cannot rely on this approach being stable across future Python versions.
  Is this a well-known approach? Is it documented somewhere or used by other big project in the Python community?
  I wonder if instead we could use the already existing code in: https://github.com/pypa/setuptools/blob/98728b103a5c5940abbc2fab60ccfc5fa265d280/setuptools/archive_util.py#L91

delijati:
> Thank you very much for the PR. I had a look on the code and it looks good. I just left a few notes bellow:

Hi thanks for looking into that PR.

> #### General remarks about the reported issue/proposed fix:
> 
>     * The `install_as_egg` interface is a non-public interface used internally by setuptools.

I know but for now it is the only solution and already used here: https://github.com/buildout/buildout/issues/604

>       This function, as the name suggests, is very likely to be deprecated, but is currently being used by `easy_install`.
> 
>     * `easy_install` despite being deprecated, is still an internal mechanism used by setuptools for editable installs.
> 
>     * In this context, this PR makes sense, since it can fix potential issues for users using editable installs.
> 
> 
> #### Review comments:
> 
>     * The CI output seems to indicate that this implementation is OS-specific. Is this approach safe to use on Windows environments? What would be the impact for the users?

I would skip the test on windows (will add that) as it does "nothing" there. [3]

>     * The implementation overrides a non-public API of `ZipFile`.
>       I understand that the method is not completely private, but this also means that we cannot rely on this approach being stable across future Python versions.
>       Is this a well-known approach? Is it documented somewhere or used by other big project in the Python community?

It is an old Python Bug that seams to be fixed in >= 3.11 [1] And fixes for it are all over the place [2]

>       I wonder if instead we could use the already existing code in: https://github.com/pypa/setuptools/blob/98728b103a5c5940abbc2fab60ccfc5fa265d280/setuptools/archive_util.py#L91

I looked into it but that would mean to change the api of all call used by `install_as_egg`. So i would rather stick with my solution.

[1] https://bugs.python.org/issue15795
[2] https://stackoverflow.com/questions/39296101/python-zipfile-removes-execute-permissions-from-binaries/54748564
[3] "Note Although Windows supports [chmod()](https://docs.python.org/2/library/os.html#os.chmod), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored."

abravalheri (approved):
Thank you very much @delijati for working in the PR and tests.

I think this change should be mostly fine since it is just applying the same existing behaviour that already exists in setuptools.

abravalheri:
Hi @delijati, would you like to add a [news fragment](https://setuptools.pypa.io/en/latest/development/developer-guide.html#adding-change-notes-with-your-prs) for the CHANGELOG?

It think this can be considered a bugfix for the fact that `scripts` were loosing their "executable mode" after installed with `pip install -e .`

delijati:
> Hi @delijati, would you like to add a [news fragment](https://setuptools.pypa.io/en/latest/development/developer-guide.html#adding-change-notes-with-your-prs) for the CHANGELOG?

Done

abravalheri:
Hi @befeleme, I was wondering if a change like this would impact OS packagers. Since you seem to be working with packaging at Red Hat, do you have any thoughts about that?

Here we tried to re-use some existing code that handles the unpacking of the wheels, which I honestly feel like it is the right thing to do.

However this existing code does [explicitly disallow](https://github.com/pypa/setuptools/blob/main/setuptools/archive_util.py#L107-L108) file names in wheels that start with `/` or `..`.

Do you think there might be any existing packages that are forcing file names inside the zip file to install files at arbitrary locations?

I had a look on the `zip` specification and it explicitly disallows names starting with `/` ([section 4.4.17](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT)), but we never know...

befeleme:
Hi @abravalheri,
we've discussed the issue briefly with @hroncok who has a far exceeding expertise regarding the Fedora ecosystem. 
It looks like the affected code is only invoked in `install_as_egg` function. In our RPM build process, we don't install wheels using setuptools but with pip, so AFAIK from the downstream's packager perspective the change shall not affect us.

> Do you think there might be any existing packages that are forcing file names inside the zip file to install files at arbitrary locations?

If such packages exist, we believe it'd better if they were forbidden to do that. So this feels as an improvement.

abravalheri:
Thank you very much for looking this up @befeleme!

I am inclined to merge this PR not exactly in the next release but in the following one, unless the other maintainers/OS re-packagers have any objections.

abravalheri:
Hi @mgorny, @FFY00 I just would like to double check with you if the changes introduced here will somehow impact negatively on Gentoo, Archlinux or any other OS/distribution you help to maintain.

This change will affect:
- uses that still rely on the legacy `setup.py install`
- uses that rely on `setup.py develop` / `pip install -e .` (for projects that use setuptools).

What the change will do:
- Preserve file mode when extracting a wheel file
- Error-out if the wheel contain absolute file paths or "../" (I am not really sure those can be be found in wheels due to the wheel and zip standards).

abravalheri:
Thank you very much @mgorny for the insights.

Regarding the confusion: I am under the impression that `setup.py install` uses `easy_install` under the hood so there might be some shared behaviour. But yes they might not be related at all as you pointed out.

FFY00:
I think a wheel is generated internally and then installed as a egg in `setup.py install`, right? Otherwise I don't see how wheel has to do with this.
Even is this is the case, the wheel would be generated by setuptools, so it has control of how it is generated.

Wasn't `easy_install` removed already?

abravalheri:
Thank you very much @FFY00 for having a look. So I think it is safe to assume at this point that the change does not affect OS distributions.

> Wasn't easy_install removed already?

`easy_install` is no longer public, but still used internally for `setup.py install/develop`.

delijati:
> I think a wheel is generated internally and then installed as a egg in `setup.py install`, right? Otherwise I don't see how wheel has to do with this. Even is this is the case, the wheel would be generated by setuptools, so it has control of how it is generated.
> 
> Wasn't `easy_install` removed already?

The internal `easy_install` is still used here https://github.com/buildout/buildout/issues/604 that's where the error occurred

FFY00:
Ah, I see! That makes much more sense to me. I am still not sure if `install_as_egg` is used by `setup.py install`, but this patch shouldn't really have any negative impact anyway since we are not working with external wheels.

Issue #2825, opened 2021-10-23
Stop vendoring packages


jaraco:
The current implementation of Setuptools has its dependencies (separately for `setuptools` and `pkg_resources`) vendored into the package. The project implemented this approach because of the bootstrapping problem.

# Bootstrapping Problem

Setuptools extended distutils, which was designed to sit at the root of a packaging system and presumed to be present and without any dependencies. Over time, Setuptools superseded distutils but inherited these constraints. In particular, when a system integrator wishes to build a system from sources, it requires a directed acyclic graph (DAG) of both build-time and runtime dependencies. As a result, Setuptools cannot depend on any package that uses Setuptools to build (or whose dependencies require Setuptools to build). This bootstrapping problem includes Setuptools itself, as it requires itself to build.

# Vendoring

As the ecosystem grew more complex, with standards-based implementations (such as packaging) appearing as third-party packages and not available in the standard library, Setuptools found itself requiring dependencies, but because of the bootstrapping problem, Setuptools adopted a vendoring strategy (copying the packages directly into the project) as a means of requiring that functionality.

However, this approach creates constraints and complications to the project:

- Not all packages can be vendored.
  - If the other package has vendored dependencies, those may not work in a vendored context.
  - Some packages have global state or modify global state or have interfaces that are reliant on the package layout (incl. pkg_resources, importlib_metadata, packaging), leading to unexpected failures when loaded in a global context.
- Refactoring functionality out of the library is difficult if not impossible due to the constraints above. In particular, this project would like to move `pkg_resources` into a separate package, but even though `pkg_resources` has no dependency on `setuptools` to run, it still must vendor its own dependencies (is this true?).
- When vendoring a package, it often is required to be rewritten to accommodate vendoring. Any absolute imports must be replaced by relative imports.
- Because vendoring is a second-class approach to dependency management (and unsustainable in the general case), it often requires specialized tooling to manage the dependencies and this management can often fall in conflict with the first-class tools.
- When vendoring dependencies, it's the responsibility of the hosting package to re-write imports to point to the vendored copies, creating non-standard usage with sometimes unclear semantics.
- Due to the constraints above, adding a new dependency can be an onerous process, requiring extra care and testing that may break in downstream environments whose workflows aren't proven.
- Because vendored dependencies are in fact de-facto satisfied, the project cannot and should not declare those dependencies as other projects do. Therefore, it's not possible to inspect the dependencies readily as one would with standard declarations.
[...]

PR #3170, opened 2022-03-13
Vendor nspektr. Utilize it in Distribution._install_dependencies.

jaraco:
## Summary of changes

In https://github.com/python/importlib_metadata/issues/368, I was exploring ways to make entry point dependency resolution available outside of Setuptools. In doing so, I explored creating the [nspektr project](/jaraco/nspektr). In creating that project and writing its tests, I discovered a flaw in how `EntryPoint.extras` resolved (https://github.com/python/importlib_metadata/issues/369), a flaw that affects [Setuptools installation logic](https://github.com/pypa/setuptools/blob/00fbad0f93ffdba0a4d5c3f2012fd7c3de9af04d/setuptools/dist.py#L874-L896).

While the flaw is fixed in a late release of `importlib_metadata`, there are likely older versions of `importlib_metadata` as well as older versions of `importlib.metadata` without the fix (at the time of this writing, the bug still exists in all Python versions). I've used nspektr to provide a workaround for the flaw (https://github.com/jaraco/nspektr/commit/7245f0716bfca38dcf2c59cbdb5196422556c4f7).

This PR replaces the inline logic in Setuptools with the more thoroughly tested logic of nspektr, adopting the workaround for the broken extras handling and simplifying Setuptools own logic.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


hroncok:
I am a little worried about yet another vendored dependency, this time even a project without a meaningful readme.

jaraco:
I respect that. I can give the project a readme.

Ultimately, I'd like to move all vendored dependencies to regular dependencies, though I'm not sure that's possible any time soon.

hroncok:
Sure, moving all vendored dependencies out is a meaningful goal. But it used to be just a handful, now we have plenty. Let's not kee adding new ones before we move all vendored dependencies to regular dependencies?

jaraco:
> Let's not keep adding new ones before we move all vendored dependencies to regular dependencies?

I'd originally set out to add this functionality to importlib.metadata, but couldn't do that because importlib.metadata can't depend on packaging. Similarly, I couldn't add it to packaging because it can't depend on importlib.metadata. So another package was necessary if the functionality was going to have use outside of setuptools (which I'd intended for it to do since I discourage use of setuptools except as a backend). Furthermore, I was in no rush to replace the behavior as found in Setuptools with the behavior in nspektr until the creation of nspektr and its tests revealed a flaw in Setuptools' own implementation.

Then I was left with the choice to either copy the fix into setuptools and maintain two copies of the same functionality or vendor the package and rely on the same copy. I've got far too little time to be able to afford maintaining multiple forks of the same thing, so vendoring it was. I really feel like I have little choice.

I don't think it's been said, though, what's the harm in having an additional package vendored?

abravalheri:
> Let's not keep adding new ones before we move all vendored dependencies to regular dependencies?

Hi @hroncok , I think I can appreciate the concerns here, however there are some extra dependencies planned to be vendored in the near future to support features like pyproject.toml metadata (e.g. `tomli`). These have been proposed a few months ago. I hope that is OK.

hroncok:
Well, I am no authority here. I've made my point, but I am not willing to die on this hill :)

jaraco:
Miro, sorry if I've upset you. I had hoped that Setuptools could add dependencies as needed without adding complications. It seems from your sentiment that adding dependencies creates concerns for you. In #2825, I laid out the reasons why Setuptools wishes not to vendor dependencies, but none of my reasons seem particularly pertinent to a user or downstream packager. I would genuinely like to hear you articulate what motivated you to express your opinion, as that could influence how we think about new dependencies going forward.

hroncok:
Nah, I'm not upset. Just a bit confused about what setuptools is turning into.

With more vendored dependencies, setuptools grows bigger and bigger every release. We had a regression test asserting the wheel is no bigger than 600 KiB and we had to recently relax that requirement. The overall license of this package gets more complicated. There are more LICENSE files that we probably should ship with the package. The overall license of Python gets more complex, as it includes setuptools wheel...

(Most of this has already happened prior to this PR, e.g. this one particular dependency does not increase the size much or does not add yet another different license, but I speak generally about new vendored deps.)

It seems to me like this project has found itself smashed between 2 different forces. On one hand, it is the default legacy way how to build packages when you install them with pip or with `python setup.py ...` invocation. On other hand, it shall be a modern PEP 517 backend with dependencies, toml support and whatnot. Maybe it shouldn't be both?

Issue #1688, opened 2019-02-17
Discussion: support for pyproject.toml configuration


pganssle:
Over the last few months, I've been using `setup.cfg` more, and I find the syntax and types and such to be super counter-intuitive. I don't find TOML to be a *breeze* per se, but it's at least a lot clearer to me how to specify a string, a list, etc. We also hear a lot of cries to move towards supporting `pyproject.toml` as the "one file" for setup configuration, in favor of `setup.cfg`. I'm not terribly opposed to this, but I want to open up the discussion for a few options:

1. Add `pyproject.toml` as the *one true way* to do declarative builds, moving all `setuptools` configuration over there and deprecating `setup.cfg` (partially what is discussed in #1160, and if we decide on that option we can move over to that issue).
2. Add `pyproject.toml` as a *second supported way* to do a declarative configuration, alongside `setup.cfg`.
3. Stay the course with `setup.cfg` indefinitely.

I am deeply concerned with the large amount of churn that we're introducing into the Python packaging space, so I don't want to make a lot of changes for the sake of changes, but I do think that at least having the *option* to use a `pyproject.toml` would be a real benefit, so I'm basically in favor of option 2.

I think the biggest downside of option 2 is that it means maintaining, testing and documenting *three separate locations* for all of our options - how to do it in `setup.cfg`, how to do it in `pyproject.toml`, and how to do it in `setup.py`. I think the hard part is going from 1 way to do everything to 2 ways. Our documentation is a mess anyway, but my beautiful vision for the future is that all our examples have "tabs" that allow you to toggle them between `setup.cfg` and `setup.py`. Adding a `pyproject.toml` in there would probably not be the worst thing.

jaraco:
I'd prefer option 1. I imagine during the deprecation period that the setup.cfg documentation could be removed, but a permalink to the old documentation made available for projects still relying on that technique.

One way to ease the transition could be for the `setup.cfg` code to _generate_ a .toml file and consume then have the `pyproject.toml` consumer just load both files. That way, a project seeking to transition would only need to do one build, manually merge the two .toml files, and delete `setup.cfg`.

My second choice would be option 3. I'd really like to avoid supporting multiple formats that do essentially the same thing.

[...]

PR #3253, opened 2022-04-06
Allow file: for requires statements in setup.cfg

### LINKED ISSUE #1951: Please allow "file:" for setup.cfg install_requires
The new declarative setup.cfg syntax is fabulous.  It's wonderful to have two-line setup.py files!

I tried to use "install_requires = file: requirements/base.in" and found that "file:" wasn't supported there.  Our pip-compile workflow works really well, and it would be great to be able to start with base.in as our source of truth, and easily use it in our setup.cfg.

As a design point, I'm not sure it makes sense to special-case which fields can use which kinds of info-gatherers.  Why can't I use "file:" for any field?  I can understand if the implementation requires it, but the more the developer can choose how to structure their own world, the better.

akx:
## Summary of changes

As requested and discussed in #1951, this PR enables using `file:` in <del>the various `requires` statements</del> `install_requires` and `extras_require` statements in setup.cfg.

As @jaraco mentioned in https://github.com/pypa/setuptools/issues/1951#issuecomment-574737840, this adds a (small) admonition advising against library packagers pinning their dependencies too tightly with this feature.

Closes #1951

### Pull Request Checklist

- [x] Changes have tests
- [x] Changes are documented 

### DIFF
--- a/docs/userguide/declarative_config.rst
+++ b/docs/userguide/declarative_config.rst
@@ -211,13 +211,13 @@ obsoletes                                          list-comma
 Options
 -------
 
-=======================  ===================================  =============== =========
+=======================  ===================================  =============== ====================
 Key                      Type                                 Minimum Version Notes
-=======================  ===================================  =============== =========
+=======================  ===================================  =============== ====================
 zip_safe                 bool
 setup_requires           list-semi                            36.7.0
-install_requires         list-semi
-extras_require           section                                              [#opt-2]_
+install_requires         file:, list-semi                                     [#opt-6]_
+extras_require           file:, section                                       [#opt-2]_, [#opt-6]_
 python_requires          str                                  34.4.0
 entry_points             file:, section                       51.0.0
 scripts                  list-comma
@@ -232,7 +232,7 @@ exclude_package_data     section
 namespace_packages       list-comma                                           [#opt-5]_
 py_modules               list-comma                            34.4.0
 data_files               section                              40.6.0          [#opt-4]_
-=======================  ===================================  =============== =========
+=======================  ===================================  =============== ====================
 
 **Notes**:
 
@@ -265,6 +265,11 @@ data_files               section                              40.6.0          [#
 .. [#opt-5] ``namespace_packages`` is deprecated in favour of native/implicit
    namespaces (:pep:`420`). Check :doc:`the Python Packaging User Guide
    <PyPUG:guides/packaging-namespace-packages>` for more information.
+
+.. [#opt-6] ``file:`` directives for reading requirements are supported since version 63.0.
+   The format for the file is basically the same as for a ``requirements.txt`` file.
+   Library developers should avoid tightly pinning their dependencies to a specific
+   version (e.g. via a "locked" requirements file).
 
 
 Compatibility with other tools
--- a/docs/userguide/declarative_config.rst
+++ b/docs/userguide/declarative_config.rst
@@ -267,7 +267,9 @@ data_files               section                              40.6.0          [#
    <PyPUG:guides/packaging-namespace-packages>` for more information.
 
 .. [#opt-6] ``file:`` directives for reading requirements are supported since version 63.0.
-   The format for the file is basically the same as for a ``requirements.txt`` file.
+   The format for the file resembles a ``requirements.txt`` file,
+   however please keep in mind that all non-comment lines must conform with :pep:`508`
+   (``pip``-specify syntaxes, e.g. ``-c/-r/-e`` flags, are not supported).
    Library developers should avoid tightly pinning their dependencies to a specific
    version (e.g. via a "locked" requirements file).
 
--- a/changelog.d/3253.change.rst
+++ b/changelog.d/3253.change.rst
@@ -1,0 +1,1 @@
+Enabled using ``file:`` for requirements in setup.cfg -- by :user:`akx`
--- a/setuptools/config/setupcfg.py
+++ b/setuptools/config/setupcfg.py
@@ -568,15 +568,27 @@ class ConfigOptionsHandler(ConfigHandler["Distribution"]):
         self.root_dir = target_obj.src_root
         self.package_dir: Dict[str, str] = {}  # To be filled by `find_packages`
 
+    @classmethod
+    def _parse_list_semicolon(cls, value):
+        return cls._parse_list(value, separator=';')
+
+    def _parse_file_in_root(self, value):
+        return self._parse_file(value, root_dir=self.root_dir)
+
+    def _parse_requirements_list(self, value):
+        # Parse a requirements list, either by reading in a `file:`, or a list.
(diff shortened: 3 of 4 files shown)


abravalheri (commented):
Hi @akx, thank you very much for analysing the previous discussion and providing the PR.

I will leave the review to the  maintainers that were participating in the original discussion, but I would like to note that this implementation does not seem to be *forward* compatible, because it only tackles `setup.cfg`.

According to https://github.com/pypa/setuptools/issues/1688#issuecomment-471212342, the long term view is to compile `setup.cfg` into `pyproject.toml`, therefore it would be necessary to support this feature in `pyproject.toml` as well.

akx:
> [...] the long term view is to compile `setup.cfg` into `pyproject.toml`, therefore it would be necessary to support this feature in `pyproject.toml` as well.

Ah, sure thing! I can also take a look at getting that in here, if the approach seems otherwise sound by maintainers :)

abravalheri:
Thank you very much for considering that @akx. Yes, it is definitely good to wait for the feedback before doing any change.

jaraco:
I’m okay to give it a try.

akx:
@abravalheri Hmm, by the looks of it enabling something like 
```
[tool.setuptools.dynamic.optional-dependencies.bonus]
file = ["requirements-bonus.txt"]
```
(along the lines of https://github.com/pypa/setuptools/blob/e5552d3c42eddc463788861ae4c8847183eeec1c/setuptools/tests/config/test_pyprojecttoml.py#L66-L71)

would require changes to the JSON Schema for pyproject.toml, which... apparently is maintained in `validate_pyproject`? Am I on the right track there?

If so, maybe all that should be done in a separate constellation of PRs..?

abravalheri:
Hi @akx, If you change the code in setuptools, I am more than happy to follow up and tackle the schemas myself.

For the time being, in the tests you can use the `ignore_option_errors` parameter to bypass the validation. Once I modify the schemas, I will also modify that.

akx:
@abravalheri Please see #3255 for the TOML case :)

docs/userguide/declarative_config.rst
```diff
@@ -192,28 +192,28 @@ obsoletes                                          list-comma
...
-=======================  ===================================  =============== =========
+=======================  ===================================  =============== ====================
 Key                      Type                                 Minimum Version Notes
-=======================  ===================================  =============== =========
+=======================  ===================================  =============== ====================
 zip_safe                 bool
-setup_requires           list-semi                            36.7.0
-install_requires         list-semi
-extras_require           section                                              [#opt-2]_
+setup_requires           file:, list-semi                     36.7.0          [#opt-5]_
```
abravalheri, line 199: Question: would it make more sense to restrict this change to `install_requires` and `extras_requires`?

`tests_require` is deprecated, so I feel like we should not touch it.

`setup_requires` is largely obsoleted by PEP 517.
In very specific use cases people might still reach for it inside `setup.py`, but not in the context of `setup.cfg`.

I understand that adding support for the `file:` directive for `setup_requires` and `tests_require` basically come for free, but I still think we should refrain from doing that.

abravalheri, line 199: Question: would it make more sense to restrict this change to `install_requires` and `extras_requires`?

`tests_require` is deprecated, so I feel like we should not touch it.

`setup_requires` is largely obsoleted by PEP 517.
In very specific use cases people might still reach for it inside `setup.py`, but not in the context of `setup.cfg`.

I understand that adding support for the `file:` directive for `setup_requires` and `tests_require` basically come for free, but I still think we should refrain from doing that.

abravalheri, line 199: Question: would it make more sense to restrict this change to `install_requires` and `extras_requires`?

`tests_require` is deprecated, so I feel like we should not touch it.

`setup_requires` is largely obsoleted by PEP 517.
In very specific use cases people might still reach for it inside `setup.py`, but not in the context of `setup.cfg`.

I understand that adding support for the `file:` directive for `setup_requires` and `tests_require` basically come for free, but I still think we should refrain from doing that.

akx: I get where you're coming from here, but various comments in #1951 say `file:` support is the last thing preventing them from moving from an imperative `setup.py` to a PEP 518 `setup.cfg` build. So, if those folks had been using files for their `setup_requires` and `tests_require`, not allowing `file:` for them would make that porting work less straightforward.

I have no other strong opinion on this, just that thought :)

abravalheri: Thank you very much @akx for pointing this out. I re-read the discussion in issue 1951 and I did not see any special request for `setup_requires` and `tests_requires`.

My view is that `tests_requires`, already does not work, so I don't think it would be helping anyone interested in the migration.

On the other hand, `setup_requires` only have value (right now) when computed dynamically via `setup.py`.

However I can see people starting to use it once it gets `file:` support just to workaround limitations in whichever GitHub bots they might be using for dependency management. In my opinion that would be an unfortunate side effect, since PEP 518 should be preferred.

---

On a side note, considering the forward compatibility with `pyproject.toml`:

- Currently `setup_requires` can be automatically converted to `[build-system] requires` by an automatic generator.
- If `setup_requires` gains the capability of being dynamically defined via `file:`, that will no longer be possible and we will have to update #3255 to handle this new use case.

akx:
@abravalheri @jaraco I rebased this to get rid of the merge conflicts. What are the next steps to get this merged..?

abravalheri:
Thank you very much @akx for rebasing the PR.

Personally, I have some strong reservations against handling dynamic `tests_require` and `setup_requires` (the main reasons are described in my previous comments https://github.com/pypa/setuptools/pull/3253#discussion_r846075228 and https://github.com/pypa/setuptools/pull/3253#discussion_r846242876).

Can we restrict the change to `install_requires` and `extras_require`?

akx:
@abravalheri Okiedoke :) Removed the `file:` stuff related to `tests_require` and `setup_requires` (the code is still gently touched in `setupcfg.py` since `parse_list_semicolon` became a classmethod).

abravalheri:
Thank you very much @akx and sorry for the trouble.
Since Jason seems to be OK with the change, I will try to merge the PRs yet this week.
(I am currently focusing on organising changes in the documentation and I had in the backlog the PEP 660, but I will try to make my best to not delay even more this merge).

abravalheri (approved):

abravalheri:
Sorry for the delay @akx.

I merged the changes and published a new version of setuptools.
For the time being, I have added the status of *beta*, then we can see the feedback from the community.

nedbat:
Thanks for this! :)

pganssle:
I'm still pretty negative on this "feature".

henryiii:
Just say my first usage of this in the wild: https://github.com/Flexget/Flexget/blob/f55a297fae18e9779d9df9e545974727ad39b07a/requirements.txt :)

PR #3255, opened 2022-04-07
Allow file: for dependencies in TOML

akx:
## Summary of changes

This is the pyproject.toml counterpart of #3253, as requested by @abravalheri in https://github.com/pypa/setuptools/pull/3253#pullrequestreview-933489932.

* The `_validate_pyproject` code was generated using https://github.com/akx/validate-pyproject/commit/61b668f

### Pull Request Checklist
- [x] Changes have tests
- [x] News fragment added in `changelog.d/`
- [x] Docs added
- [ ] JSON schema changes merged to `validate-pyproject`(https://github.com/abravalheri/validate-pyproject/pull/37)


setuptools/config/pyprojecttoml.py
```diff
@@ -365,6 +374,35 @@ def _obtain_classifiers(self, dist: "Distribution"):
...
+            optional_dependencies_map = self.dynamic_cfg["optional-dependencies"]
+            assert isinstance(optional_dependencies_map, dict)
+            return {
+                group: _parse_requirements_list(self._expand_directive(
+                    f"tool.setuptools.dynamic.optional-dependencies.{group}",
+                    directive,
+                    {},
+                ))
+                for group, directive in optional_dependencies_map.items()
+            }
```
abravalheri, line 395: Can we also have the `self._ensure_previously_set(dist, field)` before the last return in this function?

Since it does not use `self._obtain` directly, there is no contingency for the case we `optional-dependencies` is listed as `dynamic` but not provided in `tool.setuptools.dynamic`...

PEP 621 requires the following:
> Build back-ends MUST raise an error if the metadata specifies a field in dynamic but the build back-end was unable to provide the data for it.

setuptools/tests/config/test_pyprojecttoml_dynamic_deps.py
```diff
@@ -0,0 +1,75 @@
...
+    dist = apply_configuration(dist, pyproject)
+    assert dist.extras_require == {"docs": ["sphinx"]}
+
+
+@pytest.mark.xfail(reason="PEP 621 forbids mixing static and dynamic fields")
+def test_mixed_dynamic_optional_dependencies(tmp_path):
+    # This test would pass with flying colors if `validate_project_dynamic` didn't
+    # complain about the mixed dynamic/static configuration.
+    # Perhaps PEP 621 should be loosened regarding fields containing "sub-fields"
+    # (i.e. optional dependency groups in this case)?
```
abravalheri, line 54: I agree that PEP 621 is quite heavy-handed in terms of validations for `dynamic` fields,
but I don't think `setuptools` should unilaterally loosen the restrictions, that might create a non favourable reaction in the community.

All changes in the PEP behaviour can be discussed in https://discuss.python.org/c/packaging/14, but that maybe be a long and laborious process.

abravalheri, line 54: I agree that PEP 621 is quite heavy-handed in terms of validations for `dynamic` fields,
but I don't think `setuptools` should unilaterally loosen the restrictions, that might create a non favourable reaction in the community.

All changes in the PEP behaviour can be discussed in https://discuss.python.org/c/packaging/14, but that maybe be a long and laborious process.

setuptools/tests/config/test_pyprojecttoml_dynamic_deps.py
```diff
@@ -0,0 +1,75 @@
...
+[tool.setuptools.dynamic.optional-dependencies.images]
+file = ["requirements-images.txt"]
+
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+""")
+    dist = Distribution()
+    dist = apply_configuration(dist, pyproject)
+    assert dist.extras_require == {"docs": ["sphinx"], "images": ["pillow~=42.0"]}
```
abravalheri, line 75: I wonder that if instead of having the `xpass` we could do 2 checks here:
- We keep your proposed check with `apply_configuration(..., ignore_option_errors=True)`
- We add a second check to make sure PEP 621 is respected.
```python
with pytest.raises(ValueError, match="tool.setuptools.dynamic.optional-dependencies"):
    apply_configuration(dist, pyproject)
```

abravalheri, line 75: I wonder that if instead of having the `xpass` we could do 2 checks here:
- We keep your proposed check with `apply_configuration(..., ignore_option_errors=True)`
- We add a second check to make sure PEP 621 is respected.
```python
with pytest.raises(ValueError, match="tool.setuptools.dynamic.optional-dependencies"):
    apply_configuration(dist, pyproject)
```

abravalheri (commented):
Thank you very much for the very quick implementation @akx.

You went even further than I was expecting and also provided a patch for `validate-pyproject`, that is greatly appreciated!

I left some review comments. Some of them are minor, but I think the one about `_ensure_previously_set` is important to implement.

Would you also be OK allowing the maintainers to implement changes in the PR? There are minor things/nitpicks that I would edit before accepting the PR, but those are really not worth bothering you.

---
BTW: sorry for the trouble of requiring you to change the schemas in a different project. My long term vision is to move them to the setuptools repository.

Very great that you managed to find the way of re-generating the validation code even without proper documentation for `tox -e generate-validation-code` 👏 ❤️ 

---
Regarding the acceptance, let's see how Jason review the other PR for `setup.cfg`. When that one gets merged, I am very happy to merge the changes here.

setuptools/config/pyprojecttoml.py
```diff
@@ -365,6 +374,37 @@ def _obtain_classifiers(self, dist: "Distribution"):
...
+            expanded_optional_dependencies_map = {
+                group: _parse_requirements_list(self._expand_directive(
+                    f"tool.setuptools.dynamic.optional-dependencies.{group}",
+                    directive,
+                    {},
+                ))
+                for group, directive in optional_dependencies_map.items()
+            }
+            self._ensure_previously_set(dist, "optional-dependencies")
+            return expanded_optional_dependencies_map
```
abravalheri, line 397: Sorry, I was just seeing the CI errors, and it seems to be a bit more complicated than I first thought. How about?

Sorry, according to the error in the CI, it seems a bit more complicated than that...
How about:

```python
    def _obtain_optional_dependencies(self, dist: "Distribution"):
        if "optional-dependencies" not in self.dynamic:
            return None
        if "optional-dependencies" in self.dynamic_cfg:
            optional_dependencies_map = self.dynamic_cfg["optional-dependencies"]
            ...
       self._ensure_previously_set(dist, "optional-dependencies")
       return None
```

akx:
> I left some review comments. Some of them are minor, but I think the one about `_ensure_previously_set` is important to implement.

Addressed, thank you!

> Would you also be OK allowing the maintainers to implement changes in the PR? There are minor things/nitpicks that I would edit before accepting the PR, but those are really not worth bothering you.

Honestly I'd prefer if you could make review comments instead, I don't mind doing the work! That'd also give me a bit more insight to the stylistic/... choices for setuptools, for future contributions.

> Very great that you managed to find the way of re-generating the validation code even without proper documentation for `tox -e generate-validation-code`

Oh, there was a tox command..? 😁  I just looked at the `NOTICE` in the `_validate_pyproject` subdirectory. Also, `fastjsonschema` seems very interesting – I might have some work use for it at @valohai...

setuptools/config/pyprojecttoml.py
abravalheri: Here `self._ensure_previously_set` + `return None` are the fallback code path that require the user to specify `extras_require` via `setup.py` if `dynamic` contains `optional-dependencies`.

setuptools/tests/config/test_pyprojecttoml_dynamic_deps.py
```diff
@@ -0,0 +1,81 @@
...
+version = "1.0"
+dynamic = ["dependencies"]
+
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.dynamic.dependencies]
+file = ["requirements.txt"]
+""")
```
abravalheri, line 22: This is a minor comment :)

In other test files, setuptools uses `from .textwrap import DALS` to deal with the indentation. I think `from inspect import cleandoc` is also an alternative built-in Python3 stdlib.

abravalheri (commented):
Thank you very much @akx for providing the changes.
I left a few other comments regarding the previous error in the CI and the nitpicks.

Should we also include an equivalent note in the [docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) about this method not being intended for pinning?

setuptools/tests/config/test_pyprojecttoml_dynamic_deps.py
akx: Yeah, I saw DALS elsewhere, but I also noticed not all tests use it (though now that I think of it, TOML doesn't strictly speaking need to be DALS-ed). Fixed, nevertheless!

akx:
@abravalheri Review comments addressed, docs added! :)

docs/userguide/pyproject_config.rst
```diff
@@ -181,16 +181,28 @@ In the ``dynamic`` table, the ``attr`` directive [#directives]_ will read an
...
+=================          =================== =========================
+Key                        Directive           Notes
+=================          =================== =========================
+``version``                ``attr``, ``file``
+``readme``                 ``file``
+``description``            ``file``            One-line text
+``classifiers``            ``file``            Multi-line text with one classifier per line
+``entry-points``           ``file``            INI format following :doc:`PyPUG:specifications/entry-points`
+                                               (``console_scripts`` and ``gui_scripts`` can be included)
+``dependencies``           ``file``            Requirements.txt format (``#`` comments and blank lines excluded)
```
abravalheri, line 193: Sorry, this is another nitpick 😅
What do you think about using ``` ``requirements.txt`` ``` in the table description?

abravalheri, line 193: Sorry, this is another nitpick 😅
What do you think about using ``` ``requirements.txt`` ``` in the table description?

akx:
@abravalheri The CI error was just a flake8 complaint about a line length. Fixed that, as well as the `requirements.txt` formatting in the docs.

abravalheri (approved):

PR #3259, opened 2022-04-11
Trigger deprecation warning when find_packages() finds reserved name

### LINKED ISSUE #3243: [FR] Warn about top-level "tests" package in find_packages() by default

I often end up reporting bugs about packages using `find_packages()` without appropriate exclusion lists, that end up installing top-level `test` or `tests` directories. These directories are especially problematic since they're bound to cause file collisions between packages, and given their special use I don't think any specific package should be allowed to "own" them.

I think it would be best if setuptools could at least warn if `find_packages()` ends up including `test` or `tests` directories, and suggest the package maintainers to exclude them.  Perhaps they could also be skipped by default in some future version.

### Alternative Solutions

It could be possible to implicitly skip them immediately. While I honestly don't think it would realistically break anything, I'm not sure if someone isn't actually having some workflow relying on this.

### Additional context

_No response_

### Code of Conduct

mgorny:
## Summary of changes

Trigger SetuptoolsDeprecationWarning whenever find_packages() finds
a top-level package whose name is found in reserved package name list,
e.g. "tests".  Installing these packages is probably always wrong,
yet it is still a frequent mistake in packages using setuptools.  This
warning should increase the chance of the mistake being caught before
releasing.

Closes #3243

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/discovery.py
+++ b/setuptools/discovery.py
@@ -44,11 +44,14 @@ from glob import glob
 from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import Callable, Dict, Iterator, Iterable, List, Optional, Tuple, Union
+import warnings
 
 import _distutils_hack.override  # noqa: F401
 
 from distutils import log
 from distutils.util import convert_path
+
+import setuptools
 
 _Path = Union[str, os.PathLike]
 _Filter = Callable[[str], bool]
@@ -148,6 +151,13 @@ class PackageFinder(_Finder):
 
                 # Should this package be included?
                 if include(package) and not exclude(package):
+                    top_package = package.split(".")[0]
+                    if top_package in FlatLayoutPackageFinder._EXCLUDE:
+                        warnings.warn(
+                            f"{top_package!r} is a reserved package name. "
+                            "Please add it to find_package(exclude=...)",
+                            setuptools.SetuptoolsDeprecationWarning,
+                        )
                     yield package
 
                 # Keep searching subdirectories, as there may be more packages
--- a/changelog.d/3243.deprecation.rst
+++ b/changelog.d/3243.deprecation.rst
@@ -1,0 +1,3 @@
+Added a deprecation warning when `setuptools.find_packages()` includes
+a package in the reserved top-level package name list (e.g. ``tests``)
+-- by :user:`mgorny`.
--- a/setuptools/tests/test_find_packages.py
+++ b/setuptools/tests/test_find_packages.py
@@ -7,6 +7,7 @@ import platform
 
 import pytest
 
+import setuptools
 from setuptools import find_packages
 from setuptools import find_namespace_packages
 from setuptools.discovery import FlatLayoutPackageFinder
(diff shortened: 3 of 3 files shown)


mgorny:
I have to admit that I went with the "minimal change" approach. If you prefer it done some other way, please let me know.

setuptools/discovery.py
```diff
@@ -148,6 +151,13 @@ def _find_iter(cls, where: _Path, exclude: _Filter, include: _Filter) -> StrIter
 
                 # Should this package be included?
                 if include(package) and not exclude(package):
+                    top_package = package.split(".")[0]
+                    if top_package in FlatLayoutPackageFinder._EXCLUDE:
+                        warnings.warn(
+                            f"{top_package!r} is a reserved package name. "
+                            "Please add it to find_package(exclude=...)",
+                            setuptools.SetuptoolsDeprecationWarning,
+                        )
```
abravalheri, line 160: Is it really the case these patterns should be deprecated?

What if the user is the owner of the package `tasks` or `changelog` (those are perfectly valid entries for PyPI right?) and is explicitly using `[options.packages.find] include = tasks*`?

I am also not sure if this is the best place to implement the check..
This code path is traversed by the auto-discovery algorithm that will already error if the distribution contains multiple top-level packages... It is also traversed when the `src-layout` is being used (and my opinion is that if the `src-layout` is used, there should be no warning...).

I think that an alternative would be modifying `setuptools.find_packages`, `setuptools.find_namespace_packages`, `setuptools.config.setupcfg.ConfigOptionsHandler._parse_packages` and `setuptools.pyprojecttoml._ConfigExpander._expand_packages` to check for one of these patterns ***only if*** there is more than one top-level package.

abravalheri, line 160: Is it really the case these patterns should be deprecated?

What if the user is the owner of the package `tasks` or `changelog` (those are perfectly valid entries for PyPI right?) and is explicitly using `[options.packages.find] include = tasks*`?

I am also not sure if this is the best place to implement the check..
This code path is traversed by the auto-discovery algorithm that will already error if the distribution contains multiple top-level packages... It is also traversed when the `src-layout` is being used (and my opinion is that if the `src-layout` is used, there should be no warning...).

I think that an alternative would be modifying `setuptools.find_packages`, `setuptools.find_namespace_packages`, `setuptools.config.setupcfg.ConfigOptionsHandler._parse_packages` and `setuptools.pyprojecttoml._ConfigExpander._expand_packages` to check for one of these patterns ***only if*** there is more than one top-level package.

mgorny:
I'm sorry but due to health issues I'll have to limit my computer activity for some time (apparently up to 4 weeks), so I won't be able to look into it any time soon.

abravalheri:
Sorry to hear that...
I wish you a good recovery.

mgorny:
Thank you very much.

jaraco:
@mgorny I hope you're back to health. I thought I'd check in on this PR and see if you wanted to continue with it?

jaraco:
Since this was written, there have been some improvements to the build system that limit the accidental introduction of packages. The default behavior is to publish exactly one package and have it discovered automatically and fail hard if other packages are picked up. That should drastically reduce the incidence of accidental inclusion. Let's push on the projects downstream to use that instead of trying to second-guess the user's intention in code.

PR #3272, opened 2022-04-17
chore: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -59,6 +59,9 @@ env:
   SAGE_TICKET:    33288
   REMOVE_PATCHES: "*"
 
+permissions:
+  contents: read
+
 jobs:
 
   dist:


abravalheri:
Thank you very much @neilnaveen.

Do you think we also need something similar for `.github/workflows/main.yml`? (The `release` job, however, is very likely to need extra permissions).

henryiii:
This should only be needed for `pull_request_target`, see https://github.com/pypa/cibuildwheel/pull/1090, https://github.com/Textualize/rich/pull/2211#issuecomment-1120106263.

abravalheri:
Thank you very much @henryiii, since setuptools does not use `pull_request_target`, I will go ahead and also close this issue (following the example of cibuildwheel and rich).

PR #3285, opened 2022-04-25
Updated declarative_config.rst to fix user guide typo

harrysalmon:
## Summary of changes

Updated the documentation to fix a typo in docs/userguide/declarative_config.rst in the "Using a ``src/`` layout" section. 
Closes: No ticket, small docs change 

### Pull Request Checklist
- [x] Changes have tests N/A documentation only change 

[PR docs]:

### DIFF
--- a/changelog.d/3285.doc.rst
+++ b/changelog.d/3285.doc.rst
@@ -1,0 +1,1 @@
+Fixed a typo in docs/userguide/declarative_config.rst -- by user :user:`harrysalmon`
--- a/docs/userguide/declarative_config.rst
+++ b/docs/userguide/declarative_config.rst
@@ -109,8 +109,7 @@ the subdirectory like this:
     # the rest of the file as described above.
 
     [options]
-    package_dir=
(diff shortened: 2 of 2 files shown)


abravalheri:
Hi @harrysalmon, thank you very much for looking into this.

The example is not a typo 😅.
What happens is that the `package_dir=` field expects a list of "keyword pairs" in the form "key=value".

When all the packages are contained inside the same folder (e.g. `src`), the key value is an empty string.

Would you like to add a small note instead explaining this?

harrysalmon:
> The example is not a typo 😅.
> What happens is that the package_dir= field expects a list of "keyword pairs" in the form "key=value".

Thanks very much for such a quick reply! Oops 😅  that makes sense, so it's the equivalent to the dict in a setup.py `{'': 'src'}`? 

That sounds good, I'll close this and open a PR with an explanation and see if people think it's helpful.

abravalheri:
Thank you very much Harry!

PR #3289, opened 2022-04-27
Optional metadata

wimglenn:
Companion to https://github.com/pypa/distutils/pull/138

- Stop backfilling `Project-URL: homepage, https://example.org` into `Home-page: https://example.org`
- Similar with Download-URL
- Remove warnings about missing url / author / maintainer etc, these are all optional fields in the spec
- Prevent adding in extra whitespace to the long description
- Prevent "UNKNOWN" strings from creeping in when fields weren't specified by user

Related: https://github.com/pypa/warehouse/issues/11220

### DIFF
--- a/setuptools/_distutils/command/bdist_msi.py
+++ b/setuptools/_distutils/command/bdist_msi.py
@@ -231,11 +231,7 @@ class bdist_msi(Command):
         if os.path.exists(installer_name): os.unlink(installer_name)
 
         metadata = self.distribution.metadata
-        author = metadata.author
-        if not author:
-            author = metadata.maintainer
-        if not author:
-            author = "UNKNOWN"
+        author = metadata.author or metadata.maintainer
         version = metadata.get_version()
         # ProductVersion must be strictly numeric
         # XXX need to deal with prerelease versions
--- a/setuptools/_distutils/command/bdist_rpm.py
+++ b/setuptools/_distutils/command/bdist_rpm.py
@@ -399,7 +399,7 @@ class bdist_rpm(Command):
             '%define unmangled_version ' + self.distribution.get_version(),
             '%define release ' + self.release.replace('-','_'),
             '',
-            'Summary: ' + self.distribution.get_description(),
+            'Summary: ' + (self.distribution.get_description() or "UNKNOWN"),
             ]
 
         # Workaround for #14443 which affects some RPM based systems such as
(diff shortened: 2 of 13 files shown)


abravalheri:
Thank you very @wimglenn.

Let's see how the discussion in the disutils repository evolves.

jaraco:
Is there any way this change could be applied to Setuptools such that _distutils doesn't have to be altered first?

abravalheri:
My interpretation is the following:

1. If `distutils` is not changed, the user will see some warnings that will not make sense (and even be conceptually unjustified), but the build will ultimately pass. That is a bad UX, so I think it would be better to leave things as they are right now.

2. The same will happen if `distutils` changes, but the user sets `SETUPTOOLS_USE_DISTUTILS=stblib`. Maybe we can see the bad UX in this case as a compromise? The minority of users will run in this compatibility mode...

3. In theory, the alternative implementation that will work for both groups is to overwrite the check command in the setuptools layer (this approach may involve code duplication).

abravalheri:
(The last 2 bullet points in the description are fine and not part of the problem with the UX...)

wimglenn:
@jaraco  I double checked everything and I don't think there is any incompatibility. The setuptools change can go in before the distutils change, and we would not be in a worse place - however the 3rd and the 5th bullet points would remain unfixed (because they need the distutils changes).  Bullet points 1, 2, and 4 would be fixed by the setuptools change only.

Point 3 is about the warnings from the check command, they look like this:

```
$ python3 setup.py sdist -v
...
warning: check: missing required meta-data: url
warning: check: missing meta-data: either (author and author_email) or (maintainer and maintainer_email) should be supplied
...
```

The build still succeeds and makes a valid distribution, they're just bogus warnings. To fix this in setuptools directly without changing distutils you'd have to reimplement the [check command](https://github.com/pypa/distutils/blob/main/distutils/command/check.py), which is the source of the warnings, it currently exists in distutils only.

Point 5 is about the "UNKNOWN" strings, these are also coming from distutils.  Currently if you omit Summary, License, or Platform you get a PKG-INFO file with lines like this:

```
Metadata-Version: 2.1
Name: mypkg
Version: 0.1
Summary: UNKNOWN
License: UNKNOWN
Platform: UNKNOWN
```

That's because the "getters" for these fields aren't patched by setuptools, so it goes to distutils Distribution class for them:  [`get_description`](https://github.com/pypa/setuptools/blob/v62.1.0/setuptools/dist.py#L177), [`get_license`](https://github.com/pypa/setuptools/blob/v62.1.0/setuptools/dist.py#L193), and [`get_platforms`](https://github.com/pypa/setuptools/blob/v62.1.0/setuptools/dist.py#L202).  To fix that in setuptools directly we could check their return values, or re-implement them on the setuptools Distribution class, but I personally think it's better just to fix the issue where it was caused directly.

So, in summary, there is no incompatibility - I tested with `SETUPTOOLS_USE_DISTUTILS=stblib` - and the build still works but users would continue to get the bogus warnings and those UNKNOWN fields, so only 3 of the 5 points above actually get solved.

wimglenn:
@jaraco  ping.. anything more required here?

Issue #3214, opened 2022-03-26
Eventually deprecate `setup.cfg` with automatic conversion to `pyproject.toml`


abravalheri:
This is a follow up of #1688.

As discussed in the linked issue, the approach setuptools would like to take is to eventually use a single declarative format (`pyproject.toml`) instead of maintaining 2 (`pyproject.toml`/`setup.cfg`).

Originally posted by Paul Ganssle in https://github.com/pypa/setuptools/issues/1688#issue-411224984:
> Add pyproject.toml as the one true way to do declarative builds, moving all setuptools configuration over there and deprecating setup.cfg (partially what is discussed in https://github.com/pypa/setuptools/issues/1160, and if we decide on that option we can move over to that issue).

Originally posted by Jason R. Coombs in https://github.com/pypa/setuptools/issues/1688#issuecomment-471212342
> One way to ease the transition could be for the setup.cfg code to generate a .toml file and consume then have the pyproject.toml consumer just load both files. That way, a project seeking to transition would only need to do one build, manually merge the two .toml files, and delete setup.cfg.

Originally posted by Paul Ganssle in https://github.com/pypa/setuptools/issues/1688#issuecomment-570933225
>  I don't think we should warn if people use setup.cfg even when pyproject.toml is available, we should wait for pyproject.toml configuration to be stable for at least a year before actively pushing it like that.

Relevant tool: [`ini2toml`](https://ini2toml.readthedocs.io/en/latest/readme.html)

thehappycheese:
Thankyou for posting this issue. As someone tinkering on the bleeding edge of packaging tools I have been so confused by the various PEPs and documentation. I thought pyproject.toml was the future but could not find anything definitive to suggest setuptools were going to deprecate setup.cfg.

abravalheri:
Hi @thehappycheese please feel free to open a discussion if you still
have doubts, either here or in https://discuss.python.org/c/packaging/14. We will try our best to help.

thehappycheese:
Thanks @abravalheri I will take a look at the forums. Yes I understand setuptools has a big community to bring along for the journey when it happens :) I think it will be a great move though if it works out

bskinn:
@abravalheri Is [that milestone](https://github.com/pypa/setuptools/milestone/7) correctly named? I still use `setup.py` to provide specific dynamic content to `build`, and I would like to be able to keep that workflow.

[...]

PR #3295, opened 2022-04-29
Clarify that `pyproject.toml` config will not be removed

garrison:
## Summary of changes

The current documentation regarding `pyproject.toml` claims that support for using it to specify package metadata and build configuration options might be "completely removed" in a future setuptools release.

However, nowadays it is clear (https://github.com/pypa/setuptools/issues/1688#issuecomment-1079706929 and #3214) that `pyproject.toml` is not going to be removed, and is in fact intended to be the preferred format at some point in the future.

This makes an incremental change toward that future by clarifying that `pyproject.toml` support will not be removed in a future release of setuptools.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/docs/userguide/dependency_management.rst
+++ b/docs/userguide/dependency_management.rst
@@ -439,5 +439,5 @@ This can be configured as shown in the example below.
    While the ``[build-system]`` table should always be specified in the
    ``pyproject.toml`` file, support for adding package metadata and build configuration
    options via the ``[project]`` and ``[tool.setuptools]`` tables is still
-   experimental and might change (or be completely removed) in future releases.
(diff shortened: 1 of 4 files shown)


abravalheri:
Thank you very much @garrison.

I am completely fine with this change, and I do think there is no risk of removing support for (at least) the `[project]` table in `pyproject.toml` (unless there is a new and revolutionary PEP). The `[tool.setuptools]` table is still very likely to change and should not be taken as stable, but that behaviour seems to be covered by the rest of the text.

Since I am very biased 😅, I think we can leave this PR open for a little while just to see if another maintainer or member of the community has an objection.

PR #3299, opened 2022-04-30
Optional metadata sans distutils

wimglenn:
Same as https://github.com/pypa/setuptools/pull/3289 but without the changes under _distutils (companion to pypa/distutils#138).


abravalheri:
I would still wait for the changes to land in `distutils` (and then merged back in `setuptools`) before merging this PR, otherwise the end result will be warnings that don't make sense (or are even conceptually wrong) for the the user and a bad UX.

If all the merges happen, we are going with situation 2 described in https://github.com/pypa/setuptools/pull/3289#issuecomment-1113947093

PR #3311, opened 2022-05-09
Merge with pypa/distutils@a7cfb56a7

### LINKED ISSUE #132: Missing license
Originally reported by: **stefanor (Bitbucket: [stefanor](http://bitbucket.org/stefanor), GitHub: [stefanor](http://github.com/stefanor))**

---

The only mention of the license that this is distributed under, is in setup.py:

```
#!python

license="PSF or ZPL",
```

The body of the license doesn't appear to be present in any source or binary distributions. This makes it hard to know exactly what the terms of distribution are.

There are multiple versions of the Zope license in existance, but the version isn't specified. And the PSF license most commonly appears in the long essay about python's licensing history, it isn't well known as a stand-alone license. There'd be no guessing, if the license bodies were included.

Please bundle the licenses that this is distributed under, in the source. And ensure that they appear in any source and binary distributions.

Thanks

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/132

jaraco:
- Detect when a venv is created from an in-tree build. Fixes #132
- make all fields documented as optional in the spec truly optional (dont warn when omitted), and stop coercing missing values to the string "UNKNOWN"
- Extract check for 'd is None'.
- Combine logic between prefix and base_prefix.
- Adapt style
- Simply emit the expected value instead of converting to an exit code and deciphering.
- Rely on jaraco.envs for environment creation.
- Prefer tabs
- Use pass_none decorator.
- Prefer tox 3.25 with built-in support for Windows env vars.
- Run tests using tox, even on cygwin.
- Prefer tabs
- address review comments
- fix a typo
- Update changelog

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       matrix:
         python:
-        - 38
+        - 39
         platform:
         - windows-latest
     runs-on: ${{ matrix.platform }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,13 @@ jobs:
             ${{ matrix.python }}
 
   test_cygwin:
-    runs-on: windows-latest
+    strategy:
+      matrix:
+        python:
+        - 38
+        platform:
+        - windows-latest
+    runs-on: ${{ matrix.platform }}
     timeout-minutes: 75
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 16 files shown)


jaraco:
Tests for cygwin are failing because the "tox" command is not found. That's strange and surprising because I'm using the same technique for installing and invoking tox as is used by distutils.

jaraco:
Apparently Python 3.9 is required for tox to get installed in the path. Not sure why.

DWesl:
> Apparently Python 3.9 is required for tox to get installed in the path. Not sure why.

The `/usr/bin/tox` script gets installed by each of `python35-tox`, `python36-tox`, `python37-tox`, `python38-tox`, and `python39-tox` in turn, but can only be a part of one of those packages.  The maintainer usually chooses the last one, to match the shebang line in the script.  I tend to use `python -m pytest` to avoid needing to know which package the script got installed in, and updating everything when it gets moved (say, when Cygwin adds python 3.10 and makes that the default for command-line scripts).

If you want to find the discussion on the mailing list, I think the script that most recently got mentioned was `/usr/bin/sphinx-build` moving from `python3-sphinx` (replaced by `python36-sphinx`) to `python38-sphinx`.

PR #3326, opened 2022-05-18
Fix _is_gcc() for cross-compiler prefixes and handle cland and clang++

zboszor:
## Summary of changes

This patch fixes building python modules using cython for their build in the Yocto cross-compiler framework.
Compilers are not just called `gcc`, `g++` or `clang`, they may have a `<machine-triplet>-` prefix in the executable name.
Therefore, use a regex match instead of the string equality check.
Also, handle clang and clang++ along the same lines. The same `unixccompiler.py` file in the Python core already handles clang, although it doesn't handle the machine triplet prefix.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3326.change.rst
+++ b/changelog.d/3326.change.rst
@@ -1,0 +1,1 @@
+Fixed ``_is_gcc()`` to handle machine triplet prefix and handle ``clang`` and ``clang++`` -- by :user:`zboszor`
--- a/setuptools/_distutils/unixccompiler.py
+++ b/setuptools/_distutils/unixccompiler.py
@@ -260,7 +260,8 @@ class UnixCCompiler(CCompiler):
         return "-L" + dir
 
     def _is_gcc(self, compiler_name):
-        return "gcc" in compiler_name or "g++" in compiler_name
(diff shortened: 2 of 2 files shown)


zboszor:
> Thank you very much @zboszor.
> 
> Would you consider submitting the change to https://github.com/pypa/distutils?
> 
> The `setuptools/_distutils` folder is periodically synchronized with that repository, so any change merged into setuptools can potentially be overwritten in one of these periodic pulls or cause merge conflicts...

Sure, thanks for the pointer.

zboszor:
https://github.com/pypa/distutils/pull/143

abravalheri:
@zboszor I can see that the PR in `distutils` was closed because apparently it is not needed.

I understand that there might be a problem with the `python3-setuptools` package in Yocto releases. But I suppose that the solution for that would be having an up-to-date release of the `python3-setuptools` package on that platform... 

Given this scenario, I will go ahead and close this PR. But please feel free to follow up the last comment from Jason in the `pypa/distutils` thread, if you feel that something is missing.

Thank you again for the contribution and the investigation.

PR #3331, opened 2022-05-19
Replaced single backticks with double ones

codeandfire:
## Summary of changes

These two items `**` and `package_data` in the `CHANGES.rst` file were being identified as links rather than code due to the single backticks.
In turn leading to Sphinx emitting warnings about this and the docs failing to build.

Closes: No issue that I am aware of? However these warnings were preventing the docs from building on my machine.

### Pull Request Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ Deprecations
 
 Changes
 ^^^^^^^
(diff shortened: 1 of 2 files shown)


abravalheri:
Good catch! Thank you very much.

PR #3336, opened 2022-05-24
Make test_setup_install_includes_dependencies work with custom PYTHONPATH set

hroncok:
This has the same reasoning as in https://github.com/pypa/setuptools/pull/3133 -- the test imported the previously installed setuptools version in our environment and since that version did not yet use `__EASYINSTALL_INDEX`, it reached https://pypi.org/simple/ instead. That way, we recognized it as a problem. If the installed version had already used `__EASYINSTALL_INDEX`, we would probably never realize that it is the installed version that is being tested here.

The three individual commits follow my train of thoughts, so I preserved them here separated, for easier review. 

 1. The test was very hard to grasp for me, so I removed the `pytest.raises` detour -- happy to revert if there was a reason for it
 2. I changed the test to modify the `$PYTHONPATH` value instead of overriding it, making it actually work in our environment
 3. When doing it, I realized adding `install_root` to `$PYTHONPATH` is not necessary for this test, so I removed it -- happy to squash that to the previous one if you prefer that

Thanks.

### DIFF
--- a/changelog.d/3336.misc.rst
+++ b/changelog.d/3336.misc.rst
@@ -1,0 +1,1 @@
+Modified ``test_setup_install_includes_dependencies`` to work with custom ``PYTHONPATH`` –- by :user:`hroncok`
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you very much @hroncok, sorry for the delay in reviewing/merging.

I have added a small [news fragment](https://setuptools.pypa.io/en/latest/development/developer-guide.html#alright-so-how-to-add-a-news-fragment) based on the title of your PR. I hope you don't mind.

PR #3349, opened 2022-06-08
Fix documentation local build

codeandfire:
## Summary of changes

The docs were not building locally due to the following warnings:
 1. The `.gitattributes`, `.gitignore` etc. example added recently to the Data Files page was using single backticks instead of double backticks, causing these filenames to be interpreted as links rather than code.
 2. The `extlinks` extension seems have to recently started emitting the following warning:
     ```
     WARNING: extlinks: Sphinx-6.0 will require a caption string to contain exactly one '%s' and all other '%' need to be escaped as '%%'.
     ```
     Seems the issue is with the `user` entry in the `extlinks` defined in `docs/conf.py`, the caption string is `'@'`. I have fixed this by changing the caption string to `'@%s'`.

### Pull Request Checklist

[PR docs]:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -103,7 +103,7 @@ github_repo_slug = f'{github_repo_org}/{github_repo_name}'
 github_repo_url = f'{github_url}/{github_repo_slug}'
 github_sponsors_url = f'{github_url}/sponsors'
 extlinks = {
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you very much @codeandfire (sorry for the trouble with the improperly formatted examples).

PR #3355, opened 2022-06-09
Changes to the User Guide's Entry Points page

codeandfire:
## Summary of changes

Console scripts section:
 - Added `pyproject.toml` snippet
 - Modified `setup.py` snippet (corrected indentation and removed unnecessary name, version, etc. fields)
 - Changed the example to use src-layout
 - Some changes in the wording etc.

Added GUI Scripts section:
 - Included `setup.cfg` / `setup.py` / `pyproject.toml` snippets
 - Added an example Hello world function that can be invoked through a GUI script

Other changes:
 - Added note on distinction between Console and GUI scripts
 - Added note on how installers like Pip create wrapper scripts around invoked functions
 - Included few lines stating that invoked functions should not accept any arguments and parse user input within the function body

_Have not finished this yet. Will update here with further changes._

### Pull Request Checklist
- [ ] News fragment added in [`changelog.d/`].
  _(See [documentation][PR docs] for details)_

[`changelog.d/`]: https://github.com/pypa/setuptools/tree/master/changelog.d
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #3361, opened 2022-06-11
Minor corrections to the Entry Points page

codeandfire:
## Summary of changes

Hi @abravalheri, these are a few minor issues with the previous PR (#3355) that I just happened to notice. Apologies for not noticing these before, I hope these can be merged soon.

### Pull Request Checklist

[PR docs]:


codeandfire:
There are two other points that I just noticed (again, sorry for not noticing these before):

  1. The last section of the Entry Points page, which refers to extra dependencies for entry points, seems to be a deprecated functionality according to the [Dependencies Management docs page](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html). The relevant sections (on defining Extra Dependencies) in both pages seem to link to each other and share similar content, so perhaps it would be a good idea to make appropriate modifications to both pages? I wasn't sure how you would like to go about this, hence wanted your feedback first before making any changes.
  2. This point is regarding the syntax for defining entry points in the Console Scripts section. Assuming that the parts in square brackets are supposed to stand for things that are optional, it seems that the "module" part is compulsory. However in all of our examples in this page, we have been defining functions corresponding to entry points in the `__init__.py` file of the package, and therefore referring to the function without any module name, like `timmins:hello_world`. This might be confusing for some users... maybe we should modify the way this syntax has been specified?

abravalheri:
Thank you very much for the awesome improvements @codeandfire (and the level to attention to the details!).

I think we can merge these changes immediately: differently from changes in the setuptools code, the documentation changes are picked up immediately by ReadTheDocs and published as `latest`, so we don't need to create a tag release for them.

Regarding your suggestions:

1. Considering that we already have this part covered in a different document, and that this practice is deprecated (or at least `pip` does not honour it and we want to manage user's expectation), I believe we can remove [the "Dependency Management" section from the Entry Points page](https://setuptools--3361.org.readthedocs.build/en/3361/userguide/entry_point.html#dependency-management). What do you think?

2. Yeah maybe the way we define the syntax is a bit confusing. The way I see it is that everything before the `:` is what you would write in an import (or from import statement).
   What do you think about the following:

   ```rst
   The syntax for entry points is specified as follows (``[]`` denotes optionality and ``*`` repetition)::

       <name> = <package_or_module>[:<object>[.<attr>[.<nested_attr>]*]

   where ``name`` is the name for the script you want to create, the left hand side of ``:`` is the
   package or module that contains your function and the right hand side is the object you want to
   invoke (e.g. a function).

   When parsing this syntax, you can imagine that tools in the Python ecosystem will use the
   entry point value in a way that is roughly equivalent to:

   1. ``<package_or_module>`` only (can contain ``.`` in the case of sub-modules or sub-packages)::

        import <package_or_module>
        parsed_value = <package_or_module>

   2. ``<package_or_module>:<object>`` (``<object>`` does not contain ``.``)::

        from <package_or_module> import <object>
        parsed_value = <object>

   3. ``<package_or_module>:<object>.<attr>.<nested_attr>``::

        from <package_or_module> import <object>
        parsed_value = <object>.<attr>.<nested_attr>

   ```
   (Please feel free to ignore the suggestion or modify it).

codeandfire:
Thanks for your response @abravalheri! It does make a lot of sense to entirely remove the Dependency Management section from the Entry Points page. And I think the layout you have suggested for explaining the syntax bit makes things really clear. I'll amend this PR then, to incorporate these changes, and then perhaps we can merge.

codeandfire:
@abravalheri have added all changes. Let me know your feedback! Thanks!

docs/userguide/entry_point.rst
```diff
@@ -102,15 +113,12 @@ from the user, you can use regular command-line argument parsing utilities like
 `argparse <https://docs.python.org/3/library/argparse.html>`_ within the body of
```
abravalheri, line 113:
```suggestion
:mod:`argparse` within the body of
```

docs/userguide/entry_point.rst
```diff
@@ -4,15 +4,26 @@
 Entry Points
 ============
 
-Packages may provide commands to be run at the console (console scripts),
-such as the ``pip`` command. These commands are defined for a package
-as a specific kind of entry point in the ``setup.cfg`` or
-``setup.py``.
+Entry points are a very useful feature of Python packages.
+They come in use in two scenarios:
```
abravalheri, line 8: I think we can benefit here for a more specific definition of what is an entry-point. What do you think about the following change?

```suggestion
Entry points are a type of metadata that can be exposed by packages on installation.
They are a very useful feature of the Python ecosystem,
and come specially handy in two scenarios:
```

docs/userguide/entry_point.rst
```diff
@@ -462,6 +472,14 @@ is useful:
...
 
+Another point is that in this particular example, we have used plugins to
+customize the behaviour of a function (``display()``). In general, we can use entry
+points to enable plugins to not only customize the behaviour of functions, but also
+of entire classes and modules. This is unlike the case of console/GUI scripts,
+where entry points can only refer to functions. The syntax used for specifying the
+entry points remains the same as for console/GUI scripts, and is discussed in the
+`last section <#entry-points-syntax>`_.
+
 importlib.metadata
```
abravalheri, line 483: Instead of having a section arbitrarily named after a stdlib package, how about we transform this part of the document in a `.. tip::` or `.. note::`?

abravalheri (approved):
Thank you very much @codeandfire.

Sorry for the late suggestions, I was just going through the rendered version of the document and noticed a few other things that could be improved.

docs/userguide/entry_point.rst
codeandfire: Sure, makes sense!

docs/userguide/entry_point.rst
codeandfire: Ah, I see. But I think keeping it as a Tip or a Note might make users not read this part carefully - I mean, when there is some section of the document which is kept in a box and the font is smaller, people tend to skim over such things :)

How about changing the name of that section to something else like "Loading Entry Points"? What do you think about this?

codeandfire: Or I guess another option might be to let go of that section altogether, and rephrase it like this perhaps:

> Finally, note that the`importlib.metadata` module is the recommended approach for loading and importing entry points. It is a part of the standard library since Python 3.8. For older versions of Python ...

so that it seems in continuation with the text above it.

In that case I guess we can also let go of that separate Summary section.

abravalheri: I agree with your comments.

Actually the first thought I have was to rename the section to something like "How to load Entry Points?", which is inline of your proposal.

Then I realized that the previous section already introduced `importlib.metadata` in its examples and explicitly mention that it can be used to load the entry points...

My conclusion is that this part of the text is actually just a tip for users that `importlib-metadata` exists as a backport. It can likely be removed in the future (when 3.7 reaches EOL in June next year).

Please feel free to keep it as an independent section (but then we have to adequate the title to something like "Loading Entry Points in older versions of Python") or transform it in a `.. tip::/note::/important::` etc.

codeandfire: I didn't look at it this way before - that the portion about the backport is going to gradually become less relevant for users, especially given that Python 3.7 will reach EOL in the near future. In that case I think you are right and it is better to convert that section into a Tip/Note/etc. I will also remove the Summary section header since a singular subsection under the section "Entry Points for Plugins" doesn't make sense.

codeandfire: @abravalheri Have just made some changes. Let me know if you are happy with this.

abravalheri:
Thank you very much @codeandfire! I will go ahead and merge this.

On a side note, I am also working in some other improvements (motivated by all the good work you have been doing), and, I that is OK with you, I might ask your opinion on that (since you are already involved in the recent documentation changes).

codeandfire:
Sure @abravalheri, would love to be of any help!

PR #3381, opened 2022-06-17
Update build_meta.rst

cdfarrow:
Mainly grammar/style tweaks... which turned into a few more thoughts:
About this statement: "decide which program to use to 'build from source' (the default is ``setuptools``)." 
I think 'default' conveys the wrong impression here when it is explaining the new way of doing things. PEP517 says, "(build) tools should revert to the legacy behaviour of running setup.py" Maybe a footnote would be better saying it is a fallback, for backward compatibility. (But, all the same, this is interesting information that now explains to me why 'py -m build' worked before I had tried using a pyproject.toml file.)
Line 37:  ``build_meta`` implements ``setuptools``'s build system support.
This seems out of place, and doesn't make sense. build_meta hasn't been mentioned yet in this article. Maybe this sentence could go before line 61.
Line 26-28: This seems like low-level detail that is out of place in this article. If you want to keep it, maybe add a note that the default directory is dist\.
**The "bullet items" under the PEP517 section aren't formatting correctly in the Preview. Please check that I haven't messed anything up there in the final render.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/docs/build_meta.rst
+++ b/docs/build_meta.rst
@@ -25,18 +25,18 @@ modules. Under PEP 517:
 
     Two functions provided by the program, ``build_wheel(directory: str)``
     and ``build_sdist(directory: str)``, create the distribution bundle in the
-    specified ``directory``. 
+    specified ``directory``.
 
-	The program is free to use its own configuration file or extend the ``.toml`` file.
+    The program may use its own configuration file or extend the ``.toml`` file.
 
     The actual installation is done with ``pip install *.whl`` or
-	``pip install *.tar.gz``. If ``*.whl`` is available, ``pip`` will go ahead and copy
+    ``pip install *.tar.gz``. If ``*.whl`` is available, ``pip`` will go ahead and copy
     its files into the ``site-packages`` directory. If not, ``pip`` will look at
-    ``pyproject.toml`` and decide which program to use to 'build from source'
-    (the default is ``setuptools``).
+    ``pyproject.toml`` and decide which program to use to 'build from source'.
+    (Note that if there is no ``pyproject.toml`` file or the ``build-backend``
+    parameter is not defined, then the fall-back behaviour is to use ``setuptools``.)
 
-With this standard, switching between packaging tools is a lot easier. ``build_meta``
-implements ``setuptools``'s build system support.
+With this standard, switching between packaging tools is a lot easier.
 
 How to use it?
 --------------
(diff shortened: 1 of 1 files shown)


docs/build_meta.rst
```diff
@@ -159,17 +160,17 @@ requirements.
...
+   On POSIX systems, this can be done with ``tar -tf dist/*.tar.gz``
    and ``unzip -l dist/*.whl``.
-   On Windows systems you can rename the ``.whl`` to ``.zip`` to be able to
-   inspect it on the file explorer, and use the same ``tar`` command in a
-   command prompt (alternativelly there are GUI programs like `7-zip`_ that
-   handle ``.tar.gz``).
+   On Windows systems, you can rename the ``.whl`` to ``.zip`` to be able to
+   inspect it from File Explorer, or you can use the aoove ``tar`` command in a
+   command prompt. Alternatively, there are GUI programs like `7-zip`_ that
+   handle ``.tar.gz`` and ``.whl`` files.
```
abravalheri, line 170:
```suggestion
   inspect it from File Explorer. You can also use the above ``tar`` command in a
   command prompt to inspect the `.tar.gz` file. Alternatively, there are GUI programs
   like `7-zip`_ that handle ``.tar.gz`` and ``.whl`` files.
```

abravalheri (approved):
Thank you very much for the help @cdfarrow. I added a minor review comment.

I agree with most of the “extra suggestions” they make sense to me. If you like to expand the PR to include them, please feel free. Otherwise I can follow up in the future.

In general, I believe that this page itself is out of place. It is a “in dept technical” reference documentation, so it should come under some sort of “advanced” section. I also think it focuses too much in the past and how things used to be, instead of going to the point and provide the relevant information. I would like to keep all “how things used to be” description under the “Backward compatibility & deprecated practice” section.

But that is a work in progress, and we will get there eventually.

Thanks again for all the improvements and suggestions. PR like this are extremely valuable and help us to achieve a better future where setuptools documentation is actually good 😊

cdfarrow:
> In general, I believe that this page itself is out of place. It is a “in dept technical” reference documentation, so it should come under some sort of “advanced” section.

Yes, I agree.
I've made a couple more changes to the PR for you to review.
I wondered about adding '(This is normally ``dist/``.)' to the second 'bullet' item about PEP517, but I'm not sure if that is right or not. Where is the dist/ folder actually specified in the whole process?

abravalheri:
Thank you very much @cdfarrow for all the improvements. If you feel comfortable with the PR as it is right now, please let me know and  I will go ahead and merge it. (We can always follow up with more improvements in other PRs, including the docs re-organization).

> I wondered about adding '(This is normally dist/.)' to the second 'bullet' item about PEP517, but I'm not sure if that is right or not. Where is the dist/ folder actually specified in the whole process?

I don't think that the directory got specified. What seems to be happening is that `build` is just carrying out the `setuptools` tradition.

cdfarrow:
Yes, I'm happy for now; you can go ahead and merge it. Thanks.

Issue #2563, opened 2021-02-10
`AttributeError` with `version = attr:` and variable annotation syntax type hints


homeworkprod:
**Observed behavior**

`setup.cfg` contains:
```ini
[metadata]
version = attr: mypackage.VERSION
```

`mypackage/__init__.py` contains:
```py
VERSION: str = '42.0-dev'
```

`pip install -e .` results in:
```pytb
Obtaining […]
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /projectpath/venv/bin/python3.7 /projectpath/venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmpwip8xxcd
       cwd: /projectpath
  Complete output (64 lines):
  Traceback (most recent call last):
    File "/tmp/pip-build-env-tprf3sb3/overlay/lib/python3.7/site-packages/setuptools/config.py", line 35, in __getattr__
      for statement in self.module.body
  StopIteration
  
  The above exception was the direct cause of the following exception:
  
  Traceback (most recent call last):
    File "/tmp/pip-build-env-tprf3sb3/overlay/lib/python3.7/site-packages/setuptools/config.py", line 387, in _parse_attr
      return getattr(StaticModule(module_name), attr_name)
    File "/tmp/pip-build-env-tprf3sb3/overlay/lib/python3.7/site-packages/setuptools/config.py", line 43, in __getattr__
      ) from e
  AttributeError: [mypackage] has no attribute VERSION
  
  During handling of the above exception, another exception occurred:

  […]
```

**Expectation**

It works.

**Workaround**

Updating `mypackage/__init__.py` to
```py
VERSION = '42.0-dev'
```
works.

abravalheri:
Hi @homeworkprod, thank you very much for bringing this topic up.

I did a quick investigation and it seems that this problem has already been solved 👏 

I tried to verify this problem, by using one of your opensource repositories in GitHub:

```bash
> docker run -it --rm python:3.7 /bin/bash
$ git clone https://github.com/homeworkprod/syslog2irc.git
$ cd syslog2irc
$ sed -i 's/VERSION =/VERSION: str =/' src/syslog2irc/__init__.py
$ python3.7 -m pip install -U pip build
$ python3.7 -m build
...
[...]

PR #3391, opened 2022-06-19
Add support for type annotations to `attr:` ast parsing

karlotness:
## Summary of changes

This extends the ast parsing setuptools does to load attributes without evaluating a module to also process simple assignments with type annotations. Those produce a different node `ast.AnnAssign` rather than `ast.Assign` and require slightly different handling since the node has a different attribute `target` vs `targets`.

Since that seemed like it might make the generators more complicated, I reworked things into for loops and made some other tweaks. I'd be happy to take a different approach though.

The existing `ast.Assign` nodes should pass through exactly the same processing. The `AttributeError` that could result should also be the same too, except that it won't be chained with a `StopIteration` if the outer for loop is exhausted (as it would be with the `next()` call).

This doesn't close any existing open issue that I can find. There is #2563 which looks related. I think in that case the attribute loading worked because setuptools fell back to evaluating the module when the ast processing likely failed.

That fallback works in some cases, but it causes problems in when the module isn't importable until after the build (for example, extension modules that need to be compiled). The ast parsing is really helpful in these cases, and it would be nice if it could also handle type annotations.

### Pull Request Checklist
- [X] Changes have tests
- [X] News fragment added in [`changelog.d/`].

[PR docs]:

### DIFF
--- a/changelog.d/3391.change.rst
+++ b/changelog.d/3391.change.rst
@@ -1,0 +1,1 @@
+Updated ``attr:`` to also extract simple constants with type annotations -- by :user:`karlotness`
--- a/setuptools/config/expand.py
+++ b/setuptools/config/expand.py
@@ -69,24 +69,21 @@ class StaticModule:
     def __getattr__(self, attr):
         """Attempt to load an attribute "statically", via :func:`ast.literal_eval`."""
         try:
-            assignment_expressions = (
-                statement
-                for statement in self.module.body
-                if isinstance(statement, ast.Assign)
-            )
-            expressions_with_target = (
-                (statement, target)
-                for statement in assignment_expressions
-                for target in statement.targets
-            )
-            matching_values = (
-                statement.value
-                for statement, target in expressions_with_target
-                if isinstance(target, ast.Name) and target.id == attr
-            )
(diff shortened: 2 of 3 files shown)


abravalheri:
Thank you very much for the contribution @karlotness (and the awesome attention to details, I completely forgot that the AST will handle typed assignments differently).

I notice that the is one edge case that might break:

```python
attr: str
```

The AST also parses this statement as an `ast.AnnAssign`, but the value is set to `None`. If we try to parse `None` via `ast.literal_eval` we probably get an expression.

I will propose a change directly to your repository via PR (if you merge it, this PR is automatically updated). The suggestions I will send you will also include a way of having the generators back and avoid the duplication when raising the exception.

abravalheri (changes requested):
Please consider the fix suggested in https://github.com/karlotness/setuptools/pull/1 for the edge case when `ast.AnnAssign` has a `None` value

karlotness:
That looks good, thanks. I hadn't considered the split type-annotation and later assignment. I took a look and you're right, looks like that would have produced an incorrect None. Thanks for taking a look and for the quick fix.

abravalheri (approved):

PR #3407, opened 2022-06-24
[Docs] Add two footnotes about PEP 660 support

kngwyu:
It looks like that[PEP 660](https://peps.python.org/pep-0660/)  is implemented in pip 21.3 (see https://pip.pypa.io/en/stable/news/#v21-3).
Or isn't it still recommended to do `pip install -e` without `setup.py`? 🤔

### DIFF
--- a/docs/userguide/pyproject_config.rst
+++ b/docs/userguide/pyproject_config.rst
@@ -212,8 +212,7 @@ however please keep in mind that all non-comment lines must conform with :pep:`5
 .. rubric:: Notes
 
 .. [#pep660-status] Editable install without ``setup.py`` will be supported from
-   ``setuptools >= 63.0.0``, but this feature will remain experimental for a while.
-   Check https://github.com/pypa/setuptools/issues/2816 for detail.
+   ``setuptools >= 64.0.0``. Check https://github.com/pypa/setuptools/issues/2816 for detail.
 
 .. [#setupcfg-caveats] ``pip`` may allow editable install only with ``pyproject.toml``
    and ``setup.cfg``. However, this behavior may not be consistent over various build
--- a/docs/userguide/pyproject_config.rst
+++ b/docs/userguide/pyproject_config.rst
@@ -7,8 +7,8 @@ Configuring setuptools using ``pyproject.toml`` files
 .. note:: New in 61.0.0
 
 .. important::
-   For the time being, ``pip`` still might require a ``setup.py`` file
-   to support :doc:`editable installs <pip:cli/pip_install>`.
+   For the time being [#pep660-status]_, ``pip`` still might require a ``setup.py`` file
+   to support :doc:`editable installs <pip:cli/pip_install>` [#setupcfg-caveats]_.
 
    A simple script will suffice, for example:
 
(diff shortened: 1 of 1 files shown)


abravalheri:
Hi @kngwyu thank you very much for the PR.

There is one situation right now that people might still need `setup.py` regardless of the version of `pip`: if they are only using `pyproject.toml` and don't have `setup.cfg`.

Do you think the note still makes sense considering this information?

kngwyu:
> There is one situation right now that people might still need setup.py regardless of the version of pip: if they are only using pyproject.toml and don't have setup.cfg.

Ah thank you for letting me know that! I think that I was confused because I had `setup.cfg` for flake8 so the editable install worked for me without `setup.py`. 

> Do you think the note still makes sense considering this information?

No. I feel like adding a footnote is better than revising this sentence then. Writing everything precisely could make the document too heavy. I'll revise this PR next monday.

abravalheri:
Thank you very much @kngwyu for having a look on this.

kngwyu:
@abravalheri So because of https://github.com/pypa/pip/pull/10577, even an empty `setup.cfg` and `pyproject.toml` would work for editable install, but we should recommend users to use `setup.py`, right?

kngwyu:
And we don't need `setup.py/cfg` with setuptools > 63.0 since it now has PEP 660 backend?
Sorry for many questions 😅

kngwyu:
Thank you for answering!
I revised this PR to only add footnotes.

docs/userguide/pyproject_config.rst
```diff
@@ -211,6 +211,14 @@ however please keep in mind that all non-comment lines must conform with :pep:`5
 
 .. rubric:: Notes
 
+.. [#pep660-status] Editable install without ``setup.py`` will be supported from
+   ``setuptools >= 63.0.0``, but this feature will remain experimental for a while.
+   Check https://github.com/pypa/setuptools/issues/2816 for detail.
```
MuellerSeb, line 216: Since setuptools 63 was released without this support, I am a bit confused now.

abravalheri: Sorry for the confusion. There was a different major change that got in the pipeline before this one.

PEP 660 will probably get in on v64. It is currently under tests.

kngwyu: Ah, thank you for pointing it out. I misunderstood the situation.

kngwyu:
> Would you mind if we merge this PR together with the feature/pep660 branch? This way it lands on the docs together with the actual implementation.

Sounds good to me!

docs/userguide/pyproject_config.rst
```diff
@@ -7,8 +7,8 @@ Configuring setuptools using ``pyproject.toml`` files
 .. note:: New in 61.0.0
 
 .. important::
-   For the time being, ``pip`` still might require a ``setup.py`` file
-   to support :doc:`editable installs <pip:cli/pip_install>`.
+   For the time being [#pep660-status]_, ``pip`` still might require a ``setup.py`` file
+   to support :doc:`editable installs <pip:cli/pip_install>` [#setupcfg-caveats]_.
```
pradyunsg, line 11: It does not -- and this note can be removed.

`pip` will gracefully figure out whether to use PEP 660, or invoke setup.py manually. I believe this was from a pre-PEP 660 implementation days.

abravalheri:
@kngwyu #3488 is just around the corner so I decided to "clean the PR queue".
I changed a little bit the notes to avoid mentioning a specific version number, since we never know if the version gets bumped meanwhile...

PR #3436, opened 2022-07-08
Change line endings for `.cmd` file to LF

pradyunsg:
This is more compatible with users who don't transform all newlines to
CRLF on Windows.

This is a no-op in terms of user-facing changes, so this doesn't need changelog or an issue IMO. Feel free to close this, if you reckon this is not useful. :)


jaraco:
When I was a Windows user, I preferred to unify around LF for line endings. There were some editors that were hostile to files that did not include the CR in the newline, but most modern editors would handle either. In my opinion, it's better to converge rather than try to maintain divergence.

But wait, the description for this PR says "to CRLF" but when I look at the file, it currently has CRLF. Are you sure you aren't changing the line endings to LF?

jaraco:
Just to confirm, the change is to move from CRLF to LF:

```
 setuptools fix-cmd-line-endings $ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
 setuptools main $ pip-run -q jaraco.text -- -m jaraco.text.show-newlines tools/msvc-build-launcher.cmd
newline is '\r\n'
 setuptools main $ gh pr checkout 3436
Switched to branch 'fix-cmd-line-endings'
 setuptools fix-cmd-line-endings $ git-id
2d8c81ea
 setuptools fix-cmd-line-endings $ pip-run -q jaraco.text -- -m jaraco.text.show-newlines tools/msvc-build-launcher.cmd
newline is '\n'
```

PR #3472, opened 2022-07-30
docs: Fix reference to old exception name

dnicolodi:
Looks like the name of an exception from the distutils made into the doc. Small formatting update on top.


abravalheri:
Thank you very much for the contribution @dnicolodi.

PR #3476, opened 2022-08-01
Fix finding library files

TimotheusBachinger:
* this was broken after refactoring: https://github.com/pypa/setuptools/commit/25dcca2902780146a1ee298f18648334eb3cddff

## Summary of changes

Since setuptools 63.3.0, pillow couldn't be built anymore, see https://github.com/python-pillow/Pillow/issues/6471

Closes 

### Pull Request Checklist
- [x] Changes have tests: Building pillow is possible again after this change

[PR docs]:

### DIFF
--- a/changelog.d/3476.change.rst
+++ b/changelog.d/3476.change.rst
@@ -1,0 +1,1 @@
+Fixed a regression introduced in 63.3.0 which broke e.g. the build from source of Pillow. Needed libraries where not found anymore. -- by :user:`timotheus.bachinger`
--- a/setuptools/_distutils/unixccompiler.py
+++ b/setuptools/_distutils/unixccompiler.py
@@ -373,8 +373,8 @@ class UnixCCompiler(CCompiler):
 
         searched = (
             os.path.join(root, lib_name)
-            for root in map(self._library_root, dirs)
             for lib_name in lib_names
+            for root in map(self._library_root, dirs)
         )
 
         found = filter(os.path.exists, searched)


abravalheri:
Hi @TimotheusBachinger thank you very much for the contribution and fix!

Would it be possible to open the PR directly against [`pypa/distutils`](https://github.com/pypa/distutils)? Everything under the `setuptools/_distutils` folder is periodically pulled from `pypa/distutils`. I am afraid that if we fix it directly in setuptools, it might get overwritten at some point.

TimotheusBachinger:
@abravalheri sure, here we go: https://github.com/pypa/distutils/pull/163

abravalheri:
> However, I am not sure as https://github.com/pypa/distutils#to-setuptools states:
>> Simply merge the changes directly into setuptools' repo.

Oh, I always interpreted this phrase in a different way[^1], but maybe you are correct...

[^1]: As an instruction to setuptools maintainters to pull in the changes from `distutils` (via merge).

TimotheusBachinger:
@abravalheri I am also not sure. But no matter where the fix gets merged, we would need a new release of setuptools in order to get our build chain running again. 
In order to proceed, I suggest the following: 

- I've added now the changelog fragement
- merge the fix here - and also in distutils later
- make a hotfix release of setuptools (not really knowing your release cycles...)

What do you think?

nanonyme:
It would indeed be great to get clarification whether pypa/distutils is at this point supposed to be a completely obsolete repo as changes do not go into stdlib and they should be merged directly her. Perhaps that repo should be marked read-only?

abravalheri:
@nanonyme , I don't think that would be the case. If you check the list of closed PRs in `pypa/distutils`, you can see that the repository is very active and maintained.

So far my understanding is that fixes on `pypa/distutils` should be directed there, but no new features.

New features should be implemented in `setuptools`, bit without directly changing the `setuptools/_distutils` folder.

abravalheri:
@TimotheusBachinger , usually the procedure is very similar to what you described: Jason reviews the PRs in distutils and then merges then back to setuptools, creating a minor or patch release depending on the size of the changes.

nanonyme:
@abravalheri perhaps active changes but is it a project with a future? They have never done releases other than as part of CPython or setuptools.

abravalheri:
@nanonyme , new versions of `distutils`, (bugfixes and maintenance releases) are currently distributed as part of `setuptools`. The project does not release independent packages on PyPI.

But in terms of project management, it should be OK to keep the administration of the 2 code bases separated... It is just a different style of development.

In my mind the TL;DR is: if you fix bugs in the distutils code, you can provide a PR to `pypa/distutils`, if you implement a new feature, you can provide a PR to `pypa/setuptools` (the `setuptools/_distutils` folder should not be modified).

TimotheusBachinger:
@abravalheri fair enough! so let's wait for @jaraco to have a look at the PR on `distutils`

Issue #3467, opened 2022-07-26
[FR] Improve error feedback for improper one-liner requirement  with environment markers in `setup.cfg`


frenzymadness:
### setuptools version

63.2.0

### Python version

3.9 and 3.11

### OS

Fedora Linux

### Description

Project: https://github.com/thoth-station/micropipenv
The line I have a problem with: https://github.com/thoth-station/micropipenv/blob/2127293c4c9fa344fd1535c5189c39ca7ba4594d/setup.cfg#L40

It seems to me that there is a difference between this:
```
[options.extras_require]
toml = toml;python_version<"3.11"
```
and this:
```
[options.extras_require]
toml =
    toml;python_version<"3.11"
```

Micropipenv version 1.4.0 contains the [toml] definition on a single line and when you try to install micropipenv[toml], it does not work (Python 3.9.13):
```
Collecting micropipenv[toml]==1.4.0
  Using cached micropipenv-1.4.0-py3-none-any.whl (38 kB)
Requirement already satisfied: pip>=9 in /home/lbalhar/.virtualenvs/micropipenv/lib/python3.9/site-packages (from micropipenv[toml]==1.4.0) (22.2)
Collecting toml
  Using cached toml-0.10.2-py2.py3-none-any.whl (16 kB)
ERROR: Could not find a version that satisfies the requirement python-version<"3.11"; extra == "toml" (from micropipenv[toml]) (from versions: 0.0.2)
ERROR: No matching distribution found for python-version<"3.11"; extra == "toml"
```
the same happens when you use Python 3.11.

But when I use the multiline definition:
```
[options.extras_require]
toml =
    toml;python_version<"3.11"
```
it works correctly and omits toml in env with Python 3.11 and installs it if Python<3.11.

When I build a sdist from the package, there is a difference in requires.txt in egg-info directory:
single line definition:
```
pip>=9

[toml]
toml
python_version<"3.11"
```
two lines:
```
pip>=9

[toml]

[toml:python_version < "3.11"]
toml
```

### Expected behavior

I'd expect the two definitions mentioned above to behave the same way.

### How to Reproduce

You can try to install "micropipenv[toml]==1.4.0" and then you can change the extras_require definition as described above to see that it fixes the problem.

### Output

```console
$ tail -n 3 setup.cfg

[options.extras_require]
toml = toml;python_version<"3.11"

$ pip install .[toml]
Processing /home/lbalhar/Software/micropipenv
  Installing build dependencies ... done
[...]

PR #3481, opened 2022-08-03
Add warning for potential extras_require misconfiguration

frenzymadness:
Fixes: https://github.com/pypa/setuptools/issues/3467

## Summary of changes

See the linked issue, all details are there.

Closes https://github.com/pypa/setuptools/issues/3467

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/config/setupcfg.py
+++ b/setuptools/config/setupcfg.py
@@ -465,10 +465,9 @@ class ConfigHandler(Generic[Target]):
         :param dict section_options:
         """
         for (name, (_, value)) in section_options.items():
-            try:
+            with contextlib.suppress(KeyError):
+                # Keep silent for a new option may appear anytime.
                 self[name] = value
-            except KeyError:
-                pass  # Keep silent for a new option may appear anytime.
 
     def parse(self):
         """Parses configuration file items from one
--- a/setuptools/config/setupcfg.py
+++ b/setuptools/config/setupcfg.py
@@ -175,7 +175,7 @@ def parse_configuration(
     return meta, options
 
 
-def warn_accidental_env_marker_misconfig(section_name, section_options, parsed):
+def warn_accidental_env_marker_misconfig(label: str, orig_value: str, parsed: list):
     """Because users sometimes misinterpret this configuration:
 
     [options.extras_require]
@@ -192,20 +192,17 @@ def warn_accidental_env_marker_misconfig(section_name, section_options, parsed):
         leads in a valid Requirement with a valid marker
     a UserWarning is shown to inform the user about the possible problem.
     """
-
-    for name, (file, requirements) in section_options.items():
-        if "\n" not in requirements and len(parsed[name]) == 2:
-            original_requirements_str = ";".join(parsed[name])
-            try:
-                req = Requirement(original_requirements_str)
-            except InvalidRequirement:
-                pass
-            else:
-                if req.marker is None:
-                    continue
+    if "\n" not in orig_value and len(parsed) == 2:
+        original_requirements_str = ";".join(parsed)
+        try:
+            req = Requirement(original_requirements_str)
+        except InvalidRequirement:
+            pass
+        else:
+            if req.marker is not None:
                 msg = (
-                    f"One of the parsed requirements in {section_name} section "
-                    f"looks like a valid environment marker: '{parsed[name][1]}'\n"
+                    f"One of the parsed requirements in `{label}` "
+                    f"looks like a valid environment marker: '{parsed[1]}'\n"
                     "Make sure that the config is correct and check "
                     "https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#opt-2"  # noqa: E501
                 )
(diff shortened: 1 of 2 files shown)


setuptools/tests/config/test_setupcfg.py
```diff
@@ -716,6 +716,23 @@ def test_extras_require(self, tmpdir):
...
             assert dist.metadata.provides_extras == set(['pdf', 'rest'])
 
+    @pytest.mark.parametrize(
+        "config",
+        [
+            "[options.extras_require]\nfoo = bar;python_version<'3'",
+            "[options.extras_require]\nfoo = bar;os_name=='linux'"
+        ]
+    )
+    def test_extras_warns_about_potential_misconfig(self, config, tmpdir):
```
hroncok, line 726: Let's also add tests that assert it does not warn for other cases?

frenzymadness: Done, thanks.

setuptools/tests/config/test_setupcfg.py
```diff
@@ -716,6 +716,41 @@ def test_extras_require(self, tmpdir):
...
+        ]
+    )
+    def test_extras_nowarns(self, config, tmpdir, recwarn):
+        fake_env(tmpdir, config)
+
+        with get_dist(tmpdir) as dist:
+            assert len(dist.extras_require) == 2
+
+        # The examples are valid, no warnings shown
+        assert len(recwarn) == 0
```
hroncok, line 752: What about explicitly checking for our warnings in case other warnings (e.g. some DeprecationWarnings) happen?

```suggestion
        assert not any(w for w in recwarn if w.category == UserWarning)
```

frenzymadness: And what about:
```suggestion
        assert not any(w.category == UserWarning for w in recwarn)
```
This way it's without automatic `bool(warning)`.

hroncok: Even better!

setuptools/config/setupcfg.py
```diff
@@ -174,6 +175,41 @@ def parse_configuration(
     return meta, options
 
 
+def warn_about_potential_extras_misconfig(section_options, parsed):
```
abravalheri, line 178: `extras` might be a bit confusing here because it could also refer to the `extra` environment marker used inside the requirement. How about `warn_accidental_extras_require_misconfig` (or `verify_accidental_extras_require_misconfig`)?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -716,6 +716,41 @@ def test_extras_require(self, tmpdir):
             }
             assert dist.metadata.provides_extras == set(['pdf', 'rest'])
 
+    @pytest.mark.parametrize(
+        "config",
+        [
+            "[options.extras_require]\nfoo = bar;python_version<'3'",
+            "[options.extras_require]\nfoo = bar;os_name=='linux'"
```
abravalheri, line 723: It seems that the `fake_env` simply writes the given string as a `setup.cfg` file:

```python
config.write(setup_cfg.encode(encoding), mode='wb')
```

My question here is: what happens if the text file ends in `\n`? (all text files in POSIX systems should end in a `\n` character right?)

Is it worthy to change the examples to (or add them as extra examples):
```python
"[options.extras_require]\nfoo = bar;python_version<'3'\n",
"[options.extras_require]\nfoo = bar;os_name=='linux'\n"
```

Should we also add something like the following to the list of examples? (In theory, `ConfigParser` should discard that line, so when it reaches `setuptools` the comment is already not an issue. I am not sure if it is worthy adding to the test).

```ini
[options.extras_require]
foo = bar; python_version<'3'
      # I am a comment
```

setuptools/tests/config/test_setupcfg.py
```diff
@@ -716,6 +716,41 @@ def test_extras_require(self, tmpdir):
...
+                # We are testing the warning,
+                # the result should be the same
+                assert len(dist.extras_require) == 1
+                assert len(dist.extras_require["foo"]) == 2
+
+    @pytest.mark.parametrize(
+        "config",
+        [
+            "[options.extras_require]\nfoo =\n    bar;python_version<'3'",
+            "[options.extras_require]\nfoo = bar;baz\nboo = xxx;yyy"
```
abravalheri, line 742: Maybe we could also add here another valid example:

```ini
[options.extras_require]
foo =
    bar
    python_version<'3'
```
(https://pypi.org/project/python_version/)

setuptools/config/setupcfg.py
abravalheri, line 178: If we are verifying the `extras_require`, does it also make sense to reuse the same logic to verify `install_requires`? (I suppose the same problems can be found there, right?)

abravalheri (commented):
Thank you very much @frenzymadness for the work in this improvement. I added a few review comments/suggestions (but please feel free to disagree or propose alternatives).

setuptools/config/setupcfg.py
frenzymadness: In that case, I'd name the function like `warn_accidental_env_marker_misconfig` because you basically want to add an env marker and it ends up parsed as a package. Then we can use the same function when parsing extras_require as well as install_requires. What do you think?

setuptools/tests/config/test_setupcfg.py
frenzymadness: Will add some more tests as suggested.

setuptools/config/setupcfg.py
abravalheri: Sounds very good!

frenzymadness:
I believe I addressed all the comments.

frenzymadness:
I have one fixup commit there. Don't forget to squash it to the previous one before merging. Thanks for helping me with this!

abravalheri:
I was writing the following message before I noticed that I did a mistake in the refactoring and went back to correct it:

> Hi @frenzymadness, thank you very much for the modifications!
> I did some refactoring on the file to better integrate with your changes (I also refactored a little bit your code). I hope you don't mind. (The idea was to speed up the ping pong review process, please feel free to change it, if you think it is not appropriate).

But I guess you already noticed that 🤣 

> I have one fixup commit there. Don't forget to squash it to the previous one before merging. Thanks for helping me with this!

Ok, I will do it.

abravalheri:
Thank you very much @frenzymadness, I plan to merge this PR together with a few other ones in a new release soon.

PR #3491, opened 2022-08-06
Add docs section on direct URL dependencies

### LINKED ISSUE #3489: [Docs] Short explanation or example on direct URL dependencies AKA direct reference specifications
### Summary

setuptools allows specifying "direct URL dependencies" on packages that are to be pulled from a VCS (e.g. Git) repository in the form
```
package-name @ git+https://vcs-hoster.net/repository.git@branch
```
where `branch` could also be another kind of Git ref (e.g. tag). Such dependencies are not allowed in packages on PyPI but come in handy in many situations anyway.

As far as I can tell, this format isn't explicitly *spelled out* anywhere in setuptools' docs, and judging by [answers and comments about it on StackOverflow](https://stackoverflow.com/a/54794506), I'm not the only one who feels that it's unnecessarily hard to find out about. It would be great if there was a short section on it or just an example in the docs.

### OS / Environment

irrelevant

### Additional Information

Several other pieces of documentation *come close* to documenting it but don't quite put all the parts together:

- setuptools' [docs on dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html) don't contain any examples with this format, but mention that dependencies need to be specified according to PEP 508 and point out that "direct URLs" from PEP 440 are not supported by PyPI. They helpfully link to both.
- [PEP 508](https://peps.python.org/pep-0508/) specifies a grammar for dependency specifications which allow for the above format, which it calls "URL reference form" and "direct reference specification". However, it doesn't explain the semantics of the various components. Just like setuptools, PEP 508 also links to PEP 440 for more information on this form.
- [PEP 440](https://peps.python.org/pep-0440/) finally has a [section on "direct references"](https://peps.python.org/pep-0440/#direct-references) with more detailed semantics. As pointed out in an infobox there, however, this format differs slightly from the the one supported by pip (and setuptools) in that it specifies a commit hash should be included, which is not actually required by pip (and setuptools). It also doesn't specify which VCS protocols are actually supported (but gives git as an example) and only mentions tags and commit hashes as possible targets for the `@target` suffix, not branches (you could guess it in Git's case, though).
- pip has a section on [VCS support](https://pip.pypa.io/en/latest/topics/vcs-support/) where the format of these kinds of package specifications *without the commit hash requirement* is explained in detail, including which VCSs are supported, but it only mentions all this in the context of installing these packages directly via its CLI. What's missing for someone to be able to use this within a package is the information that this is also valid as a dependency specification in setuptools' `setup.cfg`/`setup.py`, which of course would be out of place in pip's docs.

So if you spend enough time and already know what you're looking for you can kind of sort of piece it together from official sources only (with some uncertainty like the commit hash requirement), but would anyone object to there being a small section on this with an example in setuptools' docs on dependencies (with another big red box above saying "THIS CAN'T BE USED ON PYPI" of course) so people can find it more easily?

Note: This is essentially the same issue as #3250 (unfortunately closed by the reporter) with more details added.

### Code of Conduct

smheidrich:
## Summary of changes

This PR adds a section on direct URL dependencies (AKA direct reference dependencies) to Setuptools' dependency management docs.
It also pulls some information out of an "attention" box at the top which I felt was more appropriate for the text's main body and moves the actual "attention" part into the newly added section where it (IMHO) belongs.

Closes #3489 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/docs/userguide/dependency_management.rst
+++ b/docs/userguide/dependency_management.rst
@@ -183,7 +183,7 @@ Direct URL dependencies
 
 .. attention::
    `PyPI`_ and other standards-conformant package indices **do not** accept
-   packages that declare dependencies using direct URLs. pip will accept them
+   packages that declare dependencies using direct URLs. ``pip`` will accept them
    when installing packages from the local filesystem or from another URL,
    however.
 
--- a/docs/userguide/dependency_management.rst
+++ b/docs/userguide/dependency_management.rst
@@ -6,13 +6,10 @@ There are three types of dependency styles offered by setuptools:
 1) build system requirement, 2) required dependency and 3) optional
 dependency.
 
-.. attention::
-   Each dependency, regardless of type, needs to be specified according to :pep:`508`.
-   This allows adding version :pep:`range restrictions <440#version-specifiers>`
-   and :ref:`environment markers <environment-markers>`.
-   Please note however that public package indexes, such as `PyPI`_
-   might not accept packages that declare dependencies using
-   :pep:`direct URLs <440#direct-references>`.
+Each dependency, regardless of type, needs to be specified according to :pep:`508`
+and :pep:`440`.
+This allows adding version :pep:`range restrictions <440#version-specifiers>`
+and :ref:`environment markers <environment-markers>`.
 
 
 .. _build-requires:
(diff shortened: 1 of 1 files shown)


docs/userguide/dependency_management.rst
```diff
@@ -181,6 +178,58 @@ detailed in :pep:`508`.
    to implement custom detection logic.
 
 
+Direct URL dependencies
+-----------------------
+
+.. attention::
+   `PyPI`_ and other standards-conformant package indices **do not** accept
+   packages that declare dependencies using direct URLs. pip will accept them
```
abravalheri, line 186:
```suggestion
   packages that declare dependencies using direct URLs. ``pip`` will accept them
```

abravalheri (approved):
Thank you very much @smheidrich.

smheidrich:
That was fast :) Thanks for merging!

PR #3516, opened 2022-08-12
Remove extraneous question mark in section heading

mhucka:
The heading "How editable installations work?" either contains a typo (the question mark should be omitted), or else the heading as written is ungrammatical because it is missing a word ("How **do** editable installations work?").

## Summary of changes

Other headings in this section are not written as questions, so this commit removes the question mark, based on the assumption that the question mark is extraneous.

### Pull Request Checklist
- <s>[ ] Changes have tests.</s> Not relevant.

This is such a small typo that it does not seem to warrant a changelog.d file.


abravalheri:
Thank you very much @mhucka

Issue #3525, opened 2022-08-15
[BUG]  invalid command 'bdist_wininst' during bdist_wheel


teije01:
### setuptools version

setuptools==65.0.0

### Python version

python 3.8

### OS

Debian GNU/Linux 11 (bullseye)

### Additional environment information

Last known working version was setuptools 63.4.3, did not test with 64.*.*

### Description

Latest setuptools causes pip install to exit with the following error:

`error: invalid command 'bdist_wininst'`

### Expected behavior

```
  Building wheel for vds-base-cli (setup.py): started
  Building wheel for vds-base-cli (setup.py): finished with status 'done'
```

### How to Reproduce

Created an environment with conda:

```bash
conda create -n test python=3.8 --yes
conda activate test
pip install "setuptools==65.0.0"
pip install vds-api-client
pip install --no-cache-dir --force --no-deps --no-binary :all: vds-api-client  # Fails
```

And then re-installing with setuptools 63.4.3 succeeds

```bash
pip install setuptools==63.4.3
pip install --no-cache-dir --force --no-deps --no-binary :all: vds-api-client
```

### Output

traceback during pip install

```console
pip install --no-cache-dir --force --no-deps --no-binary :all: vds-api-client
Looking in indexes: https://pypi.org/simple, https://__token__:****@gitlab.vandersat.com/api/v4/projects/178/packages/pypi/simple, https://pypi.prod.planet-labs.com/simple/
Collecting vds-api-client
  Downloading vds_api_client-2.1.4.tar.gz (40 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 40.9/40.9 kB 73.1 MB/s eta 0:00:00
  Preparing metadata (setup.py) ... done
Skipping wheel build for vds-api-client, due to binaries being disabled for it.
Installing collected packages: vds-api-client
  Attempting uninstall: vds-api-client
    Found existing installation: vds-api-client 2.1.4
    Uninstalling vds-api-client-2.1.4:
      Successfully uninstalled vds-api-client-2.1.4
  Running setup.py install for vds-api-client ... error
  error: subprocess-exited-with-error
  
  × Running setup.py install for vds-api-client did not run successfully.
  │ exit code: 1
  ╰─> [65 lines of output]
      /home/teije/miniconda3/envs/test/lib/python3.8/site-packages/setuptools/dist.py:771: 
        warnings.warn(
[...]

PR #3531, opened 2022-08-16
Remove code reliant on bdist_wininst

jaraco:
## Summary of changes

Ref #3525

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
I'm not confident this change is still needed, but I read through the bug and it does seem that this code isn't needed and so I'll proceed with removing it, so it's not around to cause problems. Feel free to revert if it causes any trouble.

PR #3547, opened 2022-08-19
Remove inadvertent splatting of the name attribute

### LINKED ISSUE #3545: [BUG] `ConfigDiscovery.analyse_name` incorrectly sets the `name` attribute on `Distribution`
### setuptools version

setuptools==65.1.0

### Python version

Python 3.10.2

### OS

macOS 12.5, Oracle Linux Server release 8.6

_No response_

### Description

Under certain circumstances (see reproduction steps) `setuptools.discovery.ConfigDiscovery.analyse_name` will overwrite the `name` attribute of the `Distribution` object leading to `python setup.py --name` failing.

### Expected behavior

```
$ python setup.py --name
my_package_name
```

### How to Reproduce

A project demonstrating this issue (along with a detailed description of what's going on) can be found here:

https://github.com/jeamland/setuptools-name-bug

### Output

```console
$ python setup.py --name
/Users/benno/.pyenv/versions/setuptools-name-bug/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py:108: _BetaConfiguration: Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.
  warnings.warn(msg, _BetaConfiguration)
  File "/Users/benno/work/setuptools-name-bug/setup.py", line 3, in <module>
    setup()
  File "/Users/benno/.pyenv/versions/setuptools-name-bug/lib/python3.10/site-packages/setuptools/__init__.py", line 87, in setup
  File "/Users/benno/.pyenv/versions/setuptools-name-bug/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 170, in setup
  File "/Users/benno/.pyenv/versions/setuptools-name-bug/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 471, in parse_command_line
    args = parser.getopt(args=self.script_args, object=self)
  File "/Users/benno/.pyenv/versions/setuptools-name-bug/lib/python3.10/site-packages/setuptools/_distutils/fancy_getopt.py", line 274, in getopt
    val = getattr(object, attr, 0) + 1
TypeError: can only concatenate str (not "int") to str
```

jeamland:
The `name` attribute of a `Distribution` object is used by the command-line processing system and is not intended to hold the name of the distribution itself. Setting it to the name will cause the command-line processing system to have a bad time.

Fixes #3545

### DIFF
--- a/changelog.d/3547.change.rst
+++ b/changelog.d/3547.change.rst
@@ -1,1 +1,1 @@
-Stop `ConfigDiscovery.analyse_name` from splatting the `Distribution.name` attribute -- by :user:`jeamland`
+Stop ``ConfigDiscovery.analyse_name`` from splatting the ``Distribution.name`` attribute -- by :user:`jeamland`
--- a/changelog.d/3547.change.rst
+++ b/changelog.d/3547.change.rst
@@ -1,0 +1,1 @@
+Stop `ConfigDiscovery.analyse_name` from splatting the `Distribution.name` attribute -- by :user:`jeamland`
--- a/setuptools/config/pyprojecttoml.py
+++ b/setuptools/config/pyprojecttoml.py
@@ -234,8 +234,8 @@ class _ConfigExpander:
 
         # A distribution object is required for discovering the correct package_dir
         dist = self._ensure_dist()
-
-        with _EnsurePackagesDiscovered(dist, self.setuptools_cfg) as ensure_discovered:
+        ctx = _EnsurePackagesDiscovered(dist, self.project_cfg, self.setuptools_cfg)
+        with ctx as ensure_discovered:
             package_dir = ensure_discovered.package_dir
             self._expand_data_files()
             self._expand_cmdclass(package_dir)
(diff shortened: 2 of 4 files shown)


abravalheri:
Hi @jeamland, thank you very much for providing this.

Would it be possible to add some tests to avoid regression? (Usually what I would do is create another commit with the tests, then reorder them, in such a way that the first commit demonstrates the tests failing and the second one fixes them).

I think there is another more subtle aspect that is also worthy investigating (in addition to the changes you are proposing here). Setuptools should not try to discover the distribution name, because it is given... I can have a look on that myself after we merge this PR  (but by any means feel free to investigate if you are interested).

abravalheri:
Hi @jeamland, I took advantage of this PR already being open and added:

- Tests to ensure we don't have problems setting `dist.name`
- Tests/Fixes to ensure name auto-discovery does not happen when the user explicitly sets the name

I also reorganised the order of the commits, with the tests being introduced prior to the fixes. This way we can verify that the errors happen without the fixes.

I hope you don't mind the modifications. But please feel free to modify the code to your liking, or propose different changes.

jeamland:
All looks good to me. Thanks!

PR #3554, opened 2022-08-21
[Docs] Typo in Dependency Management pyproject.toml

mfbutner:
Under the Dependency Management pyproject.toml example, requests was mistyped as requires.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 65.1.1
+current_version = 65.2.0
 commit = True
 tag = True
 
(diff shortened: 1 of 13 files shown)


abravalheri:
Thank you very much @mfbutner.

`requires` is also a valid dependency, but I guess `requests` can be less confusing for starters.

mfbutner:
Thanks for letting me know that `requires` is also valid. I thought it was a typo because the corresponding examples for setup.py and setup.cfg use `requests` and `requests` was used in the other examples throughout the page that use dependencies.

abravalheri:
Yeah, having all the dependencies on the same page being equivalent is a nice touch. I agree that using `requests` makes the docs less confusing for the users.

Thank you very much!

PR #3595, opened 2022-09-18
Catch an edge case in expand._assert_local()

mssalvatore:
## Summary of changes

Using `str.startswith()` for sandboxing file access has an edge case where someone can access files outside the root directory. For example, consider the case where the root directory is "/home/user/my-package" but some secrets are stored in "/home/user/my-package-secrets". Evaluating a check that "/home/user/my-package-secrets".startswith("/home/user/my-package") will return True, but the statement's intention is that no file outside of "/home/user/my-package" can be accessed.

Using `pathlib.Path.resolve()` and `pathlib.Path.parents` eliminates this edge case.

See https://salvatoresecurity.com/preventing-directory-traversal-vulnerabilities-in-python/ for more details.

### Caveats
1. If there is an infinite loop in the directory structure, `pathlib.Path.resolve()` will raise a `RuntimeError`.
2. `pathlib.Path.resolve()` resolves symlinks, whereas the old code using `os.pathlib.abspath()` did not.

I am not intimately familiar with setuptools and could use some guidance as to whether or not either of these two caveats would cause an issue. My guess would be that the first would not impact users and the second might, but probably won't. If either of the above caveats is an issue, I can provide an alternate implementation using `os.path.abspath()`, though that comes with its own caveats.

### Pull Request Checklist
- [x] Changes have tests
- [ ] News fragment added in [`changelog.d/`]. - This PR should not change the way users interact with setuptools, with the possible exception of the caveats above.

[PR docs]:

### DIFF
--- a/setuptools/config/expand.py
+++ b/setuptools/config/expand.py
@@ -41,6 +41,7 @@ from typing import (
     Union,
     cast
 )
+from pathlib import Path
 from types import ModuleType
 
 from distutils.errors import DistutilsOptionError
@@ -150,7 +151,9 @@ def _read_file(filepath: Union[bytes, _Path]) -> str:
 
 
 def _assert_local(filepath: _Path, root_dir: str):
-    if not os.path.abspath(filepath).startswith(root_dir):
+    # NOTE: Path.resolve() will raise RuntimeError if an infinite loop is
+    # encountered along the resolution path of root_dir or file_path.
+    if Path(root_dir).resolve() not in Path(filepath).resolve().parents:
         msg = f"Cannot access {filepath!r} (or anything outside {root_dir!r})"
         raise DistutilsOptionError(msg)
 
--- a/setuptools/config/expand.py
+++ b/setuptools/config/expand.py
@@ -151,9 +151,7 @@ def _read_file(filepath: Union[bytes, _Path]) -> str:
 
 
 def _assert_local(filepath: _Path, root_dir: str):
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you very much for working on this, @mssalvatore.

I can see that you have used `resolve` instead of `abspath`. Is there a rationale behind this choice?
I believe that this changes a little bit the semantics of the operation right? This would mean that developers cannot use symbolic links in their projects, which could mean that the change is not backwards compatible (it also introduces file-system traversing, instead of string operation, which you noted in your comments can result in recursion problems) ...

mssalvatore:
@abravalheri Using`resolve` instead of `abspath` does indeed change the semantics of the operation. `resolve` will canonicalize the path (resolving symlinks) and `abspath` won't. The docstrings for `read_files()` and a few other functions say that the "function is sandboxed and won't reach anything outside `rootdir`". For that to be true, strictly speaking, you need `resolve` instead of `abspath`.

If it's acceptable or desirable to allow symlinks to be traversed, I can reimplement this using `abspath`. Just LMK.

abravalheri:
Hi @mssalvatore thank you very much for having a look on this.

I think it is fair to assume that if the user has a symbolic link, they want it to be considered as a local file. I am also concerned about the backward compatibility...

My opinion is that it would be better to go with a loose interpretation and use `abspath` in this particular scenario. Unless there is a good reason to do otherwise.

mssalvatore:
@abravalheri I've pushed a commit to use `abspath()` and still handle the edge case.

PR #3601, opened 2022-09-21
use filename of RTD configuration

zacharyburnett:
## Summary of changes

the RTD documentation indicates a specific filename `.readthedocs.yaml`: https://docs.readthedocs.io/en/stable/config-file/v2.html

### Pull Request Checklist
- ~~[ ] Changes have tests~~

[PR docs]:


webknjaz:
`.yml` and `.yaml` are often used interchangeably. The same docs page suggests having `.yml` for other files. I don't think this is really an issue. I'd say that it's the RTDs docs that need mentioning, all the file names they recognize (and have been recognizing for years).

zacharyburnett:
> `.yml` and `.yaml` are often used interchangeably. The same docs page suggests having `.yml` for other files. I don't think this is really an issue. I'd say that it's the RTDs docs that need mentioning, all the file names they recognize (and have been recognizing for years).

Thank you for your response! I apologize, I was confused as it [mentions here specifically](https://docs.readthedocs.io/en/stable/config-file/index.html?highlight=.readthedocs.yml#configuration-file) that `.readthedocs.yml` is deprecated:
> Some other variants like `readthedocs.yaml`, `.readthedocs.yml`, etc are deprecated.

However, that might be a typo and they might have meant `readthedocs.yml` instead.

abravalheri:
Hi @zacharyburnett, thank you very much for the contribution.

My understanding is that the leading developer of setuptools usually prefers changes in common configuration files (e.g. CI or docs) to be contributed directly to the https://github.com/jaraco/skeleton repository and then merged back in setuptools (I might be wrong here, sorry if I am mistaken).

Maybe `jaraco/skeleton` is a better forum to discuss the `readthedocs.yaml` vs `.readthedocs.yml` topic?

zacharyburnett:
> Maybe `jaraco/skeleton` is a better forum to discuss the `readthedocs.yaml` vs `.readthedocs.yml` topic?

Thank you! Sorry for putting it in the wrong place!

jaraco:
The change upstream has been merged and will get merged here eventually. Thanks for the contrib!

PR #3626, opened 2022-10-06
Improve compatibility with distutils proper logging

jaraco:
Supports pypa/distutils#183.

### DIFF
--- a/setuptools/logging.py
+++ b/setuptools/logging.py
@@ -24,12 +24,6 @@ def configure():
         format="{message}", style='{', handlers=handlers, level=logging.DEBUG)
     monkey.patch_func(set_threshold, distutils.log, 'set_threshold')
 
-    # For some reason `distutils.log` module is getting cached in `distutils.dist`
-    # and then loaded again when patched,
-    # implying: id(distutils.log) != id(distutils.dist.log).
-    # Make sure the same module object is used everywhere:
-    distutils.dist.log = distutils.log
-
 
 def set_threshold(level):
     logging.root.setLevel(level*10)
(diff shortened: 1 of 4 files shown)


setuptools/logging.py
```diff
@@ -22,13 +22,8 @@ def configure():
     handlers = err_handler, out_handler
     logging.basicConfig(
         format="{message}", style='{', handlers=handlers, level=logging.DEBUG)
-    monkey.patch_func(set_threshold, distutils.log, 'set_threshold')
-
-    # For some reason `distutils.log` module is getting cached in `distutils.dist`
-    # and then loaded again when patched,
-    # implying: id(distutils.log) != id(distutils.dist.log).
-    # Make sure the same module object is used everywhere:
-    distutils.dist.log = distutils.log
```
jaraco, line 31: I think we should bring this back in the if block above.

setuptools/tests/test_distutils_adoption.py
```diff
@@ -135,24 +135,3 @@ def test_modules_are_not_duplicated_on_import(
...
-print("success")
-"""
-
-
-@pytest.mark.parametrize("distutils_version", "local stdlib".split())
-def test_log_module_is_not_duplicated_on_import(distutils_version, tmpdir_cwd, venv):
-    env = dict(SETUPTOOLS_USE_DISTUTILS=distutils_version)
-    cmd = ['python', '-c', ENSURE_LOG_IMPORT_IS_NOT_DUPLICATED]
-    output = popen_text(venv.run)(cmd, env=win_sr(env)).strip()
-    assert output == "success"
```
jaraco, line 158: Let's restore this test, but skip it when it's irrelevant.

jaraco: Restored in 177a8931.

setuptools/logging.py
jaraco: Fixed in a1a5af3b.

PR #3632, opened 2022-10-10
pyproject_config.rst: Remove beta tag for (optional) dependencies

EwoutH:
Remove the beta tag for dependencies and optional dependencies in the [Configuring setuptools using pyproject.toml files](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) documentation.

### DIFF
--- a/docs/userguide/pyproject_config.rst
+++ b/docs/userguide/pyproject_config.rst
@@ -192,8 +192,8 @@ Key                        Directive           Notes
 ``classifiers``            ``file``            Multi-line text with one classifier per line
 ``entry-points``           ``file``            INI format following :doc:`PyPUG:specifications/entry-points`
                                                (``console_scripts`` and ``gui_scripts`` can be included)
-``dependencies``           ``file``            ``requirements.txt`` format (``#`` comments and blank lines excluded) **BETA**
-``optional-dependencies``  ``file``            ``requirements.txt`` format per group (``#`` comments and blank lines excluded) **BETA**
+``dependencies``           ``file``            ``requirements.txt`` format (``#`` comments and blank lines excluded)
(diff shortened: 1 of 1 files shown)


abravalheri:
I think we can postpone this for a while since we haven't received a lot of feedback yet about using `requirements.txt` files for dependencies.

So far, the only feedback we received is of people trying to use `pip` options like `--editable`, `--index-url`, `--find-links` and being surprised with the errors (this is explicitly documented as not supported). My understanding is that people are having trouble to understand the feature and that it is confusing (so we might want to remove it in the future).

EwoutH:
@abravalheri I was looking into packaging / [Data Files Support](https://setuptools.pypa.io/en/latest/userguide/datafiles.html) in `pyproject.toml`, have you got community feedback on that? And if so, would you say that is still in beta yet?

abravalheri:
Hi @EwoutH, in general, the opportunity of moving to `pyproject.toml` is seens as achance of reimagining some complicated/problematic configuration since it does not require backwards compatibility with `setup.cfg`. The configurations for `include-package-data`, `exclude-package-data` and `package-data` are recurrently mentioned as candidates for this new hypothetical mechanism. However, I haven't been in contact yet with a design proposal.

On the other hand, simple projects should not really need to resort to these configurations: as long as the devs use a well thought directory structure maintaining the relevant files inside the package directory and use a proper `MANIFEST.in` or a VCS plugin like `setuptools-scm`, the wheel distribution will end up fine...

bkane-msft:
I'm using the following block in my `pyproject.toml` for a learning project:

```toml
[tool.setuptools.packages.find]
where = ["."]
include = ["simple_python_package*"]
exclude = ["simple_python_package.tests*", "simple_python_package.scripts*"]
namespaces = false
```

Is this the correct usage? If so, is there any kind of schedule for stabilization?

PR #3638, opened 2022-10-17
Consistently use unittest.mock in tests

hroncok:
## Summary of changes

Consistently use unittest.mock in tests

 - Some tests used unittest.mock from the standard library
 - Some tests used mock from PyPI
 - Some tests tried to import unittest.mock with a fallback to mock (the import never fails on Python 3.7+, older Pythons are not supported)

### Pull Request Checklist
- [ ] Changes have tests -- changes **are** tests

[PR docs]:


webknjaz (approved):
Personally, I'd prefer relying on the pytest's built-in `monkeypatch` fixture and the `pytest-mock` package. But I think this is already a great first step towards consistency.

PR #3671, opened 2022-11-10
Update checkout to v3

ianfhunter:
## Summary of changes

Removes warnings in github actions

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,7 +127,7 @@ jobs:
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 75
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


abravalheri:
Thank you very much @ianfhunter , I suspect that there is some overlap between this change and the other ones being proposed in #3640 , isn't the case?

ianfhunter:
> Thank you very much @ianfhunter , I suspect that there is some overlap between this change and the other ones being proposed in #3640 , isn't the case?

Yep, looks like a duplicate. I can close this one as the other is inclusive of these changes and larger

webknjaz:
FWIW @abravalheri could just merge this and rebase :man_shrugging:

PR #3712, opened 2022-11-26
Improve example docs for setup.cfg and pyproject.toml

mwtoews:
This PR improves the example documentation for both `setup.cfg` and `pyproject.toml`.

Changes to both examples:

- Add an author and email, as these are commonly used (an alternative fictional person could be nominated, if a better one exists)
- Use the SPDX short identifier [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause.html), which aligns with [PEP 639](https://peps.python.org/pep-0639/)

Other:

- With `setup.py` only, add `python_requires` to example, so that it is compatible to `pyproject.toml`

### DIFF
--- a/docs/userguide/declarative_config.rst
+++ b/docs/userguide/declarative_config.rst
@@ -25,10 +25,13 @@ boilerplate code in some cases.
     [metadata]
     name = my_package
     version = attr: my_package.VERSION
+    author = Josiah Carberry
+    author_email = josiah_carberry@brown.edu
     description = My package description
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you @mwtoews.

My understanding is that, so far, the license text is free form. Once PEP 639 is approved, then there will be a separated field for them... Nevertheless it should not hurt to have SPDX in the examples (as long as users don't expect it to work before PEP 639 is implemented/with the wrong fields).

PR #3716, opened 2022-11-29
Catching excption from utime and providing human-readable error description

### LINKED ISSUE #3667: [BUG] Running the "egg_info" command without proper rights results in a cryptic and unhelpful error message
### setuptools version

setuptools == 65.5.1

### Python version

Python 3.8, 3.9

### OS

Ubuntu Mate

I used setuptools via pip. However, the issue is not dependent on how exactly one runs the egg_info command.

### Description

Develoiping a package, I tried to install it using pip with pyproject.toml specifying setuptools as a backend. After running
`pip install -e .`
I got an output with the following error:
` running egg_info
  error: [Errno 13] Permission denied
`
The problem is that this error does not show to what the permission is denied.

I tracked it down, and the output comes from the `utime` function in this part (lines296-988) of `setuptools/command/egg_info.py`
[(Click to see the file in the repo)](https://github.com/pypa/setuptools/blob/main/setuptools/command/egg_info.py#L297)
```python
def run(self):
        self.mkpath(self.egg_info)
        os.utime(self.egg_info, None)
```
So `os.utime` raises an `OSError` which is then printed out. Unfortunately, for some weird reason, the raised error does not contain a properly filled field `filename`, hence one needs a workaround to really print the file out

### Expected behavior

As usually with this kind of error, the expected output is something like:
`error: [Errno 13] Permission denied: /foo/bar/package/src/package-egg.info`

### How to Reproduce

[dist-err.zip](https://github.com/pypa/setuptools/files/9971742/dist-err.zip)
0. I assume that `pip` is installed.
1. Unzip the above very simple package project
2. Go to the main directory of the project (in which the `pyproject.toml` is)
3. run `sudo mkdir src/dist_err_pkg.egg-info` (the quickest way of creating a directory inaccessible for a standard user in ubuntu)
4. run `pip install .`

### Output

```console
  Getting requirements to build wheel ... error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [2 lines of output]
      running egg_info
      error: [Errno 13] Permission denied
      [end of output]
  
× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

```

droodev:
## Summary of changes

An OSError raised by ``utime`` when ``egg-info`` was not writeable was not clear at all---it was only mentioning "Permision denied"; without even specifying the file. So now, when this error occurs, we provide an exception that clearly describes which file causes troubles.

The fact that the original OSError does not have the causing filename set is unclear to me. The OSError class has fields ``filename`` and ``filename2`` that should be populated in such a case but they're not for some reason. 

Closes  #3667

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3667.change.rst
+++ b/changelog.d/3667.change.rst
@@ -1,0 +1,1 @@
+Added a human-readable error description when ``.egg-info`` directory is not writeable -- by :user:`droodev`
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -295,7 +295,11 @@ class egg_info(InfoCommon, Command):
 
     def run(self):
         self.mkpath(self.egg_info)
-        os.utime(self.egg_info, None)
+        try:
+            os.utime(self.egg_info, None)
+        except OSError as e:
+            msg = f"Cannot update time stamp of directory '{self.egg_info}'"
+            raise distutils.errors.DistutilsFileError(msg) from e
         for ep in metadata.entry_points(group='egg_info.writers'):
             writer = ep.load()
             writer(self, ep.name, os.path.join(self.egg_info, ep.name))
--- a/setuptools/tests/test_egg_info.py
+++ b/setuptools/tests/test_egg_info.py
@@ -7,6 +7,7 @@ import stat
 import time
 from typing import List, Tuple
 from pathlib import Path
+from unittest import mock
 
 import pytest
 from jaraco import path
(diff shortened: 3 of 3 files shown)


droodev:
Did my best to comply with the code of conduct. Feel free to point out misbehaves regarding the code should there be any.

Deezzir:
Great fix, saved my live, thanks.

PR #3723, opened 2022-12-14
distutils.ccompiler: Make has_function work with more C99 compilers

### LINKED ISSUE #3648: Unclear how CCompiler.has_function works for functions with parameters
This issue is a continuation of python/cpython#98529.

`CCompiler.has_function` constructs the source code for a linking test for function availability like this:

https://github.com/pypa/setuptools/blob/1617a58ebb12a588d5272bef340c2596ce9c3997/setuptools/_distutils/ccompiler.py#L847-L857

Let's assume that the checked function actually exists and has one or more arguments. There are two cases to consider:

1. Include files are specified, and the function is declared in the include files (or a function-like macro is defined).
2. No include files are specified, or the function is not declared in the specified include files.

In the first case, compilation fails because not enough arguments are specified in the function call.  We never get to the linking stage.

In the second case, compilation still fails with strict C compilers (Clang 16, probably future GCC 14) because implicit function declarations are a C89 feature that was removed from C99 and future language standards.  Again, no link is attempted.

This means that `has_function` really can't detect functions successfully (at least if they expect more than one argument).

Autoconf works around this by supplying a `char foo();` declaration and calling that (still in a link-only test). That still does not allow the detection of alternative macro implementations. It is incompatible with the header file list that is allegedly supported by `has_function`: If you supply your own definition, you must not include any header files that potentially conflict with it.

(Found via the Fedora C99 porting project.)

fweimer-rh:
## Summary of changes

C99 removed support for implicit function declarations.  This means that just calling a function, without declaring the function first, can result in a compilation error.  Today, has_function works with most compilers because they issue just a warning, create an object file, and attempt a link, which then detects available of the symbol at link time, as intended.  With future compilers, compilation will already fail, and no link test is performed.

The has_function interface provides the caller with a way to supply a list of header files to include.  However, even with today's compilers, this only works if the function does not expect any parameters.  Otherwise, the function call in the C fragment created by has_function will not supply the correct argument list and fail compilation.  Therefore, this change supplies and incorrect prototype without arguments.  This is what autoconf does today in a very similar situation, so it is quite likely that compilers will support this construct in this context in the future.

The includes and include_dirs arguments are deprecated because of the parameter list mismatch issue.

Closes #3648.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/changelog.d/3723.change.rst
+++ b/changelog.d/3723.change.rst
@@ -1,0 +1,2 @@
+Updated ``CCompiler.has_function`` to work with more C99 compilers.
+Deprecated the ```includes`` and ``include_dirs`` keyword arguments.
--- a/setuptools/_distutils/ccompiler.py
+++ b/setuptools/_distutils/ccompiler.py
@@ -6,6 +6,7 @@ for the Distutils compiler abstraction model."""
 import sys
 import os
 import re
+import warnings
 
 from .errors import (
     CompileError,
@@ -824,9 +825,19 @@ class CCompiler:
         libraries=None,
         library_dirs=None,
     ):
-        """Return a boolean indicating whether funcname is supported on
-        the current platform.  The optional arguments can be used to
-        augment the compilation environment.
+        """Return a boolean indicating whether funcname is provided as
+        a symbol on the current platform.  The optional arguments can
+        be used to augment the compilation environment.
+
+        The libraries argument is a list of flags to be passed to the
+        linker to make additional symbol definitions available for
+        linking.
+
+        The includes and include_dirs arguments are deprecated.
+        Usually, supplying include files with function declarations
+        will cause function detection to fail even in cases where the
+        symbol is available for linking.
+
         """
         # this can't be included at module scope because it tries to
         # import math which might not be available at that point - maybe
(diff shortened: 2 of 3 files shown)


abravalheri:
Hi @fweimer-rh thabk you very much for the contribution.

Since the `_distutils` folder comes directly from the `pypa/distutils` repository (unaltered) I think the best would be suggesting these changes directly to that other repository (otherwise we have problems for merging/risk of overwriting it in a future change). Would you like to submit the PR to `pypa/distutils` só the distutils maintainers can have a look?

Alternatively you can try to have a look to see if it is possible to implement the changes without touching the `_distutils` folder (I don't know of it is possible...)

thesamesam:
https://github.com/pypa/distutils/pull/195 is now merged and landed in setuptools via https://github.com/pypa/setuptools/pull/3809 in release 67.2.0.

It seems a shame to not fix it in cpython for at least 3.10/3.11 but https://github.com/python/cpython/issues/98529 makes it seem like that's not going to happen.

jaraco:
Sounds like this work is complete. Thanks thesamesam.

By the way, you do get the fix in Python 3.7+ by way of the distutils hack. As long as you have setuptools installed and haven't set `SETUPTOOLS_USE_DISTUTILS=stdlib`, `import distutils` will pull in the `distutils` from Setuptools. I would avoid depending too heavily on that behavior, however, as we'd like to remove it. If you have a use-case for distutils that isn't served by the core interfaces of Setuptools (i.e. requires `import distutils`), please report that as a bug to the distutils project.

PR #3741, opened 2022-12-30
Remove msvc hack for Numpy < 1.11.2

mgorny:
## Summary of changes

Remove the msvc hack that was applied if numpy.distutils were loaded prior to Numpy 1.11.2.  These versions date back to June 2016 and no longer build with Python 3.7.  At the same time, the hack relies on packaging.version.LegacyVersion that is removed in packaging 22.0.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3741.breaking.rst
+++ b/changelog.d/3741.breaking.rst
@@ -1,0 +1,2 @@
+Removed patching of ``distutils._msvccompiler.gen_lib_options``
+for compatibility with Numpy < 1.11.2 -- by :user:`mgorny`
--- a/setuptools/monkey.py
+++ b/setuptools/monkey.py
@@ -157,9 +157,3 @@ def patch_for_msvc_specialized_compiler():
         patch_func(*msvc14('_get_vc_env'))
     except ImportError:
         pass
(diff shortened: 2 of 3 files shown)


mgorny:
Hmm, is CI broken?

abravalheri:
Let's try now with a rebase.


@Mergifyio rebase

abravalheri:
@Mergifyio rebase

jaraco:
Looking at the error message, I think this change introduces a fault. `monkey.patch_all` is no longer able to patch distutils with the removed function. Probably if the hack is no longer needed, it should also be removed from the monkeypatching of distutils.

mgorny:
> Looking at the error message, I think this change introduces a fault. `monkey.patch_all` is no longer able to patch distutils with the removed function. Probably if the hack is no longer needed, it should also be removed from the monkeypatching of distutils.

Thanks. My grep-foo has been poor there.

mgorny:
Ok, I think it's good now.

mgorny:
Thanks!

PR #3769, opened 2023-01-13
Switch to platformdirs

jaraco:
I learned recently that appdirs is abandoned and platformdirs supersedes it, so go through the motions to replace it.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
Aah. I think I see what the issue is. It tries to import `typing_extensions` on Python < 3.8, but typing_extensions isn't installed with the vendored dependencies.

abravalheri:
I think you wrap typing_extensions in a `if TYPE_CHECKING` clause. Typecheckers consider it as part of stdlib and already include it by default. This way you don't have to add an extra dependency...

jaraco:
> I think you wrap typing_extensions in a `if TYPE_CHECKING` clause. Typecheckers consider it as part of stdlib and already include it by default. This way you don't have to add an extra dependency...

I thought about this approach, but unfortunately, `Literal` is expected to be in the namespace. My preference is to honor a project's declared dependencies instead of patching around them, so I'll proceed with including the dependency for now.

PR #3770, opened 2023-01-15
Docs: Link directly to PEPs

hugovk:
## Summary of changes

Use the Sphinx `:pep:` role to link to PEPs in docs:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-pep

This means they'll link directly to their new home at https://peps.python.org/ rather than going via redirect at the old https://www.python.org/dev/peps/

And update the non-docs links to the new home.

### Pull Request Checklist
- [n/a] Changes have tests
- [n/a?] News fragment added in [`changelog.d/`].

[PR docs]:

Is a changelog needed for this?

### DIFF
--- a/docs/artwork.rst
+++ b/docs/artwork.rst
@@ -65,7 +65,7 @@ Inspiration
 This design was inspired by :user:`cajhne`'s `original proposal`_ and the
 ancient symbol of the ouroboros_.
 It features a snake moving in a circular trajectory not only as a reference to
-the Python programming language but also to the `wheel package format`_ as one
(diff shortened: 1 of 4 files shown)


jaraco (changes requested):
This PR changes code outside docs. It should avoid changing anything in distutils (or contribute those changes to /pypa/distutils). The validate project is generated code and should be changed in the upstream project.

hugovk:
Thanks, I've dropped the non-docs commit.

abravalheri (approved):

PR #3775, opened 2023-01-18
Fixed accumulating include dirs after compile

### LINKED ISSUE #3591: CCompiler `self.include_dirs` modified during each `compile()`
cf. 9f9a3e57643cb49796c1b08b5b5afb2826ecd7f6

https://github.com/pypa/setuptools/blob/ba3995e5705a22e13bb5d2231ac22c77e4417747/setuptools/_distutils/ccompiler.py#L390-L398

https://github.com/pypa/setuptools/blob/ba3995e5705a22e13bb5d2231ac22c77e4417747/setuptools/_distutils/_msvccompiler.py#L228-L233

With msvc, the class `include_dirs` is not empty, and will be appended to the instance `include_dirs` for each `compile(..., include_dirs=None)` (which is the default).  With enough calls, an impressively long argument list accumulates, resulting in an error:

```
FileNotFoundError: [WinError 206] The filename or extension is too long
```

I think that one simple fix would be:

```py
 if include_dirs is None: 
     include_dirs = self.include_dirs[:] # make a copy
```

mrbean-bremen:
## Summary of changes
- makes a copy of include dir list to avoid increasing the list in subsequent calls of compile
- fix had been proposed by @mdavidsaver in #3591

It could also make sense to merge this back into upstream `distutils`. The bug itself does not happen there (this was introduced only here in [this commit](https://github.com/pypa/setuptools/commit/9f9a3e57643cb49796c1b08b5b5afb2826ecd7f6)), but it may also be problematic if the internal include dir list is returned in this case (and may be modified unintentionally by the caller). Or maybe not, as `distutils` is deprecated anyway...

Closes #3591

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/changelog.d/3591.change.rst
+++ b/changelog.d/3591.change.rst
@@ -1,0 +1,2 @@
+Fixed gradually increasing length of include dirs after each ``compile`` call, which could lead to compile errors under
+Windows (command line too long).
--- a/setuptools/_distutils/ccompiler.py
+++ b/setuptools/_distutils/ccompiler.py
@@ -388,7 +388,7 @@ class CCompiler:
             raise TypeError("'macros' (if supplied) must be a list of tuples")
 
         if include_dirs is None:
-            include_dirs = self.include_dirs
+            include_dirs = list(self.include_dirs)
         elif isinstance(include_dirs, (list, tuple)):
             include_dirs = list(include_dirs) + (self.include_dirs or [])
         else:
--- a/setuptools/_distutils/tests/test_ccompiler.py
+++ b/setuptools/_distutils/tests/test_ccompiler.py
@@ -53,3 +53,17 @@ def test_set_include_dirs(c_file):
     # do it again, setting include dirs after any initialization
     compiler.set_include_dirs([python])
     compiler.compile(_make_strs([c_file]))
+
+
+def test_include_dirs_after_multiple_compile_calls(c_file):
+    """
+    Calling compile multiple times should not change the include dirs
+    (regression test for #3591).
+    """
+    compiler = ccompiler.new_compiler()
+    python = sysconfig.get_paths()['include']
+    compiler.set_include_dirs([python])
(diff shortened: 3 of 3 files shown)


abravalheri:
Hi @mrbean-bremen thank you very much for the contribution.

Would you like to submit this patch to https://github.com/pypa/distutils? This repository is actively maintained and periodically merged into setuptools. It is important to avoid any direct change into the `setuptools/_distutils` folder because they eventually are overwritten by the periodic merges (or may cause merge conflicts).

I can see that the comment you are mentioned was first introduced there: `https://github.com/pypa/distutils/commit/9f9a3e57643cb49796c1b08b5b5afb2826ecd7f6`.

mrbean-bremen:
Of course - I wasn't aware of this. Thanks!

mrbean-bremen:
Shall I mention the issue in setuptools in this case? Also, should I do anything to update the release notes for this? There seem to be no option for this in the `distutils` repo (or I missed it).

abravalheri:
Thank you very much @mrbean-bremen.

> Shall I mention the issue in setuptools in this case? 
Yes, when submitting the PR to distutils please mention the original issue in setuptools.

>  Also, should I do anything to update the release notes for this? There seem to be no option for this in the distutils repo (or I missed it).

I am afraid we don't have a procedure for that in distutils.

I think you can write something in the commit message so if anyone is bisecting or trying to find something via git blame they can track down easily the PRs and issues.

mrbean-bremen:
Thanks, I already did this in [the PR](https://github.com/pypa/distutils/pull/199), please check if this is sufficient.

mrbean-bremen:
I guess that this PR can be closed in this case, correct?

abravalheri:
Thank you very much @mrbean-bremen that should do the tricky :)

Yes, I agree that we can close this PR for now.

PR #3795, opened 2023-01-25
Make __file__ an absolute path in setuptools.build_meta

hauntsaninja:
## Summary of changes

This is a difference between pip's legacy wheel build (or direct invocations of setup.py) and the PEP 517 build.

While setup.py scripts that rely on this are fragile, it was quite painful to debug! Since Python 3.4, `__file__` is usually an absolute path, so this change might result in fewer surprises.

### DIFF
--- a/changelog.d/3795.change.rst
+++ b/changelog.d/3795.change.rst
@@ -1,0 +1,2 @@
+Ensured that ``__file__`` is an absolute path when executing ``setup.py`` as
+part of ``setuptools.build_meta``.
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -326,7 +326,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
     def run_setup(self, setup_script='setup.py'):
         # Note that we can reuse our build directory between calls
         # Correctness comes first, then optimization later
-        __file__ = setup_script
+        __file__ = os.path.abspath(setup_script)
         __name__ = '__main__'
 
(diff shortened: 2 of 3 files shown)


abravalheri:
Hi @hauntsaninja, thank you very much for working on this.

Would it be possible to have a separated test, instead of adding to the parameterised ones?
The reason why I am asking it is because the backend tests are very onerous and ideally we want to minimise the amount of time they take. The parameterisation works as a cartesian product making the time to run the tests increase sharply...

hauntsaninja:
Thanks for the quick review; updated

abravalheri (approved):

abravalheri:
Thank you very much @hauntsaninja, I think this looks good to me.
If I am wrong there is no way `os.path.abspath` can break right? (i.e. it is a logic operation and will work even if the file does not exist, or am I missing something here?)

hauntsaninja:
Yeah, abspath works fine if the file doesn't exist

hauntsaninja:
Hmm, it calls os.getcwd, which can fail if the cwd has been deleted. But I think that's still okay? Or at least, I see plenty of abspath and os.getcwd in the setuptools source

abravalheri:
That should be fine because CWD should contain the package code.
If I recall correctly, in PEP 517 the frontend should execute the backend entry-point with CWD set to the project root...

pradyunsg (approved):

abravalheri:
If the other maintainers have no reservations, I will include this change in the next release.

PR #3805, opened 2023-01-31
Cache supported tags for wheels.

### LINKED ISSUE #3804: [FR] Cache supported tags in Wheel.is_compatible

Calling `is_compatible` on an instance of the `Wheel` class takes mere milliseconds. But when calling this on thousands of wheels, this quickly adds up.

I think this method is called when setuptools or pip reads an index page, for example https://pypi.org/simple/setuptools/, and for each link checks if the wheel is compatible with the current interpreter and platform. (Not completely sure if this is how it works.)

My own use case is with Buildout. If this downloads a distribution, it saves it in a directory. Buildout uses this directory as an extra find-link. So the next time you call buildout, these distributions are available. This can help for the case where you have no internet, or someone has removed a distribution from PyPI (which happens a lot less these days, I am glad to say.) With thousands of wheels in there, and Buildout/setuptools calling `is_compatible` on each wheel, this takes too much time.

I created an [issue in Buildout](https://github.com/buildout/buildout/issues/626) to track this, so some more details are there. There it seems it is worse with the combination of the very latest setuptools (67.0.0) and pip (23.0.0), and extra worse on Python 3.8 compared to 3.11. But this is a bit unclear.

This is fixable in the `_is_compatible` method in [`setuptools/wheel.py`](https://github.com/pypa/setuptools/blob/v67.0.0/setuptools/wheel.py#L85-L89) by calculating the supported tags once, outside of the class.
When I checked on my system, this gives a set of 1700 supported tags. With 1000 wheels, we would calculate 1.7 million tags. A tad much. ;-)

The assumption is that calling `sys_tags` from the vendored packaging returns the same result each time.

I am preparing a PR.

### Alternative Solutions

I suppose Buildout could add another to its existing [patches](https://github.com/buildout/buildout/blob/master/src/zc/buildout/patches.py), which already includes `setuptools.package_index.PackageIndex` which is involved here.
But I think pip would benefit from a faster method as well.

### Additional context

Here is a test file to get some timings on your own system. Prerequisite: a directory with some wheel, the more the better.

```
from setuptools.wheel import Wheel
from time import time

import os

DIR = "/Users/maurits/cached-downloads/dist"
print(f"Looking for compatible wheels in {DIR}...")
wheels = 0
compatible = 0
start = time()
for filename in os.listdir(DIR):
    if not filename.endswith(".whl"):
        continue
    wheel = Wheel(os.path.join(DIR, filename))
    wheels += 1
    if wheel.is_compatible():
        compatible += 1
stop = time()
print(f"""
Processed {wheels} wheels.
There were {compatible} compatible wheels.
Time taken: {stop - start} seconds.
""")
```

Save this as `test.py`.
With a clone of the pip repo and using the main branch, this is the result:

```
$ .tox/python/bin/python test.py 
Looking for compatible wheels in /Users/maurits/cached-downloads/dist...

Processed 2284 wheels.
There were 1776 compatible wheels.
Time taken: 7.127894639968872 seconds.
```

With my branch I get the same numbers, except the time:

```
Time taken: 0.04627823829650879 seconds.
```

That is about 150 times faster.

### Code of Conduct

mauritsvanrees:
## Summary of changes

In `wheel.py:Wheel.is_compatible` we called `sys_tags` each time to get the list of supported tags. Now we call this once, and cache the result in a simple variable. This drastically speeds up checking the compatibility of hundreds of wheels.

Closes #3804

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -609,20 +609,13 @@ def test_wheel_no_dist_dir():
 
 def test_wheel_is_compatible(monkeypatch):
     def sys_tags():
-        for t in parse_tag('cp36-cp36m-manylinux1_x86_64'):
-            yield t
-    monkeypatch.setattr('setuptools.wheel.sys_tags', sys_tags)
-    # Clear the supported tags cache, otherwise the sys_tags monkeypatch
-    # has no effect.
-    setuptools.wheel._supported_tags.cache_clear()
-    try:
-        assert Wheel(
-            'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl'
-        ).is_compatible()
-    finally:
-        # Clear the cache again, otherwise the sys_tags monkeypatch
-        # is still in effect for the rest of the tests.
-        setuptools.wheel._supported_tags.cache_clear()
+        return {
+            (t.interpreter, t.abi, t.platform)
+            for t in parse_tag('cp36-cp36m-manylinux1_x86_64')
+        }
+    monkeypatch.setattr('setuptools.wheel._get_supported_tags', sys_tags)
+    assert Wheel(
+        'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
 
 
 def test_wheel_mode():
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -612,6 +612,7 @@ def test_wheel_is_compatible(monkeypatch):
         for t in parse_tag('cp36-cp36m-manylinux1_x86_64'):
             yield t
     monkeypatch.setattr('setuptools.wheel.sys_tags', sys_tags)
+    monkeypatch.setattr('setuptools.wheel._supported_tags', None)
     assert Wheel(
         'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
 
--- a/setuptools/tests/test_wheel.py
+++ b/setuptools/tests/test_wheel.py
@@ -612,9 +612,17 @@ def test_wheel_is_compatible(monkeypatch):
         for t in parse_tag('cp36-cp36m-manylinux1_x86_64'):
             yield t
     monkeypatch.setattr('setuptools.wheel.sys_tags', sys_tags)
-    monkeypatch.setattr('setuptools.wheel._supported_tags', None)
-    assert Wheel(
-        'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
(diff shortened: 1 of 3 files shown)


abravalheri:
Hi @mauritsvanrees, thank you for the contribution and sorry for delay in reviewing...

I noticed that you used a global var for the implementation and some patching/mocking/stubbing capabilities for the tests.

What do you think about `functools.lru_cache(maxsize=None)`? (I think `functools.cache` would be better, but it is not available before 3.9 - and by looking at the [implementation](https://github.com/python/cpython/blob/3.11/Lib/functools.py#L651-L653) they are equivalent).

It seems also more elegant for testing because you can call `clear_cache` and there seems to be some evidence that it is [faster](https://discuss.python.org/t/reduce-the-overhead-of-functools-lru-cache-for-functions-with-no-parameters/3956/7) because it is [written in C](https://discuss.python.org/t/reduce-the-overhead-of-functools-lru-cache-for-functions-with-no-parameters/3956/13).


---
Small note:

I don't know if `setuptools/wheel.py` is going to be maintained for long after we completely remove `install/easy_install`, but I understand that this is going to be a different discussion, which should involve the Buildout team that heavily uses it.

Since I could not find any references in the docs about it being deprecated, my assumption is that it is OK to add non-disruptive improvements.

mauritsvanrees:
Hi @abravalheri. Thanks for the review and sorry for the delay in acting on it.

> What do you think about `functools.lru_cache(maxsize=None)`?

Good point.  I should probably use that function more often.
I have added a commit to do this.  (And rebased on main and force pushed.)

This change has no effect on the timing of my small test script in issue #3804: the significant performance increase remains.

setuptools/tests/test_wheel.py
```diff
@@ -612,8 +612,17 @@ def sys_tags():
         for t in parse_tag('cp36-cp36m-manylinux1_x86_64'):
             yield t
     monkeypatch.setattr('setuptools.wheel.sys_tags', sys_tags)
-    assert Wheel(
-        'onnxruntime-0.1.2-cp36-cp36m-manylinux1_x86_64.whl').is_compatible()
+    # Clear the supported tags cache, otherwise the sys_tags monkeypatch
+    # has no effect.
+    setuptools.wheel._supported_tags.cache_clear()
```
abravalheri, line 617: The errors in the CI are probably related to the fact that the function is `_get_supported_tags` instead of `_supported_tags`.

mauritsvanrees: Strange, I saw the tests passing locally with my code.  Well, there were a few failures, but they were there on the main branch as well.  Maybe I missed something in the big test output.
Your changes are good.  Thanks!

mauritsvanrees:
Thanks for your support @abravalheri!

PR #3852, opened 2023-03-09
Removing error prone undefined late binding for Darwin builds

rickmark:
This flag causes the linker to succeed and insert dlsym for undefined values which will cause numerious run time issues that are compile time issues with ext modules

## Summary of changes

Do not request darwin (macOS) linker to ignore undefined symbols

Closes 

### Pull Request Checklist
- [NA] Changes have tests - Existing tests pass
- [NA] News fragment added in [`changelog.d/`].

[PR docs]:


rickmark:
Additionally I think setuptools may have to be modernized for `clang` on Apple platforms, including using `-bundle` instead of dynamiclib.

The additional linker stripping statement might also only be for when compiling as debug

PR #3859, opened 2023-03-15
docs: Add a note about historical `wheel` use in `requires`

mgorny:
## Summary of changes

Add a note explaining that the `wheel` requirement that used to be historically used in documentation was incorrect.  This also explains why it can frequently be found in existing projects, and what problems it causes.  I find myself repeatedly explaining it, so I think having a single official source would be better.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/docs/userguide/quickstart.rst
+++ b/docs/userguide/quickstart.rst
@@ -63,7 +63,7 @@ library will be used to actually do the packaging.
    not recommended. The backend automatically adds ``wheel`` dependency
    when it is required, and listing it explicitly causes it to be
    unnecessarily required for source distribution builds.
-   You should only include ``wheel`` in `requires` if you need to explicitly
+   You should only include ``wheel`` in ``requires`` if you need to explicitly
    access it during build time (e.g. if your project needs a ``setup.py``
    script that imports ``wheel``).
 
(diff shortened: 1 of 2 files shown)


docs/userguide/quickstart.rst
```diff
@@ -56,6 +56,14 @@ containing a ``build-system`` section similar to the example below:
 This section declares what are your build system dependencies, and which
 library will be used to actually do the packaging.
 
+.. note::
+
+   Historically this documentation has wrongly listed ``wheel``
+   in the ``requires`` list, and many projects still do that. This is
+   incorrect. The backend automatically adds ``wheel`` dependency
+   when it is required, and listing it explicitly causes it to be
+   unnecessarily required for source distribution builds.
```
abravalheri, line 65:
```suggestion
   Historically this documentation has unnecessarily listed ``wheel``
   in the ``requires`` list, and many projects still do that. This is
   not recommended. The backend automatically adds ``wheel`` dependency
   when it is required, and listing it explicitly causes it to be
   unnecessarily required for source distribution builds.
   You should only explicitly list `wheel` if you are importing it
   in your `setup.py`.
```

mgorny: Thanks. I've merged your change but fixed formatting and changed "access to it" to "access it" that I believe is more grammatically correct.

abravalheri:

Issue #1386, opened 2018-06-05
Remove dependency on wheel


agronholm:
Currently the `setuptools.command.dist_info` command relies on the `bdist_wheel` code to create a `.dist-info` directory. But since wheel has no public API, this is dangerous as that code could go away in the next release. The `dist_info` command should be natively implemented in setuptools instead.

jaraco:
This effort is going to be a lot of work. The egg_info command has a lot of cruft and deprecated functionality... and the contract between egg_info and dist_info is largely implicit.

On the other hand, it looks like the main thing that dist_info needs is the metadata, which is written in the same format as with egg_info, so probably the distutils write_pkg_info can be leveraged directly. We'll take it bit-by-bit.

agronholm:
I'm starting my summer vacation today. Hopefully I can contribute to this effort.

mattip:
As per pypa/wheel#262, maybe an easier path would be to move `bdist_wheel` here and use an as-yet-not-created official API from wheel.

agronholm:
That would be ideal.

merwok:
That sounds better than calling distutils’ write_pkg_info`, as that does not handle the most recent metadata versions with email-style separation of headers and long description.

abravalheri:
pypa/packaging#383 and pypa/packaging#498 might be relevant here.

jaraco:
Steps to integrate bdist_wheel:

- copy implementation of wheel/bdist_wheel to setuptools/commands/bdist_wheel
- define `distutils.command` for bdist_wheel in setuptools
- vendor `wheel` as `setuptools/_vendor/wheel*`
- update build backend to no longer require wheel
- prevent `distutils.command` from `wheel` package from being loaded (in https://github.com/pypa/setuptools/blob/720862807dea012f3a0e7061880691025f736f11/setuptools/dist.py#L713 and maybe others); remove after a long term

adang1345:
Question: If I'm looking to fix a bug in the `bdist_wheel` command, where should the fix go from now on? Should it go to `setuptools/command/bdist_wheel.py`, `setuptools/_vendor/wheel/bdist_wheel.py`, the https://github.com/pypa/wheel project, or multiple of these?

agronholm:
I think the first step is to report the bug. If the bug relates specifically to `bdist_wheel` and not the other vendored wheel code, then likely it should only be fixed in setuptools.

PR #3904, opened 2023-04-24
Improve core metadata compliance for `PKG-INFO`

abravalheri:
## Summary of changes

- Add `Requires-Dist` to `PKG-INFO` files and make them compatible with `.dist-info/METADATA`
- This PR extracts `PKG-INFO` logic from `setuptools.dist` to `setuptools._core_metadata`.
- Improve 'atomicity' when writing `.egg-info/PKG-INFO`.
  - It seems that there is a race condition between `importlib.metadata` reading `PKG-INFO` and `setuptools.build_meta` writing to `PKG-INFO`, which causes `importlib.metadata` to emit a deprecation warnings (which will be converted in the future into an error).
  - I have observed in the past errors in the CI behaving in a flaky way, and I think those were associated to this race condition.
 
[^1]: (used by pytest and flake8 plugins) reading 

In the future the `_core_metadata` module can be used directly to generated `.dist-info/METADATA`.

---

This is part of a series of PRs:

- #3903
- #3904
- #3905
- #3906
- #3907
- #3908

The motivation for this series of PRs is the following:

- Logic for generating `.egg-info` and `.dist-info` directories is intertwined and implicit
  (See #1386).
- Setuptools uses `pypa/wheel` API which is not stable yet and is very likely to change in the future.
- `pypa/wheel` maintainers previously described that the long term vision is to transfer `bdist_wheel`
  directly to `setuptools` (See [pypa/wheel#262](https://github.com/pypa/wheel/issues/262#issuecomment-640471412), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1306186724), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1328129928)).

---

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


abravalheri:
After some investigation and debugging it seems that the `PKG-INFO` file starts with the correct content, but ends up empty when the tests start running.

There might be a test that is not isolated modifying the `PKG-INFO` file directly...

This seems to be a flaky behaviour, and I am having trouble to replicate it locally, but I can see it in can some tests in the CI.

abravalheri:
I am using this PR to debug a series of flaky errors in the CI, so some intermediate commits might have debug code. I plan to remove the debug code once I reach a stable solution.

abravalheri:
I noticed that when I improve the isolation of `egg_info`[^1] used in `setuptools.build_meta:get_requires*`[^2], setuptools starts to get trouble to bootstrap.
It may seem that it is relying on the fact that `setuptools.build_meta:get_requires*` will generate a `setuptools.egg-info/PKG-INFO` file in the project hook that will be picked up by `importlib.metadata` in the subsequent hook calls.

[^1]: by setting `egg_base` to a temporary directory
[^2]: The isolation is necessary to avoid race conditions caused by `importlib.metadata` reading stuff from the `.egg-info` folder in the project root, while the `.egg-info` folder is being written. This situation seems to occur frequently because we use `pytest-xdist`, `pytest-checkdocs`, `pytest-perf` and other plugins that consume project metadata using `build.util.project_wheel_metadata`. I have been observing flaky tests because of this behaviour in the past months.

abravalheri:
It seems that since 2023-04-27 `virtualenv` will not include either setuptools or wheel when creating a new virtual environment for Python 3.12+: https://virtualenv.pypa.io/en/latest/changelog.html.

Some tests in the setuptools test suite assume `wheel` is installed. We need to fix that.

abravalheri:
If we try to further enhance `egg_info` isolation to avoid race conditions in the tests, and set up `egg_base` when packaging `setuptools` itself, the setuptools `sdist` will not contain the `setuptools.egg-info` folder... This means that it will not have `entry_points.txt`.

However some attributes in `setup.cfg` and `setup.py` are implemented via the `distutils.setup_keyworks` entry-point. Which means that they will not be available for being used and setting these attributes will cause errors.

abravalheri:
It seems that an easier approach is not try to change where `egg_info` writes files for now (since this task is very difficult). Instead I tried to make `write_pkg_info` a more atomic operation.

abravalheri:
@jaraco, would you be OK if I move this PR (#3904) and #3903 ahead?[^1]

The main idea of the 2 PR's is to improve `PKG-INFO` (right now `install_requires` and `extras_require` are mainly analysed by setuptools from a `*.egg-info/requires.txt` perspective, these 2 PRs shift the perspective to concern more with `PKG-INFO`).

[^1]: I know that I presented them as part of a bigger series, but for now I just would like merging these 2 PRs. The rest still depends on discussions with `pypa/wheel` and `pypa/packaging`.

jaraco:
No problem. Please proceed. I trust your instincts.

PR #3907, opened 2023-04-24
Avoid `pypa/wheel` API in `editable_wheel`

abravalheri:
## Summary of changes

- Uses `WheelBuilder` and `_core_metadata` introduced in previous PRs instead of `pypa/wheel`.

(The `pypa/wheel` APIs for building wheel files are likely to change in the future)

---

This is part of a series of PRs:

- #3903
- #3904
- #3905
- #3906
- #3907
- #3908

The motivation for this series of PRs is the following:

- Logic for generating `.egg-info` and `.dist-info` directories is intertwined and implicit
  (See #1386).
- Setuptools uses `pypa/wheel` API which is not stable yet and is very likely to change in the future.
- `pypa/wheel` maintainers previously described that the long term vision is to transfer `bdist_wheel`
  directly to `setuptools` (See [pypa/wheel#262](https://github.com/pypa/wheel/issues/262#issuecomment-640471412), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1306186724), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1328129928)).

---

### Pull Request Checklist
- [ ] Changes have tests
- [ ] News fragment added in [`changelog.d/`].
  _(See [documentation][PR docs] for details)_

[`changelog.d/`]: https://github.com/pypa/setuptools/tree/master/changelog.d
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #3908, opened 2023-04-24
Adopt `bdist_wheel` from `pypa/wheel`

abravalheri:
## Summary of changes

`pypa/wheel` maintainers previously described that the long term vision is to transfer `bdist_wheel` directly to `setuptools`.
This PR attempts to do that based on the previous changes (that untangled the way Setuptools produced core-metadata and dist-info from `pypa/wheel`).
It also relies on the `WheelBuilder` as introduced in a previous PR (the `pypa/wheel` APIs for building wheel files are likely to change in the future).

---

This is part of a series of PRs:

- #3903
- #3904
- #3905
- #3906
- #3907
- #3908

The motivation for this series of PRs is the following:

- Logic for generating `.egg-info` and `.dist-info` directories is intertwined and implicit
  (See #1386).
- Setuptools uses `pypa/wheel` API which is not stable yet and is very likely to change in the future.
- `pypa/wheel` maintainers previously described that the long term vision is to transfer `bdist_wheel`
  directly to `setuptools` (See [pypa/wheel#262](https://github.com/pypa/wheel/issues/262#issuecomment-640471412), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1306186724), [pypa/wheel#472](https://github.com/pypa/wheel/pull/472#issuecomment-1328129928)).

---

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/bootstrap.egg-info/entry_points.txt
+++ b/bootstrap.egg-info/entry_points.txt
@@ -3,6 +3,7 @@ egg_info = setuptools.command.egg_info:egg_info
 build_py = setuptools.command.build_py:build_py
 sdist = setuptools.command.sdist:sdist
 editable_wheel = setuptools.command.editable_wheel:editable_wheel
+bdist_wheel = setuptools.command.bdist_wheel:bdist_wheel
 
 [distutils.setup_keywords]
 include_package_data = setuptools.dist:assert_bool
--- a/setup.cfg
+++ b/setup.cfg
@@ -123,6 +123,7 @@ distutils.commands =
 	build_clib = setuptools.command.build_clib:build_clib
 	build_ext = setuptools.command.build_ext:build_ext
 	build_py = setuptools.command.build_py:build_py
+	bdist_wheel = setuptools.command.bdist_wheel:bdist_wheel
 	develop = setuptools.command.develop:develop
 	dist_info = setuptools.command.dist_info:dist_info
 	easy_install = setuptools.command.easy_install:easy_install
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -352,7 +352,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
             )
 
     def get_requires_for_build_wheel(self, config_settings=None):
-        return self._get_build_requires(config_settings, requirements=['wheel'])
+        return self._get_build_requires(config_settings, requirements=[])
 
     def get_requires_for_build_sdist(self, config_settings=None):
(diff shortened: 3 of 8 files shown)


agronholm:
The `macosx_libfile` module is used to extract architectural information from macos dynamic libraries in order to determine the appropriate tags for binary wheels. It could conceivably be needed outside of setuptools, if other packaging back-ends start using the new public API.

abravalheri:
Thank you very much for the clarification @agronholm.

We might do the transition in 2 steps then... In the first step we bundle `macosx_libfile` inside setuptools. Then in a second step, we might use it directly from `wheel`  when it is exposed via a public API. Does that sound good?

agronholm:
Yeah, you'd have to vendor all the necessary modules from `wheel` in the first step. I think we'd need to coordinate this so as to minimize any duplication between setuptools and wheel. What's the intended time table for this?

abravalheri:
I don't have a timeline for this yet. It is a complex change, so I tried to split it up in many small steps (e.g., fixing PKG-INFO, removing uses of private `pypa/wheel` APIs, etc ...). I am hopping that the other maintainers/members of the community will have sometime to have a look on the proposed PRs and give some feedback at some point.

Maybe on the second semester when I have a bit more time, I might start pushing a little more.
What I had in mind is:

1. Collect feedback about the proposed changes and implement all related modifications.
2. Publish a beta release and ask for beta testers on the Packaging topic in the Python discourse.
3. Implement eventual fixes
4. Create the final release (major bump version).

abravalheri:
The main idea was to get `setuptools` out of the way for `pypa/wheel` to advance the works on the public API without risking breaking existing packages. Then later come back and check for opportunities to use this public API and simplify setuptools code base.

agronholm:
I have been wondering if it makes any sense for all the PyPA stuff to be split between so many different projects. Maybe the wheel related code should be moved to `packaging` instead?

agronholm:
Yeah, I think it would be nice if the new public api for `wheel` would be implemented in `packaging` instead, and `wheel` itself was reduced to nothing more than a command line tool. `bdist_wheel` needs to be migrated to setuptools in any event. The only reason the code still lives in the `wheel` package is of course historical.

I think I'll open an issue with `packaging` to discuss this. Then there's a question of whether `auditwheel` and `wheel` should be merged together, but I think that's a separate issue.

agronholm:
https://github.com/pypa/packaging/issues/697

PR #3917, opened 2023-05-03
tests: Add extraction_filter for tarfile

s-t-e-v-e-n-k:
Python 3.12, and earlier via security backports now issue an DeprecationWarning when calling tarfile.extractall without an extraction filter set. Since the only place we've called extractall is literally right after we've created the archive, use a fully trusted filter. This can be replaced with a filter argument to extractall in future.


abravalheri:
@Mergifyio rebase

Issue #3921, opened 2023-05-11
test_cygwin job failing after adding ruff


jaraco:
In d2ec0473f8d4c25cc6f696e70ba110e1061e4dfe, I replaced pytest-flake8 with pytest-ruff (jaraco/skeleton#79). Unfortunately, adding ruff implies adding a dependency on Rust :( and the test_cygwin job is failing in CI because it doesn't have rust.

jaraco:
In 51408ac9f34f8ca97c33f7f234160c502acb11c4, I disabled the cygwin job, but then the builds started [failing on Python 3.12](https://github.com/jaraco/skeleton/actions/runs/4942599441) where [they didn't previously](https://github.com/jaraco/skeleton/actions/runs/4663048824). I don't understand why. Does the codebase no longer work with Python 3.12 in non-main branches?

jaraco:
I se I must have confused myself. I didn't mean to be looking at the pipelines on skeleton but on Setuptools. Here I'm mainly concerned about making sure the Setuptools pipelines are passing.

abravalheri:
### Regarding the errors in `setuptools.__version__`

(Ref: https://github.com/pypa/setuptools/actions/runs/4943028741/jobs/8837121715)

It is been a while that we are experiencing flaky tests erroring in things related to `importlib.metadata`.
I think they first started appearing with the change in [importlib-metadata v5.2.0](https://github.com/python/importlib_metadata/pull/417/files) (vendored in [setuptools v67.4.0](https://github.com/pypa/setuptools/blob/v67.4.0/setuptools/_vendor/importlib_metadata/_adapters.py#L54-L62)).

After some investigation, I suspect that the problem is the following:

1. The setuptools test suite runs with `pytest-xdist` (for the sake of speed)
2. During the tests `setuptools.build_meta` is called by different tests to build setuptools itself.
3. `setuptools.build_meta` calls `egg_info` a few times (for each one of the PEP 517 hooks, including `get_requires*`).
  - *a)* I suspect the `.egg-info/PKG-INFO` file is rewritten everytime `egg_info` is called
  - *b)* The `egg_info.egg_base` parameter is set to the project root (therefore there is no *isolation* for `egg_info`, if multiple builds happen concurrently, all of them will try to write the same files at the same time). This problem is related to https://github.com/pypa/setuptools/issues/3119.
[...]

PR #3923, opened 2023-05-11
Disable cygwin for now

jaraco:
Workaround for #3921.

CC @lazka : Would you be willing to look into what it will take to make a rust compiler available in the test_cygwin?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,7 +53,8 @@ testing =
 		# workaround for jaraco/skeleton#22
 		python_implementation != "PyPy"
 	pytest-enabler >= 1.3
-	pytest-ruff
+	# workaround for pypa/setuptools#3921
+	pytest-ruff; sys_platform != "cygwin"
 
 	# local
 	flake8-2020


abravalheri:
@jaraco, how about we use something like the following in `setup.cfg`:

```ini
[options.extras_require]
testing =
   # ...
   pytest-ruff; sys_platform != "cygwin"
```

(i.e. instead of disabling `cygwin` we rely on `pytest-enabler` not finding it installed and skipping the pytest plugin).

Would that work as a workaround?

jaraco:
Good idea. It's still not a precise match, because in theory it should work, but definitely a superior approach.

lazka:
(Rust doesn't support cygwin, so that's unlikely to work)

DWesl:
> what it will take to make a rust compiler available in the test_cygwin?

Add Cygwin as a target to the rust codebase, create a cross-compiler on an existing target, then use that to create a native compiler
https://github.com/rust-lang/rust/issues/79854#issuecomment-1412078987

Alternately, try to get mrustc working on Cygwin
https://github.com/thepowersgang/mrustc

PR #3932, opened 2023-05-23
Fix dependency_management.rst typo

cnpryer:


docs/userguide/dependency_management.rst
```diff
@@ -173,7 +173,7 @@ The environmental markers that may be used for testing platform types are
 detailed in :pep:`508`.
 
 .. seealso::
-   If  environment markers are not enough an specific use case,
+   If  environment markers are not enough of a specific use case,
```
cnpryer, line 176: Maybe something like this was intended?
```suggestion
   If  environment markers aren't sufficient,
```

abravalheri: Thank you very much Chris.
I think the original phrase was missing a `for`: `... for a specific use case`.

I understand that the second suggestion is more concise and clean.

However, I do like the emphasis that the wrapper solution should only be used for a small amount of specific use cases... What do you think?

cnpryer: Got it. What about "Alternatively, [the backend wrapper] can be used for specific use cases where markers aren't sufficient"

cnpryer: Done! 66d7d85cf3cbe58aa2c2d56bce07585cf93def37

abravalheri:
Thank you very much!

PR #3960, opened 2023-06-21
Align towncrier with defaults

jaraco:
As jaraco/skeleton is adopting towncrier, that upstream project is attempting to honor the default settings wherever possible.

These changes attempt to align with those defaults and re-use the functionality added upstream for managing towncrier.

## Summary of changes

- Move towncrier config to its own file.
- Use the default 'newsfragments' for news fragments.
- Use jaraco.develop.towncrier to check changes and determine the release kind.
- Revert to towncrier default types.
- Normalized headings in changelog.


jaraco:
Tests are failing on cygwin because the new jaraco.develop dependency has some logic for uploading secrets to Github, which requires PyNaCl, which is built on CFFI, which cygwin doesn't support. We'll have to find a way to exclude that dependency.

jaraco:
I'm going to push forward with merging this to unblock refreshing the skeleton. I welcome feedback, but for the most part, I expect this change to have minimal impact with one exception - existing PR changelogs will need to be updated:

- `change` -> `feature`
- `breaking` -> `removal`
- `misc` -> `bugfix` (maybe)
- The location may need to be updated (`newsfragments/`).

PR #3965, opened 2023-06-27
Bypass encoding warnings in execnet

jaraco:
Originally, I just pushed this commit to main before I realized it causes failures on Python 3.8 when something attempts to construct a requirement, but creates the wrong syntax:

```
  File "/home/runner/work/setuptools/setuptools/.tox/py/lib/python3.8/site-packages/packaging/requirements.py", line 37, in __init__
    raise InvalidRequirement(str(e)) from e
packaging.requirements.InvalidRequirement: Expected end or semicolon (after URL and whitespace)
    execnet@ git+https://github.com/jaraco/execnet@bugfix/195-encodingwarning; extra == "testing"
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

The error message is misleading. The problem is that [PEP 508](https://peps.python.org/pep-0508/) specifies `url_req` as

```
url_req       = name wsp* extras? wsp* urlspec wsp+ quoted_marker?
```

Where the urlspec needs to be separated from the quoted marker (the semicolon) with whitespace.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,8 @@ testing =
 	pytest-perf
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
+	# workaround for pytest-dev/execnet#195
+	execnet @ git+https://github.com/jaraco/execnet@bugfix/195-encodingwarning
 
 testing-integration =
 	pytest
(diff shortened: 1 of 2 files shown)


jaraco:
I can see that importlib_metadata fixed this issue in 4.8.3, so there must be some other code that's generating the invalid URL.

jaraco:
The problem [exists in build](https://github.com/pypa/build/blob/cd06da25481b9a610f846fa60cb67b5a5fa9a051/src/build/__init__.py#L150-L153) when it relies on the old, unfixed importlib.metadata found in Python 3.8 (fix was added to CPython 3.9 in https://github.com/python/cpython/issues/90263).

jaraco:
Although tests are passing, I don't believe PyPI would accept this package. Probably the overrides need to be in tox instead.

PR #3975, opened 2023-07-07
Handle `setup.py` possibly using `sys.exit`

### LINKED ISSUE #3973: [BUG] sys.exit(0) in setup.py breaks PEP 517 hooks API
### setuptools version

68.0.0

### Python version

3.11

### OS

Linux

I believe this will affect any OS and any Python version.

### Description

If a setup.py file calls `sys.exit(0)`, building it through the PEP 517 hooks interface will fail. `sys.exit()` raises a `SystemExit` exception, which propagates up out of the hook function. Normally, unhandled exceptions indicate that something went wrong, and that's OK - the frontend sees that the backend exited with an error, and displays its stdout/stderr for debugging. But in this case the exception is used even though nothing is wrong.

[PEP 517](https://peps.python.org/pep-0517/) says:

> If a hook raises an exception, or causes the process to terminate, then this indicates an error.

We could interpret that 100% literally: an exception was raised, therefore the backend has failed, but that's clearly not the intention - `sys.exit(0)` indicates that execution of `setup.py` completed successfully. **Edit:** it's like `return` for scripts. :slightly_smiling_face: 

This issue has been bounced around several packaging related projects already, most recently https://github.com/pypa/pyproject-hooks/issues/173 .

### Expected behavior

I think setuptools hooks implementation should catch and ignore `SystemExit` with an `err.code` attribute of 0 or None (the non-error cases), on the assumption that this means running `setup.py` finished successfully.

I had a go at doing this in `pyproject-hooks`, the machinery that calls the hooks, but the hooks have to return something, so catching the error outside setuptools isn't enough - we don't know what it would have returned.

### How to Reproduce

Save the following code as `setup.py` in a folder by itself and then run `python -m build path/to/folder/`:

```python
from setuptools import setup
import sys

setup(
    name="pep517-test-setup-py-support",
    version="1.0",
)
sys.exit(0)
```

### Output

```console
$ python -m build tests/samples/setup-py/
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (setuptools, wheel)
* Getting build dependencies for sdist...
running egg_info
writing pep517_test_setup_py_support.egg-info/PKG-INFO
writing dependency_links to pep517_test_setup_py_support.egg-info/dependency_links.txt
writing top-level names to pep517_test_setup_py_support.egg-info/top_level.txt
reading manifest file 'pep517_test_setup_py_support.egg-info/SOURCES.txt'
writing manifest file 'pep517_test_setup_py_support.egg-info/SOURCES.txt'

  File "/home/takluyver/.local/lib/python3.11/site-packages/pyproject_hooks/_impl.py", line 19, in read_json
    with open(path, encoding='utf-8') as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpe0ghw0d1/output.json'

ERROR Backend operation failed: FileNotFoundError(2, 'No such file or directory')
```

arcivanov:
## Summary of changes

Since `setup.py` is an arbitrary Python script legacy scripts my be complex and may use `sys.exit`.
To adhere to PEP 517 handle `SystemExit` exception by ignoring it with exit code `0` and converting it to `CalledProcessError(exit_code, 'setup.py')` in case of non-zero exit code.

fixes #3973

### Pull Request Checklist
- [X] Changes have tests

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -337,7 +337,15 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
         with _open_setup_script(__file__) as f:
             code = f.read().replace(r'\r\n', r'\n')
 
-        exec(code, locals())
+        try:
+            exec(code, locals())
+        except SystemExit as e:
+            if e.code:
+                # We pretend under PEP 517 that the implementation
+                # called a subprocess, which failed.
+                from subprocess import CalledProcessError
+                raise CalledProcessError(e.code, "setup.py")
+            # We ignore exit code indicating success
 
     def get_requires_for_build_wheel(self, config_settings=None):
         return self._get_build_requires(config_settings, requirements=['wheel'])
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -341,11 +341,15 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
             exec(code, locals())
         except SystemExit as e:
             if e.code:
-                # We pretend under PEP 517 that the implementation
-                # called a subprocess, which failed.
-                from subprocess import CalledProcessError
-                raise CalledProcessError(e.code, "setup.py")
+                raise
             # We ignore exit code indicating success
+            SetuptoolsDeprecationWarning.emit(
+                "Running `setup.py` directly as CLI tool is deprecated.",
+                "Please avoid using `sys.exit(0)` or similar statements "
+                "that don't fit in the paradigm of a configuration file.",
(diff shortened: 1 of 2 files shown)


abravalheri:
Hi @arcivanov, to avoid back and forth in the review I added a commit implementing the outcome of the issue discussion and I tried to reduce the number of tests being executed (`test_build_meta` has a few cartesian products).

I hope that is OK with you (otherwise, please feel free to revert the commits and do something different).

arcivanov:
Thank you!

PR #3976, opened 2023-07-07
Remove execnet workaround

takluyver:
The fix was just included in a release, and the branch this was installing from has been deleted, so it was causing CI failures.


takluyver:
This gets most of the CI passing again.

The remaining failure is on Cygwin, failing to find a Rust compiler while building maturin. It looks like this was triggered by the release of pydantic 2.0, which now has a core written in Rust. :scream: I'm going to leave unpicking that knot to someone else, sorry.

abravalheri:
Thank you very much @takluyver.

PR #3979, opened 2023-07-09
Fix all mypy issues

Avasam:
## Summary of changes

Progress towards #2345 
I'm not familiar with tox or this project's CI setup, could I get a hand to ensure that mypy errors are blocking on the CI now that they're all fixed with base configuration?

Changes:
- Fixed all mypy issues with base configurations
- Updated `mypy.ini` to fit the current state of type-checking in this project:
  - Added comments to explain configuration decisions
  - Explicitly target oldest supported Python version
  - Excluded vendored modules, the build folder, dynamic "extern" modules and a file that was causing "duplicate module name" issues with mypy

`mypy . --python-version=3.8|3.11 --platform=linux|win32` results:
image

A handful of separate PRs have been created and split from this one following review discussion:
- #4096
- #4097
- #4099 or #4227
- #4144
- #4231
- #4232
- https://github.com/jaraco/skeleton/issues/96
- https://github.com/python/typeshed/pull/11310
- https://github.com/python/typeshed/pull/11451
- https://github.com/python/typeshed/pull/11454
- https://github.com/pytest-dev/pytest/pull/11565
- https://github.com/pytest-dev/pytest/pull/11600

### Pull Request Checklist
- [ ] Changes have tests (see my comment/question about enforcing this in the CI)
- [x] News fragment added in [`newsfragments/`]. (should not affect the end user yet)

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -43,11 +44,7 @@
 import importlib
 from pkgutil import get_importer
 
-try:
-    import _imp
-except ImportError:
-    # Python 3.2 compatibility
-    import imp as _imp
+import _imp
```
Avasam, line 47: I'm only assuming by the comment. since the minimal supported python version is now 3.8, this should be fine right?

pkg_resources/extern/__init__.py
```diff
@@ -78,3 +79,10 @@ def install(self):
     'more_itertools',
 )
 VendorImporter(__name__, names).install()
+
+if TYPE_CHECKING:
+    from . import packaging as packaging
```
Avasam, line 84: I only type-stubs the re-imports for `packaging` because that's the only one that was needed for internal use.

tox.ini
```diff
@@ -5,7 +5,7 @@ toxworkdir={env:TOX_WORK_DIR:.tox}
 deps =
 	# Ideally all the dependencies should be set as "extras"
 	# workaround for pytest-dev/execnet#195
-	execnet @ git+https://github.com/jaraco/execnet@bugfix/195-encodingwarning
```
Avasam, line 8: I tried to run `tox -- --mypy` locally and it was failing because this branch no longer exists. That change was included in `2.0.0` according to the changelog.

jaraco:
(closed accidentally)

setup.py
```diff
@@ -88,5 +88,6 @@ def _restore_install_lib(self):
 
 if __name__ == '__main__':
     # allow setup.py to run from another directory
-    here and os.chdir(here)
+    if here:
+        os.chdir(here)
```
jaraco, line 92: Again, let's avoid changing the semantics when adding type checking, unless there's a legitimate error, and even then, better to correct it in a separate contribution.

setuptools/command/build_ext.py
```diff
@@ -3,30 +3,33 @@
...
-    # Additionally, assert that the compiler module will load
-    # also. Ref #1229.
-    __import__('Cython.Compiler.Main')
-except ImportError:
-    _build_ext = _du_build_ext
+        # Additionally, assert that the compiler module will load
+        # also. Ref #1229.
+        __import__('Cython.Compiler.Main')
+    except ImportError:
+        from distutils.command.build_ext import build_ext as _build_ext
```
jaraco, line 28: I'm not a fan of the extra nesting. Usually, I see special blocks for `if TYPE_CHECKING`, but usually other imports are unaffected. Is it possible to leave this block unconditional and top-level? If not, I'm going to be reluctant to accept the entanglement this approach requires.

setuptools/msvc.py
```diff
@@ -21,9 +21,10 @@
 import itertools
 import subprocess
 import distutils.errors
+import sys
 from setuptools.extern.more_itertools import unique_everseen
 
-if platform.system() == 'Windows':
+if sys.platform == "win32":
```
jaraco, line 27: Please don't change the semantics.  If mypy can't handle `platform.system()` then disable the checks.

setuptools/sandbox.py
```diff
@@ -263,6 +264,52 @@ def run_setup(setup_script, args):
...
 
+def _mk_dual_path_wrapper(name):
+    original = getattr(_os, name)
+
+    def wrap(self: AbstractSandbox, src, dst, *args, **kw):
+        if self._active:
+            src, dst = self._remap_pair(name, src, dst, *args, **kw)
+        return original(src, dst, *args, **kw)
+
+    return wrap
```
jaraco, line 275: I'm unsure if this refactoring is safe or why these functions were embedded. Probably it's fine to move them out, but let's do that as a separate change, either prior to adding mypy support or after by disabling the checks.

setuptools/tests/integration/test_pip_install_sdist.py
```diff
@@ -27,7 +28,12 @@
 
 pytestmark = pytest.mark.integration
 
-(LATEST,) = Enum("v", "LATEST")
+
+class v(Enum):
+    LATEST = 1
+
+
+(LATEST,) = v
```
jaraco, line 36: I'm not well-versed enough in Enums to understand what's going on here, but I'm starting not to like the smell. This new syntax adds a new name to the namespace and 5 lines to the code (3 of which are blank). Avoid changing the semantics and let's just disable mypy where it's too picky about the types.

setuptools/tests/test_bdist_egg.py
```diff
@@ -47,7 +47,7 @@ def test_bdist_egg(self, setup_context, user_override):
         assert re.match(r'foo-0.0.0-py[23].\d+.egg$', content)
 
     @pytest.mark.xfail(
-        os.environ.get('PYTHONDONTWRITEBYTECODE'),
+        os.environ.get('PYTHONDONTWRITEBYTECODE', False),
```
jaraco, line 50: I avoid adding literals where they're not needed. Let's not change the semantics. Just disable mypy if it's too picky.

setuptools/extern/__init__.py
```diff
@@ -82,3 +83,15 @@ def install(self):
...
+    import packaging as packaging
+    import packaging.version
+    import ordered_set as ordered_set
+    import more_itertools as more_itertools
+    import importlib_metadata as importlib_metadata
+    import zipp as zipp
+    import importlib_resources as importlib_resources
+    import jaraco as jaraco
+    import typing_extensions as typing_extensions
+    import tomli as tomli
```
jaraco, line 97: Same concerns as with pkg_resources/extern

.gitignore
```diff
@@ -8,6 +8,7 @@ lib
 distribute.egg-info
 setuptools.egg-info
 .coverage
+test_*
```
jaraco, line 11: This change doesn't seem right. I expect `test_*` to be tracked by git. Why the change?

jaraco (changes requested):
Thanks for the contrib! I've got some serious concerns about the checks for vendored modules, and some other suggestions about the approach, but for the most part it looks good.

>     * Use `pyproject.toml` instead of `mypy.ini` to merge tooling configurations (it also makes [exclude](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude) configuration a bit cleaner as it supports arrays in `.toml` format)
>       * If you still prefer having a standalone `mypy.ini`, I'll move it back and adapt the `exclude` config.

I do wish to retain the mypy settings in `mypy.ini` for two reasons:

- The skeleton for this project is maintained [upstream](/jaraco/skeleton), which defines the mypy settings in its own file. Moving the settings to a different file makes it difficult to compare and manage the differences with settings inherited from upstream.
- Piling all project settings in pyproject.toml is a bad idea in general. I've been meaning to write a blog post about it, but the tl;dr is that because there's no ordering or namespacing, the settings lack structure and consistency (compared with `mypy` which always appears after `docs/` and before `newsfragments/` and is clearly distinct from towncrier or tox or pytest settings.

pkg_resources/__init__.py
```diff
@@ -68,8 +65,11 @@
 from os import open as os_open
 from os.path import isdir, split
 
+importlib_machinery: Optional[types.ModuleType]
 try:
-    import importlib.machinery as importlib_machinery
+    import importlib.machinery
+
+    importlib_machinery = importlib.machinery
```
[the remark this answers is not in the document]
Avasam: In that case, I'll go back to a more concise syntax and add a single `type: ignore`.

Linking what I think is the proper mypy issue:
- https://github.com/python/mypy/issues/1393

And other relevant issues:
- https://github.com/python/mypy/issues/13123
- https://github.com/python/mypy/issues/14220
- https://github.com/python/mypy/issues/3500

pkg_resources/extern/packaging/__init__.pyi
```diff
@@ -0,0 +1,14 @@
...
+from . import _parser as _parser
+from . import _structures as _structures
+from . import _tokenizer as _tokenizer
+from . import markers as markers
+from . import metadata as metadata
+from . import requirements as requirements
+from . import specifiers as specifiers
+from . import tags as tags
+from . import utils as utils
+from . import version as version
```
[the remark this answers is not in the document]
Avasam: I'll remove the extra stubs files for this PR:
1. I completely skipped over the module name `extern`, meaning vendored. This PR should ignore issues in vendored modules.
2. If it still causes issues in non-vendored modules, I'll just ignore them for this PR, they can be looked at later.

setuptools/command/build_ext.py
Avasam: Unfortunately not, I asked about a similar situation in pyright's GitHub and received the following answer: https://github.com/microsoft/pyright/issues/6156#issuecomment-1762414590
> This is by design, not a bug.
> 
> The determination about what code to analyze and what code to completely ignore is determined in the "binding" phase of the type checker long before types are evaluated and before the code flow graph is established. As per PEP 484, a type checker must honor a specific set of conditional checks including `sys.platform == <literal>`. You're correctly using this in your code, but you're not including the non-Windows code in an `else` statement. If you want this to work in pyright or other static type checking tools, you'd need to change your code as follows:

setuptools/msvc.py
Avasam: I'll change back and just ignore related platform-specific type errors in this file for now. I also disagree about this non-obtuse platform check form not being checked by type-checkers.
Here's a mypy issue we can push forward: https://github.com/python/mypy/issues/8166
pyright almost certainly wouldn't include it until it's explicitly part of a pep (it wasn't in https://peps.python.org/pep-0484/#version-and-platform-checking) or mypy also supports it.

setuptools/sandbox.py
Avasam: Because they were not marked as static, the first parameter is inferred as the class' type. If you *do* type the first argument as something else w/o marking as static, you get `Self argument missing for a non-static method (or an invalid type for self)`.
In this case, these are not even used as static methods, so moving them out makes more sense.

Will revert, `type: ignore` and mark as TODO

(different, but similar set of issues in pyright)

setuptools/tests/integration/test_pip_install_sdist.py
Avasam: The exact error is: `"Enum" object is not iterable`

And here's why there has been no effort in supporting this form in mypy: https://github.com/python/mypy/issues/8009#issuecomment-558335186
Might be worth opening an issue for this specific feature anyway

pyright does seem to understand it just right.

setuptools/tests/test_bdist_egg.py
Avasam: I found an open pytest issue for this: https://github.com/pytest-dev/pytest/issues/10094

.gitignore
Avasam: I don't remember, must be a leftover from something.

Avasam:
```suggestion
```

pkg_resources/__init__.py
Avasam: Done in #4096 if you wanna keep it separate

pkg_resources/__init__.py
```diff
@@ -2231,7 +2231,8 @@ def resolve_egg_link(path):
 if hasattr(pkgutil, 'ImpImporter'):
     register_finder(pkgutil.ImpImporter, find_on_path)
 
-register_finder(importlib_machinery.FileFinder, find_on_path)
+if importlib_machinery:
+    register_finder(importlib_machinery.FileFinder, find_on_path)
```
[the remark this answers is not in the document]
Avasam: https://github.com/pypa/setuptools/pull/4097

pkg_resources/__init__.py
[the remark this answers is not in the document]
Avasam: https://github.com/pypa/setuptools/pull/4097

pkg_resources/extern/__init__.py
```diff
@@ -78,3 +79,10 @@ def install(self):
...
+    import platformdirs as platformdirs
+    import jaraco as jaraco
+    import importlib_resources as importlib_resources
+    import more_itertools as more_itertools
```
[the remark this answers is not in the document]
Avasam: https://github.com/pypa/setuptools/pull/4097

setup.py
Avasam: Before I open a PR, this raises the following in pyright and mypy:
- Expression value is unused Pylance([reportUnusedExpression](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUnusedExpression))
- "chdir" does not return a value Mypy([func-returns-value](https://mypy.readthedocs.io/en/latest/_refs.html#code-func-returns-value))

In other words, `os.chdir` has no return value, and that value is being used in a condition.

I understand the fluent/functional type-guard style here. I use it a lot in TypeScript and C#. Not so common in Python in my experience (I guess it's not "pythonic"?).
This is the only place in setuptool's codebase I found so far using it (at least in the code type-checked by this PR, which isn't as that much) But if you wanna allow that style, I can disable the appropriate rules in the type-checkers. Or at least ignore it for this one instance if you absolutely want to keep like as-is.

setup.cfg
```diff
@@ -52,7 +52,7 @@ testing =
 	pytest-cov; \
 		# coverage seems to make PyPy extremely slow
 		python_implementation != "PyPy"
-	pytest-mypy >= 0.9.1; \
+	pytest-mypy >= 0.10.3; \
```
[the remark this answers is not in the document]
Avasam: Idk the impacts of your skeleton project, and this isn't blocking me, so I opened https://github.com/jaraco/skeleton/issues/96

setuptools/tests/test_bdist_egg.py
Avasam: PR opened: https://github.com/pytest-dev/pytest/pull/11565

Avasam: Pytest's api doesn't contractually support None (it works because it's falsy). See: https://github.com/pytest-dev/pytest/pull/11600

Which means this will always be an error (true positive) in static type-checkers (meaning it'll always need a `# type: ignore` comment), or we can be explicit and say that when the environment variable `PYTHONDONTWRITEBYTECODE` doesn't exist, it's equivalent to passing `False` to the condition for setuptools. (which is what that change was doing)

Avasam:
@jaraco I've listed in the PR description, other related PRs that would be beneficial to handle first.

setuptools/tests/test_bdist_egg.py
jaraco: I don't feel strongly about it. Happy with whatever resolution you chose.

jaraco (approved):

jaraco:
Thanks for doing the work here. I've reviewed the dependent changes and merged the ones I think should be merged. Only #4099 I think should be closed without merging. Happy to merge this when you're ready.

setuptools/command/editable_wheel.py
```diff
@@ -600,7 +600,7 @@ def _simple_layout(
     layout = {pkg: find_package_path(pkg, package_dir, project_dir) for pkg in packages}
     if not layout:
         return set(package_dir) in ({}, {""})
-    parent = os.path.commonpath(starmap(_parent_path, layout.items()))
+    parent = os.path.commonpath(starmap(_parent_path, layout.items()))  # type: ignore[call-overload] # FIXME upstream
```
Avasam, line 603: Upstream fix: https://github.com/python/typeshed/pull/11310

Avasam:
@jaraco This PR is back up to date and ready to merge / review additional changes. But I also created https://github.com/jaraco/skeleton/pull/107 which can be done independently to reduce coverage false-positives with static-typing

Avasam:
I'm guessing those sudden test failures are not related to m changes? They also appear in #4151

setuptools/config/_validate_pyproject/__init__.py
```diff
@@ -5,7 +5,7 @@
 from .error_reporting import detailed_errors, ValidationError
 from .extra_validations import EXTRA_VALIDATIONS
 from .fastjsonschema_exceptions import JsonSchemaException, JsonSchemaValueException
-from .fastjsonschema_validations import validate as _validate
+from .fastjsonschema_validations import validate as _validate  # type: ignore[attr-defined] # mypy false-positive. Pyright is fine here
```
abravalheri, line 8: `setuptools/config/_validate_pyproject` should also be ignored because it is automatically generated using an external tool...

Avasam: Thanks for the heads-up, I missed that. Since `attr-defined` is an issue that comes up often for lots of different reasons that are not gonna be an easy fix in setuptools (and should not be a priority focus), I've disabled that error code globally.

setuptools/tests/test_egg_info.py
```diff
@@ -79,7 +79,8 @@ def run():
     @staticmethod
     def _extract_mv_version(pkg_info_lines: List[str]) -> Tuple[int, int]:
         version_str = pkg_info_lines[0].split(' ')[1]
-        return tuple(map(int, version_str.split('.')[:2]))
+        major, minor, *_ = map(int, version_str.split('.'))
```
abravalheri, line 82: Isn't this doing more computing?

(Previously we were restricting the `map` to the first 2 elements, now we are running it over everything)

I don't know, but there might be a chance the rest of the string may raise an exception when converted to `int`, isn't it?

setuptools/extension.py
```diff
@@ -27,7 +27,7 @@ def _have_cython():
 _Extension = get_unpatched(distutils.core.Extension)
 
 
-class Extension(_Extension):
+class Extension(_Extension):  # type: ignore[valid-type, misc]  # https://github.com/python/mypy/issues/14458
```
abravalheri, line 30: Would it improve things if we wrap the `get_unpatched` into the `else` branch of an `if TYPE_CHECKING`?

abravalheri (commented):
Thank you very much @Avasam. Please find some comments in the quick review I did.

setuptools/command/build_ext.py
abravalheri: @Avasam, what if `_build_ext`'s type is defined prior to the `try...except` block? Would that be enough to get rid of the conditional?

```python
_build_ext: Type[_du_build_ext]

try:
    ...
```

setuptools/command/editable_wheel.py
abravalheri: Maybe add the link to the comment?

setuptools/command/easy_install.py
```diff
@@ -2015,7 +2020,13 @@ def is_python_script(script_text, filename):
     from os import chmod as _chmod
 except ImportError:
```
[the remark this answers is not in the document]
Avasam: It's up to the checker's implementation. mypy will use the first definition, pyright will use the second.
Currently, mypy incorrectly thinks that you can pass in kwargs (so at least adding `**kwargs` to the Jython compatibility version would be an improvement)
and pyright would completely loose type safety.
This only affects setuptools internally though, so it's a "type safety risk" (a very minor one) you can take.

Pylance:
image

pyright:
`setuptools\command\easy_install.py:2030:21 - information: Type of "_chmod" is "(*args: object, **kwargs: object) -> None"`

mypy:
`setuptools\command\easy_install.py:2032: note: Revealed type is "def (path: Union[builtins.int, builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]], mode: builtins.int, *, dir_fd: Union[builtins.int, None] =, follow_symlinks: builtins.bool =)"`

setuptools/dist.py
```diff
@@ -205,7 +205,7 @@ def check_packages(dist, attr, value):
 _Distribution = get_unpatched(distutils.core.Distribution)
 
 
-class Distribution(_Distribution):
+class Distribution(_Distribution):  # type: ignore[valid-type, misc]  # https://github.com/python/mypy/issues/14458
```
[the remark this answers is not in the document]
Avasam: This is a mypy bug. And I found a better issue to describe it: https://github.com/python/mypy/issues/10962

`_Distribution` is of type `type[Distribution]`, (so technically a type alias), pyright has no issue with it. The issue it causes with mypy down the line is that some properties are not type-checked.
It's easy to work around with a `TYPE_CHECKING` block though

.gitignore
```diff
@@ -18,5 +18,6 @@ setuptools.egg-info
 .hg*
 .cache
 .idea/
+.vscode/
```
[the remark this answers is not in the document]
Avasam: Shouldn't `.idea` not be in there in that case?

It's also project-specific configs, not user specific. So far for setuptools I only had to configure `"ruff.organizeImports": false` because the rest of my user settings mostly align with the project's standards.

`.gitignore`-ing `.vscode` basically says "you're on your own if you use VSCode, we offer no support in terms of installing the right extensions and configuring them correctly for this project"

Anyway, I could try to do something similar to what you shared, but with local configs instead of something global.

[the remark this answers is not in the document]
Avasam: Even simpler I found `./.git/info/exclude` 😃 
image

setuptools/__init__.py
```diff
@@ -109,7 +109,7 @@ def setup(**attrs):
 _Command = monkey.get_unpatched(distutils.core.Command)
 
 
-class Command(_Command):
+class Command(_Command):  # type: ignore[valid-type, misc]  # https://github.com/python/mypy/issues/14458
```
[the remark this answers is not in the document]
Avasam: Yes (same as above)

mypy.ini
```diff
@@ -1,5 +1,27 @@
 [mypy]
+# CI should test for all versions, local development gets hints for oldest supported
+python_version = 3.8
+strict = False
+warn_unused_ignores = True
+# TODO: Not all dependencies are typed. setuptools itself should be typed too
+# TODO: Test environment is not yet properly configured to install all imported packages
```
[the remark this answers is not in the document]
Avasam: Mypy recently split up their errors into more subcategories and I can now more precisely disable only `import-not-found` vs `import-untyped`. I'll update the config and comments to make that clearer.

There's a couple dynamically created modules for tests, I'll disable those globally with a comment explaining why they make no sense to check for.

The following modules are found, but untyped:
- `distutils._modified` at https://github.com/pypa/setuptools/blob/main/setuptools/command/build_clib.py#L6 . However `_modified` [doesn't exist in stdlib](https://github.com/python/cpython/tree/3.11/Lib/distutils) (it was added in the vendored distutils). So I don't see why `setuptools.modified` or `.._distutils._modified` couldn't always be used directly.
- `wheel.wheelfile`
- `jaraco.develop`

IDEM, but only found in tests:
- `jaraco`
- `jaraco.envs`
- `jaraco.path`

setuptools/tests/_packaging_compat.py
```diff
@@ -1,6 +1,8 @@
+from typing import TYPE_CHECKING
+
 from packaging import __version__ as packaging_version
 
-if tuple(packaging_version.split(".")) >= ("23", "2"):
+if TYPE_CHECKING or tuple(packaging_version.split(".")) >= ("23", "2"):
```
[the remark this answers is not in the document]
Avasam: It would. But I'll try that in a different PR. I believe Jaraco prefers splitting it up that way.

setuptools/tests/test_bdist_egg.py
```diff
@@ -47,7 +47,7 @@ def test_bdist_egg(self, setup_context, user_override):
         assert re.match(r'foo-0.0.0-py[23].\d+.egg$', content)
 
     @pytest.mark.xfail(
-        os.environ.get('PYTHONDONTWRITEBYTECODE'),
+        os.environ.get('PYTHONDONTWRITEBYTECODE'),  # type: ignore[arg-type] # Unintended usage of xfail
```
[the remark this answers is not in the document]
Avasam: Passing `None` works because it's falsy, but not an intended usage (see https://github.com/pytest-dev/pytest/pull/11565#issuecomment-1801655848 and https://github.com/pytest-dev/pytest/pull/11600 )

However, I didn't notice that Jaraco last answered with https://github.com/pypa/setuptools/pull/3979#discussion_r1465272561
> I don't feel strongly about it. Happy with whatever resolution you chose.

So I'll go back to this, which I preferred
```suggestion
        os.environ.get('PYTHONDONTWRITEBYTECODE', False),
```

setuptools/tests/test_egg_info.py
Avasam: Hmm, looking back at it, you're right. I think I can get the best of both worlds to avoid
mypy: `error: Incompatible return value type (got "Tuple[int, ...]", expected "Tuple[int, int]")`
pyright:
```
Expression of type "tuple[int, ...]" cannot be assigned to return type "Tuple[int, int]"
  "tuple[int, ...]" is incompatible with "Tuple[int, int]"
    Tuple size mismatch; expected 2 but received indeterminate
```
by doing
```py
>>> version_str = "1.2.345-beta"
>>> major, minor = map(int, version_str.split('.')[:2])
>>> major, minor
(1, 2)
```

setuptools/extension.py
Avasam: Yes (same as above, that's also all 3 usages of get_unpatched)

setuptools/tests/integration/test_pip_install_sdist.py
```diff
@@ -25,10 +27,10 @@
 
 from .helpers import Archive, run
 
-
 pytestmark = pytest.mark.integration
 
-(LATEST,) = Enum("v", "LATEST")
+
+(LATEST,) = Enum("v", "LATEST")  # type: ignore[misc] # https://github.com/python/mypy/issues/8009#issuecomment-558335186
```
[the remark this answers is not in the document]
Avasam: The error here is `error: "Enum" object is not iterable  [misc]`, because we're immediately unpacking the `Enum`. As well as 9 `error: Cannot determine type of "LATEST"  [has-type]` below. The `Enum` type is essentially made anonymous, and lost.

I've linked the reasoning given for not working on allowing this anyway from the mypy devs, but I agree with you that within the scope of a module (or a class attribute, as someone else commented on that thread) is a different scenario.

Still, between that, and the lack of static typing usefulness from the functional API, I doubt this will ever be supported.

This can be done with a class API, but Jason preferred the 1-liner as mentioned here: https://github.com/pypa/setuptools/pull/3979#discussion_r1367963610

setuptools/command/build_ext.py
Avasam: Reverting this I don't see any added type error. Either this was improved, or it's pyright issues it lead to, in which case that'll show up in #4192 . But for this PR I'll just revert to what it was if the change is unnecessary.

setuptools/tests/_packaging_compat.py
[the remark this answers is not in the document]
Avasam: https://github.com/pypa/setuptools/pull/4231

setuptools/command/easy_install.py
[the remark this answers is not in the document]
abravalheri: I think it is worth doing something like the following, so that we remove the need for also defining `_FileDescriptorOrPath`

```python
try:
    from os import chmod as _chmod
except ImportError:  # type: ignore
    def _chmod(*args):
        pass

File: setuptools/tests/integration/test_pip_install_sdist.py (hunk 28 above)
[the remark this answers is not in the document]
Comment (abravalheri) [reply setuptools/tests/integration/test_pip_install_sdist.py:33]: We probably can use a link for a new issue then: https://github.com/python/mypy/issues/16936.

> Still, between that, and the lack of static typing usefulness from the functional API, I doubt this will ever be supported.

Oh, well, it is very useful... if `mypy` can get the type inference right and does not force you to be explicit when creating constants.

File: setuptools/tests/test_manifest.py [hunk 29]
```diff
@@ -75,7 +76,7 @@ def touch(filename):
 )
 
 
-translate_specs = [
+translate_specs: List[Tuple[str, List[str], List[str]]] = [
```
abravalheri, line 79: Shouldn't the typechecker be able to infer the value of this variable without the need for special annotation?

.gitignore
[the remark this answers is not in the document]
abravalheri: > Shouldn't .idea/ (IntelliJ IDEs) not be in there in that case?

Probably... it slipped under the radar.

setuptools/tests/integration/test_pip_install_sdist.py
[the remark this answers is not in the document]
abravalheri:
```suggestion
(LATEST,) = Enum("v", "LATEST")  # type: ignore[misc] # https://github.com/python/mypy/issues/16936
```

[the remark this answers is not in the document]
Avasam: Personally I'm fine with
```suggestion
v = Enum("v", "LATEST")
LATEST = v.LATEST
```
https://github.com/python/mypy/issues/16936#issuecomment-1959594042
But I'll leave that decision up to you.

setuptools/command/easy_install.py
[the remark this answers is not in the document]
Avasam: A `type: ignore` comment at that position doesn't do anything. But in my last update I did just suppress this type error. Is it satisfactory?

> Then we can also remove the `_FileDescriptorOrPath` definition, right?

Yep.

setuptools/tests/test_manifest.py
Avasam: Without the explicit definition:
mypy:
```
setuptools\tests\test_manifest.py:129: note: Revealed type is "builtins.list[Tuple[builtins.str, builtins.list[builtins.str], builtins.object]]"
setuptools\tests\test_manifest.py:143: error: No overload variant of "zip" matches argument types "repeat[str]", "object"  [call-overload]
setuptools\tests\test_manifest.py:143: note: Possible overload variants:
setuptools\tests\test_manifest.py:143: note:     def [_T_co] zip() -> zip[Any]
setuptools\tests\test_manifest.py:143: note:     def [_T_co, _T1] zip(Iterable[_T1], /) -> zip[Tuple[_T1]]
setuptools\tests\test_manifest.py:143: note:     def [_T_co, _T1, _T2] zip(Iterable[_T1], Iterable[_T2], /) -> zip[Tuple[_T1, _T2]]
setuptools\tests\test_manifest.py:143: note:     def [_T_co, _T1, _T2, _T3] zip(Iterable[_T1], Iterable[_T2], Iterable[_T3], /) -> zip[Tuple[_T1, _T2, _T3]]
setuptools\tests\test_manifest.py:143: note:     def [_T_co, _T1, _T2, _T3, _T4] zip(Iterable[_T1], Iterable[_T2], Iterable[_T3], Iterable[_T4], /) -> zip[Tuple[_T1, _T2, _T3, _T4]]
setuptools\tests\test_manifest.py:143: note:     def [_T_co, _T1, _T2, _T3, _T4, _T5] zip(Iterable[_T1], Iterable[_T2], Iterable[_T3], Iterable[_T4], Iterable[_T5], /) -> zip[Tuple[_T1, _T2, _T3, _T4, _T5]]
setuptools\tests\test_manifest.py:143: note:     def [_T_co] zip(Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], /, *iterables: Iterable[Any]) -> zip[Tuple[Any, ...]]
```

pyright:
```
setuptools\setuptools\tests\test_manifest.py
  setuptools\setuptools\tests\test_manifest.py:129:13 - information: Type of "translate_specs" is "list[Unknown]"
```

setuptools/command/easy_install.py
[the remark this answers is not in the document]
Avasam: I could also do something a bit cheeky: just leave this untyped for this PR. So mypy won't raise an issue (because I have it currently configured to not check untyped functions). Deferring these changes to #4192

setuptools/__init__.py
```diff
@@ -165,8 +169,9 @@ class Command(_Command):
     """
 
     command_consumes_arguments = False
+    distribution: Distribution  # override distutils.dist.Distribution with setuptools.dist.Distribution
 
-    def __init__(self, dist, **kw):
+    def __init__(self, dist: Distribution, **kw):
```
Avasam, line 174: I think one of these two lines is causing the doc workflow to fail, but I don't understand why.

abravalheri: Is this something that we can workaround by adding an exception to the `nitpick` configuration in `docs/conf.py`?

setuptools/command/easy_install.py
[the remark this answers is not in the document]
abravalheri: I think the way you sorted it out should just be fine...
Thank you very much @Avasam.

setuptools/tests/integration/test_pip_install_sdist.py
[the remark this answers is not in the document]
abravalheri: The way it is should be fine. It is a nice syntax, because it is a one-liner after all...

Avasam:
Btw, with over 40 commits and lots of "oops" commits. Is this gonna be squashed merged? Or should I squash my git history?

abravalheri:
> Btw, with over 40 commits and lots of "oops" commits. Is this gonna be squashed merged? Or should I squash my git history?

If you are more comfortable that way, I think you can do is a `git rebase -i` and organise the commit history in a way that makes sense to you. Then just update the PR with a push-force... Otherwise, if you tell me that it should be squashed as a whole, I am also happy to oblige.

FYI, I recently merged a bugfix contributed by a member of that community that was having problems with building. That affects primarily `build_meta`, but I see that the file is not changed in this PR, so hopefully it will not cause too much trouble 🤞.

Avasam:
I'm comfortable and experienced with interactive rebases, but for this PR you can just squash as a whole.

abravalheri:
Thank you very much @Avasam! That was a heroic piece of work.


---

I have a doubt:

How are you running `mypy`? When I try to run `mypy --python-version=3.8 --platform=linux .`, I still get back 3 errors, which I suppose should be ignored as configured in `mypy.ini`:

```php
$ mypy --python-version=3.8 --platform=linux .
setuptools/command/build_clib.py:6: error: Cannot find implementation or library stub for module named "distutils._modified"  [import-not-found]
setuptools/command/build_clib.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
setuptools/config/_validate_pyproject/formats.py:180: error: Cannot find implementation or library stub for module named "trove_classifiers"  [import-not-found]
setuptools/command/build_ext.py:19: error: Cannot find implementation or library stub for module named "Cython.Distutils.build_ext"  [import-not-found]
Found 3 errors in 3 files (checked 146 source files)
```

I can see that at least `setuptools/_distutils` and `setuptools/config/_validate_pyproject` are `excluded` in `mypy.ini`.
(I have installed `mypy==1.8.0`, `mypy-extensions==1.0.0`, `typing_extensions==4.9.0`)


Is this something that is supposed to happen?

Avasam:
@abravalheri Thank you too for your support!

The last two (`trove_classifiers` and `Cython`) are just missing installs, although I think Cython isn't installed in CI. For `distutils._modified` I don't remember why it could differ, maybe because I have an editable install of setuptools 🤷‍♂️

Since I figured out how to enable mypy in CI, I'll have a very small follow-up PR that does just that an appropriately ignores "missing-imports" that happens on CI with a comment.

PR #3995, opened 2023-07-27
Make _EditableFinder case-sensitive on case-insensitive filesystems (for certain editable installations)

### LINKED ISSUE #3994: [BUG] certain editable installs allow case-insensitive imports
### setuptools version

setuptools==68.0.0

### Python version

Python 3.10.12

### OS

macOS

I believe this is only a problem on macOS because of its case-preserving but case-insensitive filesytem. I'm not too familiar with filesystems but it may be a problem on others. See also [PEP 235 - Import on Case-Insensitive Platforms](https://peps.python.org/pep-0235/).

I think this also depends on exactly how it is installed, but I'm not familiar with what determines which editable installation method is used.

### Description

I have a project with the following structure:
```
├── pyproject.toml
├── setup.cfg
└── setuptools_editable_case_insensitive
    ├── __init__.py
    └── module_1.py
```

When the package is installed as editable, I am able to import `module_1` with arbitrary case:
```
>>> from setuptools_editable_case_insensitive import module_1
>>> from setuptools_editable_case_insensitive import MODULE_1
>>> import setuptools_editable_case_insensitive.Module_1
```

### Expected behavior

I expected to get an `ImportError` when importing a module with incorrect case, as is the behavior when installed normally:
```
>>> from setuptools_editable_case_insensitive import Module_1
ImportError: cannot import name 'Module_1' from 'setuptools_editable_case_insensitive' (/Users/aandersoniii/src/setuptools-example/setuptools_editable_case_insensitive/__init__.py)
```

### How to Reproduce

1. Clone this simplified example: `git clone git@github.com:aganders3/setuptools-editable-case-insensitive.git`
2. Create a virtual environment: `cd setuptools-editable-case-insensitive && python3.10 -m venv venv && source venv/bin/activate`
3. Install the project as editable: `pip install -e .`
4. Run the test script: `python test_imports.py`
5. Note `module_1` is successfully imported with arbitrary case

### Output

```console
❯ git clone git@github.com:aganders3/setuptools-editable-case-insensitive.git
Cloning into 'setuptools-editable-case-insensitive'...
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (6/6), done.
remote: Total 7 (delta 0), reused 7 (delta 0), pack-reused 0
Receiving objects: 100% (7/7), done.

❯ cd setuptools-editable-case-insensitive && python3.10 -m venv venv && source venv/bin/activate

❯ pip install -e .
Obtaining file:///Users/aandersoniii/src/setuptools-editable-case-insensitive
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Installing backend dependencies ... done
  Preparing editable metadata (pyproject.toml) ... done
Building wheels for collected packages: setuptools-editable-case-insensitive
  Building editable for setuptools-editable-case-insensitive (pyproject.toml) ... done
  Created wheel for setuptools-editable-case-insensitive: filename=setuptools_editable_case_insensitive-0.0.0-0.editable-py3-none-any.whl size=2787 sha256=580be31a2260f44e9e877f905ac5875526d31d0f46a433eb72b9103b55f7cb31
  Stored in directory: /private/var/folders/dd/w2cggkp91dq_xn53d6fphp8c0000gq/T/pip-ephem-wheel-cache-ik0hfo9g/wheels/a7/9d/6e/26a7dd6200d0678dcfb12a1d9447413c2a95d67d6a00fefc74
Successfully built setuptools-editable-case-insensitive
Installing collected packages: setuptools-editable-case-insensitive
Successfully installed setuptools-editable-case-insensitive-0.0.0

❯ python test_imports.py 
Hello from setuptools_editable_case_insensitive.module_1!
Hooray, I was imported with the correct name!
	__name__='setuptools_editable_case_insensitive.module_1'
	__file__='/Users/aandersoniii/src/setuptools-editable-case-insensitive/setuptools_editable_case_insensitive/module_1.py'

Hello from setuptools_editable_case_insensitive.Module_1!
Oh no, I was imported with the wrong name!
	__name__='setuptools_editable_case_insensitive.Module_1'
	__file__='/Users/aandersoniii/src/setuptools-editable-case-insensitive/setuptools_editable_case_insensitive/Module_1.py'
```

aganders3:
## Summary of changes

This updates the `_EditableFinder` that is added to `sys.meta_path` for certain editable installations to respect filename case on case-insensitive (but case-preserving) filesystems. I believe this behavior better matches normal installations and [PEP 235](https://peps.python.org/pep-0235/).

The way the change works is by checking the path is in the directory listing (which returns proper case) in addition to checking if the file exists (which is generally case-insensitive if the filesystem is). I believe this is pretty much how it is done in the `FileFinder` in the standard library (see [here](https://github.com/python/cpython/blob/983305268e2291b0a7835621b81bf40cba7c27f3/Lib/importlib/_bootstrap_external.py#L1637C28-L1637C28) and [here](https://github.com/python/cpython/blob/983305268e2291b0a7835621b81bf40cba7c27f3/Lib/importlib/_bootstrap_external.py#L1652C17-L1652C17) for example).

Closes #3994

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -772,7 +772,8 @@ class _EditableFinder:  # MetaPathFinder
         init = candidate_path / "__init__.py"
         candidates = (candidate_path.with_suffix(x) for x in module_suffixes())
         for candidate in chain([init], candidates):
-            if candidate.is_file() and _check_case(candidate, len(rest)):
+            last_n = len(rest) + len(candidate.parts) - len(candidate_path.parts)
+            if candidate.is_file() and _check_case(candidate, last_n):
                 return spec_from_file_location(fullname, candidate)
 
 
@@ -801,17 +802,17 @@ class _EditableNamespaceFinder:  # PathEntryFinder
         return None
 
 
-def _check_case(path, n):
+def _check_case(path, last_n):
     '''Verify the last `n` parts of the path have correct case.'''
     return (
         (not sys.flags.ignore_environment and "PYTHONCASEOK" in os.environ)
         or (
             # check the case of the name by listing its parent directory
             path.as_posix() in (p.as_posix() for p in path.parent.iterdir())
-            # check the case of the next n parent directories the same way
+            # check the case of the next n - 1 components the same way
             and all(
                 part.as_posix() in (p.as_posix() for p in part.parent.iterdir())
-                for part in list(path.parents)[:n]
+                for part in list(path.parents)[:last_n - 1]
             )
         )
     )
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -810,8 +810,8 @@ def _check_case(path, n):
             path.as_posix() in (p.as_posix() for p in path.parent.iterdir())
             # check the case of the next n parent directories the same way
             and all(
-                p1.as_posix() in (p.as_posix() for p in p2.iterdir())
-                for p1, p2 in list(zip(path.parents, path.parents[1:]))[:n]
+                part.as_posix() in (p.as_posix() for p in part.parent.iterdir())
+                for part in list(path.parents)[:n]
             )
         )
     )
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -756,22 +756,6 @@ NAMESPACES = {namespaces!r}
 PATH_PLACEHOLDER = {name!r} + ".__path_hook__"
 
 
-def _check_case(path, n):
-    '''Verify the last `n` parts of the path have correct case.'''
-    return (
-        (not sys.flags.ignore_environment and "PYTHONCASEOK" in os.environ)
-        or (
-            # check the case of the name by listing its parent directory
-            path.as_posix() in (p.as_posix() for p in path.parent.iterdir())
-            # check the case of the next n - 1 parent directories the same way
-            and all(
-                p1.as_posix() in (p.as_posix() for p in p2.iterdir())
-                for p1, p2 in list(zip(path.parents, path.parents[1:]))[:n - 1]
-            )
-        )
-    )
-
-
 class _EditableFinder:  # MetaPathFinder
     @classmethod
     def find_spec(cls, fullname, path=None, target=None):
@@ -815,6 +799,22 @@ class _EditableNamespaceFinder:  # PathEntryFinder
     @classmethod
     def find_module(cls, fullname):
         return None
+
+
+def _check_case(path, n):
+    '''Verify the last `n` parts of the path have correct case.'''
+    return (
+        (not sys.flags.ignore_environment and "PYTHONCASEOK" in os.environ)
+        or (
+            # check the case of the name by listing its parent directory
+            path.as_posix() in (p.as_posix() for p in path.parent.iterdir())
+            # check the case of the next n parent directories the same way
+            and all(
+                p1.as_posix() in (p.as_posix() for p in p2.iterdir())
+                for p1, p2 in list(zip(path.parents, path.parents[1:]))[:n]
+            )
+        )
+    )
 
 
 def install():
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -742,7 +742,7 @@ class _NamespaceInstaller(namespaces.Installer):
         return repr(str(self.src_root))
 
 
-_FINDER_TEMPLATE = """\
+_FINDER_TEMPLATE = '''\
 import os
 import sys
 from importlib.machinery import ModuleSpec
@@ -754,6 +754,15 @@ from pathlib import Path
 MAPPING = {mapping!r}
 NAMESPACES = {namespaces!r}
 PATH_PLACEHOLDER = {name!r} + ".__path_hook__"
+
+
+def _relax_case():
+    return not sys.flags.ignore_environment and "PYTHONCASEOK" in os.environ
+
+
+def _in_dir_case_sensitive(path, dir):
+    """Provide a case-sensitive confirmation that `path` exists in `dir`."""
+    return path.as_posix() in [f.as_posix() for f in dir.iterdir()]
 
 
 class _EditableFinder:  # MetaPathFinder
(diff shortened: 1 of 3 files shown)


aganders3:
It's not terribly surprising to see this Windows failure. Unfortunately I don't have very easy access to a Windows system but I'll see what I can do.

aganders3:
I pushed 07d0242 to hopefully address the [lint errors in the docs build](https://github.com/pypa/setuptools/actions/runs/5694679311/job/15439237394) -- I'm not sure why these errors showed up here. Maybe there was a new release of `sphinx-lint` or something. Let me know if you'd rather see that commit moved to a different PR.

setuptools/tests/test_editable_install.py
```diff
@@ -577,6 +577,27 @@ def test_similar_name(self, tmp_path):
             with pytest.raises(ImportError, match="foobar"):
                 import_module("foobar")
 
+    def test_case_sensitivity(self, tmp_path):
+        files = {
+            "foo": {
+                "__init__.py": "",
+                "lowercase.py": "x = 1",
+            },
+        }
```
abravalheri, line 586: What would happen here if I have a file structure like

```python
files = {
    "foo": {
        "__init__.py": "",
        "bar": {
            "__init__.py": "",
            "lowercase.py": "x = 1",
        }
    }
}
```

and I try to import `foo.BAR.lowercase`?

setuptools/command/editable_wheel.py
```diff
@@ -770,7 +780,9 @@ def _find_spec(cls, fullname, candidate_path):
         init = candidate_path / "__init__.py"
         candidates = (candidate_path.with_suffix(x) for x in module_suffixes())
         for candidate in chain([init], candidates):
-            if candidate.exists():
+            if candidate.is_file() and (
+                _relax_case() or _in_dir_case_sensitive(candidate, candidate.parent)
+            ):
```
abravalheri, line 785: I wonder if it would work if we the finder just handle the top level names and rely on the existing machinery for the nested modules/packages... I have the impression that `importlib.machinery.FileFinder` would first try to import the parent package and then use the parent package `__path__` to resolve the nested path...

Not sure though, I have not tested. If that works, we could skip the checks all together and possibly simplify the template, right?

abravalheri (commented):
Hi @aganders3, thank you very much for looking into this (including the changes in the docs).
Please see some comments below.

setuptools/command/editable_wheel.py
```diff
@@ -811,7 +823,7 @@ def install():
         sys.path_hooks.append(_EditableNamespaceFinder._path_hook)
     if PATH_PLACEHOLDER not in sys.path:
         sys.path.append(PATH_PLACEHOLDER)  # Used just to trigger the path hook
-"""
+'''
```
abravalheri, line 826: We probably can skip this change in the quotes right? Does black complain about it?

aganders3: Yes, I think so - this was so the docstring in `_in_dir_case_sensitive` could use triple-double quotes.

setuptools/command/editable_wheel.py
aganders3: I think you're right. I tried this a bit and got stuck with failing namespace packages, but I will try this route again.

setuptools/tests/test_editable_install.py
aganders3: Shoot, you're right this fails. I will look into an implementation with `importlib.machinery` as mentioned below.

setuptools/command/editable_wheel.py
abravalheri: Ah, sorry, I did not see that.

aganders3:
Unfortunately I still couldn't make this work with public classes from `importlib.machinery`. It mostly worked for normal imports, but needed enough special logic for namespace packages that just expanding the previous method ended up being a simpler. I can keep trying along those lines if you'd prefer, but I don't immediately see a clean solution.

I expanded the tests here to cover more of the cases you brought up. Thanks for the review!

emiratesnbd-sa-com (approved):
Emiratesnbd.sa.com

abravalheri:
Hi @aganders3 thank you very much for working on this.
Sorry I haven't had the time to review this PR in detail.
My initial/preliminary thoughts are:

- I plan to split the commits targeting the docs linter (thank you very much for that!) into a separated PR. This way I can merge the docs part and rebase this PR, which will make it a little cleaner and easier to review.
- I am concerned with the increase in complexity for the finder. I am also thinking about the loops to check file/directory existence in the file system, since they might make import statements slow (with only 1 project installed as editable, that should be fine, but maybe if the user installs multiple editable projects in the same virtualenv it might became a problem).
  - I was planning to experiment a bit with the idea I expressed in https://github.com/pypa/setuptools/pull/3995#discussion_r1278285728, and/or delegating calls to `importlib.machinery.FileFinder` or `importlib.machinery.PathFinder` (they already implement the necessary case sensitive checks and caching to not hurt performance, so it would be very nice if we can simply use them)

Once I have a bit more time I can work on this, but please feel free to bet me to that.
Regardless, thank you very much for the contribution (that includes tests ❤️), even if we don't add exactly the code as it is right now, I am sure we are going to reuse most of it in the implementation.

aganders3:
No worries - thanks again for the reviews! I may spend a _little_ more time looking again at using `importlib.machinery` but otherwise will probably leave this alone for the moment. At the very least I am happy to see this issue being addressed and I will walk away with a much deeper understanding of the import system 😅.

abravalheri:
@aganders3, I submitted a few commits trying something out...
If it does not work we can revert the commits.

setuptools/command/editable_wheel.py
abravalheri: So apparently namespace packages behave differently and the import machine does not try to find nested packages based on the `__path__` property of the parent package...

Maybe in the future it, I can try to create a reproducer and bring it up in an issue in the python/cpython repository... Not sure if that is how it is intended to work.

abravalheri:
I might squash this PR when merging because it will make it easier in the future if anyone needs to debug using `git bisect` or look for the history of the implementation using `git blame`. I hope that is OK.

(The reason being that I added a few commits "undoing" the the checks using `os.iterdir`).

abravalheri:
Thank you very much

PR #4017, opened 2023-08-14
Remove double negative in package_discovery.rst

bebound:
## Summary of changes

`find_packages()` only finds the folder which contains `__init__.py`.


bebound:
Oh, you‘re right. I miss the **dis** part. Double negative makes it harder to understand.
I change it to ignore.

abravalheri:
Thank you very much.

PR #4021, opened 2023-08-17
Include type information by default

### LINKED ISSUE #3136: [FR] Include type information by default (`*.pyi`, `py.typed`) - PEP 561

[PEP 561](https://www.python.org/dev/peps/pep-0561/) specifies how type information should be included in a distributions. Namely package authors may choose to add `py.typed` files to fully typed packages and / or include `.pyi` stub files. By default, those files won't be included in the sdist / wheel distribution packages.

To add them, a package author needs to set `include_package_data = True` and include them via the `MANIFEST.in` or with `options.package_data` (eg. with `* = py.typed, *.pyi`). Although there are some great guides out there explaining it, I believe we could make this much simpler.

https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/
https://jugmac00.github.io/blog/bite-my-shiny-type-annotated-library/

Add a new option `include_typing_files` or `include_typing_information` as a shorthand for `options.package_data`-> `* = py.typed, *.pyi`.

**How should it be used?**
Authors wanting to include typing files (`py.typed` / stub files) should set ~both `include_package_data = True` and~ `include_typing_files = True`.

**How would it work with other options to include package data files?**
Conceptually this option would add the `* = py.typed, *.pyi` pattern to `options.package_data`. This also means any file matched by `exclude_package_data` should still be excluded.

### Alternative Solutions

~Don't require `include_package_data = True`. Similar to license files, one could argue if an author sets `include_typing_files = True`, they already expressed what they want to do. No need to require an additional option.~

Include `py.typed` and `*.pyi` files by default.

### Additional context

If others agree that this is something worth adding, I could work on a PR for it.

### Code of Conduct

Danie-1:
## Summary of changes

When getting source files, include py.typed and *.pyi as well as py files.

Closes #3136

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


Danie-1:
I'm not sure where to add tests because the tests that test functionality most similar to this seem to lie in _distutils (test_sdist.py has a test_add_defaults). Should this feature be in distutils instead?

abravalheri:
Thank you very much @Danie-1 for the contribution ❤️ 💚

The errors with the docs are probably caused by the breaking changes in sphinx 7.2. I updated the main branch to temporarily cap at sphinx 7.1.2 while we fix the problems in a separated PR. You can rebase if you want to get rid of the CI errors.

> Include type information (py.typed, *.pyi) by default

I just would like to align expectations in relation to the PR title and changelog entry:

It seems to me that what the PR does is to add `.pyi` and `py.typed` files automatically to the ***`sdist`***, which is a fine contribution on its own.

But these files may not necessarily end-up on the wheel (I think that will depend on other factors like `include_package_data`, if the files are part of a folder, of if they are single file scripts at the root of the project, etc ...). I use the term **may** because that was just a strong suspicion of mine once I sat down to review the PR, but we can only really confirm that via testing and experimental data.

(I was suggesting something very similar in https://github.com/pypa/setuptools/issues/3136#issuecomment-1076305158, however upon further thought, I think both approaches would have the same shortcomings).

If you are OK with this behaviour, I think we can go ahead, but probably we will have to change the changelog entry to make this more explicit, so the other users don't create further expectations.
We also would have to leave the original issue open, since we don't fully solve the problem the user was reporting.

> I'm not sure where to add tests because the tests that test functionality most similar to this seem to lie in _distutils (test_sdist.py has a test_add_defaults).

I think these could be added to `setuptools/tests/test_sdist.py`.

> Should this feature be in distutils instead?

I don't think so. I think that the `pypa/distutils` repo accepts bugfixes, but not new features.

---

Some of the edge cases that I am not sure will not automatically add `.pyi` and `py.typed` to the wheel, even if they are in the sdist, are the following (non exhaustive list):

- What happens if the user has `include_package_data = False`?
- What happens if the user has `py_module = ["myfile"]`, with `myfile.py` and `myfile.pyi` at the root of the project?

setuptools/command/build_py.py
```diff
@@ -302,6 +302,19 @@ def _get_platform_patterns(spec, package, src_dir):
...
         )
 
+    def get_source_files(self):
+        py_files = [module[-1] for module in self.find_all_modules()]
+
+        possible_stub_files = set(os.path.splitext(f)[0] + ".pyi" for f in py_files)
+        stub_files = [f for f in possible_stub_files if os.path.isfile(f)]
+
+        possible_py_typed_files = set(
+            os.path.join(os.path.dirname(f), "py.typed") for f in py_files
```
abravalheri, line 312: I am curious, is there any particular reason why you went with `py_files` => `os.path.dirname`, instead of iterating over `self.packages` and using `self.get_package_dir`?

Can a wheel have a `py.typed` file at the root (without being inside of a package folder)?

setuptools/command/build_py.py
```diff
@@ -302,6 +302,19 @@ def _get_platform_patterns(spec, package, src_dir):
             for pattern in raw_patterns
         )
 
+    def get_source_files(self):
+        py_files = [module[-1] for module in self.find_all_modules()]
+
+        possible_stub_files = set(os.path.splitext(f)[0] + ".pyi" for f in py_files)
+        stub_files = [f for f in possible_stub_files if os.path.isfile(f)]
```
abravalheri, line 309: Probably we can use a lazy generator to collapse the 2 loops into a single one, for example:

```suggestion
        possible_stub_files = (os.path.splitext(f)[0] + ".pyi" for f in py_files)
        stub_files = set(f for f in possible_stub_files if os.path.isfile(f))
```

What do you think? This way you avoid running 2 loops (although in the original approach the second loop is potentially shorter, in the average case I don't think `set` will reduce the amount of iterations considerably).

newsfragments/3136.feat.rst
```diff
@@ -0,0 +1 @@
+Include type information (py.typed, *.pyi) by default (#3136) -- by :user:`Danie-1`
```
abravalheri, line 1: Please see my comment on "aligning expectations" on the changelog entry.

setuptools/command/build_py.py
Danie-1: This seems good. I only put the `set` in order to make the function output have no repeats rather than for performance reasons (I don't know if avoiding repeats in the output matters but it just seemed a bit wrong to include the same py.typed file loads of times in the case where there are lots of python files in the directory).

Danie-1:
Thank you very much for the detailed comments.

> It seems to me that what the PR does is to add .pyi and py.typed files automatically to the sdist, which is a fine contribution on its own.
> 
> But these files may not necessarily end-up on the wheel (I think that will depend on other factors like include_package_data, if the files are part of a folder, of if they are single file scripts at the root of the project, etc ...). I use the term may because that was just a strong suspicion of mine once I sat down to review the PR, but we can only really confirm that via testing and experimental data.

My intention with this was to make the files be included on the wheel by default (but I'll be completely honest and admit I have very little knowledge about python packaging and I don't know what the sdist is so I'll try to look into that). So at least at the moment, I would prefer to try to keep working on this until we can confirm that files are included in the wheel. In my pull request I should have said that this was intended to be work in progress (I have now marked it as draft).

About the experimental testing to confirm the behaviour, do you have any advice about how to go about this (are there any nice sources of small example packages to test with)? The way I tested these changes was to install my modified setuptools code into a venv and run `python setup.py bdist_wheel` and then pip install the created wheel and see how mypy interacted with it, but I wasn't confident at all that this was a sensible way to test it (in particular I would have liked to be able to use `pip` but I didn't know how to make it use my modified setuptools). I'll try to look into writing some unit tests as well.

> I think these could be added to `setuptools/tests/test_sdist.py`.

I'll try to add some tests there. Would `setuptools/tests/test_build_meta.py` also be a good idea? I saw this line while browsing the tests a bit and thought it might be relevant for what you mentioned.

https://github.com/pypa/setuptools/blob/5f14368eb20068e58f6d8524efc0ae70082afa89/setuptools/tests/test_build_meta.py#L420

> - What happens if the user has `include_package_data = False`?

I've tested this and the files were not included, so I will try to address that.

I should have the time to have a go at all these things as I'm a student with a lot of free time, but I will soon be going on a holiday until September, so I won't be working on it during that time but I intend to work some more after that.

setuptools/command/build_py.py
Danie-1: There wasn't any particular reason, I was just trying to get something to work before trying to refine it (I should have said this in the pull request). I'll remember to check back on this later but for now I'll leave it as is in case the code needs to be drastically changed to address the issue of files not being included in the wheel.

I tried reading https://peps.python.org/pep-0561/#packaging-type-information but couldn't tell if it addresses your question about the `py.typed` file. To me, it slightly sounds like it says that the file must be in a package folder but I don't think I fully understand what it says at the moment.

abravalheri: No problems, you can refine that at a later stage, that is completely fine.

I also have the impression that you cannot have `py.typed` outside of a package folder. Upon further read of the link you shared, I think there is evidence of that in the following paragraphs:

> This PEP does not support distributing typing information as part of module-only distributions or single-file modules within namespace packages.
> The single-file module should be refactored into a package and indicate that the package supports typing as described above.

It seems that you cannot have a top-level `.pyi`, it should always be contained inside a package dir.

abravalheri:
Hi @Danie-1, thank you very much for the comments and all the effort.

> My intention with this was to make the files be included on the wheel by default (but I'll be completely honest and admit I have very little knowledge about python packaging and I don't know what the sdist is so I'll try to look into that). So at least at the moment, I would prefer to try to keep working on this until we can confirm that files are included in the wheel. In my pull request I should have said that this was intended to be work in progress (I have now marked it as draft).

Yeah, `setuptools` is a bit complicated because it has to keep backword compatibility with a wide range of packages published in the PyPI for the last years (should I say decade?).

Please feel free to iterate over the draft as much as you want, you can ping me if you need any advice. Once you think it is good to go, I can do a review.

> About the experimental testing to confirm the behaviour, do you have any advice about how to go about this (are there any nice sources of small example packages to test with)?

If you want to make sure that a specific behaviour is observed, I recommend adding a test that captures that behaviour, so you can implement the change and check if the test passes.
There are a few examples on how we have been writing tests that check the contents of both `sdist` and `wheel`. For example https://github.com/pypa/setuptools/blob/v68.1.2/setuptools/tests/test_sdist.py#L812 or https://github.com/pypa/setuptools/blob/v68.1.2/setuptools/tests/test_build_meta.py#L315 [^1].

If you want just to investigate the existing behaviour, you can also create a separated set of scripts. For example, this is an experiment I did 2 years ago: https://github.com/abravalheri/experiment-setuptools-package-data

> (in particular I would have liked to be able to use pip but I didn't know how to make it use my modified setuptools)

This is one way of using your modified version of setuptools with both `pip` and `build` (I am supposing the working copy of setuptools is in the `modified_setuptools` directory and that you have a dummy test project in the `test_project` directory):

```bash
python3.9 -m venv .venv  # creating a virtual environment to work as a sandbox
.venv/bin/python -m pip install -U pip wheel build  # let's update pip, wheel and build
.venv/bin/python -m pip install -e ./modified_setuptools  # install the working copy of setuptools

# If we disable isolation, both `pip`/`build` will use the version of setuptools
# that is currently installed in the virtual environment
.venv/bin/python -m build ./dummy_project --no-isolation
.venv/bin/python -m pip install ./dummy_project --no-build-isolation

# You can use `unzip -Z1` to list the contents of the wheel file
# and `tar tf` to list the contents of the sdist
unzip -Z1 dummy_project/dist/*.whl
tar tf dummy_project/dist/*.tar.gz
```

> I should have the time to have a go at all these things as I'm a student with a lot of free time, but I will soon be going on a holiday until September, so I won't be working on it during that time but I intend to work some more after that.

No problems @Danie-1, it is very kind of you to be working on this and offering your contributions to the project. Please feel free to do it in your time.

[^1]: If you intend to add tests to `test_build_meta.py`, it would be better to create a separated function like https://github.com/pypa/setuptools/blob/v68.1.2/setuptools/tests/test_build_meta.py#L964 instead of directly in the `TestBuildMetaBackend` (this class is a bit complicated and there is an implicit cartesian product between the number of methods and the number of tests that need to run, which make `test_build_meta.py` a very heavy/slow test file).

Danie-1:
I've now added some tests and an implementation that passes the tests. I think it addresses the things that were mentioned previously:

- > What happens if the user has `include_package_data = False`?

  This is now checked in the tests.
- > What happens if the user has `py_module = ["myfile"]`, with `myfile.py` and `myfile.pyi` at the root of the project?
  
  PEP 561 says:
  
  > This PEP does not support distributing typing information as part of module-only distributions or single-file modules within namespace packages.

Thanks for linking your repo explaining how data files get added and showing the `--no-build-isolation` flag; after reading your comments, some PEPs and more of the setuptools source code I have a bit of a better idea of how it all works.

I spent a while considering adding something like this in `build_py.py`:
```py
        if self.packages:
            self.build_packages()
            self.build_package_data()
            self.build_type_information()  # new
```
but I was slightly hesitant because this would mean that type information files would be no longer treated as data files. I wasn't sure if this could break some things, and also I wasn't sure if this would make it harder to exclude type information files (if you really wanted).
However I think the main advantage of doing something like this is that I think it would be easier to add warnings about type information files that don't comply with PEP 561.

So the main thing I'd like to ask @abravalheri is whether you think it's worth doing the method I outlined above, or do you think the current implementation is sufficient?

setuptools/tests/test_build_meta.py
```diff
@@ -981,3 +982,99 @@ def test_system_exit_in_setuppy(monkeypatch, tmp_path):
...
+        else:
+            expected = self.get_type_files(file_spec)
+
+        yield file_spec, expected
+
+    def test_type_information_always_included(
+        self, monkeypatch, tmp_path, file_spec_and_expected
+    ):
+        """Setuptools should include type information in the wheel (py.typed, *.pyi)."""
+        file_spec, expected = file_spec_and_expected
```
abravalheri, line 1064: Thank you very much for working so thoroughly on the tests.
You came up with some very clever fixture definition 😝.

I wonder however if it would be better to have less automation in this test and more explicit expectations, just for the sake of being easier to reason about (for the test reader).

For example, you could define explicitly both the directory structure and the expected type files, something like the following:


```python
EXAMPLES = {
   "single_shallow_package": {
       "directory_structure": {
            "foo": {
                "bar.pyi": "",
                "py.typed": "",
            },
       }
       "type_files": ["foo/bar.pyi", "foo/py.typed"],
   },
   "nested_inside_namespace": {...},
   "namespace_nested_inside_regular": {...},
   ...
}

CONFIG = {
   "default": "...",
   "dont_include_package_data": "...",
   "exclude_package_data": "...",   
}
```

This way you can parametrize on the dict keys and it would be easier to figure out what is being asserted, for example:[^1]

[^1]: I like to use `dict.keys()` in Pytest fixtures because it makes the test report nicer, but please feel free to use whatever approach you prefer.

```python
@pytest.mark.parametrize("config", ["default", "dont_include_package_data"])
@pytest.mark.parametrize("example", EXAMPLES.keys())
def test_type_files_included_by_default(self, example):
    structure = self.EXAMPLE[example]["directory_structure"]
    type_files = self.EXAMPLE[example]["type_files"]
    ...
    assert type_files in wheel_contents

@pytest.mark.parametrize("config", ["exclude_package_data"]))
@pytest.mark.parametrize("example", EXAMPLES.keys())
def test_type_files_can_be_excluded(self, example):
    ...
    assert type_files not in wheel_contents
```

setuptools/tests/test_build_meta.py
```diff
@@ -981,3 +982,99 @@ def test_system_exit_in_setuppy(monkeypatch, tmp_path):
...
+            if isinstance(file_spec[key], str):
+                if self.is_type_information_file(key):
+                    output.add(key)
+            else:
+                output.update(
+                    key + "/" + f for f in self.get_type_files(file_spec[key])
+                )
+        return output
+
+    @pytest.fixture(params=itertools.product(packages_to_test, [True, False]))
```
abravalheri, line 1045: `itertools.product` is an interesting detail, it seems that that we have a large amount of tests being produced...

Using `build_backend.build_wheel(dist_dir)` in the `test_build_meta.py` file is very expensive and increases the total time necessary for running the tests (the faster the test suite runs the better it is for the developer's perspective).

I wonder if it would be better to use less expensive tests in `test_build_py`[^2]? If that is not enough to get confidence, we can keep 1 heavy test in `test_build_meta.py` and adopt light tests for the edge cases in `test_build_py`...

What do you think?

[^2]: Maybe test `find_data_files` directly (that is the "leanest" test I can think of) or alternatively check `build_py.get_outputs()`

abravalheri (commented):
Thank you very much @Danie-1, sorry for the delay in reviewing this.

> So the main thing I'd like to ask @abravalheri is whether you think it's worth doing the method I outlined above, or do you think the current implementation is sufficient?

Probably both proposals have their merit (thank you very much for investigating them). I believe that `build_type_information` would the most "semantically elegant" because it stops considering type information files as "data-files". However, as you mentioned the inclusion of the files would start to be non-optional... Therefore the approach with `data-files` seems to be the one that leaves a escape hatch for users to opt-out (and it is "backwards compatible" with the way users have been approaching the problem historically in setuptools).

We can give it a try to the implementation you are proposing and see how it goes. In the first stage we can make sure the feature is documented as "beta" (both in the docs and in the changelog), after some months of experimentation we can promote out of beta. What do you think?

abravalheri:
Regarding the CI errors, I believe you can simply rebase.

setuptools/tests/test_build_meta.py
Danie-1: I rewrote the tests now into test_build_py. There are 10 tests in total though because of the parametrisation. Should some of these be removed? Also, I wasn't sure how to address the fact that the tests have a lot of duplicated code (and I wasn't sure whether this should be considered an issue). Previously I was using a slightly complicated fixture as you mentioned, and I agree it made it hard to understand the tests. Perhaps putting the tests in a class would help (but this could also make the tests harder to understand).

setuptools/tests/test_build_meta.py
Danie-1: I removed the existing test from test_build_meta and instead simply edited one of the existing tests. Is this ok to do?

Danie-1:
I've pushed a commit that should hopefully fix the tests on windows. I can try it on windows in a few hours just to double check.
I also realised I forgot to respond to your question about documenting the feature as beta. That sounds like a good idea to me. For the docs, I suppose there should be something on the data files page mentioning it. Is that the only page where it would need mentioning?

setuptools/tests/test_build_meta.py
abravalheri: I gave it a try to refactor some of it... not the best job in the universe, but I hope it is enough...

I also removed `build_py.run()` to make them fast (to minimise the impact of having 10 new tests). If I am not wrong, `build_py.get_outputs()` don't need `.run()` only `.finalize_options()`

abravalheri:
Sorry for the delay and thank you very much @Danie-1 

> For the docs, I suppose there should be something on the data files page mentioning it. Is that the only page where it would need mentioning?

I added something quick in 111b05f. Please feel free to improve.

abravalheri:
I think this is good enough for trying out.

@Danie-1 please let me know if you want to do any other changes, otherwise I think we can merge, so it can be included in the next release (not sure if it is going to happen next week, maybe in the beginning of November).

Danie-1:
That sounds like a good plan to me. I don't have any suggestions for changes, other than a typo fix that I've just pushed.

abravalheri:
Thank you very much @Danie-1.

zhegata:
We just ran into the issue that .pyi files are not added to our packages by default. It would be relatively easy to handle on our side, but I'm glad to see this has already been solved out of the box (and looks like right on time for us). Thank you for this. 

Do you have an idea when this will be released?

abravalheri:
Hi @zhegata please have a look on the comment above: https://github.com/pypa/setuptools/pull/4021#issuecomment-1761775192

Issue #4025, opened 2023-08-23
[FR] Hash-pin GitHub workflow Actions


pnacht:
### What's the problem this feature will solve?

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

### Describe the solution you'd like

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes can be automatically updated by dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

I'll send a PR along with this issue to hash-pin the Actions and configure Dependabot to keep them updated.

### Additional context

My name is Pedro, and I work with Google and the [OpenSSF](https://openssf.org/) to improve the security of open-source projects. My colleague Joyce has made some contributions to this project already (see https://github.com/pypa/setuptools/pull/3833 and – indirectly – https://github.com/pypa/setuptools/commit/a7310562fad7a8834c9810c1edd8e00b03e1394b).

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

webknjaz:
@pnacht this repository inherits a git upstream at https://github.com/jaraco/skeleton. Any changes of such nature should go through that repo.

webknjaz:
I'm still not sure if that would work, though as it's an upstream for tens of downstream repos. And I don't know if introducing this intrusiveness wouldn't cause a disproportional amount of maintenance burden.

jaraco:
> Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

In my experience, version pinning adds a great deal of toil to the process without adding a lot of value. This project, and other skeleton-derived projects, attempt to use the "work at head" model, similar to how Google operates. Pinning works against that model and introduces manual steps and changelog noise. I find that I don't have enough time as it is to address all of the issues my projects face, so I've attempted to [reduce the friction to as low as possible](https://blog.jaraco.com/featherweight-maintenance/).

The per-repo, per-action, per-week multipliers here sound like quite a bit of work if we're talking about more than one or two repos.

[...]

PR #4026, opened 2023-08-23
Hash-pin GitHub Actions

pnacht:
## Summary of changes
Closes https://github.com/pypa/setuptools/issues/4025.

This PR hash-pins all GitHub Actions used in workflows to protect the project from supply-chain attacks. It also configures dependabot to keep the Actions up-to-date.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,7 @@ updates:
       interval: "daily"
     allow:
       - dependency-type: "all"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out ${{ env.SPKG }}
-        uses: actions/checkout@v2
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           path: build/pkgs/${{ env.SPKG }}/src
       - name: Install prerequisites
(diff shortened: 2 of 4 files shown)


pnacht:
Hey @abravalheri, thanks for the quick reply!

So, this is just about hardening your CI/CD. Since your workflows already run with minimal permissions (the `permissions.contents: read` block at the top of the workflows), the main risk from a malicious Action would be leaking credentials or poisoned releases.

`main.yml`'s `release` job seems to publish to PyPi. If one of the Actions became malicious, it could poison your environment and cause the wheel/dist to contain malicious behavior not found in the actual source code. Alternatively, they could simply steal your PyPi token and push their own malicious version.

I can't quite understand what `ci-sage.yml` does, but it seems to push a docker image which consumes an artifact created by your workflow. I suspect this is just for testing purposes, but if this docker image is actually sensitive, the Actions could also be used to poison that image.

---

But yes, this does add some overhead. setuptools currently has the following Actions:
- actions/checkout (new release every ~2 months)
- actions/upload-artifact (every ~5 months)
- actions/setup-python (~1 month)
- actions/cache (~1 month)
- codecov/codecov-action ( ~6 months)
- re-actors/alls-green (hasn't been updated since October 2022, but used to be updated every ~2 months... honestly looks abandoned now...)
- cygwin/cygwin-install-action (~6 months)

So this would likely average to something like 1 dependabot PR every 1-2 weeks.

jaraco (commented):
As proposed, this change is untenable, as it happens per-repo and increases the per-repo maintenance burden substantially. As webknjaz has described, this project is maintained alongside the [hundreds of other](https://www.dropbox.com/s/g16c8w9i7lg9dqn/projects.txt?dl=0) projects I maintain, many of which are derived from jaraco/skeleton. Before we embark on recommending the change there, I'd like to take a step back and understand better what's being proposed (why it's beneficial).

PR #4056, opened 2023-09-23
Apply assorted Pyugrade suggestions

DimitriPapadopoulos:
## Summary of changes

Apply some suggestions from [pyupgrade](https://github.com/asottile/pyupgrade).

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


Avasam:
Given that setuptools uses ruff, you can enforce and autofix this with https://docs.astral.sh/ruff/rules/#pyupgrade-up

DimitriPapadopoulos:
I have selected the pyupgrade suggestions that make most sense and are usually accepted in other projects. Would you agree to giving all pyupgrade suggestions a try, including [redundant `open` modes](https://github.com/asottile/pyupgrade#redundant-open-modes)?

Avasam:
> [...]  including [redundant `open` modes](https://github.com/asottile/pyupgrade#redundant-open-modes)?

That one is often opinionated (and one of the best examples making Ruff's configurability better than pyupgrade directly, along with string formatting). So I'd leave that up to maintainers to decide.

> Also, applying ruff-pyupgrade fixes now would be (partially?) redundant with your #4096.

#4096 does not aim at "upgrading" all code, just removing obsolete code paths/aliases.

It's also never redundant to prevent code style/quality regressions imo. (that's what linters are for)

> Perhaps we can apply these two pull requests first, then set up ruff-pyupgrade and discuss any remaining rules to apply.

My guess is that's probably what would be preferred.

DimitriPapadopoulos:
pyupgrade appears to be removing obsolete code paths too, at least partially.

abravalheri:
Not sure why the diff-cover is failing... I thought it would just measure the difference between files. This PR is not making much of a difference regarding that...

abravalheri:
Thank you very much!

I agree that an approach for automatic corrections would be nice, and I also agree that we should incrementally build up to the point we can apply it first.

PR #4060, opened 2023-09-26
GNU: use -Wl,-rpath,<dir> instead of -Wl,-R<dir>

haampie:
Closes #4060

## Summary of changes
Use `-rpath` instead of `-R` when using binutils.

The latter is supported in binutils for backwards compatibility, but in
general `-R<path>` is equivalent to `--just-symbols=<path>` when `path`
is a file; only when it's a directory, it's treated as `-rpath=<path>`.

Better avoid that ambiguity and use `-rpath`.

Also split `-Wl,--enable-new-dtags` and `-Wl,-rpath,...` into two
separate arguments, which is more common, and more likely to be parsed
correctly by compiler wrappers.

This commit does not attempt to add `--enable-new-dtags` to other
linkers than binutils ld/gold that support the flag.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/newsfragments/4060.feature.rst
+++ b/newsfragments/4060.feature.rst
@@ -1,0 +1,1 @@
+Pass more common `-Wl,--enable-new-dtags -Wl,-rpath,<path>` flags instead of `-Wl,--enable-new-dtags,-R<path>` to compilers with GNU linkers -- by :user:`haampie`
--- a/setuptools/_distutils/tests/test_unixccompiler.py
+++ b/setuptools/_distutils/tests/test_unixccompiler.py
@@ -186,7 +186,10 @@ class TestUnixCCompiler(support.TempdirManager):
                 return 'yes'
 
         sysconfig.get_config_var = gcv
-        assert self.cc.rpath_foo() == '-Wl,--enable-new-dtags,-R/foo'
+        assert self.cc.rpath_foo() == [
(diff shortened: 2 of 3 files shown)


abravalheri:
Hi @haampie, thank you very much for the contribution!

`setuptools` currently "bundles" the `setuptools/_distutils` directoty directly from https://github.com/pypa/distutils (only adding a [`MetaPathFinder`](https://docs.python.org/3/reference/import.html#finders-and-loaders) around it). Therefore we cannot accept changes on that directory directly in this repo.

If you want the code in that path to be changes, maybe it is better to discuss directly with the https://github.com/pypa/distutils maintainers and propose a PR there? Once the changes are added, pypa/distutils is pulled in the setuptools repo.

haampie:
Thanks, opened https://github.com/pypa/distutils/pull/214

haampie:
Looks like there is no activity in distutils though.

abravalheri:
We have to wait for the distutils maintainers to have the time to look into it, I don't think there is a way around that...

Avasam:
Looks like this was merged in distutils. Is this PR still relevant?

abravalheri:
Thank you for the reminder @Avasam. If it is merged into distutils, it will be incorporated soon in setuptools.

PR #4069, opened 2023-10-01
Re-export `distutils.dep_util` through `setuptools.modified`

Avasam:
## Summary of changes

Re-export `distutils.dep_util` through `setuptools.modified`

I'm looking into progressively migrating `distutil` usage in pywin32 (https://github.com/mhammond/pywin32/issues/2119), where `distutils.dep_util.newer` and `distutils.dep_util.newer_group` are used.

Migration guides (https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html and https://peps.python.org/pep-0632/#migration-advice) don't mention `distutils.dep_util`.

~~`setuptools.dep_util` already exists. It seems sensible to re-export the rest of `distutils.dep_util` as part of `setuptool`'s API.~~  
Edit: see conversations below, re-exporting thought `setuptools.modified`.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


newsfragments/4069.feature.rst
```diff
@@ -0,0 +1 @@
+Exported `distutils.dep_util` through `setuptools.dep_util` -- by :user:`Avasam`
```
Avasam, line 1:
```suggestion
Exported `distutils.dep_util` through ``setuptools.dep_util`` -- by :user:`Avasam`
```
Attempt at fixing doc test

setuptools/dep_util.py
```diff
@@ -1,4 +1,8 @@
-from distutils.dep_util import newer_group
+from distutils.dep_util import (
+    newer as newer,
+    newer_pairwise as newer_pairwise,
+    newer_group as newer_group,
+)
```
Avasam, line 5: This could also be written as
```suggestion
from distutils.dep_util import *
```

jaraco:
Before we naively expose this behavior, let's ask ourselves what these functions are used for (in pywin32) and determine where is the best place to host them. Setuptools is aiming to be primarily a build backend and not have dependencies on the library (except in setup.py). I see that [pywin32 uses two functions](https://github.com/search?q=repo%3Amhammond%2Fpywin32%20dep_util&type=code), one in setup.py and the other in a test. Moreover, the one use in setup.py has a comment "unclear why we need to check this".

I'd like to first determine what `newer_group` does, if it's needed to build pywin32, and if so, what purpose it serves. I'd still be tempted to provide it outside setuptools, or limit calls that are exposed here.

Avasam:
In pywin32 it's used to check if a file should be rebuilt/re-generated. All 3 instances of `newer` and `newer_group` seem to simply be a build optimisation.
The one `newer` in tests could be moved to test utils.
`newer_group` in `setup.py` I might have to re-implement to avoid depending on `distutils` directly.
Exploratory PR here: https://github.com/mhammond/pywin32/pull/2141

>  [...] determine where is the best place to host them. [...] I'd still be tempted to provide it outside setuptools

I think I get what you're going for here. Either these 4 file-edit-time comparison methods could be moved to their own library, or included as part of an existing file-manipulation one. Where it would belong better than in setuptools. I think it does make sense, I'm not sure where it would belong best though and where to start requesting.

jaraco:
Thanks for the exploration. I do think we want to expose these functions from setuptools.

I did some work in distutils to modernize and clean up the implementations. That should make implementing `newer_pairwise_groups` much easier.

Before we expose these new public names, I do think we should seriously consider re-naming this module. `dep_util` doesn't convey very well the purpose. In fact "util" in any module name is usually superfluous. And "dep" doesn't tell me much. I'm thinking "fresh" or "freshness" or "modified". Pick your favorite and let's expose these functions there (under `setuptools.fresh` (or similar).

I can deal with the migration to the new name.

Avasam:
Thank you for the consideration


> Before we expose these new public names, I do think we should seriously consider re-naming this module. `dep_util` doesn't convey very well the purpose. In fact "util" in any module name is usually superfluous. And "dep" doesn't tell me much.

I completely agree.

> I'm thinking "fresh" or "freshness" or "modified". Pick your favorite and let's expose these functions there (under `setuptools.fresh` (or similar).

I don't have a strong opinion on this.

Some thoughts: Functions in the module currently only compare modification date, and the name "modified" is quite telling of what the module does, but maybe that's to close to implementation?
"fresh/ness" sounds off but maybe that's just me.
Something like "file_compare" may allow more types on comparisons in the future, but maybe that's too vague and not wanted (like creation date, file size).
`setuptools.freshness.newer_group`
`setuptools.modified.newer_group`
`setuptools.file_compare.newer_group`

> I can deal with the migration to the new name.

Thanks! Feel free to close this PR as superseded by the new one.

setuptools/dep_util.py
```diff
@@ -7,7 +7,7 @@ def __getattr__(name):
     if name not in ['newer_pairwise_group']:
```
Avasam, line 7: I'm not certain the point of checking in a list rather than `if name != 'newer_pairwise_group'`. Copied boilerplate from when there's more argument?

jaraco:
Well that's weird. It used to be the case that if I pointed a PR to target a branch and then merged that branch into main, Github would repoint the original PR to main, but in this case, it closed this PR and I'm unable to reopen (the re-open is greyed out).

jaraco:
Aha! By re-pushing the feature/distutils-sync branch, I was able to re-open this PR.

setuptools/dep_util.py
jaraco: It is due to copied boilerplate (see `_distutils.dep_util`). It felt better to use a consistent syntax than to re-write the logic, especially considering this code path is deprecated.

Issue #4090, opened 2023-10-24
[BUG] setuptools.depends:get_module_constant returns None on Python 3.13.0a1


hroncok:
### setuptools version

65.6.3.post20230109

### Python version

3.13.0a1

### OS

Fedora Linux

### Additional environment information

This happens with later versions as well, but I reproduced from the main branch @ 2384d915088b960999ca74fb81ce70bffd17b082

### Description

The `get_module_constant` function from `setuptools.depends` returns `None` instead of the actual constant value on Python 3.13.0a1. This works fine on Python 3.12.0.

### Expected behavior

The function should work the same on Python 3.12 and 3.13.

### How to Reproduce

1. Create a simple Python file with a constant in it.
2. Run `setuptools.depends:get_module_constant` to retrieve it.

### Output

```console
$ cat foobar.py 
__version__ = "1"

$ venv3.12/bin/python -c 'import setuptools; print(setuptools.__version__); from setuptools.depends import get_module_constant; print(get_module_constant("foobar", "__version__"))'
65.6.3.post20230109
1

$ venv3.13/bin/python -c 'import setuptools; print(setuptools.__version__); from setuptools.depends import get_module_constant; print(get_module_constant("foobar", "__version__"))'
65.6.3.post20230109
None
```

hroncok:
By debugging the setuptools code, I figured out the bytecode numbers must have changed:

```
$ venv3.13/bin/python -c 'import setuptools; print(setuptools.__version__); from setuptools.depends import get_module_constant; print(get_module_constant("foobar", "__version__"))'
65.6.3.post20230109
RESUME 149
LOAD_CONST 83
STORE_NAME 114
RETURN_CONST 103
None
```

```
$ venv3.12/bin/python -c 'import setuptools; print(setuptools.__version__); from setuptools.depends import get_module_constant; print(get_module_constant("foobar", "__version__"))'
65.6.3.post20230109
RESUME 151
LOAD_CONST 100
STORE_NAME 90
1
```

That's from https://github.com/pypa/setuptools/blob/2384d915088b960999ca74fb81ce70bffd17b082/setuptools/depends.py#L145-L147

encukou:
This should use [`dis.opmap`](https://docs.python.org/3/library/dis.html#dis.opmap) rather than hardcoding the numeric values.
[...]

PR #4094, opened 2023-10-26
Fix setuptools.depends:get_module_constant() on Python 3.13.0a1

hroncok:
## Summary of changes

Don't hardcode opcode numbers, look them up instead.

Fixes https://github.com/pypa/setuptools/issues/4090

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


abravalheri (approved):
Thank you very much @hroncok. Probably we can add this to the next release of setuptools (which I am planning for some time around the beginning of the month, but may change if Jason has a different idea).

PR #4097, opened 2023-10-27
Remove possibly redundant `importlib_machinery` code

Avasam:
## Summary of changes

Prompted by https://github.com/pypa/setuptools/pull/3979#discussion_r1367959329 and https://github.com/pypa/setuptools/pull/3979#discussion_r1367959803
Removing obsolete code and workarounds benefit type-checkers.

The `importlib_machinery` alias seems redundant. Maybe a leftover from old code support? If the import failed, all other usages of `importlib_machinery` would error out anyway. With a less obvious `importlib_machinery` rather than the actual `ImportError`.

It seems all `importlib.machinery` usages we do here (`SourceFileLoader`, `SourcelessFileLoader` and  `FileFinder`) have been added in Python 3.3

### Pull Request Checklist
- [x] Changes have tests (the existing tests need to pass)

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -68,13 +69,9 @@
...
 
-try:
-    import importlib.machinery as importlib_machinery
 
-    # access attribute to force import under delayed import mechanisms.
-    importlib_machinery.__name__
-except ImportError:
-    importlib_machinery = None
+# access attribute to force import under delayed import mechanisms.
+importlib.machinery.__name__
```
Avasam, line 74: idk how much import order matters for this one, so I left it right there. But I'd prefer placing it after regular top-level imports if possible.

Avasam, line 74: idk how much import order matters for this one, so I left it right there. But I'd prefer placing it after regular top-level imports if possible.

My guess is that was just to trigger the `ImportError`. In which case this line can probably be removed entirely.

```suggestion
```

newsfragments/4097.feature.rst
```diff
@@ -0,0 +1 @@
+Updated `pkg_resources` to use stdlib `importlib.machinery` instead of `importlib_machinery` -- by :user:`Avasam`
```
Avasam, line 1:
```suggestion
Updated `pkg_resources` to use stdlib `importlib.machinery` instead of ``importlib_machinery`` -- by :user:`Avasam`
```

PR #4098, opened 2023-10-29
Fix - normalize cwd path on dist path check (#4088)

### LINKED ISSUE #4088: [BUG] [easy_install] bug on checking the path of the distribution location
### setuptools version

setuptools===68.2.2

### Python version

Python 3.10

### OS

Windows 11

_No response_

### Description

I can see if the package source path is already added to the PYTHONPATH then it's failing add the package source directory to the easy-install.pth when I run the "pip install -e ."
 
easy_install.py
```python
    def add(self, dist):
        """Add `dist` to the distribution map"""
        new_path = dist.location not in self.paths and (
            dist.location not in self.sitedirs
            or
            # account for '.' being in PYTHONPATH
            dist.location == os.getcwd()
        )
        if new_path:
            self.paths.append(dist.location)
            self.dirty = True
        super().add(dist)
```
In def add(self, dist) function in easy_install.py
> dist.location == os.getcwd()

This code will return False, because the dist.location is already normcased during finalize_options in develop.py
> target =  _path.normpath(self.egg_base)

This target is passed as the location in Distribution
>        # Make a distribution for the package's source
>        self.dist = pkg_resources.Distribution(
>            target,
>            pkg_resources.PathMetadata(target, os.path.abspath(ei.egg_info)),
>            project_name=ei.egg_name,
>        )

For example, if the dist.location and os.getcwd() is 
dist.location = 'c:\users\abc\work\helloworld'  
os.getcwd() = 'C:\Users\abc\work\helloworld'

The result is False, even though they are pointing same path.

it  should be updated to
> dist.location == os.path.normcase(os.getcwd()) 

easy_install.py
```python
    def add(self, dist):
        """Add `dist` to the distribution map"""
        new_path = dist.location not in self.paths and (
            dist.location not in self.sitedirs
            or
            # account for '.' being in PYTHONPATH
            # dist.location is _path.nomcased but the os.getcwd() is not
            dist.location == os.path.normcase(os.getcwd())
        )
        if new_path:
            self.paths.append(dist.location)
            self.dirty = True
        super().add(dist)
```

### Expected behavior

The path of source code should be added to the easy-install.pth file when it's installed as editable.

### How to Reproduce

Add "." (current path) to the PYTHONPATH then run the "pip install -e ." 

### Output

If the easy-install.pth is not exist. it will not dump any error.
But if there is easy-install.pth file exist, it will show this erro.

```output
      File "C:\Users\...\lib\site-packages\setuptools\command\easy_install.py", line 1690, in save
        self.dirty |= last_dirty or self.paths != self._init_paths
    TypeError: unsupported operand type(s) for |=: 'list' and 'bool'
```

spxd:
## Summary of changes

Normalize CWD path for the comparing of distribute location.

Fixes #4088 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -1256,7 +1256,7 @@ class easy_install(Command):
 
             log.info("Removing %s from easy-install.pth file", d)
             self.pth_file.remove(d)
-            if d_location in self.shadow_path:
+            if d_location in map(normalize_path, self.shadow_path):
                 self.shadow_path.remove(d_location)
 
         if not self.multi_version:
@@ -1268,7 +1268,7 @@ class easy_install(Command):
             else:
                 log.info("Adding %s to easy-install.pth file", dist)
                 self.pth_file.add(dist)  # add new entry
-                if dist_location not in self.shadow_path:
+                if dist_location not in map(normalize_path, self.shadow_path):
                     self.shadow_path.append(dist_location)
 
         if self.dry_run:
@@ -1717,8 +1717,8 @@ class PthDistributions(Environment):
     def add(self, dist):
         """Add `dist` to the distribution map"""
         dist_location = normalize_path(dist.location)
-        new_path = dist_location not in self.paths and (
-            dist_location not in self.sitedirs
+        new_path = dist_location not in map(normalize_path, self.paths) and (
+            dist_location not in map(normalize_path, self.sitedirs)
             or
             # account for '.' being in PYTHONPATH
             dist_location == normalize_path(os.getcwd())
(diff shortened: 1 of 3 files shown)


setuptools/command/easy_install.py
```diff
@@ -1717,7 +1717,7 @@ def add(self, dist):
             dist.location not in self.sitedirs
             or
             # account for '.' being in PYTHONPATH
-            dist.location == os.getcwd()
+            dist.location == normalize_path(os.getcwd())
```
abravalheri, line 1720: Will `dist.location` be always normalised? Or is it the case we should introduce an extra `or`?

spxd: Seems to not work in linux. let me check it again...

abravalheri:
@spxd thank you very much for the contribution.

The CI is currently complaining that the lines `setuptools/command/easy_install.py ` 1253-1254,1259-1260,1288 were changed but that the changes are not covered in tests (I understand that they probably were not covered before, but since this PR modifies them it would be important to add tests against regression).

Would it be possible to add a test for the `update_pth` function as well to cover those lines?

Issue #186, opened 2014-04-11
recent windows tests fail (winbot)


ghost:
Originally reported by: **Anonymous**

---

see

http://winbot.zope.org/builders/setuptools_master%20py_265_win32/builds/271/steps/test/logs/stdio

---
- Bitbucket: https://bitbucket.org/pypa/setuptools/issue/186

PR #4103, opened 2023-11-08
Sync latest distutils

jaraco:
- Rely on pytest as found in pytest-dev/pytest#11155. Fixes pypa/distutils#186.
- Clean up docstrings and remove crufty comments. Replace integer literals with booleans.
- "Refactor to newer_group to utilize higher level constructs ("any"), re-use _newer logic, and avoid complexity in branching."
- Prefer os.path.getmtime
- Inline check for target presence.
- Add test for newer_pairwise, bringing coverage in dep_util to 100%.
- Replace for/append loop with a filter function (newer_pair).
- Replace explicit list check with zip(strict=True). Allows inputs to be iterables.
- Extract a 'starfilter', similar to itertools.starmap, to generalize the concept of filtering results over a sequence of tuples.
- Replace '_starfilter' with 'jaraco.functools.splat'.
- Move dep_util to _modified and mark dep_util as deprecated.
- Extend tests for newer_pairwise and fix failed expectation when no files are newer.
- Add newer_pairwise_group (inspired by setuptools.dep_util).
- Added tests for newer_pairwise_group().
- Modernize test_newer_pairwise_group by using tmp_path and a SimpleNamespace.
- Remove latent references in docs.
- Limit sphinxlint jobs to 1. Workaround for sphinx-contrib/sphinx-lint#83.
- Replace git version with released version. Ref #186.
- Disable cygwin tests for now. Ref pypa/setuptools#3921
- Disable integration test due to known breakage from deprecation warnings.
- Allow diffcov to fail also, as it requires the tests to pass on the latest Python to succeed.
- Remove newsfragment
- Copy concurrency setting from setuptools
- Rely on distutils._modified and deprecated setuptools.dep_util.
- Add news fragment.

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
I don't understand why the pull request tests are failing but the push tests are fine... because they are both based on main so the diff between pull request and push should be zero.

abravalheri:
> I don't understand why the pull request tests are failing but the push tests are fine... because they are both based on main so the diff between pull request and push should be zero.

There are some differences in `setuptools/command/build_clib.py`, `setuptools/dep_util.py` and `setuptools/tests/test_dep_util.py`, so probably even if the same code paths are executed, a simple thing like changing the total number of lines in a file might mess with the coverage percentiles.

setuptools/dep_util.py
```diff
@@ -1,24 +1,14 @@
...
-    # build a pair of lists (sources_groups, targets) where source is newer
-    n_sources = []
-    n_targets = []
-    for i in range(len(sources_groups)):
-        if newer_group(sources_groups[i], targets[i]):
-            n_sources.append(sources_groups[i])
-            n_targets.append(targets[i])
-
-    return n_sources, n_targets
+def __getattr__(name):
```
abravalheri, line 6:
```suggestion
def __getattr__(name):  # pragma: no cover
```

setuptools/command/build_clib.py
```diff
@@ -1,7 +1,12 @@
 import distutils.command.build_clib as orig
 from distutils.errors import DistutilsSetupError
 from distutils import log
-from setuptools.dep_util import newer_pairwise_group
+
+try:
+    from distutils._modified import newer_pairwise_group
+except ImportError:
```
abravalheri, line 7:
```suggestion
except ImportError:  # pragma: no cover
```

abravalheri (commented):
Maybe the following suggestions would help with the diffcov errors?

jaraco:
Oh! Good news. The tests that were failing before stopped failing after I closed and re-opened the PR. Now all that's failing is the diffcov tests. I'm happy to ignore those failures.

Thanks for the extra set of eyes!

PR #4108, opened 2023-11-11
Import PUG guide to MANIFEST.in

jeanas:
## Summary of changes

The Packaging User Guide has a [page](https://packaging.python.org/en/latest/guides/using-manifest-in/) on using `MANIFEST.in` to control the data files in a distribution. This is very setuptools-specific since other build backends do not use `MANIFEST.in` files. Thus, I think it is better for this information to live in the setuptools documentation.

After this is merged, I will open a PR on the PUG to make https://packaging.python.org/en/latest/guides/using-manifest-in/ a redirect.

### DIFF
--- a/docs/userguide/miscellaneous.rst
+++ b/docs/userguide/miscellaneous.rst
@@ -43,8 +43,6 @@ to specify any files that the default file location algorithm doesn't catch.
 
 This file contains instructions that tell ``setuptools`` which files exactly
 should be part of the ``sdist`` (or not).
-A comprehensive guide to ``MANIFEST.in`` syntax is available at the
-:doc:`PyPA's Packaging User Guide <PyPUG:guides/using-manifest-in>`.
 
 .. attention::
    Please note that ``setuptools`` supports the ``MANIFEST.in``,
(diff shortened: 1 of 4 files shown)


abravalheri (approved):

abravalheri:
Thank you very much for the contribution @jeanas.


> I didn't keep it because it looks redundant with the start of `docs/guides/miscellaneous.rst`:

Maybe we should add the list because it feels more complete? (some minor update might be needed from https://github.com/pypa/distutils/blob/main/distutils/command/sdist.py#L233-L246).

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
+files are necessary for distributing the package. More precisely, the following
+files are included in a source distribution by default:
+
+- All Python source files implied by the ``py_modules`` and ``packages``
+  ``setup()`` arguments
+- All C source files mentioned in the ``ext_modules`` or ``libraries``
+  ``setup()`` arguments
+- Scripts specified by the ``scripts`` ``setup()`` argument
+- All files specified by the ``package_data`` and ``data_files``
+  ``setup()`` arguments
```
abravalheri, line 16:
```suggestion
- All files specified by the ``package-data`` and ``data-files``
  configuration parameters in ``pyproject.toml`` and/or equivalent in 
  ``setup.cfg``/``setup.py``.
```

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
+- The file specified by the ``license_file`` option in ``setup.cfg``
+  (setuptools 40.8.0+)
+- All files specified by the ``license_files`` option in ``setup.cfg``
+  (setuptools 42.0.0+)
```
abravalheri, line 20:
```suggestion
- All files specified by the ``license-files`` configuration parameter
  in ``pyproject.toml`` and/or equivalent in  ``setup.cfg``/``setup.py``.
  Note that if you don't explicitly set this parameter, ``setuptools``
  will include any files that match the following glob patterns:
  ``['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']``.
```

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
+
+- All Python source files implied by the ``py_modules`` and ``packages``
+  ``setup()`` arguments
+- All C source files mentioned in the ``ext_modules`` or ``libraries``
+  ``setup()`` arguments
+- Scripts specified by the ``scripts`` ``setup()`` argument
+- All files specified by the ``package_data`` and ``data_files``
+  ``setup()`` arguments
+- The file specified by the ``license_file`` option in ``setup.cfg``
+  (setuptools 40.8.0+)
```
abravalheri, line 18: I think it is not worthy to keep track of all the versions directly here.
There will be cases that a parameter is supported since one version in `setup.py`, another version in `setup.cfg` and yet another version in `pyproject.toml`...

Maybe we can add a note at the end saying something like:

> Please note that the list above is guaranteed to work in the last stable version
> of ``setuptools``. The behaviour of older versions might differ.

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
+- All files specified by the ``license_files`` option in ``setup.cfg``
+  (setuptools 42.0.0+)
+- All files matching the pattern ``test/test*.py``
+- ``setup.py`` (or whatever you called your setup script)
```
abravalheri, line 22:
```suggestion
- ``setup.py``
```

Let's remove this comment because it is not really trivial to customise the setup script name in a post-PEP517 world...

docs/userguide/miscellaneous.rst
abravalheri, line 20:
```suggestion
```

See comment above. The same is also valid for the other version notes.

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
+- ``setup.cfg``
+- ``README``
+- ``README.txt``
+- ``README.rst`` (Python 3.7+ or setuptools 0.6.27+)
+- ``README.md`` (setuptools 36.4.0+)
```
abravalheri, line 27:
```suggestion
- ``README``, ``README.txt``, ``README.rst`` or ``README.md``
```

See comment above about keeping track of versions

docs/userguide/miscellaneous.rst
```diff
@@ -4,11 +4,30 @@ Controlling files in the distribution
...
-``py_modules`` or ``packages`` configuration, and the C sources (but not C
-headers) listed as part of extensions when creating a :term:`source
-distribution (or "sdist")`.
+files are necessary for distributing the package. More precisely, the following
+files are included in a source distribution by default:
+
+- All Python source files implied by the ``py_modules`` and ``packages``
+  ``setup()`` arguments
+- All C source files mentioned in the ``ext_modules`` or ``libraries``
+  ``setup()`` arguments
```
abravalheri, line 13:
```suggestion
  ``setup()`` arguments
- Files that match the following glob patterns:
  ``['tests/test*.py', 'test/test*.py']``
```

From https://github.com/pypa/distutils/blob/main/distutils/command/sdist.py#L233-L246

abravalheri (commented):
What do you think about the following updates?

abravalheri:
Not sure, maybe some dependency released a new version recently... I will need to find some time to properly fix this

abravalheri:
Thank you very much!

PR #4113, opened 2023-11-16
Update URLs in documentation: http:// → https://

DimitriPapadopoulos:


docs/conf.py
```diff
@@ -45,15 +45,15 @@
             ),
             dict(
                 pattern=r'Old Setuptools #(?P<old_setuptools>\d+)',
-                url='http://bugs.python.org/setuptools/issue{old_setuptools}',
+                url='https://bugs.python.org/setuptools/issue{old_setuptools}',
```
DimitriPapadopoulos, line 48: I haven't been able to find examples of `https://bugs.python.org/setuptools/issue...`  example URLs out there.

Plain https://bugs.python.org/setuptools/ gives a 404.

abravalheri: O.o that is unfortunate... It seems that we are missing the history of the development.
We also had a problem with the issues and PRs in bitbucket, I believe we can no longer reach that...

I opened https://github.com/python/bugs.python.org/issues/67 to enquire about that...

PR #4125, opened 2023-11-20
Apply more assorted Pyugrade suggestions

DimitriPapadopoulos:
## Summary of changes

Apply more suggestions from [pyupgrade](https://github.com/asottile/pyupgrade).

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


abravalheri (approved):
Thank you very much @DimitriPapadopoulos.

I will not merge this immediately because I am waiting to see if any other bugfixes are necessary for v69, but I think we can merge the PR soon in the future.

DimitriPapadopoulos:
Thank you for looking into this.

Would you agree to add the pyupgrade linter once I have finished fixing the issues reported by pyupgrade?

abravalheri:
> Would you agree to add the pyupgrade linter once I have finished fixing the issues reported by pyupgrade?

If @jaraco is OK with that, I think this change would be fine (we will need to prevent/ignore changes in the `setuptools._distutils` and `setuptools._vendor` paths, though).

DimitriPapadopoulos:
By the way, I don't know how you feel about [ruff](https://astral.sh/ruff), but it lets you [disable pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up), while the original [pyupgrade](https://github.com/asottile/pyupgrade) is rather opinionated and does not let you disable [any of its rules](https://github.com/asottile/pyupgrade#implemented-features).

abravalheri:
I think we are already using `pytest-ruff` for some stuff... But I guess `pytest-ruff` requires a more verbose configuration in order to do the `pyupgrade` stuff (it would be nice if they implement profiles like `isort` did once...).

Avasam:
A sample `ruff.toml` would look like:
```toml
extend-include = [
  "UP", # pyupgrade
]
extend-ignore = [
  "UP015",  # redundant-open-modes, explicit is prefered
]
extend-exclude = [
  # Vendored
  "**/_vendor",
  "setuptools/_distutils",
] 
```

DimitriPapadopoulos:
Doesn't `pytest-ruff` search its config in `pyproject.toml` like ruff, in addition to `ruff.toml`? I am asking because we could avoid an extra config file.

Avasam:
> Doesn't pytest-ruff search its config in `pyproject.toml` like ruff, in addition to `ruff.toml`? I am asking because we could avoid an extra config file.

@jaraco 's previously written opinion on this, which I think will apply just the same to Ruff: https://github.com/pypa/setuptools/pull/3979#pullrequestreview-1691482008
> I do wish to retain the mypy settings in `mypy.ini` for two reasons:
> 
> * The skeleton for this project is maintained [upstream](/jaraco/skeleton), which defines the mypy settings in its own file. Moving the settings to a different file makes it difficult to compare and manage the differences with settings inherited from upstream.
> * Piling all project settings in pyproject.toml is a bad idea in general. I've been meaning to write a blog post about it, but the tl;dr is that because there's no ordering or namespacing, the settings lack structure and consistency (compared with `mypy` which always appears after `docs/` and before `newsfragments/` and is clearly distinct from towncrier or tox or pytest settings.

DimitriPapadopoulos:
I have added a `ruff.toml` file in 7e986ee.

I have disabled warnings about string interpolation for now, because the changes are really pervasive. Also, I find f-strings might be less readable when string interpolation involves a complex expression instead of a mere variable. In such cases, I feel keeping `.format()` or adding a temporary variable to store the result of the complex expression is preferable. Do you want me to proceed nevertheless?

abravalheri:
Let's wait to see if Jason is Ok with the approach. Worst case scenario we can remove the `ruff.toml` file and merge the PR as it is.

jaraco:
Don’t let me be the blocker here.

Normally, I’d like changes like these to be considered at jaraco/skeleton so they can apply to all projects and be kept in sync. The specific effects on this codebase, and configuration specific to this codebase, however, would need to be submitted here. So let’s move forward with this change here and see what implications it has for the general approach.

I’m pleased to see the use of Ruff over pyupgrade. Id also be interested to see Ruff replace black and implement other common formatting/linking like import sorts.

abravalheri:
@DimitriPapadopoulos, let's merge this PR as it is then (but I will not cutting a release right now, just add the changes to main, I hope that is OK).
I will just wait another day to see if we don't have any other backlash on v69 before doing the merge.

> I have disabled warnings about string interpolation for now, because the changes are really pervasive. Also, I find f-strings might be less readable when string interpolation involves a complex expression instead of a mere variable. In such cases, I feel keeping .format() or adding a temporary variable to store the result of the complex expression is preferable. Do you want me to proceed nevertheless?

If that would allow us to reduce the configuration size for Ruff, I like the idea of temporary variable. I think it is worth to be as streamline as possible. But let's do that in a follow up PR...

DimitriPapadopoulos:
> I’m pleased to see the use of Ruff over pyupgrade. Id also be interested to see Ruff replace black and implement other common formatting/linking like import sorts.

I have started working on this in https://github.com/jaraco/skeleton/pull/99.

abravalheri:
> I will just wait another day to see if we don't have any other backlash on v69 before doing the merge.

Postponed until we figure out what happened on `#4136`

DimitriPapadopoulos:
> If that would allow us to reduce the configuration size for Ruff, I like the idea of temporary variable. I think it is worth to be as streamline as possible. But let's do that in a follow up PR...

@abravalheri I have noticed that jaraco/skeleton uses the [recent `skip-string-normalization` black option](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#s-skip-string-normalization):

[**jaraco/skeleton/pyproject.toml**](https://github.com/jaraco/skeleton/blob/5732ebeeaa9480f8cd80c96a3183d7b247f27214/pyproject.toml#L5-L6)
```python
[tool.black]
skip-string-normalization = true
```
I shall instruct black to leave quotes as they are, as in jaraco/skeleton.

abravalheri:
I think that setuptools should have `ruff.toml` as identical as possible to skeleton, so that it can be merged easily.

ruff.toml
```diff
@@ -1,4 +1,3 @@
-[lint]
```
abravalheri, line 1: Should we just leave the `[lint]` section header so it differs less from the `skeleton` configuration?

DimitriPapadopoulos: Good catch. I should rather fix the [skeleton configuration](https://github.com/jaraco/skeleton/blob/main/ruff.toml) as `ignore` is part of the [**Top-level**](https://docs.astral.sh/ruff/settings/#top-level) [Ruff Settings](https://docs.astral.sh/ruff/settings/).

abravalheri: I think that the `ruff` docs are a bit outdated, the [main example](https://docs.astral.sh/ruff/configuration/#__tabbed_1_2) uses the `[lint]` section, and the detailed reference puts them the the global scope...

[Henry once told me](https://github.com/abravalheri/validate-pyproject/pull/125) that in the past Ruff docs would use the global config but now they are going towards `[lint]`...
In this [PR](https://github.com/astral-sh/ruff/pull/7732) we can see the author referencing the new `lint` sections:

> - Breaking up "Configuration" into "Configuring Ruff" (for generic configuration), and new Linter- and Formatter-specific sections.
> - Updating all example configurations to use [tool.ruff.lint] and [tool.ruff.format].

DimitriPapadopoulos: They are going towards `lint` for all options that are related to `lint`, that's for sure. However, ignore applies to both `lint` and `format`. At least that's my understanding.

abravalheri: Perfect, we got a definitive answer from the maintainers: https://github.com/astral-sh/ruff/discussions/8297#discussioncomment-8012937

DimitriPapadopoulos: I'll put back `[lint]`.

DimitriPapadopoulos: Done. But note this requires to duplicate `extend-exclude`, both under `[lint]` and `[format]`.

abravalheri:
I just squashed the first and last commit together because I want to avoid as much as I can future git commits.

ruff.toml
```diff
@@ -16,7 +16,25 @@ ignore = [
...
+]
+extend-exclude = [
+	"**/_vendor",
+	"setuptools/_distutils",
+]
 
 [format]
+extend-exclude = [
+	"**/_vendor",
+	"setuptools/_distutils",
```
abravalheri, line 37: Let's ignore this extra-code path that is automatically generated:

```suggestion
	"setuptools/_distutils",
	"setuptools/config/_validate_pyproject",
```

ruff.toml
```diff
@@ -16,7 +16,25 @@ ignore = [
...
+extend-ignore = [
+	"UP015",  # redundant-open-modes, explicit is prefered
+	"UP030",  # temporarily disabled
+	"UP031",  # temporarily disabled
+	"UP032",  # temporarily disabled
+	"UP036",  # temporarily disabled
+]
+extend-exclude = [
+	"**/_vendor",
+	"setuptools/_distutils",
```
abravalheri, line 31: Let's ignore this extra-code path that is automatically generated:

```suggestion
	"setuptools/_distutils",
	"setuptools/config/_validate_pyproject",
```

ruff.toml
abravalheri: Thank you very much @DimitriPapadopoulos.

Yeah, not ideal. But I guess it is good to go with Ruff maintainer's recommendation.

abravalheri:
Thank you very much for the contribution!

PR #4134, opened 2023-11-22
Fix incorrect PEP reference in news entry for 69.0.0

aphedges:
The entry referenced PEP 612 (Parameter Specification Variables), but it should have referenced PEP 621 (Storing project metadata in pyproject.toml) instead.


abravalheri:
Nice catch! Thabk you very much.

PR #4144, opened 2023-11-28
Mark interfaces as Protocols and add missing `self` argument

Avasam:
## Summary of changes

This was noticed in https://github.com/python/typeshed/pull/10058/files#diff-1acb87d31f86c216cde10127dbe27643bd9d83975a757d3fe75559c03fd9b934

This change allows users to use these Protocols in static-type-checking to ensure proper usage of any class implementing these Interfaces.

Doesn't close any open issue, but relates to/work towards #2345

### Pull Request Checklist
- [ ] Changes have tests (static type-checking tests are to be added later in a different PR, after #3979 )

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -544,54 +545,54 @@ def get_entry_info(dist, group, name):
...
+    def metadata_listdir(self, name):
         """List of metadata names in the directory (like ``os.listdir()``)"""
 
-    def run_script(script_name, namespace):
+    def run_script(self, script_name, namespace):
         """Execute the named script in the supplied namespace dictionary"""
 
 
-class IResourceProvider(IMetadataProvider):
+class IResourceProvider(IMetadataProvider, Protocol):
```
jaraco, line 571: Today I learned about [extending protocols](https://peps.python.org/pep-0544/#merging-and-extending-protocols). Nice work.

pkg_resources/__init__.py
```diff
@@ -544,54 +545,54 @@ def get_entry_info(dist, group, name):
     return get_distribution(dist).get_entry_info(group, name)
 
 
-class IMetadataProvider:
-    def has_metadata(name):
+class IMetadataProvider(Protocol):
```
jaraco, line 548: I have no idea if these protocols are used anywhere (such as with Zope), but I suspect not. If this change ends up causing unexpected disruption, we may have to back it out. Thanks for making it a separate PR.

PR #4148, opened 2023-11-30
Use the stdlib toml library on sufficiently new python

eli-schwartz:
Although tomli is vendored and always available, it is more clean to avoid using the backport on python 3.11. This makes it easier to automatically drop outdated branches in the future, and is a micro-optimization when user code imports a toml library in the same process space.

Additionally, setuptools.extern.VendorImporter explicitly supports removing vendored modules as long as they are globally installed and available. For distributors that rely on this, importing tomllib first permits those distributors to avoid packaging tomli on versions of python that already have tomllib.

### DIFF
--- a/setuptools/config/pyprojecttoml.py
+++ b/setuptools/config/pyprojecttoml.py
@@ -11,6 +11,7 @@ with the help of ``tomllib`` or ``tomli``.
 
 import logging
 import os
+import sys
 from contextlib import contextmanager
 from functools import partial
 from typing import TYPE_CHECKING, Callable, Dict, Mapping, Optional, Set, Union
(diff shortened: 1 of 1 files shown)


eli-schwartz:
ping? It would be very nice to stop needing this in the common case...

abravalheri:
Hi @eli-schwartz 

I am away from my pc for the entire December só I will not be able to cut any release. Once I am back in January I will have a look.

setuptools/config/pyprojecttoml.py
```diff
@@ -28,10 +29,13 @@
 
 
 def load_file(filepath: _Path) -> dict:
-    from setuptools.extern import tomli  # type: ignore
+    if sys.version_info >= (3, 11):
+        import tomllib
+    else:
```
abravalheri, line 34: Should we add 

```suggestion
    else:  # pragma: no cover
```

Just to get rid of the annoying coverage warning?

abravalheri (approved):
Thank you very much @eli-schwartz

setuptools/config/pyprojecttoml.py
eli-schwartz: If you want me to, sure.

This (among other reasons) is why I simply don't care about or use coverage. I have no opinions on it other than "if the maintainer asks me to do it then I can oblige".

eli-schwartz: Probably updating the version of python used in coverage runs is going to require repetitively reworking coverage pragma ignores, but as a coverage using project you're no doubt used to this kind of churn.

eli-schwartz: Pushed, fingers crossed.

Issue #2809, opened 2021-10-03
[FR] Prefer distlib for script generation (develop only)


jaraco:
### What's the problem this feature will solve?

There are issues with the script executable wrappers on Windows:

- https://github.com/pypa/setuptools/issues/337

And maybe others.

### Describe the solution you'd like

Setuptools should rely on [distlib](https://pypi.org/project/distlib), the same library used by pip for generating wrapper scripts, to generate its wrapper scripts when `develop` is invoked.

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

abravalheri:
My understanding is that if/whenever setuptools adopts PEP 660, the responsibility for generating the script wrappers is transferred to the installer, right? (I believe that pip would generate wrappers as it does right now for regular wheels, based on the entry-points).

If that is the case, would it also be necessary for setuptools to generate those wrappers? Is there anything else I am missing here?

PR #4161, opened 2023-12-15
Fix relocation issues with binary launchers on Windows

dimula73:
This patchset makes the launcher .exe file on Windows more usable:

1) The .exe file (and its script) is now relocatable, because it falls back to `python.exe` when shebang line of the script points to the invalid location (e.g. when the script was relocated). On Linux such problem in non-existent, since it usually sets `/usr/bin/env python`, which is relocatable.

2) When the launcher fails to run python it should return a non-zero value to pass error to upper levels. This bug makes meson basically ci-unusable on Windows, because meson sporadically returns success even when something has failed.

**Still TODO:** the patch needs binaries for x32 and arm64 platforms. I'm not sure how to build them. I'm also not sure if my build for x64 builds is suitable for this project, since I use llvm-mingw-ucrt. It might be a bit too new for the purpose of setuptools :)


jaraco:
Some work has since been done to make the scripts more relocatable in some scenarios.

This PR needs a better description, including a reproducer of the problem(s) encountered with the current behavior. It might also be nice to have a regression test capturing some of the expected behaviors.

That said, these launchers should also be considered deprecated. I want to eliminate them altogether unify around distlib for any script needs (#2809).

I'm not sure what it will take to build new executables. Someone will need to figure that out before we can accept changes. I see there is launcher/CMakeLists.txt, so maybe that's a clue.

Without someone to invest that time and energy, I'm closing this issue, though I'm happy to reopen or review a new one if the effort is revived and tracking toward mergeable.

PR #4168, opened 2024-01-02
Fix flake8-bugbear warnings

DimitriPapadopoulos:
## Summary of changes

Fix [flake8-bugbear issues](https://github.com/PyCQA/flake8-bugbear) reported by [ruff](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) 0.1.9.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -407,20 +407,18 @@ def get_provider(moduleOrReq):
     return _find_adapter(_provider_factories, loader)(module)
 
 
-def _macos_vers(_cache=None):
-    if not _cache:
-        version = platform.mac_ver()[0]
-        # fallback for MacPorts
-        if version == '':
-            plist = '/System/Library/CoreServices/SystemVersion.plist'
-            if os.path.exists(plist):
-                if hasattr(plistlib, 'readPlist'):
-                    plist_content = plistlib.readPlist(plist)
-                    if 'ProductVersion' in plist_content:
-                        version = plist_content['ProductVersion']
-
-        _cache = [version.split('.')]
-    return _cache[0]
+@functools.lru_cache(maxsize=None)
+def _macos_vers():
+    version = platform.mac_ver()[0]
+    # fallback for MacPorts
+    if version == '':
+        plist = '/System/Library/CoreServices/SystemVersion.plist'
+        if os.path.exists(plist):
+            if hasattr(plistlib, 'readPlist'):
+                plist_content = plistlib.readPlist(plist)
+                if 'ProductVersion' in plist_content:
+                    version = plist_content['ProductVersion']
+    return version.split('.')
 
 
 def _macos_arch(machine):
@@ -2422,13 +2420,9 @@ def _cygwin_patch(filename):  # pragma: nocover
     return os.path.abspath(filename) if sys.platform == 'cygwin' else filename
 
 
-def _normalize_cached(filename, _cache=None):
-    _cache = _cache or {}
-    try:
-        return _cache[filename]
-    except KeyError:
-        _cache[filename] = result = normalize_path(filename)
-        return result
+@functools.lru_cache(maxsize=None)
+def _normalize_cached(filename):
+    return normalize_path(filename)
 
 
 def _is_egg_path(path):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -407,7 +407,7 @@ def get_provider(moduleOrReq):
     return _find_adapter(_provider_factories, loader)(module)
 
 
-def _macos_vers(_cache=[]):
+def _macos_vers(_cache=None):
     if not _cache:
         version = platform.mac_ver()[0]
         # fallback for MacPorts
@@ -419,7 +419,7 @@ def _macos_vers(_cache=[]):
                     if 'ProductVersion' in plist_content:
                         version = plist_content['ProductVersion']
 
-        _cache.append(version.split('.'))
+        _cache = [version.split('.')]
     return _cache[0]
 
 
@@ -2422,7 +2422,8 @@ def _cygwin_patch(filename):  # pragma: nocover
     return os.path.abspath(filename) if sys.platform == 'cygwin' else filename
 
 
-def _normalize_cached(filename, _cache={}):
+def _normalize_cached(filename, _cache=None):
+    _cache = _cache or {}
     try:
         return _cache[filename]
     except KeyError:
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -117,11 +117,11 @@ def _file_with_extension(directory, extension):
     matching = (f for f in os.listdir(directory) if f.endswith(extension))
     try:
         (file,) = matching
-    except ValueError:
+    except ValueError as e:
         raise ValueError(
             'No distribution was found. Ensure that `setup.py` '
             'is not empty and that it calls `setup()`.'
-        )
+        ) from e
     return file
 
 
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -386,7 +386,7 @@ class FileList(_FileList):
             process_action = action_map[action]
         except KeyError as e:
             raise DistutilsInternalError(
-                "this cannot happen: invalid action '{action!s}'".format(action=action),
+                f"Invalid MANIFEST.in: unknow action {action!r} in {line!r}"
             ) from e
 
         # OK, now we know that the action is valid and we have the
--- a/setuptools/command/egg_info.py
+++ b/setuptools/command/egg_info.py
@@ -384,10 +384,10 @@ class FileList(_FileList):
 
         try:
             process_action = action_map[action]
-        except KeyError:
+        except KeyError as e:
             raise DistutilsInternalError(
                 "this cannot happen: invalid action '{action!s}'".format(action=action),
-            )
+            ) from e
 
         # OK, now we know that the action is valid and we have the
         # right number of words on the line for that action -- so we
--- a/setuptools/config/_validate_pyproject/error_reporting.py
+++ b/setuptools/config/_validate_pyproject/error_reporting.py
@@ -24,7 +24,7 @@ _SKIP_DETAILS = (
     "must not be there",
 )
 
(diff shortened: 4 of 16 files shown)


pkg_resources/__init__.py
```diff
@@ -419,7 +419,7 @@ def _macos_vers(_cache=[]):
                     if 'ProductVersion' in plist_content:
                         version = plist_content['ProductVersion']
 
-        _cache.append(version.split('.'))
+        _cache = [version.split('.')]
```
abravalheri, line 422: This would be a change in the intent of the implementation wouldn't it?
Wasn't this function "over-cleverly" written so that the `_cache` argument is modified by the first call and then it actually works as a cache on subsequent calls?

The `_` at the beginning of the argument name gives me the impression that this function is never supposed to be invoked with an explicit argument and that the mutable default argument is intentional...

To be properly re-written, we would have to use something like `functools.cache` (or `lru_cache(maxsize=None)` for Python < 3.9) I guess...

pkg_resources/__init__.py
```diff
@@ -2422,7 +2422,8 @@ def _cygwin_patch(filename):  # pragma: nocover
     return os.path.abspath(filename) if sys.platform == 'cygwin' else filename
 
 
-def _normalize_cached(filename, _cache={}):
+def _normalize_cached(filename, _cache=None):
```
abravalheri, line 2425: This is probably the same case of "over-clever" optimisation, isn't it?
I have the impression that the proposed change would defeat the optimisation, right?

setuptools/config/setupcfg.py
```diff
@@ -582,11 +582,11 @@ def _parse_version(self, value):
             # accidentally include newlines and other unintended content
             try:
                 Version(version)
-            except InvalidVersion:
+            except InvalidVersion as e:
                 raise OptionError(
                     f'Version loaded from {value} does not '
                     f'comply with PEP 440: {version}'
-                )
+                ) from e
```
abravalheri, line 589: Do we actually want to have the previous exception also shown?

Would the increased traceback bring more information for the user to debug, or is the current error message enough?

The problem with `from None` is that it also reduces the stack trace isn't it? I think that probably the way it is written right now gets the sweet spot: enough stack trace for debugging, but no extra unnecessary verboseness. Maybe it is just bugbear being overly pedantic...

setuptools/config/_validate_pyproject/error_reporting.py
```diff
@@ -24,7 +24,7 @@
     "must not be there",
 )
 
-_NEED_DETAILS = {"anyOf", "oneOf", "anyOf", "contains", "propertyNames", "not", "items"}
+_NEED_DETAILS = {"anyOf", "oneOf", "contains", "propertyNames", "not", "items"}
```
abravalheri, line 27: This code path is generated automatically, so we have to fix it first elsewhere, otherwise the changes can be overwritten in the next update...

```suggestion
_NEED_DETAILS = {"anyOf", "oneOf", "allOf", "contains", "propertyNames", "not", "items"}
```

setuptools/command/egg_info.py
```diff
@@ -384,10 +384,10 @@ def process_template_line(self, line):
 
         try:
             process_action = action_map[action]
-        except KeyError:
+        except KeyError as e:
             raise DistutilsInternalError(
                 "this cannot happen: invalid action '{action!s}'".format(action=action),
-            )
+            ) from e
```
abravalheri, line 390: We probably want to improve the error message instead... Maybe something like

```invalid action in `MANIFEST.in`: {action!r}```

... so people actually know where the problem lies.

But other than that I am a bit sceptical that having the info for the KeyError would facilitate the debugging... Also I think it would be nice to preserve the stack trace so `from None` would not work here also...

File: setuptools/build_meta.py [hunk 6]
```diff
@@ -117,11 +117,11 @@ def _file_with_extension(directory, extension):
     matching = (f for f in os.listdir(directory) if f.endswith(extension))
     try:
         (file,) = matching
-    except ValueError:
+    except ValueError as e:
         raise ValueError(
             'No distribution was found. Ensure that `setup.py` '
             'is not empty and that it calls `setup()`.'
-        )
+        ) from e
```
abravalheri, line 124: I am a bit sceptical that chaining 2 `ValueError` would facilitate the debugging...
Probably the original idea here was to preserve the stack trace but avoid unnecessary verboseness.

Using `from None` would suppress the stack trace right? (or am I mistaken on how `raise ... from None` works?) Suppressing the stack trace would make it more difficult to debug...

Maybe we should just leave this one as it is...

setuptools/monkey.py
```diff
@@ -130,7 +130,7 @@ def patch_func(replacement, target_mod, func_name):
 
 
 def get_unpatched_function(candidate):
-    return getattr(candidate, 'unpatched')
+    return candidate.unpatched
```
abravalheri, line 133: It is hard to guess what was the original intention but maybe the idea was to have a

```suggestion
    return getattr(candidate, 'unpatched', None)
```

but there was a writing mistake?

abravalheri (commented):
Thank you very much @DimitriPapadopoulos for the valuable contribution.

Please see my comments below.

pkg_resources/__init__.py
DimitriPapadopoulos: You're right of course.

Alternatively `_cache` could be moved outside the function:
```python
_cache = []

def _macos_vers():
    if not _cache:
        ...
```

setuptools/build_meta.py
DimitriPapadopoulos: https://docs.python.org/3/tutorial/errors.html#exception-chaining

As far as I can understand, you get the 2 `ValueError` with or without chaining.  Just the wording differs. From a small example similar to the first one in the above lin:

```diff
 Traceback (most recent call last):
   File "/path/to/foobar.py", line 3, in <module>
     open("database.sqlite")
 FileNotFoundError: [Errno 2] No such file or directory: 'database.sqlite'
 
-During handling of the above exception, another exception occurred:
+The above exception was the direct cause of the following exception:
 
 Traceback (most recent call last):
   File "/path/to/foobar.py", line 5, in <module>
-    raise RuntimeError("unable to handle error")
+    raise RuntimeError("unable to handle error") from exc
 RuntimeError: unable to handle error
```

setuptools/config/_validate_pyproject/error_reporting.py
DimitriPapadopoulos: So do I commit the above suggestion or not? If the code is generated automatically, is it worth it?

pkg_resources/__init__.py
abravalheri: If that is the selected approach, we should use a more specific name.

Alternatively we can use something like:

```python
@functools.lru_cache(maxsize=None)
def _macos_vers():
    version = platform.mac_ver()[0]
    # fallback for MacPorts
    if version == '':
        plist = '/System/Library/CoreServices/SystemVersion.plist'
        if os.path.exists(plist):
            if hasattr(plistlib, 'readPlist'):
            plist_content = plistlib.readPlist(plist)
                if 'ProductVersion' in plist_content:
                    version = plist_content['ProductVersion']
    return version.split('.')
```

and get rid of the variable all together (similar approach for the other overly-clever function).

setuptools/config/_validate_pyproject/error_reporting.py
abravalheri: I think you can commit, I checked `validate-pyproject` and I think it was already corrected there. So if later we overwrite, that would not be problematic.

I am holding back a bit to re-generate the code because some minor changes were introduced. They should not cause problems, but you know how it is... Hyrum's Law hits heavily on the setuptools code base 😅.

setuptools/build_meta.py
abravalheri: Oh well, if that is the case, then the proposed changes should be fine, they don't change that much anyway.

setuptools/config/setupcfg.py
abravalheri: For future reference: see resolution in https://github.com/pypa/setuptools/pull/4168#discussion_r1440649502.

setuptools/command/egg_info.py
```diff
@@ -384,10 +384,10 @@ def process_template_line(self, line):
 
         try:
             process_action = action_map[action]
-        except KeyError:
+        except KeyError as e:
             raise DistutilsInternalError(
                 "this cannot happen: invalid action '{action!s}'".format(action=action),
```
abravalheri, line 389: I think something like this would make it easier to identify and correct mistakes:

```suggestion
                f"Invalid MANIFEST.in: unknow action {action!r} in {line!r}"
```

But I am not sure if this change requires further changes in the tests...

setuptools/command/egg_info.py
abravalheri: Future reference: see resolution for `raise ... from ...`  in https://github.com/pypa/setuptools/pull/4168#discussion_r1440649502.

setuptools/build_meta.py
abravalheri: Thank you btw for clarifying my misunderstanding :)

setuptools/command/egg_info.py
DimitriPapadopoulos: Let's give it a try... I should be able to modify the tests accordingly.

pkg_resources/__init__.py
DimitriPapadopoulos: By the way, function [`plistlib.readPlist`](https://docs.python.org/3.8/library/plistlib.html#plistlib.readPlist) has been [removed from Python  ≥ 3.9:](https://github.com/python/cpython/pull/15615)
> Deprecated since version 3.4: Use [load()](https://docs.python.org/3.8/library/plistlib.html#plistlib.load) instead.

DimitriPapadopoulos: The cache mechanism is now based on `@functools.lru_cache`.

pkg_resources/__init__.py
DimitriPapadopoulos: Same, the cache mechanism is now based on `@functools.lru_cache`.

PR #4169, opened 2024-01-02
Fix flake8-return warnings

DimitriPapadopoulos:
## Summary of changes

Fix [flake8-return](https://pypi.org/project/flake8-return/) issues reported by [ruff](https://docs.astral.sh/ruff/rules/#flake8-return-ret) 0.1.9.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2832,9 +2832,7 @@ class Distribution:
 
     def _get_version(self):
         lines = self._get_metadata(self.PKG_INFO)
-        version = _version_from_file(lines)
-
-        return version
+        return _version_from_file(lines)
 
     def activate(self, path=None, replace=False):
         """Ensure distribution is importable on `path` (default=sys.path)"""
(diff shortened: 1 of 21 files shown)


pkg_resources/__init__.py
```diff
@@ -1141,8 +1141,7 @@ def obtain(self, requirement, installer=None):
         None is returned instead.  This method is a hook that allows subclasses
         to attempt other ways of obtaining a distribution before falling back
         to the `installer` argument."""
-        if installer is not None:
-            return installer(requirement)
+        if installer return installer(requirement) else None
```
abravalheri, line 1144: This is probably one of the things that is causing problems (syntax error).

DimitriPapadopoulos: Fixed.

abravalheri:
Thank you very much @DimitriPapadopoulos .


Should `flake8-return` be added to the ruff configuration as well after these fixes?

DimitriPapadopoulos:
Yes, it could be added. Some rules should be disabled though. I have not fixed such issues:
``` python
if ...:
    return ...
else:
    return ...
```
I find the symmetry makes them more readable than:
``` python
if ...:
    return ...
return ...
```

abravalheri:
Thank you very much @DimitriPapadopoulos, let's merge this for now...
Later we can add the `flake8-return`.

> Some rules should be disabled though.

I am not sure about this. I am more keen to keep the linter configuration as dry as possible, ideally just working out of the box.

PR #4189, opened 2024-01-22
Point to a live documentation

befeleme:
## Summary of changes

The previous link, https://packaging.python.org/en/latest/specifications/declaring-project-metadata/, ends with 404.
Update the link to point to the actual documentation


abravalheri:
Thank you @befeleme.

PR #4192, opened 2024-01-25
Initial pyright config

Avasam:
## Summary of changes
Work towards step 2 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833 . Non-trivial fixes should go in different PRs.
Merge https://github.com/pypa/setuptools/pull/3979 and #4257 first.
`npx -y pyright@latest . --pythonversion=3.8|3.12 --pythonplatform=Linux|Windows` results:
`0 errors, 418 warnings, 0 informations`

### Pull Request Checklist
- [x] Changes have tests (this is the test!)

[PR docs]:

### DIFF
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -75,7 +75,7 @@ jobs:
           echo 'To run this test locally with npm pre-installed, run:'
           echo '> npx -y pyright@${{ env.PYRIGHT_VERSION }} --threads'
           echo 'You can also instead install "Pyright for Python" which will install npm for you:'
-          if [$PYRIGHT_VERSION == 'latest'] then
+          if [$PYRIGHT_VERSION == 'latest']; then
             echo '> pip install -U'
           else
             echo '> pip install pyright==${{ env.PYRIGHT_VERSION }}'
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -70,7 +70,7 @@ jobs:
           python-version: ${{ matrix.python }}
           allow-prereleases: true
       - name: Install typed dependencies
-        run: python -m pip install .[core][type]
+        run: python -m pip install -e .[core][type]
       - name: Inform how to run locally
         run: |
           echo 'To run this test locally with npm pre-installed, run:'
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -64,7 +64,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Setup Python
-        id: python-install
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
@@ -86,3 +85,4 @@ jobs:
         uses: jakebailey/pyright-action@v2
         with:
           version: ${{ env.PYRIGHT_VERSION }}
+          extra-args: --threads
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -44,21 +44,6 @@ jobs:
           - "3.12"
         platform:
           - ubuntu-latest
-          - macos-latest
-          - windows-latest
-        include:
-          - python: "3.9"
-            platform: ubuntu-latest
-          - python: "3.10"
-            platform: ubuntu-latest
-          - python: "3.11"
-            platform: ubuntu-latest
-          # Python 3.8, 3.9 are on macos-13 but not macos-latest (macos-14-arm64)
-          # https://github.com/actions/setup-python/issues/850
-          # https://github.com/actions/setup-python/issues/696#issuecomment-1637587760
-          - {python: "3.8", platform: "macos-13"}
-        exclude:
-          - {python: "3.8", platform: "macos-latest"}
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 10
     steps:
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -21,10 +21,10 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  # pin pyright so a new version doesn't suddenly cause the CI to fail,
+  # pin pyright version so a new version doesn't suddenly cause the CI to fail,
   # until types-setuptools is removed from typeshed.
   # For help with static-typing issues, or pyright update, ping @Avasam
-  PYRIGHT_VERSION: "latest"
+  PYRIGHT_VERSION: "1.1.377"
 
   # Environment variable to support color support (jaraco/skeleton#66)
   FORCE_COLOR: 1
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -75,7 +75,7 @@ jobs:
           echo 'To run this test locally with npm pre-installed, run:'
           echo '> npx -y pyright@${{ env.PYRIGHT_VERSION }} --threads'
           echo 'You can also instead install "Pyright for Python" which will install npm for you:'
-          if ['$PYRIGHT_VERSION' == 'latest']; then
+          if [ '$PYRIGHT_VERSION' == 'latest' ]; then
             echo '> pip install -U'
           else
             echo '> pip install pyright==${{ env.PYRIGHT_VERSION }}'
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -69,7 +69,7 @@ jobs:
           python-version: ${{ matrix.python }}
           allow-prereleases: true
       - name: Install typed dependencies
-        run: python -m pip install -e .[core][type]
+        run: python -m pip install -e .[core,type]
       - name: Inform how to run locally
         run: |
           echo 'To run this test locally with npm pre-installed, run:'
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -24,7 +24,7 @@ env:
   # pin pyright so a new version doesn't suddenly cause the CI to fail,
   # until types-setuptools is removed from typeshed.
   # For help with static-typing issues, or pyright update, ping @Avasam
-  PYRIGHT_VERSION: "1.1.377"
+  PYRIGHT_VERSION: "latest"
 
   # Environment variable to support color support (jaraco/skeleton#66)
   FORCE_COLOR: 1
@@ -75,12 +75,13 @@ jobs:
           echo 'To run this test locally with npm pre-installed, run:'
           echo '> npx -y pyright@${{ env.PYRIGHT_VERSION }} --threads'
           echo 'You can also instead install "Pyright for Python" which will install npm for you:'
-          if [$PYRIGHT_VERSION == 'latest']; then
+          if ['$PYRIGHT_VERSION' == 'latest']; then
             echo '> pip install -U'
           else
             echo '> pip install pyright==${{ env.PYRIGHT_VERSION }}'
           fi
           echo 'pyright --threads'
+        shell: bash
       - name: Run pyright
         uses: jakebailey/pyright-action@v2
         with:
--- a/.github/workflows/pyright.yml
+++ b/.github/workflows/pyright.yml
@@ -1,0 +1,88 @@
+# Split workflow file to not interfere with skeleton
+name: pyright
+
+on:
+  merge_group:
+  push:
+    branches-ignore:
+    # temporary GH branches relating to merge queues (jaraco/skeleton#93)
+    - gh-readonly-queue/**
+    tags:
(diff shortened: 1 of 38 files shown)


.github/workflows/main.yml
```diff
@@ -34,6 +34,24 @@ env:
...
+        - ubuntu-latest
+        - macos-latest
+        - windows-latest
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jakebailey/pyright-action@v2
+        with:
+          python-version: ${{ matrix.python }}
+
```
Avasam, line 54: I know this isn't how you usually do your tests. I'm using it as a placeholder. I need a hand ensuring mypy and pyright are run as part of your testing setup.

setuptools/command/build.py
```diff
@@ -138,3 +143,4 @@ def get_output_mapping(self) -> Dict[str, str]:
         Destination files should be strings in the form of
         ``"{build_lib}/destination/file/path"``.
         """
+        ...
```
Avasam, line 146: Refer to https://github.com/microsoft/pyright/issues/6487

.github/workflows/main.yml
abravalheri: Hi @Avasam , does simply adding `pytest-mypy` to the `testing` `extras_require` config (in `setup.cfg` I suspect) does not work for now?

abravalheri (commented):
Hi @Avasam , thank you very much for the efforts in this field. I understand that this is a wipe, but I left a few comments, hopefully that is fine with you.

setuptools/command/build.py
abravalheri: I see the problem here is because we are talking specifically about prototypes, and the spec says:

> Bodies of protocol methods are type checked.

That is a bit disappointing... These pesky ellipsis can be annoying

.github/workflows/main.yml
Avasam: It's already there https://github.com/pypa/setuptools/blob/main/setup.cfg#L50 , so if it was used, it should already have been failing.
Unless the CI is running on PyPy ?

Avasam: I figured it out, it was indeed never run, you need to explicitly specify the `--mypy` flag to pytest. Also this installs and runs fine on PyPy now (dropping Python 3.7 support and AST from Mypy probably helped). Other than https://github.com/abravalheri/validate-pyproject/pull/159

Avasam: On the note of pytest integration:
I just tried out `pytype-pyright` but it didn't fail where `mypy`, `pytest-mypy`, `jakebailey/pyright-action` and `pyright` did, even after setting `--pyright-dir .` on the pytest command line. So I'm not sure I can trust it for this use (it was also developed for internal use in a different project)

Avasam:
The hundreds of `information: Analysis of function "..." is skipped because it is unannotated` are probably quite undesirable (we just started an error to remove spam in logs from `EncodingWarning`...) so I've opened https://github.com/microsoft/pyright/issues/7415

mypy.ini
```diff
@@ -23,9 +23,10 @@ disable_error_code = attr-defined
 
 # - Avoid raising issues when importing from "extern" modules, as those are added to path dynamically.
 #   https://github.com/pypa/setuptools/pull/3979#discussion_r1367968993
+# - distutils doesn't exists on Python 3.12 (stdlib) and mypy won't see the distutils hack.
```
Avasam, line 26:
```suggestion
```

setup.cfg
```diff
@@ -73,7 +73,9 @@ testing =
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
-	mypy==1.9  # pin mypy version so a new version doesn't suddenly cause the CI to fail
+  # pin type-checkers so a new version doesn't suddenly cause the CI to fail
```
Avasam, line 76:
```suggestion
	# pin type-checkers so a new version doesn't suddenly cause the CI to fail
```

pyrightconfig.json
```diff
@@ -0,0 +1,36 @@
...
+	"enableTypeIgnoreComments": true,
+	"typeCheckingMode": "basic",
+	// Avoid raising issues when importing from "extern" modules, as those are added to path dynamically.
+	// https://github.com/pypa/setuptools/pull/3979#discussion_r1367968993
+	"reportMissingImports": "none",
+	// Too many issues caused by vendoring and dynamic patching, still worth fixing when we can
+	"reportAttributeAccessIssue": "warning",
+	// Dynamically defined module-level variables declared as None in pkg_resources lead to a lot of false-positives
+	// TODO: Fix how these variables are typed
+	"reportOptionalCall": "warning",
```
Avasam, line 25: The following PRs work towards this:
- https://github.com/pypa/setuptools/pull/4258
- https://github.com/pypa/setuptools/pull/4258

setuptools/_reqs.py
```diff
@@ -28,15 +28,13 @@ def parse_strings(strs: _StrOrIter) -> Iterator[str]:
...
 @overload
 def parse(strs: _StrOrIter) -> Iterator[Requirement]: ...
-
-
 @overload
 def parse(strs: _StrOrIter, parser: Callable[[str], _T]) -> Iterator[_T]: ...
-
-
-def parse(strs, parser=parse_req):
+def parse(strs: _StrOrIter, parser: Callable[[str], _T] = parse_req) -> Iterator[_T]:  # type: ignore[arg-type]
```
Avasam, line 37: xref https://github.com/python/mypy/issues/3737

pyrightconfig.json
```diff
@@ -0,0 +1,31 @@
...
+	"enableTypeIgnoreComments": true,
+	"typeCheckingMode": "basic",
+	// Too many issues caused by dynamic patching, still worth fixing when we can
+	"reportAttributeAccessIssue": "warning",
+	// Fails on Python 3.12 due to missing distutils and on cygwin CI tests
+	"reportMissingImports": "warning",
+	// FIXME: A handful of reportOperatorIssue spread throughout the codebase
+	"reportOperatorIssue": "warning",
+	// Almost all of these are caused by jaraco.path.build's spec argment not being a Mapping
+	"reportArgumentType": "warning",
```
Avasam, line 23: Relates to https://github.com/jaraco/jaraco.path/issues/2
Although a targeted fix is sufficient. It doesn't need to be fully typed.

pyrightconfig.json
```diff
@@ -0,0 +1,32 @@
...
+	"reportAttributeAccessIssue": "warning",
+	// Fails on Python 3.12 due to missing distutils and on cygwin CI tests
+	"reportAssignmentType": "warning",
+	"reportMissingImports": "warning",
+	"reportOptionalCall": "warning",
+	// FIXME: A handful of reportOperatorIssue spread throughout the codebase
+	"reportOperatorIssue": "warning",
+	// Deferred initialization (initialize_options/finalize_options) causes many "potentially None" issues
+	// TODO: Fix with type-guards or by changing how it's initialized
+	"reportArgumentType": "warning", // A lot of these are caused by jaraco.path.build's spec argment not being a Mapping https://github.com/jaraco/jaraco.path/pull/3
```
Avasam, line 26: Relates to https://github.com/jaraco/jaraco.path/issues/2
Although a targeted fix is sufficient. It doesn't need to be fully typed.

I opened https://github.com/jaraco/jaraco.path/pull/3 to resolve this comment

setuptools/command/build.py
Avasam: Further discussion: https://github.com/astral-sh/ruff/issues/8756#issuecomment-1817575511 . It does seem to be more accurate to spec. Ruff was also updated to not strip ellipses in Protocols in https://github.com/astral-sh/ruff/releases/tag/v0.1.6

pyrightconfig.json
```diff
@@ -0,0 +1,32 @@
...
+	"reportAttributeAccessIssue": "warning",
+	// Fails on Python 3.12 due to missing distutils and on cygwin CI tests
+	"reportAssignmentType": "warning",
+	"reportMissingImports": "warning",
+	"reportOptionalCall": "warning",
+	// FIXME: A handful of reportOperatorIssue spread throughout the codebase
+	"reportOperatorIssue": "warning",
+	// Deferred initialization (initialize_options/finalize_options) causes many "potentially None" issues
+	// TODO: Fix with type-guards or by changing how it's initialized
+	"reportArgumentType": "warning", // A lot of these are caused by jaraco.path.build's spec argument not being a Mapping https://github.com/jaraco/jaraco.path/pull/3
```
abravalheri, line 26: I am assuming that `pyright` is OK with syntactically incorrect JSON files (the JSON spec does not include any forms of comment).

I think they are fine and useful. As long as we don't get an error for using them I am completely OK.

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -275,7 +275,7 @@ def _valid_command_options(cmdclass: Mapping = EMPTY) -> dict[str, set[str]]:
     return valid_options
 
 
-def _load_ep(ep: metadata.EntryPoint) -> tuple[str, type] | None:
+def _load_ep(ep: metadata.EntryPoint) -> tuple[str, ModuleType] | None:
```
abravalheri, line 278: In general, `_load_ep` could return `tuple[str, Any]` actually, because they can point to any attribute inside of any module, right?

But in this particular file `_load_ep` will always load a class (which should be a subclass of `Command` or at least mimic the same protocol).

tox.ini
```diff
@@ -6,6 +6,7 @@ setenv =
 	PYTHONWARNDEFAULTENCODING = 1
 	SETUPTOOLS_ENFORCE_DEPRECATION = {env:SETUPTOOLS_ENFORCE_DEPRECATION:1}
 commands =
+	pyright --threads
```
abravalheri, line 9: I am really sorry @Avasam, I had a change of heart at this.

It seems that when you install `pyright`, it [automatically install NodeJS](https://github.com/RobertCraigie/pyright-python?tab=readme-ov-file#how-pyright-for-python-works). That is a huge undertake: it is a separated programming language/environment and it does seem excessive to automatically add that to the contributors' machines (or mine).

In this scenario, I am reconsidering the first approach that you suggested (adding a new workflow to the CI). That might work better.

I am thinking that if we have a separated workflow file, it would not interfere with `skeleton`. That might be the best scenario.

setuptools/config/_apply_pyprojecttoml.py
Avasam: Ah, I see how `importlib_metadata.EntryPoint.load`'s `functools.reduce` led me astray
`importlib.metadata.EntryPoint.load` is correctly typed as `Any` in typeshed, but `importlib_metadata.EntryPoint.load` kept the `ModuleType` inference

tox.ini
Avasam: > I am reconsidering the first approach that you suggested (adding a new workflow to the CI). That might work better.
> I am thinking that if we have a separated workflow file, it would not interfere with skeleton. That might be the best scenario.

I am 100% for this solution and is my preference as well !

Avasam: Ok so here's a workflow file without touching `tox.ini` or `.github/workflows/main.yml`, or referencing pyright in `pyproject.toml`.

Incidentally it also won't be picked-up by the `collateral` job.

.github/workflows/pyright.yml
```diff
@@ -0,0 +1,89 @@
...
+          - python: "3.10"
+            platform: ubuntu-latest
+          - python: "3.11"
+            platform: ubuntu-latest
+          # Python 3.8, 3.9 are on macos-13 but not macos-latest (macos-14-arm64)
+          # https://github.com/actions/setup-python/issues/850
+          # https://github.com/actions/setup-python/issues/696#issuecomment-1637587760
+          - {python: "3.8", platform: "macos-13"}
+        exclude:
+          - {python: "3.8", platform: "macos-latest"}
```
abravalheri, line 61: Thank you very much @Avasam for putting this togheter.

Would it be too bad if we simplify this for the sake of resource usage? (the CI is shared by the PyPA organisation)

I was thinking first and last supported versions and the cheapest platform:

```suggestion
        python:
          - "3.8"
          - "3.12"
        platform:
          - ubuntu-latest
```

Avasam: I considered it. Oldest and newest should cover using things that are too new or removed (like distutils before/after 3.12)

abravalheri:
Thank you very much, let's get this one going.

PR #4217, opened 2024-02-13
Append arbitrary args instead of prepending.

jameshilliard:
## Summary of changes

Restore original argument append behavior which was changed in #4079.

Fixes:
```
error: option --dist-dir not recognized

ERROR Backend subprocess exited when trying to invoke build_wheel
```

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -369,7 +369,12 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
         return self._bubble_up_info_directory(metadata_directory, ".dist-info")
 
     def _build_with_temp_dir(
-        self, setup_command, result_extension, result_directory, config_settings
+        self,
+        setup_command,
+        result_extension,
+        result_directory,
+        config_settings,
+        arbitrary_args=(),
     ):
         result_directory = os.path.abspath(result_directory)
 
@@ -384,6 +389,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
                 *setup_command,
                 "--dist-dir",
                 tmp_dist_dir,
+                *arbitrary_args,
             ]
             with no_install_setup_requires():
                 self.run_setup()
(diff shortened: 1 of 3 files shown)


jameshilliard:
Closing in favor of #4218.

setuptools/build_meta.py
```diff
@@ -402,10 +404,11 @@ def build_wheel(
     ):
         with suppress_known_deprecation():
             return self._build_with_temp_dir(
-                ['bdist_wheel', *self._arbitrary_args(config_settings)],
+                ['bdist_wheel'],
                 '.whl',
                 wheel_directory,
                 config_settings,
+                True,
```
abravalheri, line 411: We probably can avoid the conditional all together this way:

```suggestion
                self._arbitrary_args(config_settings),
```

(untested suggestion)

setuptools/build_meta.py
```diff
@@ -385,6 +385,8 @@ def _build_with_temp_dir(
                 "--dist-dir",
                 tmp_dist_dir,
             ]
+            if arbitrary_args:
+                sys.argv += self._arbitrary_args(config_settings)
```
abravalheri, line 389: We probably can avoid the conditional all together this way:

```suggestion
                tmp_dist_dir,
                *arbitrary_args,
            ]
```

(untested suggestion)

setuptools/build_meta.py
```diff
@@ -369,7 +369,7 @@ def prepare_metadata_for_build_wheel(
         return self._bubble_up_info_directory(metadata_directory, ".dist-info")
 
     def _build_with_temp_dir(
-        self, setup_command, result_extension, result_directory, config_settings
+        self, setup_command, result_extension, result_directory, config_settings, arbitrary_args=False
```
abravalheri, line 372: The CI error is probably a formatting error in this line:

```suggestion
        self,
        setup_command,
        result_extension,
        result_directory,
        config_settings,
        arbitrary_args=(),
```

(untested suggestion)
(type of `arbitrary_args` changed to match other suggestions)

setuptools/build_meta.py
jameshilliard: applied as suggested

setuptools/build_meta.py
jameshilliard: applied as suggested

setuptools/build_meta.py
jameshilliard: applied as suggested

abravalheri:
Thank you very much @jameshilliard for working on this.

Would you like to add a regression test so that we don't accidentally break the use case in the future?

jameshilliard:
> Would you like to add a regression test so that we don't accidentally break the use case in the future?

Test added that should catch this issue.

abravalheri:
Thank you very much @jameshilliard!

Issue #2491, opened 2020-12-17
Setuptools does not pass config_settings through backend


di:
The `build` project now has a `--config-setting` flag to pass options to the backend:

```
$ python -m build --help
usage: python -m build [-h] [--version] [--sdist] [--wheel] [--outdir dir] [--skip-dependencies]
                       [--no-isolation] [--config-setting CONFIG_SETTING]
                       [srcdir]

- - - >8 - - -

  --config-setting CONFIG_SETTING, -C CONFIG_SETTING
                        pass option to the backend
 ```

I wanted to use this to set `wheel`'s `--plat-name` argument like so:

```
$ python -m build --wheel --config-setting plat-name=foo
```

I found that this worked as expected until the settings got to setuptool's `_build_with_temp_dir` function:

https://github.com/pypa/setuptools/blob/d368be2b5be8676ba8a3d55045b45b55090f6982/setuptools/build_meta.py#L190-L200

In this example, `_build_with_temp_dir` gets called with `config_settings={'plat-name': 'foo'}`, which then gets 'fixed' into `config_settings={'plat-name': 'foo', '--global-option': []}`, and then setuptools ignores everything in `config_settings` and only considers `--global-option`.

It almost seems like `--global-option` itself could be used for this, but unfortunately it doesn't work as it's value is passed to _lots_ of things that don't accept the arguments I'm hoping to pass:

```
$ python -m build --wheel --config-setting="--global-option=--plat-name" --config-setting="--global-option=foo" -n
usage: _in_process.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: _in_process.py --help [cmd1 cmd2 ...]
   or: _in_process.py --help-commands
   or: _in_process.py cmd --help

error: option --plat-name not recognized
...
```

I created a commit that "fixes" this for me as a proof of concept: https://github.com/pypa/setuptools/commit/fc95b3b83d6d5b561dc0a356995edf4c99785a6f

Possibly related issues:
* this is very similar to the request in https://github.com/pypa/setuptools/issues/1816, but @gaborbernat is asking for the ability to pass options _before_ the setup command, rather than after.
* https://github.com/pypa/setuptools/issues/1928 may also be related

jaraco:
[...]

PR #4218, opened 2024-02-13
Revert build option only bdist wheel

jameshilliard:
## Summary of changes

Reverts https://github.com/pypa/setuptools/pull/4217 which broke the ability to pass `build_ext` arguments via config args.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -289,6 +289,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
             *sys.argv[:1],
             *self._global_args(config_settings),
             "egg_info",
+            *self._arbitrary_args(config_settings),
         ]
         try:
             with Distribution.patch():
@@ -376,7 +377,6 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
         # Build in a temporary directory, then copy to the target.
         os.makedirs(result_directory, exist_ok=True)
         temp_opts = {"prefix": ".tmp-", "dir": result_directory}
-
         with tempfile.TemporaryDirectory(**temp_opts) as tmp_dist_dir:
             sys.argv = [
                 *sys.argv[:1],
@@ -384,6 +384,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
                 *setup_command,
                 "--dist-dir",
                 tmp_dist_dir,
+                *self._arbitrary_args(config_settings),
             ]
             with no_install_setup_requires():
                 self.run_setup()
@@ -402,10 +403,7 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
     ):
         with suppress_known_deprecation():
             return self._build_with_temp_dir(
-                ['bdist_wheel', *self._arbitrary_args(config_settings)],
-                '.whl',
-                wheel_directory,
-                config_settings,
+                ['bdist_wheel'], '.whl', wheel_directory, config_settings
             )
 
     def build_sdist(self, sdist_directory, config_settings=None):
--- a/setuptools/tests/test_build_meta.py
+++ b/setuptools/tests/test_build_meta.py
@@ -706,6 +706,25 @@ class TestBuildMetaBackend:
         for file in files:
             assert file.is_symlink() or os.stat(file).st_nlink > 0
 
+    @pytest.mark.filterwarnings("ignore::setuptools.SetuptoolsDeprecationWarning")
+    # Since the backend is running via a process pool, in some operating systems
+    # we may have problems to make assertions based on warnings/stdout/stderr...
+    # So the best is to ignore them for the time being.
+    def test_editable_with_global_option_still_works(self, tmpdir_cwd):
+        """The usage of --global-option is now discouraged in favour of --build-option.
+        This is required to make more sense of the provided scape hatch and align with
+        previous pip behaviour. See pypa/setuptools#1928.
+        """
+        path.build({**self._simple_pyproject_example, '_meta': {}})
+        build_backend = self.get_build_backend()
+        assert not Path("build").exists()
+
+        cfg = {"--global-option": ["--mode", "strict"]}
+        build_backend.prepare_metadata_for_build_editable("_meta", cfg)
+        build_backend.build_editable("temp", cfg, "_meta")
+
+        self._assert_link_tree(next(Path("build").glob("__editable__.*")))
+
     def test_editable_without_config_settings(self, tmpdir_cwd):
         """
         Sanity check to ensure tests with --mode=strict are different from the ones
@@ -720,7 +739,13 @@ class TestBuildMetaBackend:
         build_backend.build_editable("temp")
         assert not Path("build").exists()
 
-    @pytest.mark.parametrize("config_settings", [{"editable-mode": "strict"}])
+    @pytest.mark.parametrize(
+        "config_settings",
+        [
+            {"--build-option": ["--mode", "strict"]},
+            {"editable-mode": "strict"},
+        ],
+    )
     def test_editable_with_config_settings(self, tmpdir_cwd, config_settings):
         path.build({**self._simple_pyproject_example, '_meta': {}})
         assert not Path("build").exists()


abravalheri:
Hi @jameshilliard , thank you very much for the contribution. 

There is a problem of consensus here.
According to the discussion the solution of passing the arguments to all hooks also breakes the build process for some users in a different way...

The best here would be working towards a more prevalent solution.

jameshilliard:
> There is a problem of lack of consensus here.
> According to the previous discussion the solution of passing the arguments to all hooks also breakes the build process for some other users in a different way...

The impression I got from the [various threads](https://github.com/pypa/setuptools/issues/2491) is that changing the behavior of `--build-option` like this would cause a regression and it seems nobody from that thread noticed #4079 to comment on it. It's unclear why the behavior for this flag which which was previously working correctly for a number of use cases was changed to do something very different without any usable replacement option for those use cases.

abravalheri:
> The impression I got from the `various threads` is that changing the behavior of `--build-option` like this would cause a regression and it seems nobody from that thread noticed `#4079` to comment on it.

Hi @jameshilliard, I don't think this assessment is entirely accurate.

In the thread linked, there is a lot of people asking for a long term solution, which is the point being discussed in #4083.

When working on the error that prompted #4079, I proposed the patch (describing the approach I would use for it) in the same thread, so even if the participants did not had the time review #4079, they were informed about the proposed changes before they were applied.

Moreover, after the patch proposal, there was one comment "against", but my interpretation is that it was a comment in the nature "we need a proper/major fix or long-term solution" other than against the particular proposed patch (I did point that out in the thread).

In the same thread, I got feedback from one of `pip`'s maintainers, that agreed the patch would be good enough for a quick-and-dirty solution that does not solve the overall problem, but does improve the error being discussed.

#4079 makes it clear that the change is not the final solution for the `config_settings` saga, but just a stopgap. Handling `config_settings` is a hard challenge because the way PEP 517 runs each build hook in a brand new process and because evaluating `setup.py` imposes a strong level of decoupling between `setuptools.build_meta` and the rest of `setuptools`. #4083 aims at discussing this challenge and trying to find solutions for it.

---

Now, my problem with this PR is that it fixes problems for one group of users, but re-introduces problems for another group of users... We are between a rock and a hard place here, so I don't think adopting this PR exactly how it is right now would be beneficial, we would only be caught in the middle of a tug of war...

The best approach in my opinion is proceeding towards #4083, which would fix the problem for all groups of users.

---


### Suggested Workaround

Please also note that the way it is right now `config_settings` is at best supported in a `BETA`/best-case basis, for all intents and purposes... It is not something we can really consider an stable interface on the `setuptools` side.

The only known "stable" mechanism so far for passing configuration parameters for individual commands is via [`setup.cfg`](https://setuptools.pypa.io/en/latest/deprecated/distutils/configfile.html).

If no other tool is relying on `DIST_EXTRA_CONFIG`[^1], you can also use it to avoid patching an existing `setup.cfg` and instead add an extra one.

This is an arbitrary example that does not correspond to the reproducer mentioned in the PR original text, but gives a good idea on what is possible to do:

```ini
export DIST_EXTRA_CONFIG=/tmp/setuptools-build.cfg
echo -e '[build_ext]\nparallel = 8\n[bdist_wheel]\npy_limited_api = cp311' > $DIST_EXTRA_CONFIG
python -m build
```

[^1]: I know that `cibuildwheel` takes over `DIST_EXTRA_CONFIG` for their needs to communicate with `setuptools`.

jameshilliard:
> When working on the error that prompted #4079, I proposed the patch (describing the approach I would use for it) in the same thread, so even if the participants did not had the time review #4079, they were informed about the proposed changes before they were applied.

I'm mostly referring to [this comment](https://github.com/pypa/setuptools/issues/2491#issuecomment-1743297257) which seems to quite clearly indicate that this would be a significant regression.

In that comment a stopgap approach for not breaking existing use cases is suggested as well and it's unclear to me why that approach was not taken instead of this one which breaks existing use cases without providing a usable replacement for those use cases.

See:
> The easiest thing for setuptools to do in my mind is to introduce a set of “legacy” namespaced build options, one each for egg_info, sdist and wheel. Paper over the cracks while we hash out what is to become of setuptools’ config settings.

> In the same thread, I got feedback from one of pip's maintainers, that agreed the patch would be good enough for a quick-and-dirty solution that does not solve the overall problem, but does improve the error being discussed.

When it comes to python packaging there seems to be an issue where non-pip based use cases are an afterthought.

> Now, my problem with this PR is that it fixes problems for one group of users, but re-introduces problems for another group of users... We are between a rock and a hard place here, so I don't think adopting this PR exactly how it is right now would be beneficial, we would only be caught in the middle of a tug of war...

So why not just offer a new flag for the new behavior and preserve the previous behavior under the old flag as a stopgap?

> Please also note that the way it is right now `config_settings` is at best supported in a `BETA`/best-case basis, for all intents and purposes... It is not something we can really consider an stable interface on the `setuptools` side.

So setuptools warns quite loudly that directly invoking `setup.py` is deprecated, which kinda implies that one should be able to transition to pep517. This approach of deprecating/breaking features before having working replacements are ready has made transitioning to pep517 quite difficult in general for distros, this was a major issue with even having a pep517 frontend working in the first place as it wasn't originally possible to bootstrap a pep517 toolchain properly as it was incorrectly assumed everyone could just use pip.

> If no other tool is relying on DIST_EXTRA_CONFIG[1](https://github.com/pypa/setuptools/pull/4218#user-content-fn-1-8786f16451255739e7bef78631ca25e4), you can also use it[2](https://github.com/pypa/setuptools/pull/4218#user-content-fn-2-8786f16451255739e7bef78631ca25e4) to avoid modifying an existing setup.cfg and instead add an extra one.

Hmm, I'll give this a try and see if it's workable for us at all.

I'll reopen #4217 for now since it seems to at least partially reduce the breakage from this change.

layday:
> I'm mostly referring to [this comment](https://github.com/pypa/setuptools/issues/2491#issuecomment-1743297257) which seems to quite clearly indicate that this would be a significant regression.

I wasn't aware that `$DIST_EXTRA_CONFIG` existed when I made that comment.  You might be able to generate a throw-away config file using process substitution for a more direct alternative to config settings - unsure if that's supported by the env var.

abravalheri:
Hi @jameshilliard, thank you for taking the time to go through my comments and add your toughts.

I just want to highlight that I am definitely not against fixing this bug, and that when the patch you reported as problematic was introduced, to the best of my knowledge, it was not a given that it would end up causing exceptions (you were the first one to openly report crashes that to us, after a couple of months that the PR was merged). I honestly apologize for the inconvenience.

For the sake of increasing productivity and avoid further stress, the criteria that I am considering wheatear to accept PRs targeting the issue is:

- The PR should not introduce other problems (or re-introduce)
- The PR should not introduce significant changes in user interface that has not been previously discussed (e.g., adding a `config_settings["--temporary-legacy-options"]` would need to be first discussed).

... + the usual stuff (pass the tests, add relevant test cases, etc...)

---

> I'm mostly referring to [this comment](https://github.com/pypa/setuptools/issues/2491#issuecomment-1743297257) which seems to quite clearly indicate that this would be a significant regression.

You probably already saw this, but these concerns and viability were discussed in:

https://github.com/pypa/setuptools/issues/2491#issuecomment-1743723820
https://github.com/pypa/setuptools/issues/2491#issuecomment-1745126670

> In that comment a stopgap approach for not breaking existing use cases is suggested as well and it's unclear to me why that approach was not taken instead of this one which breaks existing use cases without providing a usable replacement for those use cases.

You mentioned a proposed stopgap approach that would not break existing use cases. Please note that the proposal was not accompanied by a PR and that adding more namespaces/flags/options/structure to the interface would require more discussion... My honest interpretation of this comment was: "this is not enough users have the need for passing `config_settings` to all build hooks"; to which my line of thought was: "look passing `config_settings` to all hooks is already broken right now, it does not work. Not doing anything does not help. Let's make `--global-option` and `--build-option` partially work as they were intended to be legacy backward compatible with `pip`, while we keep looking for a long-term solution".

> When it comes to python packaging there seems to be an issue where non-pip based use cases are an afterthought.

This is a very fair comment, and I agree with that. Ideally we should involve people that have stakes of different use cases, so everyone has a say.

In Setuptools we are open to work with everyone and we try to find compromises that also would work for non-pip based use cases. There were changes in both `pypa/distutils` and `pypa/setuptools` code bases that were prompted and developed together with people from the fedora/debian/arch/nix/conda/... communities. We just ask everyone to please bear with us because `setuptools`/`distutils` is complex.

In this particular instance, it is clear that both `config_settings["--build-option"]` and `config_settings["--global-option"]` were added as a temporary backward-compatibility hack to interoperate with `pip`'s legacy options. Not only they use the same name as the `pip`'s options, but also they are hilariously convoluted to write in the CLI, which cannot be justified otherwise. So for better or worse, `pip` is the primary target. What happens is that people have other needs that, so out of necessity they have to resort to using this interface as well.

But equally in other places in the code base, you can find compatibility specific hacks for non-`pip` use cases.

> So why not just offer a new flag for the new behavior and preserve the previous behaviour under the old flag as a stopgap?

Compounding more fragile APIs on top of an already fragile API does not sound like a good idea to me.

I might to sound like a broken record but: it was not immediately a given that the change would break other workflows. It took around 3 months for us to be reported otherwise - it there was an immediate report before `pip` have changed their code base, I would probably have reverted the change on the spot.

The reason why we are having the discussion in the first place is because `config_settings` implementation when PEP 517 was first introduced in setuptools was seriously lacky, if not completely unusable. In hindsight, probably it would be best to add a warning every time that `config_settings` was passed as a non-empty dict, so people should be at least aware that they are not stable (if anyone is interested on that, please feel free to open a PR).

> So setuptools warns quite loudly that directly invoking setup.py is deprecated, which kinda implies that one should be able to transition to pep517. This approach of deprecating/breaking features before having working replacements are ready has made transitioning to pep517 quite difficult in general for distros, this was a major issue with even having a pep517 frontend working in the first place as it wasn't originally possible to bootstrap a pep517 toolchain properly as it was incorrectly assumed everyone could just use pip.

Please note that the whole transition to PEP 517 was extremely difficult to setuptools itself as well. We still have problems nowadays with it, and we still have spend tons of energy on it. We are all kind of in the same boat.

Since deprecation warnings and documentation changes are the only means setuptools have to let their users know that something will change at some point in the future, our approach is to add them as soon as possible. However, we still keep deprecated features around for months in a row (and more often *years*) to avoid major breaks...

> I'll reopen https://github.com/pypa/setuptools/pull/4217 for now since it seems to at least partially reduce the breakage from this change.

Thank you very much. I think that is a good idea. If we can fix this problem without re-introducing other issues, that would be the best.

abravalheri:
We probably can close this now, in favour of #4217.

PR #4228, opened 2024-02-19
Avoid using private `_config_vars`

Avasam:
## Summary of changes
This PR attempts to avoid using a private global variables, to instead use the available public API. This resolves a `noqa` and upcomming `type: ignore` in https://github.com/pypa/setuptools/pull/3979/files#diff-820ceee125a584dd996725a342b2aaec47c5e96399faf426348dfeebe242ea4bR31  
All 3 implementations of `get_config_vars` (in `sysconfig`, `distutils.sysconfig`, and `setuptools._distutils.sysconfig`) return the global dictionary directly (not a copy), so usage in `build_ext` should be equivalent.
`get_config_vars` also already ensures `_config_vars` is initialized

I'm not sure about the test change however. 

### Pull Request Checklist
- [x] Changes have tests
- [ ] News fragment added in [`newsfragments/`]. (I don't think this counts as user-facing changes?)

[PR docs]:

### DIFF
--- a/setuptools/_distutils/tests/test_util.py
+++ b/setuptools/_distutils/tests/test_util.py
@@ -33,7 +33,7 @@ def environment(monkeypatch):
     monkeypatch.setattr(os.path, 'join', os.path.join)
     monkeypatch.setattr(os.path, 'isabs', os.path.isabs)
     monkeypatch.setattr(os.path, 'splitdrive', os.path.splitdrive)
-    monkeypatch.setattr(sysconfig, '_config_vars', copy(sysconfig._config_vars))
+    monkeypatch.setattr(sysconfig, 'get_config_vars', sysconfig.get_config_vars)
 
 
 @pytest.mark.usefixtures('save_env')
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -8,7 +8,7 @@ from pathlib import Path
 
 from distutils.command.build_ext import build_ext as _du_build_ext
 from distutils.ccompiler import new_compiler
-from distutils.sysconfig import customize_compiler, get_config_var
+from distutils.sysconfig import customize_compiler, get_config_var, get_config_vars
 from distutils import log
 
 from setuptools.errors import BaseError
@@ -26,7 +26,8 @@ except ImportError:
 
 # make sure _config_vars is initialized
 get_config_var("LDSHARED")
-from distutils.sysconfig import _config_vars as _CONFIG_VARS  # noqa
+
+_CONFIG_VARS = get_config_vars()
 
 
 def _customize_compiler_for_shlib(compiler):
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -24,9 +24,6 @@ try:
 except ImportError:
     _build_ext = _du_build_ext
 
-# make sure _config_vars is initialized
-get_config_var("LDSHARED")
-
 _CONFIG_VARS = get_config_vars()
 
 


setuptools/_distutils/tests/test_util.py
```diff
@@ -33,7 +33,7 @@ def environment(monkeypatch):
     monkeypatch.setattr(os.path, 'join', os.path.join)
     monkeypatch.setattr(os.path, 'isabs', os.path.isabs)
     monkeypatch.setattr(os.path, 'splitdrive', os.path.splitdrive)
-    monkeypatch.setattr(sysconfig, '_config_vars', copy(sysconfig._config_vars))
+    monkeypatch.setattr(sysconfig, 'get_config_vars', sysconfig.get_config_vars)
```
abravalheri, line 36: Changes in `setuptools/_distutils/tests/test_util.py` are overwritten every time `setuptools` syncs up with `pypa/distutils`, so probably they have to be proposed in the `pypa/distutils` repository first.

abravalheri:
> All 3 implementations of `get_config_vars` (in `sysconfig`, `distutils.sysconfig`, and `setuptools._distutils.sysconfig`) return the global dictionary directly (not a copy), so usage in `build_ext` should be equivalent.
> `get_config_vars` also already ensures `_config_vars` is initialized

Hi @Avasam, I don't understand this comment, could you please expland a little bit...

Isn't `disutils.get_config_vars` using `copy()` of `sysconfig.get_config_vars()` in https://github.com/pypa/distutils/blob/e651e536ef1fbafbe48856815602127df0371835/distutils/sysconfig.py#L539 ?

setuptools/_distutils/tests/test_util.py
Avasam: Thanks! I can go make the changes upstream to the right place.

Avasam:
> Isn't `disutils.get_config_vars` using `copy()` of `sysconfig.get_config_vars()` in [pypa/distutils@`e651e53`/distutils/sysconfig.py#L539](https://github.com/pypa/distutils/blob/e651e536ef1fbafbe48856815602127df0371835/distutils/sysconfig.py#L539) ?

Only if it was `None` (ie: not already initialized), in which case setuptools would've failed anyway with an `AttributeError`

> Also this is an implementation detail that can change at any point, right?
Possibly? But setuptools is also in control of that now that distutils isn't part of the stdlib. (and I doubt a security update to Python 3.8 - 3.11 would change this behaviour)

> Other than that, `build_ext` is clearly attempting to modify the global state in the `_customize_compiler_for_shlib`... As ugly as an attempt of using a private variable may look, my personal opinion is the following:
> 
> * The code still reads more clear and intentional to me the way it is now: I can clearly see that `build_ext` is doing something shady and that `_customize_compiler_for_shlib` only works because it patches the internals of `distutils`[1](#user-content-fn-1-9ba6ff18fc76016ea5fc216464a4ed9d).
> * If at some point `distutils` stops using `_config_vars`, the tests will fail immediately when we try to access an attribute that is no longer defined, and we will be forced to re-evaluate the patching technique. By using a public API, we loose this "free sanity check".

Whilst I'm challenging the use of a variable that is both private and not exposed in stubs, if you truly believe this is what setuptools should be doing, then I can expose `_config_vars` in the distutils stubs to help type checkers in #3979 & #4192 . This conversation will serve as the rationale for exposing it in typeshed.

abravalheri:
Thank you very much @Avasam for having a deeper look on this.

> Whilst I'm challenging the use of a variable that is both private and not exposed in stubs, if you truly believe this is what setuptools should be doing, then I can expose _config_vars in the distutils stubs to help type checkers.

My personal opinion is that `setuptools` is doing something shady and using the public API of `distutils` would only give a false sense of security.

The reality is that the patch is entangled to depend on that particular private variable... The moment the implementation changes everything falls apart. By importing the private attribute, at least we have a small "sanity check" giving hints about the global change in the state being necessary for the patching to have the desirable effect.

Unless Jason have requested this change, I think we can just be happy with our lives and add an `type: ignore` comment where it is needed. We probably don't have to change the `distutils` stubs (because frankly we are doing something shady anyway). Or if we really want to put more effort on this, add a `cast` to "reinterpret" the type somewhere inside of the file?

Avasam:
A type-ignore with an explicit type declaration (or cast if it's really needed) with a documented resolution is fine by me.

Avasam:
CLosed by 55eeabd46fb7f1d4d6796a66fa29150b3aa47153

PR #4233, opened 2024-02-23
Using vendored distutils._modified directly in build_clib

Avasam:
## Summary of changes
Noticed here https://github.com/pypa/setuptools/pull/3979#discussion_r1498494873
>  `_modified` [doesn't exist in stdlib](https://github.com/python/cpython/tree/3.11/Lib/distutils) (it was added in the vendored distutils). And `distutils.dep_util.newer_pairwise_group` doesn't exist either. So I don't see why `setuptools.modified` or `.._distutils._modified` couldn't always be used directly.

Work towards #2345 by having 1 less untyped import

### Pull Request Checklist
- [x] Changes have tests (existing `SETUPTOOLS_USE_DISTUTILS` tests should pass)

[PR docs]:

### DIFF
--- a/setuptools/command/build_clib.py
+++ b/setuptools/command/build_clib.py
@@ -3,7 +3,7 @@ from distutils.errors import DistutilsSetupError
 from distutils import log
 
 # Using vendored version directly because distutils._modified doesn't exist in stdlib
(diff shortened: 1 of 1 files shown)


Avasam:
Well this didn't work the way I thought it would.

Issue #3810, opened 2023-02-07
[debt] Many ResourceWarnings when `PYTHONWARNDEFAULTENCODING = 1`


jaraco:
In jaraco/skeleton, I've rolled out PYTHONWARNDEFAULTENCODING=1. Merging that change into Setuptools causes the test suite to fail due to `pytest.ini:filterwarnings[error]`.

Avasam:
The rule is still in "preview", but this can be checked for setuptool's own code using https://docs.astral.sh/ruff/rules/unspecified-encoding/

Avasam:
Now that Python 3.9 support is dropped, there's a comment linking to this issue that may be addressable:
https://github.com/pypa/setuptools/blob/84ed5913724df5a12dc804e1d5efe12508e706d2/pytest.ini#L11-L13
```ini
	# Workarounds for pypa/setuptools#3810
	# Can't use EncodingWarning as it doesn't exist on Python 3.9.
	# These warnings only appear on Python 3.10+
```

PR #4234, opened 2024-02-23
Massively reduce `EncodingWarning` spam in tests

Avasam:
## Summary of changes
~~Explicitly add `encoding=locale.getpreferredencoding(False)` to files opened as string. Starting Python 3.10 this can become `encoding="locale"`~~
~~Done by running `ruff check . --select=PLW1514 --fix --unsafe-fixes` (no manual code change)~~
~~I also added the check to Ruff config~~ 
Edit: This was changed by a constant that's `"locale"` when supported. `PLW1514` also wasn't added to the Ruff config because it requires the preview flag which itself causes new lint errors unrelated to this PR

I didn't go for "utf-8" because it *could* be a breaking change for some projects on Windows.

Works toward https://github.com/pypa/setuptools/issues/3810#issuecomment-1830794636 (I don't think it closes, idk if there's more warnings, probably from vendored packages)

### Pull Request Checklist
- [x] Changes have tests (all existing tests should pass, but with a lot less warnings now)

[PR docs]:


abravalheri (commented):
Hi @Avasam, I would like to thank you for all the contributions you are doing to `setuptools`. You have been very nice to the project and put a lot of effort in tackling very hard and time consuming tasks.

---

My minor comment about this are the following:

- When dealing with `.pth` files specifically, the advice we got from CPython maintainers is to omit the `encoding` parameter and ignore the warning specifically, instead of `getpreferredencoding`: https://github.com/python/cpython/issues/77102#issuecomment-1670160349.

   I while ago I left a [comment](https://github.com/pypa/setuptools/blob/v69.1.0/setuptools/command/editable_wheel.py#L531-L539) in `editable_wheel` to remind myself about that in the future.

- `.py` files should always use `UTF-8` right? (Because we only support Python3)

- In the `tests` folder, for everything else that is not `.pth` or is explicitly trying to test encoding, we probably can risk `utf-8`.

Avasam:
Alright so doing this made me notice another issue that leads me to believe ignoring these warnings is the only non-breaking way to stop to spam in tests (until Python 3.9 support is dropped so `encoding="locale"` can be used everywhere):
In diffcov (which runs Python 3.12), the warning is simply changed to `EncodingWarning: UTF-8 Mode affects locale.getpreferredencoding(). Consider locale.getencoding() instead`. But since I'm only using `getpreferredencoding` for `open`, may as well use `encoding="locale"` at that point. (both available at Python 3.10 iirc). Which defeats the purpose of my attempted changes here.

> * In the `tests` folder, for everything else that is not `.pth` or is explicitly trying to test encoding, we probably can risk `utf-8`.

I was thinking that too. Some or most tests could probably still be changed to use "utf-8" directly. But we could also just blanket ignore this warning (if that's doable) given the special situation setuptools always has. And it would achieve the same result of improving test results readability.

> * `.py` files should always use `UTF-8` right? (Because we only support Python3)

I have come across python files with non-utf-8 characters in the wild, specifically this one with Windows 1252 chars that I remember tripping up Ruff:
https://github.com/mhammond/pywin32/blob/main/Pythonwin/pywin/test/_dbgscript.py

And pywin32 also has these files with a different top-of-the-file encoding declaration:
https://github.com/mhammond/pywin32/blob/main/com/win32com/demos/iebutton.py
https://github.com/mhammond/pywin32/blob/main/com/win32com/demos/ietoolbar.py

```py
>>> open("./Pythonwin/pywin/test/_dbgscript.py").readlines()
['# -*- coding: latin-1 -*-\n', '#\n', '# Script for testing pywin.debugger & interactive exec from test_pywin\n', '\n', '# Umlauts for encoding test: áéúäöü\n', '\n', 'aa = 11\n', 'aa = 22\n', '\n', '\n', 'class CC:\n', '    cc = 44\n', '\n', '\n', 'def ff(bb=55):\n', '    global aa\n', '    aa = 77\n', '    return aa + bb\n', '\n', '\n', 'ff()\n', 'aa = 33\n']
>>> open("./Pythonwin/pywin/test/_dbgscript.py", encoding="utf-8").readlines()
  File "C:\Program Files\Python39\lib\codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe1 in position 129: invalid continuation byte
```

Granted these are Python 2 & Internet Explorer era files, and the one with non-UTF-8 chars is in a test. But apparently they exist?

Avasam:
All tests passes, and you can see using the diffcov test (becuase they only appear in failing test) that there's only 7 `EncodingWarning` left down from over 100! (github search in action page stopped counting after 100).

That doesn't mean this is done though, I still need to:
1. Validate with you, the maintainers, whether the name and location of the new utilities is good
2. Manually search the code for `.pth` to try to set the `encoding` explicitly to `None` (unless you tell me that passing tests are enough)
3. Should I use `encoding_for_open_for_mode` *everywhere* where the `mode` argument to `open` is dynamic and could potentially be `b` (bytes read/write)? Rn I only added it where tests fail.

pytest.ini
```diff
@@ -14,17 +14,21 @@ filterwarnings=
...
-	# pypa/build#615
-	ignore:'encoding' argument not specified::build.env
+	# pytest-dev/pytest # TODO: Raise issue upstream 
+	ignore:'encoding' argument not specified::_pytest
 
-	# dateutil/dateutil#1284
-	ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning:dateutil.tz.tz
+	# pypa/distutils # TODO: Raise issue upstream 
+	ignore:'encoding' argument not specified::distutils
+	ignore:'encoding' argument not specified::setuptools._distutils
```
jaraco, line 31: These changes shouldn't go in "## upstream" - those are handled in [jaraco/skeleton](/jaraco/skeleton).  Let's add any setuptools-specific ones below "## end upstream".

pytest.ini
```diff
@@ -14,17 +14,21 @@ filterwarnings=
...
-	ignore:'encoding' argument not specified::platform
-
-	# pypa/build#615
-	ignore:'encoding' argument not specified::build.env
+	# pytest-dev/pytest # TODO: Raise issue upstream 
+	ignore:'encoding' argument not specified::_pytest
 
-	# dateutil/dateutil#1284
-	ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning:dateutil.tz.tz
+	# pypa/distutils # TODO: Raise issue upstream
```
jaraco, line 29: See pypa/distutils#150 (pending merge into Setuptools).

setuptools/compat/encoding.py
```diff
@@ -0,0 +1,31 @@
+import sys
+from typing import Optional
+
+
+locale_encoding = "locale" if sys.version_info >= (3, 10) else None
```
jaraco, line 5: It's my understanding that `encoding='locale'` is essentially locking in the legacy behavior, which is UTF-8 on most systems, but cp1252 or similar on Windows. I believe the intention of the EncodingWarnings is to [eventually default to UTF-8](https://peps.python.org/pep-0597/#prepare-to-change-the-default-encoding-to-utf-8) regardless of the platform or environment. It would be my preference too if Setuptools could adopt UTF-8 and avoid the platform variance or at the very least minimize the variance to places where it's absolutely necessary.
When I made similar changes to distutils (https://github.com/pypa/distutils/issues/232), I simply hard-coded "utf-8".
I still need to absorb the discussion, which I haven't the time right now, but I wanted to get this review in.

jaraco (changes requested):
Thanks so much for doing this work. I know it's a lot.

I'd like to explore adopting the future default UTF-8 if possible.

setuptools/compat/encoding.py
Avasam: Your understanding of `encoding='locale'` seems to align with mine. As a point of precision for this PR's goal: It's not to be a long-term fix that will be compatible with Python 3.15 (current ETA of UTF-8 as default), but rather just to stop the warning spam whilst keeping the current behaviour + backwards compatibility (no possible breaking change) and keeping it easy to find where the encoding is set explicitly for an eventual update to UTF-8.

As far as going full utf-8 goes, @abravalheri suggested starting with tests, where we couldn't directly affect any user of setuptools.

pytest.ini
Avasam: Since `setuptools._distutils` is vendored, I understood "upstream" as "this warning is raised from code that would need to be changed upstream".

pytest.ini
Avasam: I'm not sure how that PR relates to the following EncodingWarnings:

```
/home/runner/work/setuptools/setuptools/setuptools/_distutils/dist.py:398: EncodingWarning: 'encoding' argument not specified
  parser.read(filename)
    
/home/runner/work/setuptools/setuptools/setuptools/_distutils/text_file.py:118: EncodingWarning: 'encoding' argument not specified
  self.file = open(self.filename, errors=self.errors)
```

Maybe you meant https://github.com/pypa/distutils/pull/236 or https://github.com/pypa/distutils/issues/232 ?

setuptools/compat/encoding.py
jaraco: If the goal is to suppress the warnings, I would simply suppress them in the pytest.ini rather than coding in the less desirable behavior. At the very least, if this code is undesirable/transitional, I'd like to see a comment (probably linked to an open issue) indicating that there's anticipated work.

Avasam: I discovered pretty late into doing this how the `pytest.ini` file could be used to suppress these warnings. If you're fine suppressing them there, with an appropriate comment, to make current failed tests more readable, then I'll try to make it work that way instead.

Avasam: Something like this https://github.com/pypa/setuptools/pull/4255 ?
If so, I'll immediately close this PR in favor of that one.

abravalheri:
In https://github.com/pypa/setuptools/pull/4261, I am testing out `"utf-8"` for `setuptools/tests` discussed in https://github.com/pypa/setuptools/pull/4234#pullrequestreview-1897799732.

Avasam:
As mentioned in https://github.com/pypa/setuptools/pull/4255#discussion_r1518358144 , the vast majority of the spam has been greatly reduced. And about half of what's left should be fixed upstream (and can be ignored by that PR). Let's try to use UTF-8 where applicable for non-test changes.

PR #4244, opened 2024-02-28
Return an empty `list` by default in `pkg_resources.ResourceManager.cleanup_resources`

Avasam:
## Summary of changes

Subclasses of are expected to return a `list`. There's no mention of this method potentially returning `None` in the docstring. If this is intended, then the docstring should be updated, and the return type be `Optional[list[str]]`

Work towards #2345 and merging `pkg_resources`

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


abravalheri:
Thank you very much @Avasam.

PR #4245, opened 2024-02-28
Prevent `pkg_resources._find_adapter` from ever returning `None`

Avasam:
## Summary of changes

Prevent `pkg_resources._find_adapter` from ever returning `None`, whilst keeping the exact same public API behaviour (the message of the error changes, but the error will be the same).
Doesn't immediately fix anything, but will be necessary to type-check `pkg_resources` w/o errors, which will be needed to validate annotations from merging typeshed stubs.

### Pull Request Checklist
- [x] Changes have tests (all existing test should pass as-is, type-checking tests are coming in a different PR)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -3207,7 +3207,9 @@ def _find_adapter(registry, ob):
     for t in types:
         if t in registry:
             return registry[t]
-    return None
+    # _find_adapter would previously return None, and immediatly be called.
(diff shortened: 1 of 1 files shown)


abravalheri:
I see... Previously you would have a situation like:

```python
finder = _find_adapter(_distribution_finders, importer) # previously could return `None`
return finder(importer, path_item, only)  # `None` is not callable, so this can raise `TypeError`
```

In the end of the day there is a `TypeError` anyway... So it should be fine.

Thank you very much @Avasam.

PR #4246, opened 2024-02-28
Merge and update variable and params annotations from typeshed

Avasam:
## Summary of changes

Merge and update (because they were not all correct or exact) variable and params annotations from typeshed. Missing public annotations are missing from typeshed too.
This PR purposefully avoids adding TypeVars, overloads and return types. Those will all be tackled in a follow-up PR. Return types are inferred at this step.

Step 3.1 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833 but only for `pkg_resources`

### Pull Request Checklist
- [x] Changes have tests (type-checking tests are coming right after #3979 )

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -108,6 +125,19 @@
...
+if TYPE_CHECKING:
+    from _typeshed import StrPath
+    from types import _LoaderProtocol
+
+# Type aliases
+_NestedStr = Union[str, Iterable[Union[str, Iterable["_NestedStr"]]]]
+_InstallerType = Callable[["Requirement"], Optional["Distribution"]]
+_PkgReqType = Union[str, "Requirement"]
+_EPDistType = Union["Distribution", _PkgReqType]
+_MetadataType = Optional["IResourceProvider"]
```
Avasam, line 137: These aliases come from the typeshed stub. Please lmk if you'd prefer different names

jaraco (approved):

pkg_resources/__init__.py
```diff
@@ -82,8 +104,8 @@
 
 # declare some globals that will be defined later to
 # satisfy the linters.
-require = None
-working_set = None
+require = cast("Callable[..., Any]", None)
+working_set = cast("WorkingSet", None)
 add_activation_listener = None
 cleanup_resources = None
 resource_stream = None
```
Avasam, line 111: I feel like the way it's been done in
https://github.com/pypa/setuptools/pull/4242/files#diff-e3d3d454fa3a072c9f46f8affa27513892fbc2d245e87f57a5927a7be851de05R3332-R3349
is better than casting all these with forward references and redefining callable types.

pkg_resources/__init__.py
```diff
@@ -1660,7 +1739,10 @@ def _parents(path):
 class EggProvider(NullProvider):
     """Provider based on a virtual filesystem"""
 
-    def __init__(self, module):
+    egg_name: str
+    egg_info: str
+
```
Avasam, line 1744:
```suggestion
```
Actually there's no assurance that an egg is even found.

Avasam:
As draft until https://github.com/pypa/setuptools/pull/4262 is completed

pkg_resources/__init__.py
```diff
@@ -2023,7 +2114,7 @@ def __init__(self, importer):
 ] = _declare_state('dict', '_distribution_finders', {})
 
 
-def register_finder(importer_type, distribution_finder):
+def register_finder(importer_type: type, distribution_finder: _AdapterType):
```
abravalheri, line 2117: Instead of `_AdapterType`, it would be more "sound" here to have a `callable that, passed a path item and the importer instance, yields ``Distribution`` instances found on that path item` or something that matches the same type as the `_distribution_finders` variable above, would it?

pkg_resources/__init__.py
```diff
@@ -2201,7 +2296,7 @@ def resolve_egg_link(path):
 )
 
 
-def register_namespace_handler(importer_type, namespace_handler):
+def register_namespace_handler(importer_type: type, namespace_handler: _AdapterType):
```
abravalheri, line 2299: Instead of `_AdapterType`, it would be more "sound" here to have a callable like this:

```python
        def namespace_handler(importer, path_entry, moduleName, module):
            # return a path_entry to use for child packages
```

wouldn't it?

pkg_resources/__init__.py
```diff
@@ -3183,7 +3333,7 @@ def _always_object(classes):
     return classes
 
 
-def _find_adapter(registry, ob):
+def _find_adapter(registry: Mapping[type, _AdapterType], ob: object):
```
abravalheri, line 3336: Here the way to go is probably have 3 definitions using `typing.overload` for the 3 different types of adapters and compatible `ob` (something like `_ProviderFactory`+ `_LoaderProtocol`; `_DistributionFider`+`_Importer`; `_NamespaceHandler`+`_Importer`, with `_Importer` a protocol that is compatible with `PathEntryFinder` and/or `MetaPathFinder`? ... considering the limitations in https://github.com/python/typeshed/issues/11541 and https://github.com/python/typeshed/issues/2468)

pkg_resources/__init__.py
```diff
@@ -108,6 +125,19 @@
...
+
+# Type aliases
+_NestedStr = Union[str, Iterable[Union[str, Iterable["_NestedStr"]]]]
+_InstallerType = Callable[["Requirement"], Optional["Distribution"]]
+_PkgReqType = Union[str, "Requirement"]
+_EPDistType = Union["Distribution", _PkgReqType]
+_MetadataType = Optional["IResourceProvider"]
+# Any object works, but let's indicate we expect something like a module (optionally has __loader__ or __file__)
+_ModuleLike = Union[object, types.ModuleType]
+_AdapterType = Callable[..., Any]  # Incomplete
```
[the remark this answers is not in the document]
Avasam: I plan on updating with:
```py
_ProviderFactoryType = Callable[[_ModuleLike], "IResourceProvider"]
_DistFinderType = Callable[[_T, str, bool], Iterable["Distribution"]]
_NSHandlerType = Callable[[_T, str, str, types.ModuleType], Optional[str]]
_AdapterT = TypeVar(
    "_AdapterT", _DistFinderType[Any], _ProviderFactoryType, _NSHandlerType[Any]
)
```

in a follow-up PR, but I can add it to this one if you'd like.

pkg_resources/__init__.py
```diff
@@ -353,7 +390,7 @@ class UnknownExtra(ResolutionError):
     """Distribution doesn't have an "extra feature" of the given name"""
 
 
-_provider_factories = {}
+_provider_factories: Dict[Type[_ModuleLike], _AdapterType] = {}
```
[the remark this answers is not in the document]
Avasam: See comment https://github.com/pypa/setuptools/pull/4246/files#r1596038268

pkg_resources/__init__.py
```diff
@@ -99,7 +123,20 @@
     stacklevel=2,
 )
 
+if TYPE_CHECKING:
+    from _typeshed import StrPath
+    from types import _LoaderProtocol
```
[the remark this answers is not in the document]
Avasam: That comes from typeshed. It's private to help prevent accidental runtime imports because typeshed didn't start using `@type_check_only` until very recently (and I believe we'll keep these names private anyway). In other words, `_LoaderProtocol` doesn't exist in the `types` module.

We could potentially ask to move `_LoaderProtocol` from `types.pyi` to `typeshed/__init__.pyi` and see what other typeshed maintainers think.

pkg_resources/__init__.py
```diff
@@ -363,7 +400,9 @@ class UnknownExtra(ResolutionError):
 DEVELOP_DIST = -1
 
 
-def register_loader_type(loader_type, provider_factory):
+def register_loader_type(
+    loader_type: Type[_ModuleLike], provider_factory: _AdapterType
```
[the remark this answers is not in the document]
Avasam: See comment https://github.com/pypa/setuptools/pull/4246/files#r1596038268
This would become
```py
def register_loader_type(
    loader_type: Type[_ModuleLike],
    provider_factory: _ProviderFactoryType,
) -> None: ...
```

pkg_resources/__init__.py
Avasam: See comment https://github.com/pypa/setuptools/pull/4246/files#r1596038268
This would become:
```py
def register_finder(
    importer_type: Type[_T], distribution_finder: _DistFinderType[_T]
) -> None:
```

pkg_resources/__init__.py
Avasam: See comment https://github.com/pypa/setuptools/pull/4246/files#r1596038268
This would become:
```py
def register_namespace_handler(
    importer_type: Type[_T],
    namespace_handler: _NSHandlerType[_T],
) -> None: ...
```

pkg_resources/extern/__init__.py
```diff
@@ -48,16 +56,22 @@ def load_module(self, fullname):
...
+    def find_spec(
+        self,
+        fullname: str,
+        path: Optional[Sequence[str]] = None,
+        target: Optional[ModuleType] = None,
+    ):
         """Return a module spec for vendored names."""
         return (
-            importlib.util.spec_from_loader(fullname, self)
+            # FIXME: VendorImporter doesn't subclass importlib.abc.Loader
```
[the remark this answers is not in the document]
Avasam:
```suggestion
            # https://github.com/python/typeshed/issues/11882
```
Whatever the result now linking to this issue and its resolution should be correct.

pkg_resources/__init__.py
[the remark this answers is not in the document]
abravalheri: If `_LoaderProtocol` is undocumented, I think we should not import it.

If `typeshed` maintainers are happy to make an utility definition of `LoaderProtocol` in [`_typeshed`](https://github.com/python/typeshed/tree/main/stdlib/_typeshed) accessible for public usage that would be great. Otherwise if the protocol is small, declaring the protocol inline should also be fine.

(Brain dump: probably having an `_typeshed.importlib` module that helps to address the concerns in https://github.com/python/typeshed/issues/11882#issuecomment-2104253881 would be even greater).

pkg_resources/__init__.py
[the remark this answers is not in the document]
abravalheri: If you are planning that as a follow-up PR, that should be fine.

Thank you very much.

Avasam:
@abravalheri in-code comments have been updated to reflect that fixes have been done upstream. I don't think we need to wait for the next mypy release. `type-ignore` can be removed when updating to mypy 1.11

abravalheri:
Thank you very much.

PR #4249, opened 2024-02-29
Make `pkg_resources.find_nothing` return an (empty) `Generator` of `Distribution`

Avasam:
## Summary of changes

This aligns with `find_eggs_in_zip` and `find_on_path` and ensures `find_distributions` always returns a generator.

This was noticed when attempting to type the `pkg_resoruces` package for #2345 

xref https://github.com/python/typeshed/pull/11512#discussion_r1508147634

### Pull Request Checklist
- [x] Changes have tests (existing tests should pass, type-checking tests will be added after #3979 )

[PR docs]:

### DIFF
--- a/newsfragments/4249.bugfix.rst
+++ b/newsfragments/4249.bugfix.rst
@@ -1,0 +1,1 @@
+Ensured ``pkg_resources.find_distributions`` always returns a `Generator` as per its docstring by making ``pkg_resources.find_nothing`` yield nothing rather than returning an empty `tuple` -- by :user:`Avasam`
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -27,7 +27,7 @@ import io
 import time
 import re
 import types
-from typing import Protocol
+from typing import Generator, Protocol
 import zipfile
 import zipimport
 import warnings
@@ -2089,8 +2089,11 @@ def find_eggs_in_zip(importer, path_item, only=False):
 register_finder(zipimport.zipimporter, find_eggs_in_zip)
 
 
-def find_nothing(importer, path_item, only=False):
-    return ()
+def find_nothing(
+    importer, path_item, only=False
+) -> Generator["Distribution", None, None]:
+    return
+    yield
 
 
 register_finder(object, find_nothing)


abravalheri:
Hi @Avasam, would it make more sense that instead of changing the implementation we just consider that the `distribution_finder` in the `register_finder` function returns as type an `Iterable` instead of `Generator`?

If we consider the implementation, that seems to be the requirement for the functions to work, isn't it?

The documentation probably uses the term `yield` with a more general way, rather than to direct imply the use of the `yield` keyword in Python.

Avasam:
@abravalheri Yes it makes perfect sense, I'd just update the docstrings as well. In fact going with `Iterable` is what I did in that typeshed PR comment.

abravalheri:
@Avasam would something like the following be an appropriate documentation change?

```diff
diff --git i/pkg_resources/__init__.py w/pkg_resources/__init__.py
index 625d1f83d..8ee988114 100644
--- i/pkg_resources/__init__.py
+++ w/pkg_resources/__init__.py
@@ -2049,7 +2049,7 @@ def register_finder(importer_type, distribution_finder):

     `importer_type` is the type or class of a PEP 302 "Importer" (sys.path item
     handler), and `distribution_finder` is a callable that, passed a path
-    item and the importer instance, yields ``Distribution`` instances found on
+    item and the importer instance, iterates over ``Distribution`` instances found on
     that path item.  See ``pkg_resources.find_on_path`` for an example."""
     _distribution_finders[importer_type] = distribution_finder
```

Avasam:
> @Avasam would something like the following be an appropriate documentation change?
> 
> ```diff
> diff --git i/pkg_resources/__init__.py w/pkg_resources/__init__.py
> index 625d1f83d..8ee988114 100644
> --- i/pkg_resources/__init__.py
> +++ w/pkg_resources/__init__.py
> @@ -2049,7 +2049,7 @@ def register_finder(importer_type, distribution_finder):
> 
>      `importer_type` is the type or class of a PEP 302 "Importer" (sys.path item
>      handler), and `distribution_finder` is a callable that, passed a path
> -    item and the importer instance, yields ``Distribution`` instances found on
> +    item and the importer instance, iterates over ``Distribution`` instances found on
>      that path item.  See ``pkg_resources.find_on_path`` for an example."""
>      _distribution_finders[importer_type] = distribution_finder
> ```

That's still not quite right, it "returns an Iterable of ``Distribution``" rather than "iterates over ``Distribution``"

Avasam:
Since the entire premise of this PR is wrong, and that the only change wouldn't result in a typing change anyway. I'll avoid doing a ship of Theseus out of it and I'll just finish typing the `register_*` methods with the established `Iterable` return type for distribution finders. After #4246 is merged.

Avasam:
Since the entire premise of this PR is wrong, and that the only change wouldn't result in a typing change anyway. I'll avoid doing a ship of Theseus out of it and I'll just finish typing the `register_*` methods with the established `Iterable` return type for distribution finders. After #4246 is merged.

PR #4251, opened 2024-03-03
Simplify FileList._safe_path for PathLike support

jaraco:
- **In FileList.append**
- **Refactor _safe_path into three functions. Remove support for bytes. Allow for path to be `PathLike`.**

## Summary of changes

Supports pypa/distutils#237.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
I wonder if anyone is still dependent on the support for bytes decoding. I guess we should either retain compatibility or declare the change as backward-incompatible. I'll do the latter; we can back out the change (restore support for bytes) if it turns out to cause any trouble.

PR #4257, opened 2024-03-05
Enable mypy on CI

Avasam:
## Summary of changes
Follow-up to #3979. Extracted from #4192. Enables mypy type-checking in the CI.

Example failure: https://github.com/pypa/setuptools/actions/runs/8158516102/job/22300634747?pr=4257#step:9:69

### Pull Request Checklist
- [x] Changes have tests (this is enabling the tests! 😄  )
- [ ] News fragment added in [`newsfragments/`]. (no user-facing change)

[PR docs]:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,10 @@ testing =
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
+	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
+	tomli
+	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
+	importlib_metadata
 
 testing-integration =
 	pytest
--- a/setuptools/config/_validate_pyproject/fastjsonschema_validations.py
+++ b/setuptools/config/_validate_pyproject/fastjsonschema_validations.py
@@ -1,5 +1,4 @@
 # noqa
-# type: ignore
 # flake8: noqa
 # pylint: skip-file
 # mypy: ignore-errors
--- a/mypy.ini
+++ b/mypy.ini
@@ -13,30 +13,21 @@ exclude = (?x)(
 	| ^setuptools/_distutils/ # Vendored
 	| ^setuptools/config/_validate_pyproject/ # Auto-generated
 	)
-disable_error_code = 
-	# TODO: Test environment is not yet properly configured to install all imported packages
-	# import-not-found, # This can be left commented out for local runs until we enforce running mypy in the CI
-	# TODO: Not all dependencies are typed. Namely: distutils._modified, wheel.wheelfile, and jaraco.*
-	import-untyped,
-	# Ignoring attr-defined because setuptools wraps a lot of distutils classes, adding new attributes,
-	# w/o updating all the attributes and return types from the base classes for type-checkers to understand
-	# Especially with setuptools.dist.command vs distutils.dist.command vs setuptools._distutils.dist.command
-	# *.extern modules that actually live in *._vendor will also cause attr-defined issues on import
-	attr-defined,
+# Ignoring attr-defined because setuptools wraps a lot of distutils classes, adding new attributes,
+# w/o updating all the attributes and return types from the base classes for type-checkers to understand
+# Especially with setuptools.dist.command vs distutils.dist.command vs setuptools._distutils.dist.command
+# *.extern modules that actually live in *._vendor will also cause attr-defined issues on import
+disable_error_code = attr-defined
 
-# Avoid raising issues when importing from "extern" modules, as those are added to path dynamically.
-# https://github.com/pypa/setuptools/pull/3979#discussion_r1367968993
-[mypy-pkg_resources.extern.*,setuptools.extern.*]
+# - Avoid raising issues when importing from "extern" modules, as those are added to path dynamically.
+#   https://github.com/pypa/setuptools/pull/3979#discussion_r1367968993
+# - distutils._modified has different errors on Python 3.8 [import-untyped], on Python 3.9+ [import-not-found]
+# - All jaraco modules are still untyped
+[mypy-pkg_resources.extern.*,setuptools.extern.*,distutils._modified,jaraco.*]
 ignore_missing_imports = True
 
-[mypy-pkg_resources.tests.*,setuptools.tests.*]
-disable_error_code =
-	# Tests include creating dynamic modules that won't exists statically before the test is run.
-	# Let's ignore all "import-not-found", as if an import really wasn't found, then the test would fail.
-	import-not-found,
-	# mmany untyped "jaraco" modules
-	import-untyped,
-
-# Mypy issue, this vendored module is already excluded!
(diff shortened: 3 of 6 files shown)


setup.cfg
```diff
@@ -47,9 +47,7 @@ testing =
 	pytest-cov; \
 		# coverage seems to make PyPy extremely slow
 		python_implementation != "PyPy"
-	pytest-mypy >= 0.9.1; \
-		# workaround for jaraco/skeleton#22
-		python_implementation != "PyPy"
+	pytest-mypy
```
Avasam, line 50: https://github.com/jaraco/skeleton/pull/111

tox.ini
```diff
@@ -7,7 +7,7 @@ setenv =
 	SETUPTOOLS_ENFORCE_DEPRECATION = {env:SETUPTOOLS_ENFORCE_DEPRECATION:0}
 	# ^-- Temporarily disable enforcement so CI don't fail on due dates
 commands =
-	pytest {posargs}
+	pytest --mypy {posargs}
```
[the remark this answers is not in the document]
Avasam: Oh is this it? https://github.com/pypa/setuptools/blob/main/pyproject.toml
How should I proceed? Copy the line you shared? Or does something vendored need to be pulled in?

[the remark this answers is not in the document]
abravalheri: I see, this is the reason:

https://github.com/pypa/setuptools/blob/v69.1.1/pyproject.toml#L8-L9

Removing these lines, we should not have to manually add this flag.

setuptools/config/_validate_pyproject/fastjsonschema_validations.py
```diff
@@ -1,5 +1,4 @@
 # noqa
-# type: ignore
```
Avasam, line 2: This is vendored code, but the fix is already upstream: https://github.com/abravalheri/validate-pyproject/pull/159

setup.cfg
abravalheri: Should we wait to see how that issue goes.

Even if we don't change this dependency, I believe the tests will still run on the CI (for everything else that is not PyPy).

setup.cfg
```diff
@@ -66,13 +64,16 @@ testing =
...
+	tomli
 	pytest-timeout
 	pytest-perf; \
 		# workaround for jaraco/inflect#195, pydantic/pydantic-core#773 (see #3986)
 		sys_platform != "cygwin"
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
+	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
+	importlib_metadata; python_version >= "3.12"
```
abravalheri, line 76: You mean in this line: https://github.com/pypa/setuptools/blob/v69.1.1/setuptools/_importlib.py#L12?

Is there any other place we import it directly?

Avasam: I do believe that's the only place.
Others indirect or similar imports include `from setuptools.extern import importlib_metadata` found a few lines lower
And this one in code template https://github.com/pypa/setuptools/blob/v69.1.1/setuptools/command/easy_install.py#L2141

The reason this line is needed for Python 3.12 is because for older versions `importlib_metadata` is a dependency of `build`, `twine` and `keyring` (itself a dependency of `jaraco.develop`, `twine`). Which seemingly all dropped it in 3.12.

You can't add a Python 3.12 check in code because users could have `importlib_metadata` on Python 3.12.
You can't type-ignore for a specific platform_version or os either.
Sensibly, the tests should install it to validate code that uses it.

The `; python_version >= "3.12"` part could be considered redundant.
```suggestion
	importlib_metadata
```

setup.cfg
Avasam: I don't think this change is necessary anymore since I'm no longer unconditionally adding the `--mypy` flag. As you said, it can be brought back from upstream later.
```suggestion
	pytest-mypy >= 0.9.1; \
		# workaround for https://github.com/jaraco/skeleton/issues/22
		python_implementation != "PyPy"
```

setup.cfg
abravalheri: > Sensibly, the tests should install it to validate code that uses it.

This is a good point. By installing `importlib_metadata` we are exercising more code paths in `disable_importlib_metadata_finder` and improving coverage.

setup.cfg
```diff
@@ -66,13 +66,16 @@ testing =
 	filelock>=3.4.0
 	ini2toml[lite]>=0.9
 	tomli-w>=1.0.0
+	tomli
```
abravalheri, line 69: I suppose that this is to cover https://github.com/pypa/setuptools/blob/0435c72fe989230f2549b15b0fc2b0a52f776e8f/setuptools/tests/integration/test_pip_install_sdist.py#L203?

If we are going for that approach, would it make more sense to remove the [`testing-integration`](https://github.com/pypa/setuptools/blob/0435c72fe989230f2549b15b0fc2b0a52f776e8f/setup.cfg#L77) extra and just use `testing` for everything?

Avasam: I don't see an issue with having less dependencies in `testing-integration`. And if you wanna merge them together that's up to you.

If you want to reduce duplication, then I think my latest commit should help with that (assuming it works)

abravalheri: Hi @Avasam, I was just asking your option about the matter, sorry if it prompted a change.

I think what we do in the PR is to just leave as you wrote it before: not touching `testing-integration` and just adding the extras in `testing`.

In a separated PR, I think we just remove the `testing-integration` and use `testing` everywhere...

Although it might add a bit of overhead for the integration tests, it will simplify and streamline the setup... This way we reduce the divergence with `skeleton`. The additional overhead in the integration tests should not be too much, and the integration tests just run before the releases anyway.

setup.cfg
```diff
@@ -41,29 +41,21 @@ exclude =
...
+
 	# upstream
-	pytest >= 6
 	pytest-checkdocs >= 2.4
 	pytest-cov; \
 		# coverage seems to make PyPy extremely slow
 		python_implementation != "PyPy"
 	pytest-mypy >= 0.9.1; \
-		# workaround for jaraco/skeleton#22
+		# workaround for https://github.com/jaraco/skeleton/issues/22
```
Avasam, line 52:
```suggestion
		# workaround for jaraco/skeleton#22
```
This line didn't need to change

setup.cfg
Avasam: > sorry if it prompted a change.

I actually have the deduplication commit already ready, so it's more of a coincidence ^^

> I think what we do in the PR is to just leave as you wrote it before

np, I'll revert that change and leave modifications to `setup.cfg` that are not absolutely necessary to another PR.

> The additional overhead in the integration tests should not be too much

Especially if the pypi downloads are cached anyway (are they?)

abravalheri: > Especially if the pypi downloads are cached anyway (are they?)

I don't remember now if the CI is keeping the pip cache 😅. But it should not matter that much.

I just hope someone is not installing `setuptools[testing-integration]`. But that is unlikely in my opinion.

abravalheri:
Thank you very much @Avasam!

PR #4258, opened 2024-03-06
Type `pkg_resources._declare_state` and make it work statically

Avasam:
## Summary of changes

After playing around with previous PRs with whether `_distribution_finders`, `_namespace_handlers` and `_namespace_packages` definitions should be moved closer to declarations, I noticed that, since we have to define the variables for other reasons anyway, there's no point to `_declare_state`'s global namespace shenanigans. This PR makes it work for static analysis and, after #4246 is merged, will reduce changes in #4242

### Pull Request Checklist
- [ ] Changes have tests (existing tests should pass, no intended behaviour change)
- [ ] News fragment added in [`newsfragments/`]. (even typing-wise there's no user-facing change here, it's all privates)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -27,7 +27,7 @@ import io
 import time
 import re
 import types
-from typing import List, Protocol
+from typing import Any, Dict, List, Protocol
 import zipfile
 import zipimport
 import warnings
@@ -94,9 +94,6 @@ iter_entry_points = None
 resource_listdir = None
 resource_filename = None
 resource_exists = None
-_distribution_finders = None
-_namespace_handlers = None
-_namespace_packages = None
 
 
 warnings.warn(
@@ -120,11 +117,10 @@ class PEP440Warning(RuntimeWarning):
 parse_version = packaging.version.Version
 
 
-_state_vars = {}
+_state_vars: Dict[str, Any] = {}
 
 
-def _declare_state(vartype, **kw):
-    globals().update(kw)
+def _declare_state(vartype: str, **kw: object) -> None:
     _state_vars.update(dict.fromkeys(kw, vartype))
 
 
@@ -2025,7 +2021,8 @@ class EggMetadata(ZipProvider):
         self._setup_prefix()
 
 
-_declare_state('dict', _distribution_finders={})
+_distribution_finders = {}
+_declare_state('dict', _distribution_finders=_distribution_finders)
 
 
 def register_finder(importer_type, distribution_finder):
@@ -2198,8 +2195,10 @@ if hasattr(pkgutil, 'ImpImporter'):
 
 register_finder(importlib.machinery.FileFinder, find_on_path)
 
-_declare_state('dict', _namespace_handlers={})
-_declare_state('dict', _namespace_packages={})
+_namespace_handlers = {}
+_declare_state('dict', _namespace_handlers=_namespace_handlers)
+_namespace_packages = {}
+_declare_state('dict', _namespace_packages=_namespace_packages)
 
 
 def register_namespace_handler(importer_type, namespace_handler):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -27,7 +27,7 @@ import io
 import time
 import re
 import types
-from typing import Any, Dict, List, Protocol
+from typing import Any, Callable, Dict, Iterable, List, Protocol, Optional
 import zipfile
 import zipimport
 import warnings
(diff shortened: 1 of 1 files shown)


abravalheri (approved):
Thank you very much @Avasam. I think this make sense for me, and it is not a significative change.

The downside is that it does add some "duplication", that the code base so far has been trying to avoid. However, the `_declare_state` DSL would not be very comfortable for declaring types anyway...

The advantage of this approach is that later we can use to improve the documentation, like proposed in https://github.com/pypa/setuptools/pull/4242/files#diff-e3d3d454fa3a072c9f46f8affa27513892fbc2d245e87f57a5927a7be851de05R100-R106 [^1], and that has already proven useful to clarify tricky parts like https://github.com/pypa/setuptools/pull/4249 and improve the understanding on how the code works.

Overall, I think that after we add the correct type hints [as in #4242](https://github.com/pypa/setuptools/pull/4242/files#diff-e3d3d454fa3a072c9f46f8affa27513892fbc2d245e87f57a5927a7be851de05R100-R106) we are going to gain more than we loose (better clarity/documentation vs. repetition).

(BTW, would it make sense to add the annotations in this PR since it is already introducing the variable declaration?) That would probably showcase the benefits of the approach even more.

---

@jaraco do you have any thoughts?



[^1]: Regarding your [comment](
https://github.com/pypa/setuptools/pull/4242/files#r1505157594) in #4242, I think it does make sense to keep the variable type declaration and `_declare_state` togheter outside of the `if TYPE_CHECKING`.

abravalheri:
@Avasam, I am assuming this change does not affect https://github.com/pypa/setuptools/blob/b4b622e2100f0c2ec739f61e1113a94c40e47466/pkg_resources/__init__.py#L3320, because https://github.com/pypa/setuptools/blob/b4b622e2100f0c2ec739f61e1113a94c40e47466/pkg_resources/__init__.py#L3340 is already ensuring `working_set` is a global variable, right?

Avasam:
> @Avasam, I am assuming this change does not affect
> 
> https://github.com/pypa/setuptools/blob/b4b622e2100f0c2ec739f61e1113a94c40e47466/pkg_resources/__init__.py#L3320
> 
> , because
> https://github.com/pypa/setuptools/blob/b4b622e2100f0c2ec739f61e1113a94c40e47466/pkg_resources/__init__.py#L3340
> 
> is already ensuring `working_set` is a global variable, right?

Yep, I validated for `working_set` too

> (BTW, would it make sense to add the annotations in this PR since it is already introducing the variable declaration?) That would probably showcase the benefits of the approach even more.

I know you've already approved the PR, but I added that change since it's kinda small anyway and doesn't change any additional lines. After #4246, as part of completely typing the `register_*` methods, I'll introduce some aliases (same that are already in typeshed) to avoid duplicating those `Callable` type definition

Avasam:
This change will also reduce errors in https://github.com/pypa/setuptools/pull/4246 to bring it closer to passing

pkg_resources/__init__.py
```diff
@@ -2200,8 +2199,12 @@ def resolve_egg_link(path):
...
 register_finder(importlib.machinery.FileFinder, find_on_path)
 
-_declare_state('dict', _namespace_handlers={})
-_declare_state('dict', _namespace_packages={})
+_namespace_handlers: Dict[
+    type, Callable[[object, str, str, types.ModuleType], Optional[str]]
+] = {}
+_declare_state('dict', _namespace_handlers=_namespace_handlers)
+_namespace_packages: Dict[Optional[str], List[str]] = {}
+_declare_state('dict', _namespace_packages=_namespace_packages)
```
abravalheri, line 2207: To be honest, given this approach of pre-declaring the global variables, it seems that `_declare_state` could be simplified to something like:

```python
_declare_state(vartype: str, varname: str) -> None:
    _state_vars[varname] = vartype
```

In turn that would allow us to simplify:

```python
_distribution_finders: Dict[
    type, Callable[[object, str, bool], Iterable["Distribution"]]
] = {}
_declare_state('dict', '_distribution_finders')
```

OR if we want to have a "oneliner"/"decorator-style-ish":

```python
_declare_state(vartype: str, varname: str, value: T) -> T:
    _state_vars[varname] = vartype
    return value

...

_distribution_finders: Dict[
    type, Callable[[object, str, bool], Iterable["Distribution"]]
] = _declare_state('dict', '_distribution_finders', {})

# it is a pity Python don't have an equivalent of decorators for assignment statements...
# (so that we wouldn't need to write the name of the variable as an argument).
```

But that is a refactoring that might require further discussion and maybe a separated PR, so just posting here for the sake of brain-storming.

abravalheri: TODO: I hope I remember this suggestion in the future :P

Avasam: I don't remember at which python version this becomes supported (with `from __future__ import annotations`), but one could even do:
```py
_distribution_finders = _declare_state(
    'dict',
    '_distribution_finders',
    dict[type, Callable[[object, str, bool], Iterable["Distribution"]]](),
)
```

Avasam: https://github.com/pypa/setuptools/pull/4346

PR #4262, opened 2024-03-06
`pkg_resources`: Runtime fixes to make typing annotations work

Avasam:
## Summary of changes

Should fix the last errors in #4246. These are all actual code changes, please review carefully.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -2825,7 +2825,7 @@ def activate(self, path=None, replace=False):
         if path is None:
             path = sys.path
         self.insert_on(path, replace=replace)
-        if path is sys.path:
+        if path is sys.path and self.location:
```
abravalheri, line 2828: I wonder if it is possible that `self.location` is an empty string...
The empty string is a valid entry for `sys.path` (and holds special meaning).

Avasam: Just in case:
```suggestion
        if path is sys.path and self.location is not None:
```

pkg_resources/__init__.py
```diff
@@ -2492,7 +2492,7 @@ def resolve(self):
             raise ImportError(str(exc)) from exc
 
     def require(self, env=None, installer=None):
-        if self.extras and not self.dist:
+        if not self.dist:
```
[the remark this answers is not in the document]
Avasam: The issue here is the code `self.dist.requires` below, which isn't type-safe id `self.dist` is potentially `None`.
The change of behaviour is that now a `UnknownExtra` will be raised if you have no distribution. Which I guess can be weird if you have no extras. The message is still appropriate though. A change like this would keep the `AttributeError` from before, be type-safe, explicit about value validation, and have an appropriate error message. (from the user's pov, only the message would change)
```suggestion
        if not self.dist:
            error_cls = UnknownExtra if self.extras else AttributeError
            raise error_cls("Can't require() without a distribution", self)
```

pkg_resources/__init__.py
```diff
@@ -1636,7 +1636,7 @@ def _validate_resource_path(path):
         )
 
     def _get(self, path) -> bytes:
-        if hasattr(self.loader, 'get_data'):
+        if self.loader and hasattr(self.loader, 'get_data'):
```
Avasam, line 1639:
```suggestion
        if hasattr(self.loader, 'get_data') and self.loader:
```
Micro-optimisation. `hasattr` was already safe enough at runtime. truthyness of `self.loader` would never be checked, but satisfy static type-checking.

pkg_resources/__init__.py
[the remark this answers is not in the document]
abravalheri: That sounds reasonable. Thanks @Avasam.

pkg_resources/__init__.py
```diff
@@ -1636,7 +1636,7 @@ def _validate_resource_path(path):
         )
 
     def _get(self, path) -> bytes:
-        if hasattr(self.loader, 'get_data'):
+        if hasattr(self.loader, 'get_data') and self.loader:
```
abravalheri, line 1639:
```suggestion
        if self.loader and hasattr(self.loader, 'get_data'):
```
Might make sense to reverse the short-circuit?

abravalheri: Never mind, just saw https://github.com/pypa/setuptools/pull/4262#discussion_r1518238830

abravalheri (approved):

abravalheri:
Thank you very much @Avasam.

PR #4267, opened 2024-03-08
`pkg_resources` type the declared global variables

Avasam:
## Summary of changes

Moved declarations of `@_call_aside` defined globals closer to where they're defined. And after `ResourceManager` & `WorkingSet` to avoid forward reference issues
Extracted from #4242 Works towards #2345 , resolves a few `reportOptionalCall` issues that #4192 would raise

### Pull Request Checklist
- [x] Changes have tests (existing runtime tests should pass, unblocks more type-checking PRs)

[PR docs]:


pkg_resources/__init__.py
```diff
@@ -3323,6 +3311,26 @@ def _initialize_master_working_set():
...
+    resource_listdir = __resource_manager.resource_listdir
+    set_extraction_path = __resource_manager.set_extraction_path
+    cleanup_resources = __resource_manager.cleanup_resources
+
+    working_set = WorkingSet()
+    require = working_set.require
+    iter_entry_points = working_set.iter_entry_points
+    add_activation_listener = working_set.subscribe
+    run_script = working_set.run_script
+    run_main = run_script
```
abravalheri, line 3331: My question here is:

Instead of "re-declaring" the global variables, wouldn't it make more sense to ditch the `@_call_aside` and just run the majority of the body of `_initialize*` functions directly on the global scope?

@jaraco, do you know if `_call_aside()`/`_initialize_master_working_set()`/`_initialize()` is used in `pkg_resources` just for the sake of organising the code, or is this very small delay on the execution a trick that solves a specific problem?

(I do like the way the code is organised right now, but if changes are needed for type-checking, I would prefer not having double book keeping. Of course we can decide that the type-checking here is not worth the trouble).

Avasam: > I would prefer not having double book keeping

FWIW, you already have to for other checkers (just as a note, these variables were moved from the top of the file, so this PR doesn't introduce double book-keeping).

That being said, I agree it'd be nicer if we didn't have to, but I wouldn't want to revamp `pkg_resources` too much outside of making it a fully typed package to take it out of typeshed.

abravalheri (approved):
Thank you very much @Avasam.

I think it is probably better to get this one going, we can address concerns in future PRs.

If we want the code to be "type-safe" at some point, we will probably need some degree typing gymnastics.

PR #4268, opened 2024-03-08
Update and pin mypy to 1.9

Avasam:
## Summary of changes

- pin mypy version so a new version doesn't suddenly cause the CI to fail
- Fix new failures in 1.9
- Explain why we're not running mypy on Python 3.12 in the CI yet
- Ignore `.egg` folder for faster local runs

### Pull Request Checklist
- [x] Changes have tests (these are test fixes)
- [x] News fragment added in [`newsfragments/`]. (no user facing changes)

[PR docs]:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,7 @@ testing =
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
+	mypy==1.9
 	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
 	tomli
 	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ testing =
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
-	mypy==1.9
+	mypy==1.9  # pin mypy version so a new version doesn't suddenly cause the CI to fail
 	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
 	tomli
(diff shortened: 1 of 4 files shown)


Avasam:
cygwin failure is unrelated:
```
HTTP status 404 fetching http://mirrors.kernel.org/sourceware/cygwin/x86_64/setup.ini.sig
mbox note: Unable to get http://mirrors.kernel.org/sourceware/cygwin/x86_64/setup.ini.sig from <http://mirrors.kernel.org/sourceware/cygwin/>
unattended_mode is set at mbox: returning default value
mbox note: Unable to get setup from <http://mirrors.kernel.org/sourceware/cygwin/>
unattended_mode is set at mbox: returning default value
```
https://mirrors.kernel.org/sourceware/cygwin/x86_64/setup.bz2.sig
image

setuptools/_core_metadata.py
```diff
@@ -62,7 +62,7 @@ def _read_list_from_msg(msg: Message, field: str) -> Optional[List[str]]:
 
 
 def _read_payload_from_msg(msg: Message) -> Optional[str]:
-    value = msg.get_payload().strip()
+    value = str(msg.get_payload()).strip()
```
Avasam, line 65: idk if this needs to be publicised as it all seems pretty internal. but this now supports `Message` whose payload is also a single-part `Message`. (multi-part Message support is undefined here, and will use a list repr)

setup.cfg
```diff
@@ -73,6 +73,7 @@ testing =
 	# for tools/finalize.py
 	jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"
 	pytest-home >= 0.5
+	mypy==1.9
```
abravalheri, line 76:
```suggestion
	mypy==1.9  # pin mypy version so a new version doesn't suddenly cause the CI to fail
```

setuptools/_core_metadata.py
abravalheri: I believe that `is_multipart()` should always be false in this case, so according to the docs the output is a string... So in terms of runtime the function should be safe.

Adding a explicit call to `str` should be OK to convince the type checker.

PR #4273, opened 2024-03-11
Update codecov-action to v4

Avasam:
## Summary of changes

Update codecov-action to v4
Updating actions separately just in case one happens to have breaking changes that affects setuptools and/or requires further changes.

### Pull Request Checklist
- [ ] Changes have tests
- [ ] News fragment added in [`newsfragments/`]. (no user facing changes)

[PR docs]:


webknjaz (changes requested):
This patch is incomplete: v4 *requires* an API token to be passed to the action (except for PRs from forks — which is the only reason it doesn't explode here).

Avasam:
> This patch is incomplete: v4 *requires* an API token to be passed to the action (except for PRs from forks — which is the only reason it doesn't explode here).

Thank you! I must've glanced over that. I guess that's why your PR is still in draft.

Reference: https://docs.codecov.com/docs/adding-the-codecov-token#github-actions

abravalheri:
I added the token as a secret

webknjaz:
@abravalheri I learned recently that it's possible to have an org-global secret. That sounds like less maintenance. Just need to ask somebody with org privileges to grab the right one from codecov and set it up..

webknjaz:
@Avasam the reason that PR is in draft is that nothing worked when I started and there were some fixes that I haven't tested, I think. Though, I don't think they are going to be ported to v3 which is entirely different code base so this bump is fine. But be careful. I didn't update to v4 because there were a million bugs when it was launched and I didn't get to verify what's fixed by now.

Avasam:
Thanks for the heads up @webknjaz . At worst this is an easy revert and we can reference any issue we encounter.

.github/workflows/main.yml
```diff
@@ -100,7 +100,9 @@ jobs:
         run: pipx run coverage xml --ignore-errors
       - name: Publish coverage
         if: hashFiles('coverage.xml') != ''  # Rudimentary `file.exists()`
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
webknjaz, line 105: Please, use an input instead of an env var.

Avasam: I followed codecov's documentation. Do you have an example of how this should be done instead to be improved?

webknjaz: https://github.com/codecov/codecov-action#usage

.github/workflows/main.yml
```diff
@@ -190,7 +192,9 @@ jobs:
         shell: C:\cygwin\bin\env.exe CYGWIN_NOWINPATH=1 CHERE_INVOKING=1 C:\cygwin\bin\bash.exe -leo pipefail -o igncr {0}
       - name: Publish coverage
         if: hashFiles('coverage.xml') != ''  # Rudimentary `file.exists()`
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
Avasam, line 197:
```suggestion
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
```

.github/workflows/main.yml
Avasam:
```suggestion
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
```

Avasam, line 105:
```suggestion
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
```

webknjaz:
@Avasam there's already a `with:`. You can't use duplicate keys in YAML. Put all inputs under one key. In both invocations.

.github/workflows/main.yml
Avasam: I guess the doc page I was looking at was outdated as it only offered using `env` rather than passing it as an input

webknjaz:
It's also a good idea to keep the inputs sorted by name.

PR #4283, opened 2024-03-16
Deduplicate testing dependencies by dropping `[testing-integration]`

Avasam:
## Summary of changes

In https://github.com/pypa/setuptools/pull/4257#discussion_r1514754970, @abravalheri mentioned that:
> [...] I think we just remove the `testing-integration` and use `testing` everywhere...
> 
> Although it might add a bit of overhead for the integration tests, it will simplify and streamline the setup... This way we reduce the divergence with `skeleton`. The additional overhead in the integration tests should not be too much, and the integration tests just run before the releases anyway.

This PR is doing just that.

Accepting this closes #4282 , which was my original idea.

### Pull Request Checklist
- [x] Changes have tests (these are test changes)

[PR docs]:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ pass_env =
 
 [testenv:integration]
 deps = {[testenv]deps}
-extras = testing
+extras = {[testenv]extras}
 pass_env =
 	{[testenv]pass_env}
 	DOWNLOAD_PATH
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ pass_env =
 
 [testenv:integration]
 deps = {[testenv]deps}
-extras = testing-integration
+extras = testing
 pass_env =
 	{[testenv]pass_env}
 	DOWNLOAD_PATH
--- a/conftest.py
+++ b/conftest.py
@@ -24,6 +24,7 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     config.addinivalue_line("markers", "integration: integration tests")
     config.addinivalue_line("markers", "uses_network: tests may try to download files")
+    _IntegrationTestSpeedups.disable_plugins_already_run(config)
 
 
 collect_ignore = [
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -23,7 +23,7 @@ pass_env =
 
 [testenv:integration]
 deps = {[testenv]deps}
-extras = testing-integration
+extras = testing
```
abravalheri, line 26:
```suggestion
extras = {[testenv]extras}
```

abravalheri (approved):
Thank you very much, @Avasam .

I think this is the simplest approach...
The next 2 weeks I don't know if I will have the time to contribute here, but my plan is to download the PR and rub the integration tests manually to see if everything works as expected (I think the CI do not run them on PR).

abravalheri:
@Avasam, when running the tests I was having an error[^1]:

```
setuptools/config/_validate_pyproject/formats.py:180: error: Cannot find implementation or library stub for module named "trove_classifiers"  [import-not-found]
setuptools/config/_validate_pyproject/formats.py:180: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 3 source files)
```

As a workaround, I added d98a9d92a.
Does it make sense?
Do you know why https://github.com/pypa/setuptools/blob/6ee23bf0579c52e1cbe7c97fc20fd085ff2a25c7/mypy.ini#L16 does not work?


[^1]: `pipx run --python python3.11 tox -e integration`

Avasam:
@abravalheri I remember having trove_classifier be picked up as an import previously, but I don't remember if/how I fixed it. Or if I just fixed it locally by installing trove_classifier then the CI did something different that didn't need it install.

Either way, the difference that I notice now is that integration tests will also run checkdocks, cov, mypy, ruff and perf whereas previously it didn't.

abravalheri:
@Avasam https://github.com/pypa/setuptools/pull/4283/commits/ded26346f7927c3e1f369a64c0dbd95b3f23b3f8 is a way of disabling those tests. Does it make sense for you, or should I revert that commit?

I believe that `mypy`, `ruff`, `checkdocs` already run on other tests so there is no need for repeating it... The `cov` plugin is more complicated. There is a discussion in https://github.com/pypa/setuptools/pull/3643, but the TL;DR of my problem is that `pytest-cov` is not reporting the real coverage of those tests, because we use nested subprocess (and therefore enabling the flag would bring a false sense of completeness). I haven't found a good way around of it.

Avasam:
Hmm, in a way, I wonder if we're just moving the duplication from one file to another.

Duplicating the declaration of dependencies not not be repeated, instead of those that are shared. 

Whilst leaving the door open to forgetting to disable a test. Not that I see new tooling being added too often, so it'd probably be caught eventually.

---

Other than that, it does bring setup cfg closer to the skeleton and dedupes dependencies in that file.

abravalheri:
Yeah, that is true... well, it is an optimisation thing, we can drop it at some point. I think that if the `cov` problem with integration tests is ever solved, it would be a nice moment to drop it.

PR #4286, opened 2024-03-20
Support PEP 625

### LINKED ISSUE #3593: [FR] Implement PEP 625 - File Name of a Source Distribution

Conform to accepted standards, make it possible to reliably determine a project's (canonical form) name and version from the source distribution filename.

See https://peps.python.org/pep-0625/

When creating sdist files, normalise the project name and version parts according to the specification, documented [here](https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name).

### Alternative Solutions

Continue as at present, which will leave sdist consumers with no reliable way of knowing the filename and version of a sdist short of either extracting the metadata from the sdist (if the sdist conforms to [PEP 643](https://www.python.org/dev/peps/pep-0643/)) or actually building the distribution.

### Additional context

Code that wants the project's formal name will still need to read the distribution metadata - that is understood and this specification doesn't affect that.

### Code of Conduct

di:
## Summary of changes

Support [PEP 625](https://peps.python.org/pep-0625/) by canonicalizing package name and version in filenames.

Closes #3593.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/_distutils/dist.py
+++ b/setuptools/_distutils/dist.py
@@ -23,6 +23,7 @@ from .errors import (
     DistutilsArgError,
     DistutilsClassError,
 )
+from setuptools.extern.packaging.utils import canonicalize_name, canonicalize_version
 from .fancy_getopt import FancyGetopt, translate_longopt
 from .util import check_environ, strtobool, rfc822_escape
 from ._log import log
@@ -1189,7 +1190,9 @@ class DistributionMetadata:
         return self.version or "0.0.0"
 
     def get_fullname(self):
-        return "{}-{}".format(self.get_name(), self.get_version())
+        return "{}-{}".format(
+            canonicalize_name(self.get_name()), canonicalize_version(self.get_version())
+        )
 
     def get_author(self):
         return self.author
--- a/setuptools/_distutils/dist.py
+++ b/setuptools/_distutils/dist.py
@@ -23,7 +23,6 @@ from .errors import (
     DistutilsArgError,
     DistutilsClassError,
 )
-from setuptools.extern.packaging.utils import canonicalize_name, canonicalize_version
 from .fancy_getopt import FancyGetopt, translate_longopt
 from .util import check_environ, strtobool, rfc822_escape
 from ._log import log
@@ -1190,9 +1189,7 @@ class DistributionMetadata:
         return self.version or "0.0.0"
 
     def get_fullname(self):
-        return "{}-{}".format(
-            canonicalize_name(self.get_name()), canonicalize_version(self.get_version())
-        )
+        return "{}-{}".format(self.get_name(), self.get_version())
 
     def get_author(self):
         return self.author
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -973,7 +973,7 @@ class Distribution(_Distribution):
     def _override_get_fullname(self):
         def _get_fullname_canonicalized(self):
             return "{}-{}".format(
-                canonicalize_name(self.get_name()),
+                canonicalize_name(self.get_name()).replace('-', '_'),
                 canonicalize_version(self.get_version()),
             )
 
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -7,7 +7,6 @@ import numbers
 import os
 import re
 import sys
-import contextlib
 from contextlib import suppress
 from glob import iglob
 from pathlib import Path
@@ -27,7 +26,6 @@ from .extern.more_itertools import partition, unique_everseen
 from .extern.ordered_set import OrderedSet
 from .extern.packaging.markers import InvalidMarker, Marker
 from .extern.packaging.specifiers import InvalidSpecifier, SpecifierSet
-from .extern.packaging.utils import canonicalize_name, canonicalize_version
 from .extern.packaging.version import Version
 
 from . import _entry_points
@@ -966,28 +964,7 @@ class Distribution(_Distribution):
         # Postpone defaults until all explicit configuration is considered
         # (setup() args, config files, command line and plugins)
 
-        with self._override_get_fullname():
-            super().run_command(command)
-
-    @contextlib.contextmanager
-    def _override_get_fullname(self):
-        def _get_fullname_canonicalized(self):
-            return "{}-{}".format(
-                canonicalize_name(self.get_name()).replace('-', '_'),
-                canonicalize_version(self.get_version()),
-            )
-
-        class NoValue:
-            pass
-
-        orig_val = getattr(self, 'get_fullname', NoValue)
-        self.get_fullname = _get_fullname_canonicalized.__get__(self)
-
-        try:
-            yield
-        finally:
-            if orig_val is not NoValue:
-                self.get_fullname = orig_val
+        super().run_command(command)
 
 
 class DistDeprecationWarning(SetuptoolsDeprecationWarning):
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -7,6 +7,7 @@ import numbers
 import os
 import re
 import sys
+import contextlib
 from contextlib import suppress
 from glob import iglob
 from pathlib import Path
@@ -26,6 +27,7 @@ from .extern.more_itertools import partition, unique_everseen
 from .extern.ordered_set import OrderedSet
 from .extern.packaging.markers import InvalidMarker, Marker
 from .extern.packaging.specifiers import InvalidSpecifier, SpecifierSet
+from .extern.packaging.utils import canonicalize_name, canonicalize_version
 from .extern.packaging.version import Version
 
 from . import _entry_points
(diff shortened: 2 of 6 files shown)


di:
Hmm, the [failing test](https://github.com/pypa/setuptools/actions/runs/8365722198/job/22904237174?pr=4286) doesn't fail for me locally.

di:
Looks like it's because this job is testing against stdlib distutils and not vendored distutils? What's the right path forward here?

abravalheri (commented):
Thank you very much for the contribution, @di!

I left some comments about the specific implementation.

Maybe @jaraco would like to have a look on the PR?

setuptools/_distutils/dist.py
```diff
@@ -23,6 +23,7 @@
     DistutilsArgError,
     DistutilsClassError,
 )
+from setuptools.extern.packaging.utils import canonicalize_name, canonicalize_version
```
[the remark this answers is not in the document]
di: I see. Am I understanding correctly that https://github.com/pypa/distutils has no dependencies, so we can't use `packaging` there at all?

> Maybe it would be better to go down in the class hierarchy to avoid modifying this directory directly?

Annoying, but makes sense I guess!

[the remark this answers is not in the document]
abravalheri: > we can't use `packaging` there at all?

I believe this assessment is correct 😅.

di:
@abravalheri Thanks! Everything is passing now.

setuptools/dist.py
```diff
@@ -964,7 +966,28 @@ def run_command(self, command):
         # Postpone defaults until all explicit configuration is considered
         # (setup() args, config files, command line and plugins)
 
-        super().run_command(command)
+        with self._override_get_fullname():
```
abravalheri, line 969: Hi @di, thank you very much for the update.

I have a question here (more out of curiosity really).

I am supposing that you needed to adopt this more complex approach because probably you found issues when trying a simpler one, e.g. something like the following:

```python
       def get_fullname(self):
            return "{}-{}".format(
                canonicalize_name(self.get_name()),
                canonicalize_version(self.get_version()),
            )
```

(i.e. simply overwrite via inheritance the method definition).

Do you remember which problems the simpler approach caused?

di: @abravalheri Which class inherits from `DistributionMetadata` here that I would add that function to? I didn't see one, and I believe it gets instantiated within `distutils`, not within `setuptools`.

I'm following the pattern here: https://github.com/pypa/setuptools/blob/6ee23bf0579c52e1cbe7c97fc20fd085ff2a25c7/setuptools/command/sdist.py#L80-L99

But maybe I should do something like https://github.com/pypa/setuptools/blob/main/setuptools/monkey.py instead?

It's unclear to me what would be preferred.

abravalheri: If I am not mistaken `get_fullname` is "dynamically" defined in the `Distribution` class via https://github.com/pypa/setuptools/blob/v69.2.0/setuptools/_distutils/dist.py#L151-L153. So I was supposing that we could just overwrite it in the `setuptools.dist.Distribution`.

But you are right, when the line https://github.com/pypa/setuptools/blob/v69.2.0/setuptools/dist.py#L307 runs, it gets overwritten anyway.

So yeah, now I can see what would be the problem.

Maybe we can do something like `self.get_fullname = self._get_fullname_canonicalized` just after the line https://github.com/pypa/setuptools/blob/v69.2.0/setuptools/dist.py#L307 (instead of the context manager)?

There is also the monkeypatch solution that you mentioned... Jason would be the best person to advise on that.

di: Yes, that would work, although I don't really see any precedent for doing that. It also would only patch that instance, and not all `Distribution`s that might get instantiated, although it seems like that might not be an issue.

I'm happy to switch to the `monkey.py` solution instead if that makes more sense.

setuptools/dist.py
```diff
@@ -964,7 +966,28 @@ def run_command(self, command):
...
+            super().run_command(command)
+
+    @contextlib.contextmanager
+    def _override_get_fullname(self):
+        def _get_fullname_canonicalized(self):
+            return "{}-{}".format(
+                canonicalize_name(self.get_name()),
```
dimbleby, line 976: This is the wrong canonicalization, as the test case later confirms.  

The hyphens were supposed to be replaced by underscores.

setuptools/dist.py
jaraco: My instinct is that the code should be first and foremost implemented in the canonical class. That is, what happens if we alter `distutils.dist.DistributionMetadata.get_fullname` to have the new behavior? Of course, that would need to be implemented in [pypa/distutils](/pypa/distutils), but since we maintain that, it should be straightforward to get that behavior ported here. That wouldn't fix the issue if distutils is used from stdlib, but that mode is soon to be deprecated, and if it needs to be supported, the `monkey` technique already provides forward ports of that functionality, so we should use that.

And if `get_fullname` is used elsewhere to mean non-canonicalized names, we can add a new method `get_filename` (or `get_canonical_full_name` or similar) and plumb that through.

setuptools/tests/test_config_discovery.py
```diff
@@ -255,7 +255,7 @@ def test_py_modules_when_wheel_dir_is_cwd(self, tmp_path):
 
 
 class TestNoConfig:
-    DEFAULT_VERSION = "0.0.0"  # Default version given by setuptools
+    CANONICAL_DEFAULT_VERSION = "0"  # Canonical default version given by setuptools
```
jaraco, line 258: TestNoConfig doesn't feel like the right place to assert expectations about file naming. I suspect this test probably needs to be adjusted to account for the change, but it also feels like there should be another test asserting specifically the intention about how sdists are named. It's possible there are tests in distutils that cover this behavior and also need to be adjusted (but aren't run as part of setuptools).

setuptools/dist.py
jaraco: In cd57ac3f1, I explored simply altering get_fullname and the tests pass. That patch isn't viable because it alters `_distutils` (and also reaches back into setuptools for packaging), but it at least demonstrates that all the Setuptools tests pass. Next, I'll explore what happens if this change is introduced into distutils.

setuptools/dist.py
di: Indeed! Thanks for the catch, fixed in bf33f79fee5ba88dba5dde8beb57ba03d856dc31.

setuptools/dist.py
di: > In https://github.com/pypa/setuptools/commit/cd57ac3f1c00479d70c1db8991bac5f3c0eda5df, I explored simply altering get_fullname and the tests pass. That patch isn't viable because it alters _distutils (and also reaches back into setuptools for packaging), but it at least demonstrates that all the Setuptools tests pass. Next, I'll explore what happens if this change is introduced into distutils.

Yes, this is what I had originally done in the first commit (b0135f5), which @abravalheri said wasn't viable.

jaraco: I applied a similar diff to the distutils project and most of the tests pass and the 5 failures look like intentional changes to align with the PEP:

```
FAILED distutils/tests/test_sdist.py::TestSDist::test_prune_file_list - AssertionError: assert ['fake-1.zip'] == ['fake-1.0.zip']
FAILED distutils/tests/test_sdist.py::TestSDist::test_make_distribution - AssertionError: assert ['fake-1.tar'...ake-1.tar.gz'] == ['fake-1.0.ta...e-1.0.tar.gz']
FAILED distutils/tests/test_sdist.py::TestSDist::test_add_defaults - AssertionError: assert ['fake-1.zip'] == ['fake-1.0.zip']
FAILED distutils/tests/test_sdist.py::TestSDist::test_manual_manifest - FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/f2/2plv6q2n7l932m2x004jlw340000gn/T/tmpa_mhw7wg/dist/fake-1.0.tar.gz'
FAILED distutils/tests/test_sdist.py::TestSDist::test_make_distribution_owner_group - FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/f2/2plv6q2n7l932m2x004jlw340000gn/T/tmpoypx9oc9/dist/fake-1.0.tar.gz'
```

So it seems to me the best approach will be to apply this change to distutils. A few things will need to happen before that can occur:

- distutils is currently ahead of Setuptools, so let's get those changes merged into Setuptools and released, so they're not conflated with this change
- distutils currently doesn't have access to anything in packaging. We'll want to vendor packaging or at least these functions from packaging to implement the canonicalization


Additionally, I'd like to see more tests in distutils covering expectations about these names (probably just docstrings on get_fullname covering more cases).

jaraco: In https://github.com/pypa/distutils/pull/244, I've proposed applying the changes in distutils.

Next, we'll want to adapt this change to provide a forward-port of that functionality in monkey.

PR #4296, opened 2024-04-11
Merge changes from skeleton

jaraco:

PR #4297, opened 2024-04-11
Fix ruff.toml

carljm:
https://github.com/pypa/setuptools/pull/4296 broke `ruff.toml`, by adding a second `extend-select` section. Merge them to fix ruff checking.


carljm:
Already fixed in https://github.com/pypa/setuptools/commit/230bde5008fbc7b0764649f39aa8640befd9ec0b

PR #4304, opened 2024-04-14
NEWS: Put releases in numerical order

hugovk:
## Summary of changes

I thought some releases were missing, but I guess they were added in chronological order rather than numerical.

Let's keep numerical, especially as no dates are shown.

### Pull Request Checklist
- [n/a] Changes have tests
- [n/a] News fragment added in [`newsfragments/`].

[PR docs]:

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -4,15 +4,6 @@ v69.5.1
 No significant changes.
 
 
-v69.4.2
-=======
-
-Bugfixes
---------
-
-- Merged bugfix for pypa/distutils#246 (#27489545)
-
-
 v69.5.0
 =======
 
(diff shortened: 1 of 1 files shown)


abravalheri:
Thank you very much

NEWS.rst
```diff
@@ -4,23 +4,23 @@ v69.5.1
...
 
-v69.4.2
+v69.5.0
 =======
 
-Bugfixes
+Features
 --------
 
-- Merged bugfix for pypa/distutils#246 (#27489545)
```
webknjaz, line 13: Oh, apparently the old issue number was broken before this reshuffle..

jaraco:
I'm not sure putting the changes in numerical order is correct. Having them in temporal order provides better fidelity about what changes are included in a given release. In the current ordering, it looks like v69.5.1 includes no changes from v69.5.0, when in fact, 69.4.2 was released between 69.5.0 and 69.5.1, so the changelog now seemingly incorrectly signals that v69.5.0 includes v69.4.2 and that v69.5.1 had no significant changes. Same applies for v69.3.1 between v69.4.0 and v69.4.1.

Also, dates are shown in the [rendered changelog](https://setuptools.pypa.io/en/latest/history.html).

I think we should revert this change, despite the confusion it might cause. If someone wants a changelog sorted by version number, they should re-render the changelog that way. You can mechanically convert from logic oldering to sorted versions, but not vice-versa.

PR #4312, opened 2024-04-19
Uses RST substitution to put badges in 1 line

shenxianpeng:
## Summary of changes

It might be some update to Github's RST renderer caused badges to display in multiple lines. uses RST substitution to put badges in 1 line.

Before the change

image

After the change

image

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,31 +1,33 @@
-.. image:: https://img.shields.io/pypi/v/setuptools.svg
+.. |pypi-version| image:: https://img.shields.io/pypi/v/setuptools.svg
    :target: https://pypi.org/project/setuptools
 
-.. image:: https://img.shields.io/pypi/pyversions/setuptools.svg
+.. |py-version| image:: https://img.shields.io/pypi/pyversions/setuptools.svg
 
-.. image:: https://github.com/pypa/setuptools/actions/workflows/main.yml/badge.svg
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you very much @shenxianpeng! That was a good catch, I haven't noticed this before your PR.

I think that is a fine change, but I wonder if the best approach would not be to notify GitHub about this regression? (The rationale is that with this change we fix setuptools, but there is a lot of other repositories out there with the same setup; so if GitHub fixes it at the source, everyone benefits).

Should we notify GitHub first and then see their feedback before merging the change?

abravalheri:
Thanks @shenxianpeng, I found an open discussion on the topic and reported the problem description there: https://github.com/orgs/community/discussions/114690

webknjaz:
> It might be some update to Github's RST renderer caused badges to display in multiple lines

Yes, I noticed this in many other of my projects. It also looks like they've broken rendering of other things and the overall support of RST on GitHub has always been quite limited.
So I think this is a good change for now but in the future it might make sense to migrate to MyST and regular Markdown...

abravalheri:
Thank you @shenxianpeng, let's merge it for now... I don't know how long the discussion on the github site will take.

PR #4325, opened 2024-04-24
Add `__pycache__/` to `.gitignore`

Avasam:
## Summary of changes

Running `tox` locally leaves a lot of files behind in `__pycache__` folders.
image
This PR adds the cache folder to the .gitignore, since this is specific to this project (using tox) and not my personal environment.

### Pull Request Checklist
- [ ] Changes have tests
- [ ] News fragment added in [`newsfragments/`]. (no user facing changes)

[PR docs]:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ setuptools/tests/bdist_wheel_testdata/*/*.egg-info/
 .cache
 .pytest_cache/
 .mypy_cache/
+__pycache__/


abravalheri:
https://blog.jaraco.com/skeleton/#ignoring-artifacts does mention `__pycache__` in the suggested `.gitignore_global`, so by that logic, this line would not be necessary...

However there is precedent (the currently ignored `*.pyc`, `.mypy_cache`, `.pytest_cache`, etc...). So I will defer this one to Jason 😅.

Avasam:
CC @jaraco

jaraco:
My preference would be to eliminate non-project-specific ignores from the .gitignore (and the toil that comes with maintaining such a file across the multitudes of projects that exist). I'm pretty sure none of those exclusions currently in the .gitignore are both specific to setuptools and not covered by the recommended config.

What do you think about instead trying to adopt that approach? I think you'll find it's much nicer for everyone involved. I'll update it to link to my latest config.

jaraco:
We're almost certainly going to decline this change in favor of #4453. Thanks for the suggestion regardless.

PR #4347, opened 2024-05-07
Removed `typing_extensions` from vendored dependencies

### LINKED ISSUE #4324: Re-evaluate if `typing_extensions` is still needed as a vendored dependency for `pkg_resources`
> This (`typing_extensions`) was not listed. The `vendored.txt` file says the following:
>
> ```
> # required for platformdirs on Python < 3.8
> typing_extensions==4.4.0
> ```
>
> But now we removed support to Python < 3.8, so I guess we can remove this vendored dependency?

_Originally posted by @abravalheri in https://github.com/pypa/setuptools/pull/4320#discussion_r1574742424_

Avasam:
## Summary of changes
Removed `typing_extensions` from vendored dependencies.

All vendored dependencies put `typing_extensions` behind a `TYPE_CHECKING` block, in a stub file, or behind a Python 3.8+ check. Hence it is no longer necessary to vendor `typing_extensions`. (and if a vendored library re-introduce a problem importing it at runtime, we can either raise an issue upstream and wait for the next version, or re-introduce a patch to make it conditional rather than use a vendor version).

Closes #4324

### Pull Request Checklist
- [x] Changes have tests (expecting existing tests to cover this change)

[PR docs]:


newsfragments/4324.removal.rst
```diff
@@ -0,0 +1 @@
+Removed `typing_extensions` from vendored dependencies -- by :user:`Avasam`
```
Avasam, line 1:
```suggestion
Removed ``typing_extensions`` from vendored dependencies -- by :user:`Avasam`
```

abravalheri:
Thank you very much @Avasam. I am in favour of simplifying as much as we can the vendoring.

There is a risk it might reappear when we update the vendored dependencies, but should be good for now.

PR #4348, opened 2024-05-07
Update dynamic module imports in `pkg_resources` to direct static imports & fix attribute error

Avasam:
## Summary of changes
Update dynamic module imports in `pkg_resources` to direct static imports as per https://github.com/pypa/setuptools/pull/4330/files#r1582361952

This also fixes an attribute error that was spotted thanks to the mypy config change.

### Pull Request Checklist
- [x] Changes have tests (this is static type checking update + existing checks should validate the import changes work)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -74,7 +74,7 @@ from pkg_resources.extern.packaging import markers as _packaging_markers
 from pkg_resources.extern.packaging import requirements as _packaging_requirements
 from pkg_resources.extern.packaging import utils as _packaging_utils
 from pkg_resources.extern.packaging import version as _packaging_version
-from pkg_resources.extern.platformdirs import user_cache_dir
+from pkg_resources.extern.platformdirs import user_cache_dir as _user_cache_dir
 
 # declare some globals that will be defined later to
 # satisfy the linters.
@@ -1340,7 +1340,7 @@ def get_default_cache():
     or a platform-relevant user cache dir for an app
     named "Python-Eggs".
     """
-    return os.environ.get('PYTHON_EGG_CACHE') or user_cache_dir(appname='Python-Eggs')
+    return os.environ.get('PYTHON_EGG_CACHE') or _user_cache_dir(appname='Python-Eggs')
 
 
 def safe_name(name):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -70,15 +70,11 @@ from pkg_resources.extern.jaraco.text import (
     drop_comment,
     join_continuation,
 )
-
-from pkg_resources.extern import platformdirs
-from pkg_resources.extern import packaging
-
-__import__('pkg_resources.extern.packaging.version')
-__import__('pkg_resources.extern.packaging.specifiers')
-__import__('pkg_resources.extern.packaging.requirements')
-__import__('pkg_resources.extern.packaging.markers')
-__import__('pkg_resources.extern.packaging.utils')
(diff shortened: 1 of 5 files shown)


pkg_resources/__init__.py
```diff
@@ -70,15 +70,14 @@
...
-__import__('pkg_resources.extern.packaging.markers')
-__import__('pkg_resources.extern.packaging.utils')
+from pkg_resources.extern.packaging.markers import Marker, InvalidMarker
+from pkg_resources.extern.packaging.requirements import (
+    InvalidRequirement as _packaging_InvalidRequirement,
+    Requirement as _packaging_Requirement,
+)
+from pkg_resources.extern.packaging.utils import canonicalize_name
+from pkg_resources.extern.packaging.version import Version, InvalidVersion
+from pkg_resources.extern.platformdirs import user_cache_dir
```
abravalheri, line 80: I know that `pkg_resources` does have `__all__` set, but I think I would worry less of people trying to import these names directly from `pkg_resources` if they had the `_` char?

Maybe something like:

```diff
-__import__('pkg_resources.extern.packaging.version')
-__import__('pkg_resources.extern.packaging.specifiers')
-__import__('pkg_resources.extern.packaging.requirements')
-__import__('pkg_resources.extern.packaging.markers')
-__import__('pkg_resources.extern.packaging.utils')
+from .extern.packaging import version as _packaging_version
+from .extern.packaging import specifiers as _packaging_specifiers
+from .extern.packaging import requirements as _packaging_requirements
+from .extern.packaging import markers as _packaging_markers
+from .extern.packaging import utils as _packaging_utils
```

(The rest we probably can leave as it is, right?)

Avasam: I see no issue aliasing packaging imports to private names.

pkg_resources/__init__.py
```diff
@@ -70,15 +70,11 @@
...
-__import__('pkg_resources.extern.packaging.version')
-__import__('pkg_resources.extern.packaging.specifiers')
-__import__('pkg_resources.extern.packaging.requirements')
-__import__('pkg_resources.extern.packaging.markers')
-__import__('pkg_resources.extern.packaging.utils')
+from pkg_resources.extern.packaging import markers as _packaging_markers
+from pkg_resources.extern.packaging import requirements as _packaging_requirements
+from pkg_resources.extern.packaging import utils as _packaging_utils
+from pkg_resources.extern.packaging import version as _packaging_version
+from pkg_resources.extern.platformdirs import user_cache_dir
```
abravalheri, line 77:
```suggestion
from pkg_resources.extern.platformdirs import user_cache_dir as _user_cache_dir
```

pkg_resources/__init__.py
```diff
@@ -1344,9 +1340,7 @@ def get_default_cache():
     or a platform-relevant user cache dir for an app
     named "Python-Eggs".
     """
-    return os.environ.get('PYTHON_EGG_CACHE') or platformdirs.user_cache_dir(
-        appname='Python-Eggs'
-    )
+    return os.environ.get('PYTHON_EGG_CACHE') or user_cache_dir(appname='Python-Eggs')
```
abravalheri, line 1343:
```suggestion
    return os.environ.get('PYTHON_EGG_CACHE') or _user_cache_dir(appname='Python-Eggs')
```

Just because I am afraid someone will try to use these names from `pkg_resources` even if it is not listed in `__all__`...

abravalheri (approved):

PR #4352, opened 2024-05-09
Type-check on all Python versions

Avasam:
## Summary of changes

Actually type-check targeting all Python versions in CI, including 3.12 & 3.13

I'd rather if it was possible for the CI to tell mypy which python version to target to override `python_version` config.

### Pull Request Checklist
- [x] Changes have tests (these are the test changes)
- [x] News fragment added in [`newsfragments/`]. (not public-facing)

[PR docs]:

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,7 +1,7 @@
 [mypy]
 # CI should test for all versions, local development gets hints for oldest supported
-# Some upstream typeshed distutils stubs fixes are necessary before we can start testing on Python 3.12
-python_version = 3.8
+# But our testing setup doesn't allow passing CLI arguments, so local devs have to set this manually.
+# python_version = 3.8
 strict = False
 warn_unused_ignores = True
 warn_redundant_casts = True
@@ -30,15 +30,19 @@ disable_error_code = attr-defined
 [mypy-pkg_resources.tests.*]
 disable_error_code = import-not-found
 
-# - distutils._modified has different errors on Python 3.8 [import-untyped], on Python 3.9+ [import-not-found]
+# - distutils doesn't exist on Python 3.12, unfortunately, this means typing
+#   will be missing for subclasses of distutils on Python 3.12 until either:
+#     - support for `SETUPTOOLS_USE_DISTUTILS=stdlib` is dropped (#3625)
+#       for setuptools to import `_distutils` directly
+#     - or non-stdlib distutils typings are exposed
 # - All jaraco modules are still untyped
 # - _validate_project sometimes complains about trove_classifiers (#4296)
 # - wheel appears to be untyped
-[mypy-distutils._modified,jaraco.*,trove_classifiers,wheel.*]
+[mypy-distutils.*,jaraco.*,trove_classifiers,wheel.*]
 ignore_missing_imports = True
 
 # Even when excluding a module, import issues can show up due to following import
 # https://github.com/python/mypy/issues/11936#issuecomment-1466764006
-[mypy-setuptools.config._validate_pyproject.*]
+[mypy-setuptools.config._validate_pyproject.*,setuptools._distutils.*]
 follow_imports = silent
 # silent => ignore errors when following imports
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -1,4 +1,9 @@
 """Extensions to the 'distutils' for large or complex distributions"""
+# mypy: disable_error_code=override
+# Command.reinitialize_command has an extra **kw param that distutils doesn't have
+# Can't disable on the exact line because distutils doesn't exists on Python 3.12
+# and mypy isn't aware of distutils_hack, causing distutils.core.Command to be Any,
+# and a [unused-ignore] to be raised on 3.12+
 
 from __future__ import annotations
 
@@ -114,8 +119,10 @@ def setup(**attrs):
 setup.__doc__ = distutils.core.setup.__doc__
 
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Command = distutils.core.Command
+    _Command: TypeAlias = distutils.core.Command
 else:
     _Command = monkey.get_unpatched(distutils.core.Command)
 
@@ -207,7 +214,7 @@ class Command(_Command):
                     "'%s' must be a list of strings (got %r)" % (option, val)
                 )
 
-    @overload  # type:ignore[override] # Extra **kw param
+    @overload
     def reinitialize_command(
         self, command: str, reinit_subcommands: bool = False, **kw
     ) -> _Command: ...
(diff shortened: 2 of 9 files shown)


Avasam:
I'm gonna leave this one for later. I won't have a choice but to tackle it to merge setuptools stubs from typeshed, but it doesn't affect pkg_resources which I'm focusing on for now.

Avasam:
Looks like I'll have to handle a handful of the typeshed-related TODOs first. Because `distutils` imports become `Any` in Python 3.12, a handful of type-ignores become unused.

Avasam:
Hopefully I didn't forget to add anything to typeshed. Waiting for https://github.com/python/mypy/issues/17285

setuptools/command/install.py
```diff
@@ -6,7 +6,8 @@
 import glob
 import platform
 import distutils.command.install as orig
-from typing import Any, ClassVar, cast
+from typing import Any, cast
+from typing import ClassVar
```
Avasam, line 10:
```suggestion
from typing import Any, ClassVar, cast
```

setuptools/__init__.py
```diff
@@ -1,4 +1,9 @@
 """Extensions to the 'distutils' for large or complex distributions"""
+# mypy: disable_error_code=override
+# Command.reinitialize_command has an extra **kw param that distutils doesn't have
+# Can't disable on the exact line because distutils doesn't exists on Python 3.12
+# and mypy isn't aware of distutils_hack, causing distutils.core.Command to be Any,
+# and a [unused-ignore] to be raised on 3.12+
```
Avasam, line 6: Alternatively:
```suggestion
# See comment in mypy.ini about distutils on Python 3.12
```

mypy.ini
```diff
@@ -1,7 +1,7 @@
 [mypy]
 # CI should test for all versions, local development gets hints for oldest supported
-# Some upstream typeshed distutils stubs fixes are necessary before we can start testing on Python 3.12
-python_version = 3.8
+# But our testing setup doesn't allow passing CLI arguments, so local devs have to set this manually.
+# python_version = 3.8
```
Avasam, line 4: Would love if there was a way to tell the CI to pass the `--python-version` argument to override this config so that IDEs and maybe default local runs can test against the oldest supported Python version.

mypy.ini
```diff
@@ -30,15 +30,19 @@ disable_error_code = attr-defined
 [mypy-pkg_resources.tests.*]
 disable_error_code = import-not-found
 
-# - distutils._modified has different errors on Python 3.8 [import-untyped], on Python 3.9+ [import-not-found]
+# - distutils doesn't exist on Python 3.12, unfortunately, this means typing
+#   will be missing for subclasses of distutils on Python 3.12 until either:
+#     - support for `SETUPTOOLS_USE_DISTUTILS=stdlib` is dropped (#3625)
+#       for setuptools to import `_distutils` directly
+#     - or non-stdlib distutils typings are exposed
```
Avasam, line 37: This includes potentially moving `_distutils` to a path where the last folder ends in `distutils` so it can be added to the [`MYPY_PATH`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-mypy_path).

Or typeshed accepting to publish a non-stdlib `types-distutils` for Python 3.12+

Until then, Python 3.12+ users won't be able to have complete typing without `types-setuptools` since anything inheriting or re-exported from `distutils` will be typed as `Any`.

setuptools/config/setupcfg.py
```diff
@@ -108,7 +110,8 @@ def _apply(
     filenames = [*other_files, filepath]
 
     try:
-        _Distribution.parse_config_files(dist, filenames=filenames)  # type: ignore[arg-type] # TODO: fix in distutils stubs
+        # TODO: Temporary cast until mypy 1.12 is released with upstream fixes from typeshed
+        _Distribution.parse_config_files(dist, filenames=cast(List[str], filenames))
```
Avasam, line 114: https://github.com/python/typeshed/commit/ac8720b99d41ca1c7340e18ff235bab679d42a8f#diff-d648cd8f6aa0211e1f3e96439ba4a67ec8b7474fbf0aba2459b894cb5022a0bcR16

mypy.ini
abravalheri: `pytest-mypy` mentions the following:

> Mypy supports [reading configuration settings](http://mypy.readthedocs.io/en/latest/config_file.html) from a mypy.ini file. Alternatively, the plugin can be configured in a conftest.py to invoke mypy with extra options:
> 
> ```python
> def pytest_configure(config):
>     plugin = config.pluginmanager.getplugin('mypy')
>     plugin.mypy_argv.append('--check-untyped-defs')
> ```

Not sure if that is useful to achieve what you want via `conftest.py`.

setuptools/dist.py
```diff
@@ -195,8 +195,10 @@ def check_packages(dist, attr, value):
 
 
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Distribution = distutils.core.Distribution
+    _Distribution: TypeAlias = distutils.core.Distribution
```
abravalheri, line 201: I think this one we can do:

```suggestion
    from distutils.core import Distribution as _Distribution
```

and remove the import for `TypeAlias`.

(`errors.py` should be fine using the `TypeAlias` though - I am indifferent on the approach for that file).

setuptools/extension.py
```diff
@@ -27,8 +27,10 @@ def _have_cython():
 # for compatibility
 have_pyrex = _have_cython
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Extension = distutils.core.Extension
+    _Extension: TypeAlias = distutils.core.Extension
```
abravalheri, line 33:
```suggestion
    from distutils.core import Extension as _Extension
```

And then we can drop the `TypeAlias`.

abravalheri (approved):

mypy.ini
Avasam: I think it would, something like:
```py
def pytest_configure(config):
    plugin = config.pluginmanager.getplugin('mypy')
    plugin.mypy_argv.append(f'--python-version={sys.version_info.major}.{sys.version_info.minor}')
```

setuptools/dist.py
Avasam: Since you already approved and these are changed in https://github.com/pypa/setuptools/pull/4504 , I'll keep it as-is for this PR

mypy.ini
abravalheri: > Since you approved, mind if I tackle your comments in a follow-up ?

Sure!

----

> `plugin.mypy_argv.append(f'--python-version={sys.version_info.major}.{sys.version_info.minor}')`

Is that going to behave differently than just omitting `python_version` in `mypy.ini`?

Avasam: Honestly, I never run pytest-mypy locally. But yes that would mean that the mypy-test commands would always check for the installed python. A more complete snippet of code would probably check if this is being run on the CI and whether the dev already passed `--python-version` as to not overwrite it (let's say they run `pytest-mypy` directly w/o tox)

PR #4358, opened 2024-05-15
Write a `compile_commands.json` from `build_ext`

### LINKED ISSUE #1975: Option to Generate Compilation Databases
The [Compilation Database Format](
https://clang.llvm.org/docs/JSONCompilationDatabase.html) describes a way for build systems to inform tools such as linters and autocompleters of the compilation commands used for a specific translation unit. This is especially important for CPython extensions, as `#include <Python.h>` will not work without special options. This means that most editors are unable to provide any completions or errors in-editor for Python extensions compiled with setuptools.

I have been writing compilation databases for my setuptools projects by hand by copying from the stdout of `setup.py build_ext`, but it would be nice to be able to generate them directly from setup.py.

This issue is mainly to gauge interest and establish what would be required to implement this. This might also be a task for `distutils` and bpo, but I am not familiar enough with the setuptools/distutils ecosystem to tell.

godlygeek:
## Summary of changes

Produce a JSON database of the compiler commands executed while building extension modules as `build/compile_commands.json`. This is usable by various C and C++ language servers, linters, and IDEs, like `clangd`, `clang-tidy`, and CLion. These tools need to understand the header search path and macros passed on the compiler command line in order to correctly interpret source files. In the case of Python extension modules, the developer might not even know all of the compiler flags that are being used, since some are inherited from the interpreter via `sysconfig`.

Closes #1975

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/_distutils/unixccompiler.py
+++ b/setuptools/_distutils/unixccompiler.py
@@ -145,6 +145,10 @@ class UnixCCompiler(CCompiler):
     if sys.platform == "cygwin":
         exe_extension = ".exe"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.compile_commands = []
+
     def preprocess(
         self,
         source,
@@ -185,7 +189,11 @@ class UnixCCompiler(CCompiler):
     def _compile(self, obj, src, ext, cc_args, extra_postargs, pp_opts):
         compiler_so = compiler_fixup(self.compiler_so, cc_args + extra_postargs)
         try:
-            self.spawn(compiler_so + cc_args + [src, '-o', obj] + extra_postargs)
+            cmd = compiler_so + cc_args + [src, '-o', obj] + extra_postargs
+            self.spawn(cmd)
+            self.compile_commands.append(
+                {"directory": os.getcwd(), "arguments": cmd, "file": src}
+            )
         except DistutilsExecError as msg:
             raise CompileError(msg)
 
--- a/newsfragments/4358.feature.rst
+++ b/newsfragments/4358.feature.rst
@@ -1,0 +1,1 @@
+A ``compile_commands.json`` is now written to the ``build/`` directory when building extension modules.
--- a/setuptools/command/build_ext.py
+++ b/setuptools/command/build_ext.py
@@ -1,9 +1,10 @@
+import json
 import os
 import sys
 import itertools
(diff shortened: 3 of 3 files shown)


godlygeek:
Let's call this a proof-of-concept. There are no tests yet. This implementation changes both the setuptools `build_ext` command and the vendored `distutils.unixccompiler` module.

If this seems like a reasonable approach to the problem, I guess the changes to `unixccompiler` would need to be done in https://github.com/pypa/distutils/ instead, and tested there.

Perhaps it should also apply to non-Unix compilers as well, but I'm not sure. `compile_commands.json` originated from the Clang/LLVM world, and I think most tools that use it expect clang- or gcc-compatible flags.

I'm not very familiar with the architecture of `setuptools`, so if this approach seems fundamentally wrong or there's a cleaner way to do it (especially one that doesn't require modifying the vendored `distutils`), please point me in the right direction.

setuptools/_distutils/unixccompiler.py
```diff
@@ -145,6 +145,10 @@ class UnixCCompiler(CCompiler):
     if sys.platform == "cygwin":
         exe_extension = ".exe"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.compile_commands = []
```
jaraco, line 150: Let's see if we can figure out an integration that doesn't require multiple touch points, maybe by wrapping spawn. Ideally, the implementation would be one line added somewhere to the existing implementation for integration and the rest of the behavior would be encapsulated in a function or class.

jaraco:
Thanks for the contrib.

> If this seems like a reasonable approach to the problem, I guess the changes to `unixccompiler` would need to be done in https://github.com/pypa/distutils/ instead, and tested there.

Yes. That's correct.

Let's start by proposing an interface there to collect the commands in distutils. Some things to think about:

- Should it only affect one compiler? If so, what should the behavior be on the unsupported compilers? My preference would be to provide the behavior universally or at least provide a degenerate behavior on compilers that are not relevant or where the implementation is not yet obvious.
- Let's see if we can create a clean hook point to the existing behavior so there's just one touch point.
- We'll want to document and capture in unit tests what the guaranteed interface is meant to be (i.e. what setuptools can rely on).

I'm going to close this for now, but we can re-open or revive in a new PR.

PR #4375, opened 2024-05-21
Remove calls to `typing.cast` with better type narrowing

Avasam:
## Summary of changes

Remove calls to `typing.cast` with better type narrowing. Better assurance of actual type.

### Pull Request Checklist
- [ ] Changes have tests
- [ ] News fragment added in [`newsfragments/`]. (not user facing)

[PR docs]:


setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -171,6 +173,10 @@ def _long_description(dist: "Distribution", val: _DictOrStr, root_dir: StrPath):
         _set_config(dist, "long_description_content_type", ctype)
 
     if file:
+        # Due to bad type narrowing, mypy sees file as `str | tuple[Any, ...]`
+        # and flipping the condition mypy sees val as `str | None`.
+        # Pyright gets both w/o issue. Can't use `tuple[()]()` until Python 3.9. So casting instead.
+        # https://github.com/python/mypy/issues/17273
```
Avasam, line 179: This still offers a path to remove the cast once Python 3.8 support is dropped, or if mypy improves its type narrowing for empty tuples (https://github.com/python/mypy/issues/17273)

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -153,15 +153,17 @@ def _guess_content_type(file: str) -> Optional[str]:
...
 
-def _long_description(dist: "Distribution", val: _DictOrStr, root_dir: StrPath):
+def _long_description(
+    dist: "Distribution", val: _LongDescriptionValue, root_dir: StrPath
+):
     from setuptools.config import expand
 
     if isinstance(val, str):
-        file: Union[str, list] = val
+        file: Union[str, tuple[()]] = val
```
abravalheri, line 162: Curiosity question: what does this notation mean? `tuple of empty tuple`?

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -153,15 +153,17 @@ def _guess_content_type(file: str) -> Optional[str]:
...
         text = expand.read_files(file, root_dir)
         ctype = _guess_content_type(val)
     else:
-        file = val.get("file") or []
+        file = val.get("file") or ()
```
abravalheri, line 166: In this case the runtime type of `val["file"]`[^1] is `Union[str, list[str]]`, because `tomlkit` maps TOML arrays into lists (and never produces tuples...).

Would it be better to define `file: Union[str, list[str]]` or `file: Union[str, Sequence[str]]`?

[^1]: when `"file" in val`

abravalheri (approved):
Thank you very much @Avasam.

setuptools/config/_apply_pyprojecttoml.py
Avasam: It means an empty tuple `()`

setuptools/config/_apply_pyprojecttoml.py
Avasam: The reason I used an empty tuple was to have a statically falsy value that matches an `Iterable[str]` for `expand.read_files`.

Atm it didn't end up changing anything as per the comment.

After looking at it again... isn't this actually showing that there's an issue?
`file` could be truthy with a list of files, in which case this would add a `list` to the `_referenced_files` set instead of a `str`

abravalheri: Umm.. Good question.

I think I will need to have a look on this tomorrow morning.

I might be wrong about the whole `list[str]` here (but I need to double check, if this code regards the `project.readme` field in `pyproject.toml` and or `tool.setuptools.dynamic.readme`, if it is just the first it should only `str`, if it includes the second, then it is `str | List[str]`...)

Avasam: If it's the first one, then calling `expand.read_files` also seems overkill.

Documentation ref: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

But I'll let you validate the code.

abravalheri: So it seems to be only for `project.readme`, which means `str` if present.

In that case it should be fine to use `()`, in runtime the value will never be a `list[str]`.
I believe that `()` is better than `""` because I am using a comprehension/for-loop... In the comprehension, `()` would have no effect, but I don't know exactly what the effect of `""` would be. Will it try to open a file named `""`?

I used `read_files` because I did not want to re-implement the logic for checking for existing and local files.

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -153,15 +152,17 @@ def _guess_content_type(file: str) -> Optional[str]:
...
 
     if isinstance(val, str):
-        file: Union[str, list] = val
+        file = val
         text = expand.read_files(file, root_dir)
-        ctype = _guess_content_type(val)
+        ctype = _guess_content_type(file)
     else:
-        file = val.get("file") or []
+        file = val.get("file", "")
```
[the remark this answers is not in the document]
abravalheri: Maybe something like the following?

```diff
from typing import Sequence

# ...

    if isinstance(val, str):
-       file: Union[str, list] = val
+       file: Union[str, Sequence[str]]  = val
        text = expand.read_files(file, root_dir)
        ctype = _guess_content_type(file)
    else:
-       file = val.get("file") or []
+       file = val.get("file") or ()
```

setuptools/dist.py
```diff
@@ -272,6 +272,7 @@ class Distribution(_Distribution):
     }
 
     _patched_dist = None
+    namespace_packages: List[str]
```
abravalheri, line 275: Personally, I would not add this one here, because setuptools augments many instance attributes dynamically via entry-points (it might be a lost cause to try to statically define all of them).

But if that is required to make the type checker happy, I think we should at least annotate something like the following.

```suggestion
    namespace_packages: List[str]  #: :meta private: DEPRECATED
```

setuptools/config/_apply_pyprojecttoml.py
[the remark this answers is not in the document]
Avasam: Agreed. I think I just never ended up updating this following your comment.

setuptools/dist.py
Avasam: It's for https://github.com/pypa/setuptools/pull/4375/files#diff-6c888d2f35e4149ad56292334032e9a216d1ab2b32a4df3c791d864c42dfad04L522-R523

```suggestion
    # Used by setuptools/command/editable_wheel.py for legacy namespaces
    namespace_packages: List[str]  #: :meta private: DEPRECATED
```

Avasam: Alternatively, we can even make it statically depcrecated using Python 3.13 decoractor (backported in `typing_extensions`) with something like:
```suggestion
    if TYPE_CHECKING:
        # Used by setuptools/command/editable_wheel.py for legacy namespaces
        @property
        @deprecated("Some deprecated message about what to use instead.")
        def namespace_packages(self) -> List[str]: ...  #: :meta private: DEPRECATED
```

image

setuptools/dist.py
```diff
@@ -272,6 +272,8 @@ class Distribution(_Distribution):
     }
 
     _patched_dist = None
+    # Used by setuptools/command/editable_wheel.py for legacy namespaces
```
Avasam, line 275:
```suggestion
    # Used by build_py, editable_wheel and install_lib commands for legacy namespaces
```

setuptools/dist.py
abravalheri: Thank you very much @Avasam. The last change should be enough (always good to avoid the hassle of conditionally importing `typing_extensions`).

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -153,15 +152,18 @@ def _guess_content_type(file: str) -> Optional[str]:
...
+    file: Union[str, tuple[()]]
     if isinstance(val, str):
-        file: Union[str, list] = val
+        file = val
         text = expand.read_files(file, root_dir)
-        ctype = _guess_content_type(val)
+        ctype = _guess_content_type(file)
     else:
-        file = val.get("file") or []
+        file = val.get("file", ())
```
abravalheri, line 166: This is better right? (e.g. `{"file": ""}`)

```suggestion
        file = val.get("file") or ()
```

setuptools/dist.py
Avasam: Yeah and it results in a weird non-runtime property.

setuptools/config/_apply_pyprojecttoml.py
Avasam: Ah oops. Yes, didn't double check against the original implementation before my PR.

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -153,15 +152,18 @@ def _guess_content_type(file: str) -> Optional[str]:
...
+):
     from setuptools.config import expand
 
+    file: Union[str, tuple[()]]
     if isinstance(val, str):
-        file: Union[str, list] = val
+        file = val
         text = expand.read_files(file, root_dir)
-        ctype = _guess_content_type(val)
+        ctype = _guess_content_type(file)
```
Avasam, line 164: To future readers, if you're wondering why not
```suggestion
    if isinstance(val, str):
        file: Union[str, tuple[()]] = val
        text = expand.read_files(file, root_dir)
        ctype = _guess_content_type(val)
```
This is why: https://github.com/pypa/setuptools/pull/4375#discussion_r1608733218
image

PR #4379, opened 2024-05-22
Enforce ruff/pyupgrade rules (UP)

DimitriPapadopoulos:
## Summary of changes

Enforce ruff/pyupgrade rules (`UP`).

Follow-up of #4125.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -3049,7 +3049,7 @@ class Distribution:
             version = self._get_version()
             if version is None:
                 path = self._get_metadata_path_for_display(self.PKG_INFO)
-                msg = (f"Missing 'Version:' header and/or {self.PKG_INFO} file at path: {path}")
+                msg = f"Missing 'Version:' header and/or {self.PKG_INFO} file at path: {path}"
                 raise ValueError(msg, self) from e
 
             return version
@@ -3105,9 +3105,7 @@ class Distribution:
             try:
                 deps.extend(dm[safe_extra(ext)])
             except KeyError as e:
-                raise UnknownExtra(
-                    f"{self} has no such extra feature {ext!r}"
-                ) from e
+                raise UnknownExtra(f"{self} has no such extra feature {ext!r}") from e
         return deps
 
     def _get_metadata_path_for_display(self, name):
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -200,7 +200,7 @@ def get_supported_platform():
     m = macosVersionString.match(plat)
     if m is not None and sys.platform == "darwin":
         try:
-            plat = 'macosx-%s-%s' % ('.'.join(_macos_vers()[:2]), m.group(3))
+            plat = 'macosx-{}-{}'.format('.'.join(_macos_vers()[:2]), m.group(3))
         except ValueError:
             # not macOS
             pass
@@ -492,7 +492,7 @@ def compatible_platforms(provided: str | None, required: str | None) -> bool:
             provDarwin = darwinVersionString.match(provided)
             if provDarwin:
                 dversion = int(provDarwin.group(1))
-                macosversion = "%s.%s" % (reqMac.group(1), reqMac.group(2))
+                macosversion = "{}.{}".format(reqMac.group(1), reqMac.group(2))
                 if (
                     dversion == 7
                     and macosversion >= "10.3"
@@ -1316,7 +1316,7 @@ class Environment:
                 for dist in other[project]:
                     self.add(dist)
         else:
-            raise TypeError("Can't add %r to environment" % (other,))
+            raise TypeError("Can't add {!r} to environment".format(other))
         return self
 
     def __add__(self, other: Distribution | Environment) -> Self:
@@ -2018,7 +2018,7 @@ class ZipProvider(EggProvider):
             return ''
         if fspath.startswith(self.zip_pre):
             return fspath[len(self.zip_pre) :]
-        raise AssertionError("%s is not a subpath of %s" % (fspath, self.zip_pre))
+        raise AssertionError("{} is not a subpath of {}".format(fspath, self.zip_pre))
 
     def _parts(self, zip_path):
         # Convert a zipfile subpath into an egg-relative path part list.
@@ -2026,7 +2026,7 @@ class ZipProvider(EggProvider):
         fspath = self.zip_pre + zip_path
         if fspath.startswith(self.egg_root + os.sep):
             return fspath[len(self.egg_root) + 1 :].split(os.sep)
-        raise AssertionError("%s is not a subpath of %s" % (fspath, self.egg_root))
+        raise AssertionError("{} is not a subpath of {}".format(fspath, self.egg_root))
 
     @property
     def zipinfo(self):
@@ -2729,15 +2729,15 @@ class EntryPoint:
         self.dist = dist
 
     def __str__(self) -> str:
-        s = "%s = %s" % (self.name, self.module_name)
+        s = "{} = {}".format(self.name, self.module_name)
         if self.attrs:
             s += ':' + '.'.join(self.attrs)
         if self.extras:
-            s += ' [%s]' % ','.join(self.extras)
+            s += ' [{}]'.format(','.join(self.extras))
         return s
 
     def __repr__(self) -> str:
-        return "EntryPoint.parse(%r)" % str(self)
+        return "EntryPoint.parse({!r})".format(str(self))
 
     @overload
     def load(
@@ -3108,7 +3108,7 @@ class Distribution:
                 deps.extend(dm[safe_extra(ext)])
             except KeyError as e:
                 raise UnknownExtra(
-                    "%s has no such extra feature %r" % (self, ext)
+                    "{} has no such extra feature {!r}".format(self, ext)
                 ) from e
         return deps
 
@@ -3150,7 +3150,7 @@ class Distribution:
 
     def egg_name(self):
         """Return what this distribution's standard .egg filename should be"""
-        filename = "%s-%s-py%s" % (
+        filename = "{}-{}-py{}".format(
             to_filename(self.project_name),
             to_filename(self.version),
             self.py_version or PY_MAJOR,
@@ -3162,7 +3162,7 @@ class Distribution:
 
     def __repr__(self) -> str:
         if self.location:
-            return "%s (%s)" % (self, self.location)
+            return "{} ({})".format(self, self.location)
         else:
             return str(self)
 
@@ -3172,7 +3172,7 @@ class Distribution:
         except ValueError:
             version = None
         version = version or "[unknown version]"
-        return "%s %s" % (self.project_name, version)
+        return "{} {}".format(self.project_name, version)
 
     def __getattr__(self, attr: str):
         """Delegate all unrecognized public attributes to .metadata provider"""
@@ -3200,9 +3200,9 @@ class Distribution:
     def as_requirement(self):
         """Return a ``Requirement`` that matches this distribution exactly"""
         if isinstance(self.parsed_version, packaging.version.Version):
-            spec = "%s==%s" % (self.project_name, self.parsed_version)
+            spec = "{}=={}".format(self.project_name, self.parsed_version)
         else:
-            spec = "%s===%s" % (self.project_name, self.parsed_version)
+            spec = "{}==={}".format(self.project_name, self.parsed_version)
 
         return Requirement.parse(spec)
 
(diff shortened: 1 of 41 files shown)


Avasam:
You should ignore `UP038` too ! https://github.com/astral-sh/ruff/issues/7871

ruff.toml
```diff
@@ -26,10 +28,6 @@ ignore = [
 ]
 extend-ignore = [
 	"UP015",  # redundant-open-modes, explicit is preferred
```
Avasam, line 30: This can be moved in `ignore`. `extend-ignore` is deprecated https://docs.astral.sh/ruff/settings/#lint_extend-ignore

pkg_resources/__init__.py
```diff
@@ -2699,15 +2699,15 @@ def __init__(
...
         if self.attrs:
             s += ':' + '.'.join(self.attrs)
         if self.extras:
-            s += ' [%s]' % ','.join(self.extras)
+            s += ' [{}]'.format(','.join(self.extras))
         return s
 
     def __repr__(self):
-        return "EntryPoint.parse(%r)" % str(self)
+        return f"EntryPoint.parse({str(self)!r})"
```
Avasam, line 2710: Not an automated change, but this can be simplified further

```suggestion
        return f"EntryPoint.parse('{self}')"
```

pkg_resources/__init__.py
```diff
@@ -3470,7 +3462,7 @@ def __hash__(self):
         return self.__hash
 
     def __repr__(self):
-        return "Requirement.parse(%r)" % str(self)
+        return f"Requirement.parse({str(self)!r})"
```
Avasam, line 3465: Not an automated change, but this can be simplified further

```suggestion
        return f"Requirement.parse('{self}')"
```

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -219,8 +219,8 @@ def test_get_metadata__bad_utf8(tmpdir):
         "codec can't decode byte 0xe9 in position 1: "
         'invalid continuation byte in METADATA file at path: '
     )
-    assert expected in actual, 'actual: {}'.format(actual)
-    assert actual.endswith(metadata_path), 'actual: {}'.format(actual)
+    assert expected in actual, f'actual: {actual}'
+    assert actual.endswith(metadata_path), f'actual: {actual}'
```
Avasam, line 223: Technically changes the string, to simplify further, but it's an assert info message


```suggestion
    assert expected in actual, f'{actual=}'
    assert actual.endswith(metadata_path), f'{actual=}'
```

setuptools/command/easy_install.py
```diff
@@ -1112,8 +1110,9 @@ def install_wheel(self, wheel_path, tmpdir):
             self.execute(
                 wheel.install_as_egg,
                 (destination,),
-                ("Installing %s to %s")
-                % (os.path.basename(wheel_path), os.path.dirname(destination)),
+                (
+                    f"Installing {os.path.basename(wheel_path)} to {os.path.dirname(destination)}"
+                ),
```
Avasam, line 1115: I think those parenthesis are redundant and lead to longer line

```suggestion
                f"Installing {os.path.basename(wheel_path)} to {os.path.dirname(destination)}",
```

setuptools/package_index.py
```diff
@@ -720,20 +719,15 @@ def gen_setup(self, filename, fragment, tmpdir):
             with open(os.path.join(tmpdir, 'setup.py'), 'w', encoding="utf-8") as file:
                 file.write(
                     "from setuptools import setup\n"
-                    "setup(name=%r, version=%r, py_modules=[%r])\n"
-                    % (
-                        dists[0].project_name,
-                        dists[0].version,
-                        os.path.splitext(basename)[0],
-                    )
+                    f"setup(name={dists[0].project_name!r}, version={dists[0].version!r}, py_modules=[{os.path.splitext(basename)[0]!r}])\n"
```
Avasam, line 722: Another case where the line is so long, splitting or using `.format` is imo much more readable

setuptools/package_index.py
```diff
@@ -1014,7 +1004,7 @@ def __iter__(self):
         yield self.password
 
     def __str__(self):
-        return '%(username)s:%(password)s' % vars(self)
+        return '{username}:{password}'.format(**vars(self))
```
Avasam, line 1007: Not an automated change, but this can be cleaned up further. Statically checkable too.

```suggestion
        return f'{self.username}:{self.password}'
```

setuptools/tests/config/test_setupcfg.py
```diff
@@ -85,24 +85,24 @@ def test_basic(self, tmpdir):
...
-        config_dict = read_configuration('%s' % config)
+        config_dict = read_configuration(f'{config}')
         assert config_dict['metadata']['version'] == '10.1.1'
         assert config_dict['metadata']['keywords'] == ['one', 'two']
         assert config_dict['options']['scripts'] == ['bin/a.py', 'bin/b.py']
 
     def test_no_config(self, tmpdir):
         with pytest.raises(DistutilsFileError):
-            read_configuration('%s' % tmpdir.join('setup.cfg'))
+            read_configuration('{}'.format(tmpdir.join('setup.cfg')))
```
Avasam, line 95: The format call here seems quite extra
```suggestion
            read_configuration(str(tmpdir.join('setup.cfg')))
```
(Can't double check from my phone, but if `tmpdir.join` already returns a `str`):

```suggestion
            read_configuration(tmpdir.join('setup.cfg'))
```

pkg_resources/__init__.py
```diff
@@ -3118,19 +3114,15 @@ def activate(self, path: list[str] | None = None, replace: bool = False):
 
     def egg_name(self):
         """Return what this distribution's standard .egg filename should be"""
-        filename = "%s-%s-py%s" % (
-            to_filename(self.project_name),
-            to_filename(self.version),
-            self.py_version or PY_MAJOR,
-        )
+        filename = f"{to_filename(self.project_name)}-{to_filename(self.version)}-py{self.py_version or PY_MAJOR}"
```
Avasam, line 3117: One of those cases I'd argue might be cleaner as `format` call. Especially with the length of the line and every f string values.

pkg_resources/__init__.py
```diff
@@ -3295,8 +3287,8 @@ def check_version_conflict(self):
             ):
                 continue
             issue_warning(
-                "Module %s was already imported from %s, but %s is being added"
-                " to sys.path" % (modname, fn, self.location),
+                f"Module {modname} was already imported from {fn}, but {self.location} is being added"
+                " to sys.path",
```
Avasam, line 3291: I'd either split the line back under the line limit. Or merge it back since rule E501 isn't enabled.

Currently it's the worst of both worlds (102 characters and split)

Avasam (commented):
Some comments and imo potential non-automated improvements I noticed from also having applied UP03 in other projects

setuptools/package_index.py
DimitriPapadopoulos: That , or perhaps use additional variables with meaningful names:
```python
                    name = dists[0].project_name
                    version = dists[0].version
                    modules = os.path.splitext(basename)[0]
                    f"setup(name={name!r}, version={version!r}, py_modules=[{modules!r}])\n"
```

pkg_resources/__init__.py
DimitriPapadopoulos: Perhaps add extra `name` and `version` variables?

pkg_resources/__init__.py
DimitriPapadopoulos: Detection at least can be automated:
```console
$ grep -R '{str(' --exclude-dir=_distutils --exclude-dir=_vendor
setuptools/tests/test_sdist.py:                                depends=[{str(dep_path)!r}]
pkg_resources/__init__.py:        return f"EntryPoint.parse({str(self)!r})"
pkg_resources/__init__.py:                raise _packaging_version.InvalidVersion(f"{str(ex)} {info}") from None
pkg_resources/__init__.py:            {str(ex)}\n{notes}
pkg_resources/__init__.py:        return f"Requirement.parse({str(self)!r})"
$ 
```

pkg_resources/__init__.py
DimitriPapadopoulos: I automated the detection of such explicit conversions in f-strings.

setuptools/tests/config/test_setupcfg.py
DimitriPapadopoulos: According to the [pytest documentation](https://docs.pytest.org/en/8.2.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures):
> The `tmpdir` and `tmpdir_factory` fixtures are similar to `tmp_path` and `tmp_path_factory`, but use/return legacy [`py.path.local`](https://py.readthedocs.io/en/latest/path.html) objects rather than standard [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) objects.

```python
>>> import py.path
>>> 
>>> tmpdir = py.path.local("/tmp")
>>> x = tmpdir.join('setup.cfg')
>>> 
>>> type(x)
<class 'py._path.local.LocalPath'>
>>> 
>>> str(x)
'/tmp/setup.cfg'
>>> 
>>> x
local('/tmp/setup.cfg')
>>> 
```

pkg_resources/__init__.py
DimitriPapadopoulos: I chose to merge it back, in the spirit of the [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#breaking-long-lines-and-strings):
> However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them.

setuptools/package_index.py
DimitriPapadopoulos: I have added the extra variables, and also changed `{name!r}` to `'{name}'`. I find the intent is clearer this way.

pkg_resources/__init__.py
DimitriPapadopoulos: Actually, I have seen `str(self)!r}` elsewhere, especially in _distutils_, and I believe this is on purpose. From the [`repr()`](https://docs.python.org/fr/dev/library/functions.html#repr) documentation:
> For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to [eval()](https://docs.python.org/fr/dev/library/functions.html#eval);

We precisely want an expression that yields `str(self)` when executed. Of course, `{str(self)!r})` is equivalent to `'{self}'`, but the former expression conveys a specific intent.

setuptools/package_index.py
DimitriPapadopoulos: You're right, both output the same result. Again, I suspect the author wanted to convey the very precise idea that `{str(self)!r}` printed into Python code will yield `str(self)`  upon execution. Both are equivalent in practice and your suggestion more concise, but I didn't feel like changing the code that much.

pkg_resources/__init__.py
Avasam: Yeah it's not the same as a standalone `str` or `repr` in a fstring. The former being truly redundant, and the later changeable to `!r`

setuptools/package_index.py
DimitriPapadopoulos: I've reverted to `{str(self)!r}`, let's see whether Windows tests pass.

setuptools/dist.py
```diff
@@ -46,7 +46,7 @@ def check_importable(dist, attr, value):
         assert not ep.extras
```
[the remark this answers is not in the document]
DimitriPapadopoulos: What's the benefit of the current template? I'd rather leave such manual fixes for a later PR.

DimitriPapadopoulos:
Restarted from scratch, it's easier and less error-prone than rebasing.

setuptools/dist.py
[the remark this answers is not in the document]
DimitriPapadopoulos: I've eventually pushed a commit with manual fixes for the very few easy cases.

Avasam:
Might wanna disable UP027 too (I've done so here: https://github.com/pypa/setuptools/pull/4558)

Avasam:
CI failure looks completely unrelated, I got the same thing here: https://github.com/pypa/setuptools/actions/runs/11113018762/job/30876386043?pr=4556

Avasam:
@jaraco and/or @abravalheri Can this be reviewed? Finishing to enable all `UP` rules is a PR I would've done as well if this one didn't already exists.

jaraco:
Looks great. Thanks!

PR #4381, opened 2024-05-22
Use set instead of True-only dict for non-public names

Avasam:
## Summary of changes

Extracted from #4366 to only include internal values.

### Pull Request Checklist
- [ ] Changes have tests (existing tests should pass)
- [ ] News fragment added in [`newsfragments/`]. (no public-facing changes)

[PR docs]:


abravalheri (approved):
Thank you very much!

PR #4382, opened 2024-05-22
Prevent a `TypeError` in `compat.py311.shutil_rmtree`

Avasam:
## Summary of changes

Noticed while adding pyright checks to setuptools (https://github.com/pypa/setuptools/pull/4192)
Also typed `shutil_rmtree` (xref https://github.com/python/typeshed/pull/12002 )

### Pull Request Checklist
- [ ] Changes have tests (#4192 would add the changes that caught this)

[PR docs]:

### DIFF
--- a/setuptools/compat/py311.py
+++ b/setuptools/compat/py311.py
@@ -1,12 +1,26 @@
-import sys
+from __future__ import annotations
+
 import shutil
+import sys
+from typing import Any, Callable, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from _typeshed import StrOrBytesPath, ExcInfo
+
+# Same as shutil._OnExcCallback from typeshed
+_OnExcCallback = Callable[[Callable[..., Any], str, BaseException], object]
 
 
-def shutil_rmtree(path, ignore_errors=False, onexc=None):
+def shutil_rmtree(
+    path: StrOrBytesPath,
+    ignore_errors: bool = False,
+    onexc: _OnExcCallback | None = None,
+) -> None:
     if sys.version_info >= (3, 12):
         return shutil.rmtree(path, ignore_errors, onexc=onexc)
 
-    def _handler(fn, path, excinfo):
-        return onexc(fn, path, excinfo[1])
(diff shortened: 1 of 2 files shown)


setuptools/compat/py311.py
```diff
@@ -1,12 +1,26 @@
...
+
+if TYPE_CHECKING:
+    from _typeshed import StrOrBytesPath, ExcInfo
 
 
-def shutil_rmtree(path, ignore_errors=False, onexc=None):
+def shutil_rmtree(
+    path: StrOrBytesPath,
+    ignore_errors: bool = False,
+    onexc: shutil._OnExcCallback | None = None,
```
abravalheri, line 16: I think we should not depend on a private member of the stdlib.

abravalheri (commented):
Thank you very much for working on the type improvements of Setuptools.

I am a bit shocked with the amount of overhead this particular PR adds, is there anyway we can simplify it?

Is there anyway we can simplify this?

setuptools/compat/py311.py
Avasam: Since it's a type alias, I can simply expand it.

setuptools/compat/py311.py
```diff
@@ -1,12 +1,26 @@
-import sys
+from __future__ import annotations
+
 import shutil
+import sys
+from typing import TYPE_CHECKING
+from collections.abc import Callable
+from typing import Any
```
Avasam, line 7:
```suggestion
from typing import Any, Callable, TYPE_CHECKING
```

setuptools/compat/py311.py
```diff
@@ -1,12 +1,24 @@
-import sys
+from __future__ import annotations
+
 import shutil
+import sys
+from typing import Any, Callable, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from _typeshed import StrOrBytesPath, ExcInfo
```
Avasam, line 8: > Is there anyway we can simplify this?

The runtime fix is just the added conditional. Everything else is type hints.

I just combined imports, otherwise I could omit annotating the internal function and move `StrOrBytesPath` to a runtime alias with `StrPath` in `setuptools._path`.

PR #4369, opened 2024-05-19
Adopted the bdist_wheel command from "wheel"

agronholm:
## Summary of changes

This adopts the ``bdist_wheel`` command and vendors most of the `wheel` code.

Closes #1386.

### Pull Request Checklist
- [X] Changes have tests
- [X] News fragment added in [`newsfragments/`].
  _(See [documentation][PR docs] for details)_

[`newsfragments/`]: https://github.com/pypa/setuptools/tree/master/newsfragments
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

**NOTE**: Coverage reports missing lines in `test_bdist_wheel` but those would be handled with combined coverage reporting.

PR #4383, opened 2024-05-22
`bdist_wheel` typing improvement

Avasam:
## Summary of changes

A couple things I noticed from #4369 CC @agronholm

### Pull Request Checklist
- [ ] Changes have tests (existing tests + mypy checks newly completed methods)

[PR docs]:


agronholm:
Yeah, I didn't really focus on the type annotations because I feel that the whole things needs a rewrite, preferably on top of the `packaging` APIs I'm currently working on.

setuptools/command/bdist_wheel.py
```diff
@@ -232,40 +235,47 @@ class bdist_wheel(Command):
...
+        self.egginfo_dir: str | None = None
         self.root_is_pure: bool | None = None
-        self.skip_build = None
+        self.skip_build = False
         self.relative = False
         self.owner = None
         self.group = None
         self.universal: bool = False
-        self.compression: str | int = "deflated"
+        self.compression: int = ZIP_DEFLATED
```
abravalheri, line 251: Umm... that seems to be a runtime "lie".

We know for sure that `self.compression` will have an `str` value temporarily assigned to it (if the `--compression` flag is given in `sys.argv`).

If a developer believes this info, it may generate problems later?

setuptools/command/bdist_wheel.py
```diff
@@ -232,40 +235,47 @@ class bdist_wheel(Command):
...
-            self.compression = self.supported_compressions[self.compression]
-        except KeyError:
-            raise ValueError(f"Unsupported compression: {self.compression}") from None
+        # Handle compression not being an int or a supported value
+        if not (
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            try:
+                self.compression = self.supported_compressions[str(self.compression)]
```
abravalheri, line 274: I can see how the `str(...)` cast is deceiving the typechecker to make it happy: so it believes that `self.compression` is an integer in runtime... but we know it can actually be a string...

setuptools/command/bdist_wheel.py
```diff
@@ -295,21 +305,21 @@ def finalize_options(self):
...
-            components += (self.build_number,)
+            components.append(self.build_number)
         return "-".join(components)
 
     def get_tag(self) -> tuple[str, str, str]:
         # bdist sets self.plat_name if unset, we should only use it for purepy
         # wheels if the user supplied it.
-        if self.plat_name_supplied:
-            plat_name = cast(str, self.plat_name)
+        if self.plat_name_supplied and self.plat_name:
```
abravalheri, line 321: Umm... that is a bit intriguing isn't it.

It seems that `self.plat_name_supplied` is only used once in the whole file, so why do we need it to cache the result of such simple information like `if self.plat_name`?

@agronholm do you have any thoughts about this?

agronholm: All of `bdist_wheel` code has been yucky, and there are lots of things that don't make sense, so I've refrained from refactoring it myself. This is just one of those examples. I don't really see why this extra member is needed, so your suggestion makes sense.

setuptools/command/bdist_wheel.py
agronholm: By developer, do you mean someone working with `bdist_wheel` code?

I feel that this change is ill advised. If anything, we should use a separate member for the "translated" compression option.

abravalheri: > By developer, do you mean someone working with bdist_wheel code?

Yes.

But I agree with you a better change would be separating into 2 different properties, or some arrangement similar to that.

setuptools/command/bdist_wheel.py
abravalheri: On further inspection, I see that the else statement do a separated check on `self.plat_name`.

So I suppose this variable distinguishes between "given by the user" and "programmatically set".

Avasam: > So I suppose this variable distinguishes between "given by the user" and "programmatically set".

That's what I figured looking at it too

setuptools/command/bdist_wheel.py
Avasam: It looks like it intended to *allow* using a `str` as the initial value for human readability, that gets converted into its equivalent compression type `int`.

The problem is that you can set `compression` manually after `finalize_options` but before `run`, which would result in an error if you use a `str`. Hence `compression` *cannot* be a str\* unless you expect the code using to to reconvert/recheck everytime.

The idea of allowing a string for human readability is kinda obsoleted by `zipfile` constants (ZIP_DEFLATED, ZIP_STORED, etc.).

Allowing a `compression` *parameter* as a string (like in `__init__` ) is fine, because you can convert it in that method before it ever reached the `compression` attribute. So I kept that feature in.

\* Whilst it can't be a str and have sane & safe typing, what it *could* be is a `Descriptor` (or `property` with setter):
- setter that accepts both an int or str and does the conversion imediatly
- getter that returns the int value

setuptools/command/bdist_wheel.py
Avasam: I think the idea of using a Descriptor or property would clean this up
https://github.com/pypa/setuptools/pull/4383/files#r1646513416

setuptools/command/bdist_wheel.py
abravalheri: > This would also move that logic out of `finalize_options`.

I think that is fine... it is only used once and it is a relatively simple computation, so there is no need for caching...

Maybe something like the following:
```diff
diff --git i/setuptools/command/bdist_wheel.py w/setuptools/command/bdist_wheel.py
index 7bfb3fb02..97aed9428 100644
--- i/setuptools/command/bdist_wheel.py
+++ w/setuptools/command/bdist_wheel.py
@@ -248,7 +248,7 @@ class bdist_wheel(Command):
         self.owner = None
         self.group = None
         self.universal: bool = False
-        self.compression: int = ZIP_DEFLATED
+        self.compression: str | int = "deflated"
         self.python_tag: str = python_tag()
         self.build_number: str | None = None
         self.py_limited_api: str | Literal[False] = False
@@ -265,18 +265,6 @@ class bdist_wheel(Command):
         self.data_dir = self.wheel_dist_name + ".data"
         self.plat_name_supplied = bool(self.plat_name)
 
-        # Handle compression not being an int or a supported value
-        if not (
-            isinstance(self.compression, int)
-            and self.compression in self.supported_compressions.values()
-        ):
-            try:
-                self.compression = self.supported_compressions[str(self.compression)]
-            except KeyError:
-                raise ValueError(
-                    f"Unsupported compression: {self.compression}"
-                ) from None
-
         need_options = ("dist_dir", "plat_name", "skip_build")
 
         self.set_undefined_options("bdist", *zip(need_options, need_options))
@@ -304,6 +292,19 @@ class bdist_wheel(Command):
         if self.build_number is not None and not self.build_number[:1].isdigit():
             raise ValueError("Build tag (build-number) must start with a digit.")
 
+    def _zip_compression(self) -> int:
+        # Handle compression not being an int or a supported value
+        if (
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            return self.compression
+
+        try:
+            return self.supported_compressions[str(self.compression)]
+        except KeyError:
+            raise ValueError(f"Unsupported compression: {self.compression}") from None
+
     @property
     def wheel_dist_name(self) -> str:
         """Return distribution full name with - replaced with _"""
@@ -443,7 +444,7 @@ class bdist_wheel(Command):
             os.makedirs(self.dist_dir)
 
         wheel_path = os.path.join(self.dist_dir, archive_basename + ".whl")
-        with WheelFile(wheel_path, "w", self.compression) as wf:
+        with WheelFile(wheel_path, "w", self._zip_compression()) as wf:
             wf.write_files(archive_root)
 
         # Add to 'Distribution.dist_files' so that the "upload" command works
```

Avasam: That works too. As you said, the only difference with the Descriptor/property approach I was suggesting is caching the value.

setuptools/command/bdist_wheel.py
```diff
@@ -597,3 +595,18 @@ def adios(p: str) -> None:
...
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            return self.compression
+
+        try:
+            if isinstance(self.compression, str):
+                return self.supported_compressions[self.compression]
+        except KeyError:
+            pass
```
Avasam, line 610: This could be done w/o a try-except.

```suggestion
        if isinstance(self.compression, str):
            compression = self.supported_compressions.get(self.compression)
            if compression is not None:
                return compression
```

And just for fun, using the walrus operator (which I don't like)
```suggestion
        if isinstance(self.compression, str) and (
            compression := self.supported_compressions.get(self.compression)
        ):
            return compression
```

setuptools/command/bdist_wheel.py
```diff
@@ -597,3 +595,18 @@ def adios(p: str) -> None:
...
+    def _zip_compression(self) -> int:
+        if (
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            return self.compression
+
+        try:
+            if isinstance(self.compression, str):
+                return self.supported_compressions[self.compression]
```
Avasam, line 608: I had previously done
```suggestion
            return self.supported_compressions[str(self.compression)]
```
To make the type-checker happy (even through this is wrapped in a try-except anyway). Which technically allows using something else than a str as long as it's stringified to a valid compression str. But that's extremely niche.

setuptools/command/bdist_wheel.py
Avasam, line 610: I had previously done

```suggestion
        compression = self.supported_compressions.get(self.compression)
        if compression is not None:
            return compression
```

To make the type-checker happy (even through this is wrapped in a try-except anyway). Which technically allows using something else than a str as long as it's stringified to a valid compression str. But that's extremely niche. Might be desired if that *feels* cleaner, up to you 🤷

mypy.ini
```diff
@@ -15,7 +15,7 @@ exclude = (?x)(
 	| ^setuptools/_vendor/ # Vendored
 	| ^setuptools/_distutils/ # Vendored
 	| ^setuptools/config/_validate_pyproject/ # Auto-generated
-    | ^setuptools/tests/bdist_wheel_testdata/  # Duplicate module name
+	| ^setuptools/tests/bdist_wheel_testdata/  # Duplicate module name
```
abravalheri, line 18: This directory no longer exists.

```suggestion
```

setuptools/command/bdist_wheel.py
abravalheri: Ok, so we do need the `if` guard here according to the last result of the CI.

setuptools/command/bdist_wheel.py
```diff
@@ -595,3 +593,16 @@ def adios(p: str) -> None:
...
+    def _zip_compression(self) -> int:
+        if (
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            return self.compression
+
+        compression = self.supported_compressions.get(self.compression)
+        if compression is not None:
+            return compression
```
abravalheri, line 606: Ok, so let's walrus our way out of it (I can see that the walrus is used in `distutils` and other code bases maintained by Jason).

```suggestion
        if isinstance(self.compression, str) and (
            compression := self.supported_compressions.get(self.compression)
        ):
            return compression
```

setuptools/command/bdist_wheel.py
```diff
@@ -595,3 +593,17 @@ def adios(p: str) -> None:
...
+        if (
+            isinstance(self.compression, int)
+            and self.compression in self.supported_compressions.values()
+        ):
+            return self.compression
+
+        if isinstance(self.compression, str) and (
+            compression := self.supported_compressions.get(self.compression)
+        ):
+            return compression
```
abravalheri, line 607: I am really puzzled by the error in the CI:

```
>       raise ValueError(f"Unsupported compression: {self.compression!r}")
E       ValueError: Unsupported compression: 'stored'
```

Maybe ZIP_STORED is 0?

Let's see if the following change fixes it:

```suggestion
        compression = self.supported_compressions.get(str(self.compression))
        if compression is not None:
            return compression
```

abravalheri:
Thank you very much!

PR #4386, opened 2024-05-23
Enforce ruff/refurb rules 

DimitriPapadopoulos:
## Summary of changes

Enforce ruff/refurb rules (`FURB`).

Follow-up of #4172.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


Avasam:
Looks like you're missing imports for `operator`

setuptools/tests/test_wheel.py
```diff
@@ -115,11 +115,11 @@ def build_wheel(extra_file_defs=None, **kwargs):
...
-    contents = set()
-    for dirpath, dirnames, filenames in os.walk(root):
-        for filename in filenames:
-            contents.add(os.path.join(os.path.relpath(dirpath, root), filename))
-    return contents
+    return set(
+        os.path.join(os.path.relpath(dirpath, root), filename)
+        for dirpath, dirnames, filenames in os.walk(root)
+        for filename in filenames
+    )
```
Avasam, line 122: From personal experience, list comprehensions are just faster than generators if you know you'll iterate over all items.

```suggestion
    return set([
        os.path.join(os.path.relpath(dirpath, root), filename)
        for dirpath, dirnames, filenames in os.walk(root)
        for filename in filenames
    ])
```

DimitriPapadopoulos: I have applied **C4** rules as well in this PR.

Avasam: Aren't you concerned enabling all of `flake-8 comprehensions` is gonna further delay the review of this PR?
I'm pretty sure @jaraco would prefer [`lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments = true`](https://docs.astral.sh/ruff/settings/#lint_flake8-comprehensions_allow-dict-calls-with-keyword-arguments) for [unnecessary-collection-call (C408)](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/#unnecessary-collection-call-c408)

Avasam: Aren't you concerned enabling all of `flake-8 comprehensions` is gonna further delay the review of this PR?
For isntance, I'm pretty sure @jaraco would prefer [`lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments = true`](https://docs.astral.sh/ruff/settings/#lint_flake8-comprehensions_allow-dict-calls-with-keyword-arguments) for [unnecessary-collection-call (C408)](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/#unnecessary-collection-call-c408)

DimitriPapadopoulos: I'm not in a hurry, but I probably went overboard :smile: Should I apply `C401` and `C403` only from the C4 set?

Avasam: Imo it's probably safer to just recognize that this line can be improved further. Either with a single manual change, or as a follow-up to keep all changes here automated.

Especially if we provide a PR to enable `C4` separatly, this will kinda just fix itself.

But that's just my opinion from previous contributions, I'm not a maintainer ^^"

DimitriPapadopoulos: Agreed. I'll remove `C4` rules and keep suboptimal automatic `FURB` fixes — they will be fixed by an upcoming PR enforcing `C4` rules.

Avasam:
@DimitriPapadopoulos I think you're missing a `ruff.toml` change adding `FURB`

Avasam:
Now that this is back to being a simpler PR. @abravalheri / @jaraco refurb would also have been one of my suggestions with improvements in readability and simplifications.

setuptools/command/egg_info.py
```diff
@@ -475,8 +475,7 @@ def global_exclude(self, pattern):
         return self._remove_files(match.match)
 
     def append(self, item) -> None:
-        if item.endswith('\r'):  # Fix older sdists built on Windows
-            item = item[:-1]
+        item = item.removesuffix('\r')
```
Avasam, line 478: Might wanna keep this comment, I'm guessing it was removed by an unsafe autofix
```suggestion
        item = item.removesuffix('\r')  # Fix older sdists built on Windows
```

DimitriPapadopoulos:
Rebased.

DimitriPapadopoulos:
* The Python 3.14 CI errors might be related to https://github.com/python/cpython/issues/133653.
* Not sure about the PyPy 3.10 CI error, but it seems unrelated to this PR as well.

PR #4355, opened 2024-05-12
Generic typing for register methods in `pkg_resources`

Avasam:
## Summary of changes

*Review and merge after https://github.com/pypa/setuptools/pull/4246*
This is an immediate follow-up but with more precise and complex typing.
works towards step 3.2 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

Once this #4391 and #4390 are merged, `pkg_setuptools` will be on par with Typeshed

### Pull Request Checklist
- [x] Changes have tests (static type checking)
- [ ] News fragment added in [`newsfragments/`]. (no user facing changes yet until pkg_resources is considered typed)
  _(See [documentation][PR docs] for details)_

[`newsfragments/`]: https://github.com/pypa/setuptools/tree/master/newsfragments
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #4390, opened 2024-05-23
pkg_resources: Merge `@overload` and `TypeVar` annotations from typeshed

Avasam:
## Summary of changes

Step 3.2 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

Once this #4355 and #4391 are merged, `pkg_setuptools` will be on par with Typeshed

### Pull Request Checklist
- [x] Changes have tests (type-checking)
- [x] News fragment added in [`newsfragments/`]. (no user facing changes yet until pkg_resources is considered typed)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -122,8 +122,9 @@ _MetadataType = Union["IResourceProvider", None]
 _ResolvedEntryPoint = Any  # Can be any attribute in the module
 _ResourceStream = Any  # TODO / Incomplete: A readable file-like object
 # Any object works, but let's indicate we expect something like a module (optionally has __loader__ or __file__)
-_ModuleLike = Any
-_ProviderFactoryType = Callable[[_ModuleLike], "IResourceProvider"]
+_ModuleLike = Union[object, types.ModuleType]
+# Any: Should be _ModuleLike but we end up with issues where _ModuleLike doesn't have _ZipLoaderModule's __loader__
+_ProviderFactoryType = Callable[[Any], "IResourceProvider"]
 _DistFinderType = Callable[[_T, str, bool], Iterable["Distribution"]]
 _NSHandlerType = Callable[[_T, str, str, types.ModuleType], Union[str, None]]
 _AdapterT = TypeVar(
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -34,6 +34,8 @@ import re
 import types
 from typing import (
     Any,
+    Iterator,
+    Literal,
     Mapping,
     MutableSequence,
     NamedTuple,
@@ -98,7 +100,7 @@ from pkg_resources.extern.packaging import version as _packaging_version
 from pkg_resources.extern.platformdirs import user_cache_dir as _user_cache_dir
 
 if TYPE_CHECKING:
-    from _typeshed import StrPath, StrOrBytesPath, BytesPath
+    from _typeshed import BytesPath, StrPath, StrOrBytesPath
 
 warnings.warn(
     "pkg_resources is deprecated as an API. "
@@ -109,14 +111,18 @@ warnings.warn(
 
 
 _T = TypeVar("_T")
+_DistributionT = TypeVar("_DistributionT", bound="Distribution")
 # Type aliases
 _NestedStr = Union[str, Iterable[Union[str, Iterable["_NestedStr"]]]]
+_InstallerTypeT = Callable[["Requirement"], "_DistributionT"]
 _InstallerType = Callable[["Requirement"], Optional["Distribution"]]
 _PkgReqType = Union[str, "Requirement"]
 _EPDistType = Union["Distribution", _PkgReqType]
 _MetadataType = Optional["IResourceProvider"]
+_ResolvedEntryPoint = Any  # Can be any attribute in the module
+_ResourceStream = Any  # TODO / Incomplete: A readable file-like object
 # Any object works, but let's indicate we expect something like a module (optionally has __loader__ or __file__)
-_ModuleLike = Union[object, types.ModuleType]
+_ModuleLike = Any
 _ProviderFactoryType = Callable[[_ModuleLike], "IResourceProvider"]
 _DistFinderType = Callable[[_T, str, bool], Iterable["Distribution"]]
 _NSHandlerType = Callable[[_T, str, str, types.ModuleType], Optional[str]]
@@ -128,6 +134,10 @@ _AdapterT = TypeVar(
 # Use _typeshed.importlib.LoaderProtocol once available https://github.com/python/typeshed/pull/11890
 class _LoaderProtocol(Protocol):
     def load_module(self, fullname: str, /) -> types.ModuleType: ...
+
+
+class _ZipLoaderModule(Protocol):
+    __loader__: zipimport.zipimporter
 
 
 _PEP440_FALLBACK = re.compile(r"^v?(?P<safe>(?:[0-9]+!)?[0-9]+(?:\.[0-9]+)*)", re.I)
@@ -403,7 +413,11 @@ def register_loader_type(
     _provider_factories[loader_type] = provider_factory
 
 
-def get_provider(moduleOrReq: str | Requirement):
+@overload
+def get_provider(moduleOrReq: str) -> IResourceProvider: ...
+@overload
+def get_provider(moduleOrReq: Requirement) -> Distribution: ...
+def get_provider(moduleOrReq: str | Requirement) -> IResourceProvider | Distribution:
     """Return an IResourceProvider for the named module or requirement"""
     if isinstance(moduleOrReq, Requirement):
         return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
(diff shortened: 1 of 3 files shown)


pkg_resources/__init__.py
```diff
@@ -126,6 +133,10 @@ class _LoaderProtocol(Protocol):
     def load_module(self, fullname: str, /) -> types.ModuleType: ...
 
 
+class ZipLoaderModule(Protocol):
```
abravalheri, line 136: If we are introducing a new name here, better to make explicitly private, as not everyone pays attention to `__all__` (unfortunately Hyrum's law hits hard on setuptools/pkg_resources).

pkg_resources/__init__.py
```diff
@@ -109,13 +112,17 @@
...
 
 
-T = TypeVar("T")
+_T = TypeVar("_T")
+_DistributionT = TypeVar("_DistributionT", bound="Distribution")
 # Type aliases
 _NestedStr = Union[str, Iterable[Union[str, Iterable["_NestedStr"]]]]
-_InstallerType = Callable[["Requirement"], Optional["Distribution"]]
+_StrictInstallerType = Callable[["Requirement"], "_DistributionT"]
+_InstallerType = Optional[Callable[["Requirement"], Optional["Distribution"]]]
```
abravalheri, line 120: It feels like we should defer adding `Optional` to the end usage.
This way the resulting type annotation would be more obvious.

```suggestion
_InstallerType = Callable[["Requirement"], Optional["Distribution"]]
```

pkg_resources/__init__.py
```diff
@@ -1118,13 +1214,29 @@ def add(self, dist: "Distribution"):
...
+        installer: _InstallerType = None,
+        replace_conflicting: bool = False,
+    ) -> Optional["Distribution"]: ...
+    def best_match(
+        self,
+        req: "Requirement",
+        working_set: WorkingSet,
+        installer: Union[_InstallerType, "_StrictInstallerType[_DistributionT]"] = None,
+        replace_conflicting: bool = False,
+    ) -> Optional[Union["Distribution", _DistributionT]]:
```
abravalheri, line 1239: Doesn't `_DistritubtionT` already covers `Distribution`?

pkg_resources/__init__.py
```diff
@@ -768,14 +802,42 @@ def add(
...
+        extras: Optional[Tuple[str, ...]] = None,
+    ) -> List["Distribution"]: ...
+    def resolve(
+        self,
+        requirements: Iterable["Requirement"],
+        env: Optional["Environment"] = None,
+        installer: Union[_InstallerType, "_StrictInstallerType[_DistributionT]"] = None,
+        replace_conflicting: bool = False,
+        extras: Optional[Tuple[str, ...]] = None,
+    ) -> Union[List["Distribution"], List[_DistributionT]]:
```
abravalheri, line 840: Would it be easier to make `_DistributionT` covariant?

abravalheri (commented):
Thank you very much for working on this @Avasam.

pkg_resources/__init__.py
```diff
@@ -109,14 +111,18 @@
...
+_InstallerTypeT = Callable[["Requirement"], "_DistributionT"]
 _InstallerType = Callable[["Requirement"], Optional["Distribution"]]
 _PkgReqType = Union[str, "Requirement"]
 _EPDistType = Union["Distribution", _PkgReqType]
 _MetadataType = Optional["IResourceProvider"]
+_ResolvedEntryPoint = Any  # Can be any attribute in the module
+_ResourceStream = Any  # TODO / Incomplete: A readable file-like object
 # Any object works, but let's indicate we expect something like a module (optionally has __loader__ or __file__)
-_ModuleLike = Union[object, types.ModuleType]
+_ModuleLike = Any
```
abravalheri, line 125: This annotation looks worse than the previous one, no?

pkg_resources/__init__.py
```diff
@@ -1582,6 +1703,8 @@ def run_script(self, script_name: str, namespace: dict[str, Any]):
                     **locals()
                 ),
             )
+        if not self.egg_info:
+            raise TypeError("Provider is missing egg_info", self.egg_info)
```
abravalheri, line 1707: This if never runs, right?

The reasoning is that when `self.egg_info` is falsey, then `self.has_metadata(script)` is also falsey, so the first exception always run and the if is never reached...

Avasam: This check could be hit if a subclass overwrote `has_metadata` in a way that doesn't garrantee `self.egg_info` is truthy or that `has_metadata` called by `run_script` will always fail (like `FileMetadata` could hit it if it didn't have the `name == 'PKG-INFO'` check since the base `run_script` forces `name` to start with `scripts/`.

There's just no way to statically communicate that `has_metadata` checks for `self.egg_info` (it could be a type guard if `egg_info` was a parameter of `has_metadata`).

Funnily enough, the `Metadata` test class in `pkg_resources/tests/test_resources.py` shows exactly how this can be hit:
```py
>>> from pkg_resources.tests.test_resources import Metadata
>>> Metadata(["scripts/",""]).run_script("", {})
  File "setuptools\pkg_resources\__init__.py", line 1708, in run_script
    script_filename = self._fn(self.egg_info, script)
  File "setuptools\pkg_resources\__init__.py", line 1744, in _fn
    return os.path.join(base, *resource_name.split('/'))
  File "C:\Program Files\Python39\lib\ntpath.py", line 78, in join
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```


I can alternatively do `if not self.has_metadata(script) or not self.egg_info:`, but I went with a `TypeError` to keep the exact same behaviour just in case someone downstream did override `has_metadata` in a way that triggers this error.

pkg_resources/__init__.py
abravalheri: Ummm 🤔.

I would understand if `ZipProvider` was a subclass of `IResourceProvider`... however `IResourceProvider` is a protocol. That should eliminate the problem with the variance, no?

Avasam: Pyright gives us a more detailed error here:
```
Argument of type "type[ZipProvider]" cannot be assigned to parameter "provider_factory" of type "_ProviderFactoryType" in function "register_loader_type"
  Type "type[ZipProvider]" is incompatible with type "_ProviderFactoryType"
    Parameter 1: type "_ModuleLike" is incompatible with type "_ZipLoaderModule"
      Type "_ModuleLike" is incompatible with type "_ZipLoaderModule"
        "object" is incompatible with protocol "_ZipLoaderModule"
          "__loader__" is not present [reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

So the following might indeed be more appropriate:
```py
_ModuleLike = Union[object, types.ModuleType]
# Any: Should be _ModuleLike but we end up with variance issues
_ProviderFactoryType = Callable[[Any], "IResourceProvider"]
```

pkg_resources/__init__.py
abravalheri: Ok, I see, the problem is related to the way the implementation is forcing `os.path.join(self.egg_info, ...)` because in theory if `has_metadata/get_metadata` are implemented by a subclass in a way that does not depend `egg_info`, that would be fine...

I suppose that if the subclass overwrites `_fn` so that `os.path.join` is not called, then it would also be OK.

pkg_resources/__init__.py
abravalheri: That would also affect the `_provider_factories` variable right?

Would it make sense to add `_ZipLoaderModule` in the `_ModuleType` union?

pkg_resources/__init__.py
Avasam: This also brings the question, could `egg_info` be initialized to an empty string instead of None ? All of setuptool's code checks for Truthynes of it, never for an explicit `None`. It's also an issue that will come up again as much code that currently isn't type-checked by mypy (functions are missing annotations) or my pyright branch (error currently disabled to pass as-is initially) makes the assumption that `egg_info` is not None w/o checking.

That would technically count as a breaking change downstream though (if someone did `x.egg_info is not None`). So that can be a conversation for later.

abravalheri: I feel that we should not add this exception in this part of the code.

If the major concern is that inheritance can be used so that `self.egg_info` is `None`. Then inheritance also can be used to change `def _fn` to handle the `None` case right? (and all the other methods that rely on `self.egg_info`).

Effectively adding this hard check here is a backwards incompatible change...

---
If inheritance is out of the table, then `self.egg_info` is ensure to not be `None` there.

> This also brings the question, could egg_info be initialized to an empty string instead of None ?

Haven't tested that. Effectively Python functions treat the `""` directory as CWD right? So maybe it would work... but I don't know.

Avasam: Last commit applies your suggestion.

We could consider coercing `None` to `""` in `_fn` since that seems to follow the logic where most code checks for an empty string first, and `run_script` was already letting and empty string pass. But is technically a change in behaviour.

Avasam: Alternatively if this is the only point of contention left for this PR, I can undo the parameter change for `_fn` since it's not public and will get the checker passing. Splitting off the conversation about what to do with a none `egg_info` in `run_script` or `None` base in `_fn`

pkg_resources/__init__.py
```diff
@@ -1613,12 +1736,12 @@ def _isdir(self, path) -> bool:
...
         )
 
-    def _listdir(self, path):
+    def _listdir(self, path) -> list[str]:
         raise NotImplementedError(
             "Can't perform this operation for unregistered loader type"
         )
 
-    def _fn(self, base, resource_name: str):
+    def _fn(self, base: str, resource_name: str):
```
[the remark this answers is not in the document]
Avasam: https://github.com/pypa/setuptools/pull/4390#discussion_r1649195623

pkg_resources/__init__.py
abravalheri: Thank you!

I think we can leave the implementation as per your last commit indefinitely.

PR #4391, opened 2024-05-23
pkg_resources: fix incorrect implicit return types

Avasam:
## Summary of changes

Checked all public methods (`def ([^_]|__)`) in `pkg_resources` for their implied return type using pyright/pylance. Updated annotations for those that seemed incorrect or returned a partial `Any` even after #4390 .

Step 3.3 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

Once this #4355 and #4390 are merged, `pkg_setuptools` will be on par with Typeshed

### Pull Request Checklist
- [x] Changes have tests (type-checking and existing basic tests for syntax errors)
- [x] News fragment added in [`newsfragments/`]. (no user facing changes yet until pkg_resources is considered typed)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -34,6 +34,8 @@ import re
 import types
 from typing import (
     Any,
+    Dict,
+    Iterator,
     Mapping,
     MutableSequence,
     NamedTuple,
@@ -45,8 +47,6 @@ from typing import (
     Callable,
     Iterable,
     TypeVar,
-    Optional,
-    Dict,
     overload,
 )
 import zipfile
@@ -98,6 +98,7 @@ from pkg_resources.extern.packaging import version as _packaging_version
 from pkg_resources.extern.platformdirs import user_cache_dir as _user_cache_dir
 
 if TYPE_CHECKING:
+    from typing_extensions import Self
     from _typeshed import StrPath, StrOrBytesPath, BytesPath
 
 warnings.warn(
@@ -111,15 +112,15 @@ warnings.warn(
 _T = TypeVar("_T")
 # Type aliases
 _NestedStr = Union[str, Iterable[Union[str, Iterable["_NestedStr"]]]]
-_InstallerType = Callable[["Requirement"], Optional["Distribution"]]
+_InstallerType = Callable[["Requirement"], Union["Distribution", None]]
 _PkgReqType = Union[str, "Requirement"]
 _EPDistType = Union["Distribution", _PkgReqType]
-_MetadataType = Optional["IResourceProvider"]
+_MetadataType = Union["IResourceProvider", None]
 # Any object works, but let's indicate we expect something like a module (optionally has __loader__ or __file__)
 _ModuleLike = Union[object, types.ModuleType]
 _ProviderFactoryType = Callable[[_ModuleLike], "IResourceProvider"]
 _DistFinderType = Callable[[_T, str, bool], Iterable["Distribution"]]
-_NSHandlerType = Callable[[_T, str, str, types.ModuleType], Optional[str]]
+_NSHandlerType = Callable[[_T, str, str, types.ModuleType], Union[str, None]]
 _AdapterT = TypeVar(
     "_AdapterT", _DistFinderType[Any], _ProviderFactoryType, _NSHandlerType[Any]
 )
@@ -151,7 +152,7 @@ def _declare_state(vartype: str, varname: str, initial_value: _T) -> _T:
     return initial_value
 
 
-def __getstate__():
+def __getstate__() -> dict[str, Any]:
     state = {}
     g = globals()
     for k, v in _state_vars.items():
@@ -159,7 +160,7 @@ def __getstate__():
     return state
 
 
-def __setstate__(state):
+def __setstate__(state: dict[str, Any]) -> dict[str, Any]:
     g = globals()
     for k, v in state.items():
         g['_sset_' + _state_vars[k]](k, g[k], v)
@@ -314,11 +315,11 @@ class VersionConflict(ResolutionError):
     _template = "{self.dist} is installed but {self.req} is required"
 
     @property
-    def dist(self):
+    def dist(self) -> Distribution:
         return self.args[0]
 
     @property
-    def req(self):
+    def req(self) -> Requirement:
         return self.args[1]
 
     def report(self):
@@ -344,7 +345,7 @@ class ContextualVersionConflict(VersionConflict):
     _template = VersionConflict._template + ' by {self.required_by}'
 
     @property
-    def required_by(self):
+    def required_by(self) -> set[str]:
         return self.args[2]
 
 
@@ -357,11 +358,11 @@ class DistributionNotFound(ResolutionError):
     )
 
     @property
-    def req(self):
+    def req(self) -> Requirement:
         return self.args[0]
 
     @property
-    def requirers(self):
+    def requirers(self) -> set[str] | None:
         return self.args[1]
 
     @property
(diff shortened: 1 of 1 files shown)


pkg_resources/__init__.py
```diff
@@ -99,7 +101,16 @@
...
+    from typing_extensions import Self
+    from _typeshed import StrPath, StrOrBytesPath
+    from itertools import chain
+
+    # yield_lines is exported, we should type it, but extern is too dynamic
+    def yield_lines(iterable: "_NestedStr") -> chain[str]: ...
+
+    # Trick type-checkers into seeing the original instances for type comparisons
+    from packaging import version as _packaging_version
+    from packaging import requirements as _packaging_requirements
```
abravalheri, line 113: I think we should defer any changes like this until we have a decision about how to reconcile `extern` and type-cheking.

```python
if TYPE_CHECKING:
    # ...
    def yield_lines(iterable: "_NestedStr") -> chain[str]: ...

    # Trick type-checkers into seeing the original instances for type comparisons
    from packaging import version as _packaging_version
    from packaging import requirements as _packaging_requirements
```

pkg_resources/__init__.py
```diff
@@ -691,7 +702,9 @@ def find(self, req: "Requirement"):
             raise VersionConflict(dist, req)
         return dist
 
-    def iter_entry_points(self, group: str, name: Optional[str] = None):
+    def iter_entry_points(
+        self, group: str, name: Optional[str] = None
+    ) -> Generator["EntryPoint", None, None]:
```
abravalheri, line 707: Should we do `Iterator["EntryPoint"]` here?

I find the `Generator[..., None, None]` annotation very distracting (personal opnion).
Also the name of the method is `iter_*`...

In general I find it conceptually easier to deal with `Iterator[...]`.
It is immediately obvious what to do with the value.

`Generator[...]` on the other hand needs you to remember which position regards the return type, which position regards the yield type and which position regards the send type.

(Again this is a personal opinion)

pkg_resources/__init__.py
```diff
@@ -2230,13 +2247,13 @@ def safe_listdir(path):
     return ()
 
 
-def distributions_from_metadata(path):
+def distributions_from_metadata(path) -> Generator["Distribution", None, None]:
```
abravalheri, line 2250: Same as previous comment.

pkg_resources/__init__.py
```diff
@@ -3351,7 +3371,9 @@ def _bypass_ensure_directory(path):
             pass
 
 
-def split_sections(s: _NestedStr):
+def split_sections(
+    s: _NestedStr,
+) -> Generator[Tuple[Optional[str], List[str]], None, None]:
```
abravalheri, line 3376: Same as previous comment.

abravalheri (commented):
Thank you very much @Avasam.

Other than the open question of what to do with `extern`, I think all the other changes are good.

pkg_resources/__init__.py
Avasam: For `Generator[type, None None]` sure.
There's precedence even in typeshed and enforced in Flake8-PYI to consider `__iter__` methods that return `Generators` as an implementation detail and to use `Iterator` instead.

Would you like me to change all of the explicit generator return types that have that signature?

pkg_resources/__init__.py
Avasam: Even if we did, this part would be useful. At least the two imports.

Even if static type-checkers understood the `extern` imports, they would be treated as a different type than the originals:

```py
from pkg_resources.parse_version
from packaging import Version

def foo(bar: Version): ...
foo(parse_version()):... # type error
```
(I'm on my phone typing from memory, will update this comment with actual type issue soon).

The re-exported `yield_lines` method isn't typed either iirc. (Although that can be fixed upstream)

Or would you prefer keeping even that for later, adding on top of the `extra` discussion ?

pkg_resources/__init__.py
Avasam: (for the sake of this PR I'll split it off, but there's almost nothing left to do to mark `pkg_resources` as py.typed, so that's definitely something I'd like to handle as soon as all PRs linked in this PR's description are merged)

abravalheri (approved):

pkg_resources/__init__.py
```diff
@@ -557,7 +558,7 @@ def metadata_isdir(self, name: str) -> bool:
         """Is the named metadata a directory?  (like ``os.path.isdir()``)"""
 
     def metadata_listdir(self, name: str):
-        """List of metadata names in the directory (like ``os.listdir()``)"""
+        """list of metadata names in the directory (like ``os.listdir()``)"""
```
abravalheri, line 561: This one and the next are a bit unexpected 😅

pkg_resources/__init__.py
```diff
@@ -780,7 +781,7 @@ def resolve(
         replace_conflicting: bool = False,
         extras: tuple[str, ...] | None = None,
     ):
-        """List all distributions needed to (recursively) meet `requirements`
+        """list all distributions needed to (recursively) meet `requirements`
```
abravalheri, line 784: Also unexpected 😅

pkg_resources/__init__.py
```diff
@@ -1257,7 +1258,7 @@ def resource_string(
         )
 
     def resource_listdir(self, package_or_requirement: _PkgReqType, resource_name: str):
-        """List the contents of the named resource directory"""
+        """list the contents of the named resource directory"""
```
abravalheri, line 1261: Same as previous comment

pkg_resources/__init__.py
```diff
@@ -1363,7 +1364,7 @@ def postprocess(self, tempname: StrOrBytesPath, filename: StrOrBytesPath):
             os.chmod(tempname, mode)
 
     def set_extraction_path(self, path: str):
-        """Set the base path where resources will be extracted to, if needed.
+        """set the base path where resources will be extracted to, if needed.
```
abravalheri, line 1367: Same as previous.

pkg_resources/__init__.py
```diff
@@ -2924,9 +2926,9 @@ def _build_dep_map(self):
         return dm
 
     def requires(self, extras: Iterable[str] = ()):
-        """List of Requirements needed for this distro if `extras` are used"""
+        """list of Requirements needed for this distro if `extras` are used"""
```
abravalheri, line 2929: Same as previous.

pkg_resources/__init__.py
```diff
@@ -3430,7 +3436,7 @@ def _call_aside(f, *args, **kwargs):
 
 @_call_aside
 def _initialize(g=globals()):
-    "Set up global resource manager (deliberately not state-saved)"
+    "set up global resource manager (deliberately not state-saved)"
```
abravalheri, line 3439: Same as previous.

pkg_resources/__init__.py
Avasam:
```suggestion
        """List of metadata names in the directory (like ``os.listdir()``)"""
```

pkg_resources/__init__.py
```diff
@@ -590,7 +591,7 @@ def resource_isdir(self, resource_name: str):
         """Is the named resource a directory?  (like ``os.path.isdir()``)"""
 
     def resource_listdir(self, resource_name: str):
-        """List of resource names in the directory (like ``os.listdir()``)"""
+        """list of resource names in the directory (like ``os.listdir()``)"""
```
Avasam, line 594:
```suggestion
        """List of resource names in the directory (like ``os.listdir()``)"""
```

pkg_resources/__init__.py
Avasam:
```suggestion
        """List all distributions needed to (recursively) meet `requirements`
```

pkg_resources/__init__.py
Avasam:
```suggestion
        """List the contents of the named resource directory"""
```

pkg_resources/__init__.py
Avasam:
```suggestion
        """List of Requirements needed for this distro if `extras` are used"""
```

pkg_resources/__init__.py
Avasam:
```suggestion
        """Set the base path where resources will be extracted to, if needed.
```

pkg_resources/__init__.py
Avasam:
```suggestion
    "Set up global resource manager (deliberately not state-saved)"
```

abravalheri:
Thank you!

Issue #4137, opened 2023-11-24
Deprecate distutils_hack and support for 'import distutils' before 'import setuptools'


jaraco:
With the release of Python 3.12 without distutils in the stdlib, it's now a good time to consider dropping support for `import distutils` and all reliance of distutils from the stdlib (i.e. `SETUPTOOLS_USE_DISTUTILS=stdlib`). Let's start by deprecating that code path.

PR #4392, opened 2024-05-23
Sort imports for more consistent diff (ruff/isort)

Avasam:
## Summary of changes

As title says.
I've made sure that `_distutils_hack` should be sorted before `_distutils`, and that `setuptools` gets sorted before `distutils` in tests as long as the hack is supported (ref: https://github.com/pypa/setuptools/issues/4137 )

### Pull Request Checklist
- [x] Changes have tests
- [x] News fragment added in [`newsfragments/`]. (no user-facing changes)

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -22,68 +22,66 @@ This module is deprecated. Users are directed to :mod:`importlib.resources`,
 
 from __future__ import annotations
 
-from abc import ABC
 import sys
+from abc import ABC
 
 if sys.version_info < (3, 8):  # noqa: UP036 # Check for unsupported versions
     raise RuntimeError("Python 3.8 or later is required")
 
-import os
+import _imp
+import collections
+import email.parser
+import errno
+import functools
+import importlib
+import importlib.abc
+import importlib.machinery
+import inspect
 import io
-import time
+import ntpath
+import operator
+import os
+import pkgutil
+import platform
+import plistlib
+import posixpath
 import re
+import stat
+import tempfile
+import textwrap
+import time
 import types
+import warnings
+import zipfile
+import zipimport
+from pkgutil import get_importer
 from typing import (
+    TYPE_CHECKING,
     Any,
     BinaryIO,
-    Literal,
+    Callable,
     Dict,
+    Iterable,
     Iterator,
+    Literal,
     Mapping,
     MutableSequence,
     NamedTuple,
     NoReturn,
-    Tuple,
-    Union,
-    TYPE_CHECKING,
     Protocol,
-    Callable,
-    Iterable,
+    Tuple,
     TypeVar,
+    Union,
     overload,
 )
-import zipfile
-import zipimport
-import warnings
-import stat
-import functools
-import pkgutil
-import operator
-import platform
-import collections
-import plistlib
-import email.parser
-import errno
-import tempfile
-import textwrap
-import inspect
-import ntpath
-import posixpath
-import importlib
-import importlib.abc
-import importlib.machinery
-from pkgutil import get_importer
-
-import _imp
 
 sys.path.extend(((vendor_path := os.path.join(os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor')) not in sys.path) * [vendor_path])  # fmt: skip
 # workaround for #4476
 sys.modules.pop('backports', None)
 
 # capture these to bypass sandboxing
-from os import utime
-from os import open as os_open
-from os.path import isdir, split
+from os import open as os_open, utime  # isort: skip
+from os.path import isdir, split  # isort: skip
 
 try:
     from os import mkdir, rename, unlink
@@ -93,20 +91,16 @@ except ImportError:
     # no write support, probably under GAE
     WRITE_SUPPORT = False
 
-from jaraco.text import (
-    yield_lines,
-    drop_comment,
-    join_continuation,
-)
 import packaging.markers
 import packaging.requirements
 import packaging.specifiers
 import packaging.utils
 import packaging.version
+from jaraco.text import drop_comment, join_continuation, yield_lines
 from platformdirs import user_cache_dir as _user_cache_dir
 
 if TYPE_CHECKING:
-    from _typeshed import BytesPath, StrPath, StrOrBytesPath
+    from _typeshed import BytesPath, StrOrBytesPath, StrPath
     from typing_extensions import Self, TypeAlias
 
 warnings.warn(
--- a/ruff.toml
+++ b/ruff.toml
@@ -14,6 +14,7 @@ extend-select = [
 	"ANN2", # missing-return-type-*
 	"FA", # flake8-future-annotations
 	"F404", # late-future-import
+	"I", # isort
 	"PYI", # flake8-pyi
 	"UP", # pyupgrade
 	"TRY",
@@ -55,6 +56,16 @@ ignore = [
 # Suppress nuisance warnings about module-import-not-at-top-of-file (E402) due to workaround for #4476
 "setuptools/__init__.py" = ["E402"]
 "pkg_resources/__init__.py" = ["E402"]
+
+[lint.isort]
+combine-as-imports = true
+split-on-trailing-comma = false
+# Force Ruff/isort to always import setuptools before distutils in tests as long as distutils_hack is supported
+# This also ensures _distutils_hack is imported before distutils
+# https://github.com/pypa/setuptools/issues/4137
+section-order = ["future", "standard-library", "eager", "third-party", "first-party", "local-folder", "delayed"]
+sections.eager = ["_distutils_hack"]
+sections.delayed = ["distutils"]
 
 [lint.flake8-annotations]
 ignore-fully-untyped = true
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -10,6 +10,23 @@ from glob import iglob
 from pathlib import Path
 from typing import TYPE_CHECKING, MutableMapping
 
+from more_itertools import partition, unique_everseen
+from ordered_set import OrderedSet
+from packaging.markers import InvalidMarker, Marker
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
+from packaging.version import Version
+
+from . import (
+    _entry_points,
+    _reqs,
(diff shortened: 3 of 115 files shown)


ruff.toml
```diff
@@ -42,6 +43,17 @@ ignore = [
...
+[lint.isort]
+combine-as-imports = true
+split-on-trailing-comma = false
+# Force Ruff/isort to always import setuptools before distutils in tests as long as distutils_hack is supported
+# This also ensures _distutils_hack is imported before distutils
+# https://github.com/pypa/setuptools/issues/4137
+known-third-party = ["setuptools"]
+known-first-party = ["distutils"]
+# Once support is dropped, move distutils to first-party as setuptools repo now owns it
+# known-first-party = ["distutils"]
```
Avasam, line 55:
```suggestion
```

pkg_resources/__init__.py
```diff
@@ -27,61 +27,59 @@
...
 sys.path.extend(((vendor_path := os.path.join(os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor')) not in sys.path) * [vendor_path])  # fmt: skip
 # workaround for #4476
 sys.modules.pop('backports', None)
 
 # capture these to bypass sandboxing
-from os import utime
-from os import open as os_open
-from os.path import isdir, split
+from os import open as os_open, utime  # isort: skip
+from os.path import isdir, split  # isort: skip
```
Avasam, line 82: This change is not automated

setuptools/dist.py
```diff
@@ -20,23 +26,17 @@
...
-from packaging.specifiers import InvalidSpecifier, SpecifierSet
-from packaging.version import Version
-
-from . import _entry_points
-from . import _reqs
-from . import command as _  # noqa  -- imported for side-effects
+from . import (
+    _entry_points,
+    _reqs,
+    command as _,  # noqa: F401 # imported for side-effects
```
Avasam, line 32: This change is not automated (made the noqa more specific)

setuptools/tests/config/test_pyprojecttoml.py
```diff
@@ -4,21 +4,19 @@
 
 import jaraco.path
 import pytest
+import setuptools  # noqa: F401 -- force distutils.core to be patched
```
Avasam, line 7: This change is not automated (more precise noqa)

setuptools/tests/config/test_pyprojecttoml.py
```diff
@@ -4,21 +4,19 @@
 
 import jaraco.path
 import pytest
+import setuptools  # noqa: F401 # force distutils.core to be patched
```
Avasam, line 7: This change is not automated (more precise noqa)

setuptools/tests/test_config_discovery.py
```diff
@@ -3,18 +3,17 @@
 from configparser import ConfigParser
 from itertools import product
 
+import jaraco.path
+import pytest
+import setuptools  # noqa: F401 # force distutils.core to be patched
```
Avasam, line 8: This change is not automated (more precise noqa)

ruff.toml
```diff
@@ -56,6 +57,15 @@ ignore = [
...
 "pkg_resources/__init__.py" = ["E402"]
 
+[lint.isort]
+combine-as-imports = true
+split-on-trailing-comma = false
+# Force Ruff/isort to always import setuptools before distutils in tests as long as distutils_hack is supported
+# This also ensures _distutils_hack is imported before distutils
+# https://github.com/pypa/setuptools/issues/4137
+known-third-party = ["setuptools"]
+known-first-party = ["distutils"]
```
abravalheri, line 67: Would this make sense?

```suggestion
section-order = ["future", "standard-library", "eager", "third-party", "first-party", "local-folder", "delayed"]
sections.eager = ["_distutils_hack"]
sections.delayed = ["distutils"]
```

It is more verbose, but it does is more compatible with what isort would do by default, special-casing only `distutils` and `_distutils_hack`

abravalheri (commented):
I think in the case of this PR it would be interesting in separating 2 commits: 1 for the configuration changes (like `ruff.toml` and `.pre-commit-config.yaml`) and another one for the bulk of the automated changes. If necessary [`git commit --no-verify`](https://pre-commit.com/#temporarily-disabling-hooks) can be used to allow the first commit to be made without changing the rest of the codebase.

Because squashing is not viable if we want to keep the 2 commits separated, it would make more sense to use `rebase` instead of `merging` if it is necessary to update the codebase.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
-- repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.4.9
-  hooks:
-  - id: ruff
-  - id: ruff-format
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.9
+    hooks:
+      - id: ruff
```
abravalheri, line 5: Does it make sense to add `args: [--fix]` here?

ruff.toml
Avasam: I didn't realize I could create custom sections. That is so much better yes !

.pre-commit-config.yaml
Avasam: Absolutely, that's kinda the point of using pre-commit isn't it? Although yml formatting here is accidental and doesn't really belong in this PR. I'll extract that and add `--fix`

Avasam: https://github.com/pypa/setuptools/pull/4552

Avasam:
> I think in the case of this PR it would be interesting in separating 2 commits

Certainly, now that I have confirmation this is indeed wanted, I'll be cautious about keeping it into 2 clean commits.

ruff.toml
```diff
@@ -56,6 +57,16 @@ ignore = [
 "setuptools/__init__.py" = ["E402"]
 "pkg_resources/__init__.py" = ["E402"]
 
+[lint.isort]
+combine-as-imports = true
+split-on-trailing-comma = false
+# Force Ruff/isort to always import setuptools before distutils in tests as long as distutils_hack is supported
+# This also ensures _distutils_hack is imported before distutils
+# https://github.com/pypa/setuptools/issues/4137
+section-order = ["future", "standard-library", "eager", "third-party", "first-party", "delayed", "local-folder"]
```
abravalheri, line 66: I proposed we leave `"delayed"` as the last item in this array.

```suggestion
section-order = ["future", "standard-library", "eager", "third-party", "first-party", "local-folder", "delayed"]
```

The reasoning is the following:

Relative import statements like `from . import ..` seem to be considered "local-folder", but they still refer to `setuptools`. This means that `distutils` imports will come before setuptools imports, which we are trying to avoid.

For example this change in setuptools/command/bdist_wheel.py:

```python
from packaging import tags, version as _packaging_version
from wheel.metadata import pkginfo_to_metadata
from wheel.wheelfile import WheelFile

from distutils import log

from .. import Command, __version__
from .egg_info import egg_info as egg_info_cls
```

abravalheri (changes requested):
Thank you very much @Avasam for putting so much work on this.

I am interested in merging this one before anything else, to get it out of the way and rip off the metaphorical band-aid. Unless you have a different suggestion.

ruff.toml
Avasam: Here's my reasoning:
For non-setuptools folders, like tools and tests, it doesn't change anything and it's closer to what a user of setuptools would have.

For setuptools itself: AFAIK, to import a submodule, the top module/package gets imported too. So `_distutils_hack` should still always run first. Unless I'm missing something:

image
image
(overly zealous prints but I'm trying to remove assumptions from my proof)

pkg_resources/__init__.py
```diff
@@ -93,20 +91,16 @@
     # no write support, probably under GAE
```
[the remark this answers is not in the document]
abravalheri: Google App Engine? I don't know the [change](https://github.com/pypa/setuptools/commit/33e3ec5eadb59ece949d4cad1231ff23780e4c02) is from 2009, before setuptools moved to Github and unfortunatelly all references to PR have been messed up.

ruff.toml
abravalheri: `_distutils_hack` will not be affected by putting `delayed` last, right? (Because it goes to its own group `eager`).

Avasam: As long as `eager` is before `first-party`, `local-folder` and `delayed` yeah

Just to be clear, swapping the order of "local-folder" and "delayed" works in both cases. I'm arguing for keeping the expectation and habit of putting relative imports last.
(prefering left side in this image)
image

pkg_resources/__init__.py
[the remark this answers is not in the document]
Avasam: That tracks https://code.google.com/archive/p/appengine-monkey/issues/25 / https://github.com/quankim277/appengine-monkey/issues/25

ruff.toml
abravalheri: I see, thank you very much for clarifying!

We currently have a recommendation that `distutils` should be imported after `setuptools` expressed in https://github.com/pypa/setuptools/blob/a6f7484b25628ae351d4b96fa6254ada387ece10/_distutils_hack/__init__.py#L17-L24

So I think that honouring that is more important than the traditional approach.

I understand that indirectly we can verify that both approaches work. But the implicit approach might be fragile in the future[^1]. So in principle, I am in favour of directly ensuring that any import of `distutils` always come explicitly after the imports to `setuptools`[^2].

[^1]: For example to solve cycles, Python import machinery may start evaluating the parent module, then interrupt this evaluation in favour of children module, then proceed to finish the evaluation of the parent module.

[^2]: Unless we verify an unintended consequence and we need to revert.

abravalheri: We can change that once `SETUPTOOLS_USE_DISTUTILS=stdlib` is fully removed.

abravalheri (approved):

abravalheri:
Thank you very much.

This might add some extra work for the currently opened PRs, but I think it is good in the long term.

Avasam:
> some extra work for the currently opened PRs

The good news is it shouldn't be much: accept both on merge conflict concerning imports, then let Ruff do the rest.

PR #4397, opened 2024-05-24
Add Type annotations developer guideline

Avasam:
## Summary of changes

Add Type annotations developer guideline
CC @abravalheri feel free to wordsmith. I figured these things are probably better documented with sources than left in PR discussions.

### Pull Request Checklist
- [x] Changes have tests (N/A)
- [x] News fragment added in [`newsfragments/`]. (N/A)

[PR docs]:

### DIFF
--- a/docs/development/developer-guide.rst
+++ b/docs/development/developer-guide.rst
@@ -131,3 +131,35 @@ simple Python script ``tools/vendor.py``.
 To refresh the dependencies, run the following command::
 
     $ tox -e vendor
+
+----------------
+Type annotations
+----------------
+
+Most standards and best practices are enforced by
+`Ruff <https://docs.astral.sh/ruff/rules/>`_'s ``ANN2``, ``FA``, ``PYI``, ``UP``
+and ``YTT`` rules.
+
+Explicit return types have to be added for typed public functions whose
+parameters are *all* annotated. This is enforced by ``ANN2``, but it's worth noting
+that this is due to mypy inferring ``Any`` even for simple return types. Mypy also
+doesn't count functions with missing parameter annotations as "typed". (see
+`python/mypy#4409 <https://github.com/python/mypy/issues/4409>`_,
+`python/mypy#10149 <https://github.com/python/mypy/issues/10149>`_ and
+`python/mypy#6646 <https://github.com/python/mypy/issues/6646>`_).
+Otherwise, return annotations can be omitted to reduce verbosity,
+especially for complex return types.
+
+Instead of typing an explicit return type annotation as
+``Generator[..., None, None]``, we'll prefer using an ``Iterator`` as it is more
+concise and conceptually easier to deal with. Returning a ``Generator`` with no
+``yield`` type or ``send`` type can sometimes be considered as exposing
+implementation details. See
+`Y058 <https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md#Y058>`_.
+
+Avoid importing private type-checking-only symbols. These are often
+`typeshed <https://github.com/python/typeshed>`_ internal details and are not
+guaranteed to be stable.
+Importing from ``_typeshed`` or ``typing_extensions`` is fine, but if you find
+yourself importing the same symbol in ``TYPE_CHECKING`` blocks a lot, consider
+implementing an alias directly in ``setuptools``.


docs/development/developer-guide.rst
```diff
@@ -131,3 +131,22 @@ simple Python script ``tools/vendor.py``.
...
     $ tox -e vendor
+
+----------------
+Type annotations
+----------------
+
+Setuptools makes use of inferred return annotations to reduce verbosity,
+especially for complex return types. Explicit return types can be added for
+functions whose inferred return type contains ``Any``, or that are not checked by
+mypy (ie.: contains no parameter *and* no return type, see
```
abravalheri, line 142: I have a curiosity question:

In which instances `mypy` can correctly infer return types without degenerating to `Any`?
For example, I tried the following:

```python
def fn(a: str):
    return int(a)

reveal_type(fn)
```

```console
main.py:4: note: Revealed type is "def (a: builtins.str) -> Any"
Success: no issues found in 1 source file
```

To me the return value should be `int`, but `mypy` infers a "degenerate" `Any`.
Is there any circumstance that it can actually get it right without explicit marks?

abravalheri (approved):

docs/development/developer-guide.rst
Avasam: Oh my... I've always (wrongly) assumed that mypy at least infers the return type for functions it does type-check...

Maybe we should instead turn on some `ANN2` rules for public methods so this can be checked automatically instead of having to add it as a guideline.
- https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/
- https://docs.astral.sh/ruff/rules/missing-return-type-special-method/
- https://docs.astral.sh/ruff/rules/missing-return-type-static-method/
- https://docs.astral.sh/ruff/rules/missing-return-type-class-method/

https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_incomplete_defs is also an option.

Avasam: Even w/o enforcement, running `ruff check pkg_resources --fix --unsafe-fixes --select=ANN2 --ignore=ANN202,ANN204` (ignoring tests) is pretty valuable.

Let's merge #4390 and #4391 first.

Avasam:
@abravalheri Updated and added a few more details !

docs/development/developer-guide.rst
```diff
@@ -131,3 +131,35 @@ simple Python script ``tools/vendor.py``.
 To refresh the dependencies, run the following command::
 
     $ tox -e vendor
+
+----------------
+Type annotations
+----------------
+
+Most standards and best practices are enforced by
+[Ruff](https://docs.astral.sh/ruff/rules/)'s ``ANN2``, ``FA``, ``PYI``, ``UP``
```
Avasam, line 140: `ANN2` added in https://github.com/pypa/setuptools/pull/4409/

docs/development/developer-guide.rst
```diff
@@ -131,3 +131,35 @@ simple Python script ``tools/vendor.py``.
...
+and ``YTT`` rules.
+
+Explicit return types have to be added for typed public functions whose
+parameters are *all* annotated. This is enforced by ``ANN2``, but it's worth noting
+that this is due to mypy inferring `Any` even for simple return types. Mypy also
```
Avasam, line 145:
```suggestion
that this is due to mypy inferring ``Any`` even for simple return types. Mypy also
```

docs/development/developer-guide.rst
```diff
@@ -131,3 +131,35 @@ simple Python script ``tools/vendor.py``.
...
+``Generator[..., None, None]``, we'll prefer using an ``Iterator`` as it is more
+concise and conceptually easier to deal with. Returning a ``Generator`` with no
+``yield`` type or ``send`` type can sometimes be considered as exposing
+implementation details. See
+[Y058](https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md#Y058).
+
+Avoid importing private type-checking-only symbols. These are often
+[typeshed](https://github.com/python/typeshed) internal details and are not
+guaranteed to be stable.  
+Importing from ``_typeshed`` or ``typing_extensions`` is fine, but if you find
```
Avasam, line 163: It's annoying, I can't find how to create a newline w/o creating a new paragraph in RST.

abravalheri: Yeah `\n` is threated in RST as a space and `\n\n` creates a new paragraph, no other option.

docs/development/developer-guide.rst
abravalheri, line 140: I think this is markdown syntax for link, but setuptools uses RST.
There is a couple of other occurrences below.

abravalheri: I did not find the ``FA`` rules, is that PyFlake (`F`)?

Avasam: Not used to RST, but I think I got it right

abravalheri: Thank you.

Reminder for my future self, probably adding something like https://github.com/sloria/sphinx-issues is useful.

abravalheri:
Thank you very much.

PR #4406, opened 2024-05-30
PERF: remove isdir() check in copy_file(), about 10% of the run time was checking if the file to copy is a directory.

morotti:
Hello,

I was doing a bit of profiling on my CI builds. The step to build package (python setup.py bdist_wheel) is taking more time than I would like.

Going in with a commercial profiler, I am finding that nearly 10% of the time is taken in `isdir()` calls, mostly in the `copy_file()` function.
Which is quite surprising IMO because the copy file function should not be given directories anyway :D

Can I offer to remove the isdir() calls to get a speedup?
I've found 2 calls that were passing directories instead of a filepath and corrected them.

By the way, the whole `_copy_file_contents()` could be replaced by `shutil.copyfile()` for another little boost. The code goes back to year 2000 or before, which predates optimized shutil functions.

Screenshot of the profiler run. The whole run is about 20+ seconds. 
image

Regards.

### DIFF
--- a/setuptools/_distutils/command/install_data.py
+++ b/setuptools/_distutils/command/install_data.py
@@ -54,7 +54,8 @@ class install_data(Command):
                         "setup script did not provide a directory for "
                         f"'{f}' -- installing right in '{self.install_dir}'"
                     )
-                (out, _) = self.copy_file(f, self.install_dir)
+                dst = os.path.join(self.install_dir, os.path.basename(f))
+                (out, _) = self.copy_file(f, dst)
                 self.outfiles.append(out)
             else:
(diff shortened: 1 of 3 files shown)


abravalheri:
Hi @morotti, thank you for having a look on this.
It is nice to have performance optimisations!

The only thing that I would say is that the development of the `setuptools/_distutils` folder is not done in this repository.
PRs that change those files first need to be merged in https://github.com/pypa/distutils and then we pull them in `setuptools`.

So I would recommend submitting the PR to https://github.com/pypa/distutil and discussing with the maintainers of that repo.

I enabled the CI on this PR so you can also use as an argument that "it does not break anything".

PR #4409, opened 2024-05-31
Enforce return types for typed public functions & mark `pkg_resources` as `py.typed`

Avasam:
### ~~Merge #4390 and #4391 first !~~
Addresses https://github.com/pypa/setuptools/pull/4397#discussion_r1622757650

This also marks `pkg_resources` as `py.typed` as it is now more on par (or even improved) over typeshed's stubs. ~~With the exception of `extern` imports causing a loss of typing, but that's a completely different discussion and `pkg_resources` as a deprecated module, isn't affected too badly.~~ #4457

Step 4 & 5 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


Avasam:
@abravalheri @jaraco out of all my currently open PRs, this is the one I'd like to prioritise most, as it marks `pkg_resources` as `py.typed`, launching the 6 months countdown to remove `pkg_resources` from typeshed.

I've also been thinking of maybe not enforcing `ANN2` on the setuptools side yet, since I'd probably disable it later to progressively type `setuptools` like I did `pkg_resources`

pkg_resources/__init__.py
```diff
@@ -102,7 +103,7 @@
 from platformdirs import user_cache_dir as _user_cache_dir
 
 if TYPE_CHECKING:
-    from _typeshed import BytesPath, StrPath, StrOrBytesPath
+    from _typeshed import StrPath, BytesPath, StrOrBytesPath
```
jaraco, line 106: Unsure why this order has changed. Maybe something was removed and re-added. At some point I want to introduce `ruff check --select I`, which suggests:

```suggestion
    from _typeshed import BytesPath, StrOrBytesPath, StrPath
```

jaraco (approved):
LGTM. I didn't see any changes in the implementation. Please confirm that's the case.

The only risk with merging this is if we have to back out #4457, probably some of the type checking will need to back off, but that outcome seems unlikely at this point.

pkg_resources/__init__.py
Avasam: Probably removed/added, merge+autofix, or something like that. I have prepped #4392 for automatic import sorting.

Avasam:
> I didn't see any changes in the implementation. Please confirm that's the case.

I re-reviewed the PR and I don't see any change in implementation either.

> if we have to back out https://github.com/pypa/setuptools/pull/4457, probably some of the type checking will need to back off, but that outcome seems unlikely at this point.

Even that I feel would be minor.

PR #4411, opened 2024-06-02
Enforce ruff/flake8-implicit-str-concat rules (ISC)

DimitriPapadopoulos:
## Summary of changes

ISC001 Implicitly concatenated string literals on one line

ISC003 Explicitly concatenated string should be implicitly concatenated

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/bdist_egg.py
+++ b/setuptools/command/bdist_egg.py
@@ -382,8 +382,9 @@ def scan_module(egg_dir, base, name, stubs):
         for bad in [
             'getsource',
             'getabsfile',
+            'getfile',
             'getsourcefile',
-            'getfile' 'getsourcelines',
+            'getsourcelines',
             'findsource',
             'getcomments',
             'getframeinfo',
--- a/setuptools/tests/test_find_packages.py
+++ b/setuptools/tests/test_find_packages.py
@@ -180,7 +180,8 @@ class TestFlatLayoutPackageFinder:
             [
                 "pkg/__init__.py",
                 "examples/__init__.py",
-                "examples/file.py" "example/other_file.py",
+                "examples/file.py",
+                "example/other_file.py",
                 # Sub-packages should always be fine
                 "pkg/example/__init__.py",
                 "pkg/examples/__init__.py",


pkg_resources/tests/test_resources.py
```diff
@@ -257,7 +257,7 @@ def test_marker_evaluation_with_extras(self):
             "/foo_dir/Foo-1.2.dist-info",
             metadata=Metadata((
                 "METADATA",
-                "Provides-Extra: baz\n" "Requires-Dist: quux; extra=='baz'",
+                "Provides-Extra: baz\nRequires-Dist: quux; extra=='baz'",
```
DimitriPapadopoulos, line 260: Use triple-quoted strings instead?

setuptools/command/bdist_egg.py
```diff
@@ -382,8 +382,9 @@ def scan_module(egg_dir, base, name, stubs):
         for bad in [
             'getsource',
             'getabsfile',
+            'getfile',
             'getsourcefile',
-            'getfile' 'getsourcelines',
+            'getsourcelines',
```
DimitriPapadopoulos, line 387: That was a real bug!

setuptools/tests/config/test_expand.py
```diff
@@ -83,7 +83,7 @@ def test_read_attr(self, tmp_path, monkeypatch):
             "pkg/__init__.py": "",
             "pkg/sub/__init__.py": "VERSION = '0.1.1'",
             "pkg/sub/mod.py": (
-                "VALUES = {'a': 0, 'b': {42}, 'c': (0, 1, 1)}\n" "raise SystemExit(1)"
+                "VALUES = {'a': 0, 'b': {42}, 'c': (0, 1, 1)}\nraise SystemExit(1)"
```
DimitriPapadopoulos, line 86: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -35,7 +35,7 @@ def fake_env(
     tmpdir, setup_cfg, setup_py=None, encoding='ascii', package_path='fake_package'
 ):
     if setup_py is None:
-        setup_py = 'from setuptools import setup\n' 'setup()\n'
+        setup_py = 'from setuptools import setup\nsetup()\n'
```
DimitriPapadopoulos, line 38: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -97,7 +97,7 @@ def test_no_config(self, tmpdir):
     def test_ignore_errors(self, tmpdir):
         _, config = fake_env(
             tmpdir,
-            '[metadata]\n' 'version = attr: none.VERSION\n' 'keywords = one, two\n',
+            '[metadata]\nversion = attr: none.VERSION\nkeywords = one, two\n',
```
DimitriPapadopoulos, line 100: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -171,22 +171,22 @@ def test_license_cfg(self, tmpdir):
     def test_file_mixed(self, tmpdir):
         fake_env(
             tmpdir,
-            '[metadata]\n' 'long_description = file: README.rst, CHANGES.rst\n' '\n',
+            '[metadata]\nlong_description = file: README.rst, CHANGES.rst\n\n',
```
DimitriPapadopoulos, line 174: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -171,22 +171,22 @@ def test_license_cfg(self, tmpdir):
...
         )
 
         tmpdir.join('README.rst').write('readme contents\nline2')
         tmpdir.join('CHANGES.rst').write('changelog contents\nand stuff')
 
         with get_dist(tmpdir) as dist:
             assert dist.metadata.long_description == (
-                'readme contents\nline2\n' 'changelog contents\nand stuff'
+                'readme contents\nline2\nchangelog contents\nand stuff'
```
DimitriPapadopoulos, line 182: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -171,22 +171,22 @@ def test_license_cfg(self, tmpdir):
...
             )
 
     def test_file_sandboxed(self, tmpdir):
         tmpdir.ensure("README")
         project = tmpdir.join('depth1', 'depth2')
         project.ensure(dir=True)
-        fake_env(project, '[metadata]\n' 'long_description = file: ../../README\n')
+        fake_env(project, '[metadata]\nlong_description = file: ../../README\n')
```
DimitriPapadopoulos, line 189: Use triple-quoted strings instead?

setuptools/tests/config/test_setupcfg.py
```diff
@@ -253,7 +253,7 @@ def test_dict(self, tmpdir):
 
     def test_version(self, tmpdir):
         package_dir, config = fake_env(
-            tmpdir, '[metadata]\n' 'version = attr: fake_package.VERSION\n'
+            tmpdir, '[metadata]\nversion = attr: fake_package.VERSION\n'
```
DimitriPapadopoulos, line 256: Use triple-quoted strings instead?

setuptools/tests/test_find_packages.py
```diff
@@ -180,7 +180,8 @@ class TestFlatLayoutPackageFinder:
             [
                 "pkg/__init__.py",
                 "examples/__init__.py",
-                "examples/file.py" "example/other_file.py",
+                "examples/file.py",
+                "example/other_file.py",
```
DimitriPapadopoulos, line 184: Again a real bug!

Avasam:
ISC001 definitively as there's no valid reason to ISC on a single line. It just makes the line longer and is likely to accidentally create bugs (which you did catch!)

ISC002 vs ISC003 is more opinionated, so up to the maintainers. Personally I prefer disallowing multiline implicit string concatenation as it helps from a quick read understanding that two lines are not two parameters.

Also here's a feature request to control `ISC003` with `[lint.flake8-implicit-str-concat] allow-multiline` (would be nice if we can just select `ISC` https://github.com/astral-sh/ruff/issues/11582 )

pkg_resources/tests/test_resources.py
abravalheri: I think it is not necessary to use triple-quotes here, specially in the context of quick tests...

setuptools/command/bdist_egg.py
abravalheri: Thank you very much @DimitriPapadopoulos for working on this.

Given that this PR for the time being is failing the CI, would you like to extract this bug fix in a PR of its own?

pkg_resources/tests/test_resources.py
DimitriPapadopoulos: But then tests are as important and should be as maintainable as the source code.

setuptools/command/bdist_egg.py
DimitriPapadopoulos: I can do that, but what's the intent? As far as I can tell, this change is not (directly) responsible for the CI failures.

DimitriPapadopoulos:
About `ISC003`, I tend to agree with the [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#breaking-long-lines-and-strings):
> ### 2) Breaking long lines and strings
> [...]
> The preferred limit on the length of a single line is 80 columns.
> [...]
> However, never break user-visible strings such as printk messages because that breaks the ability to grep for them.

I think it  make sense to unfold user-visible strings, I seem to recall _ruff_ does allow literal strings longer than the admissible line length, up to a larger limit at least.

DimitriPapadopoulos:
I've changed this PR to only fix actual bugs uncovered by `ISC001`.

setuptools/command/bdist_egg.py
abravalheri: Great, if the CI is no longer failure that is OK. We can proceed with this PR.

pkg_resources/tests/test_resources.py
abravalheri: I don't think that adding an explicit `\n` makes short snippets unreadable.
For tests, it even makes sense, because we are explicitly saying there should be a `\n` there...

On the other hand, using triple quotes for short strings either results in distracting code (e.g. when the indentation is broken inside the triple-quotes), or requires you to bring out the big guns (e.g. `inspect.cleandoc`).

abravalheri:
> About `ISC003`, I tend to agree with the [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#breaking-long-lines-and-strings):

We should probably favour `ISC002` in setuptools to align with [skeleton](https://github.com/jaraco/skeleton/blob/main/ruff.toml#L20-L21).

abravalheri:
Thank you very much @DimitriPapadopoulos.

DimitriPapadopoulos:
> > About `ISC003`, I tend to agree with the [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#breaking-long-lines-and-strings):
> 
> We should probably align with [skeleton](https://github.com/jaraco/skeleton/blob/main/ruff.toml#L20-L21).

It clearly a bad idea to ignore `ISC001`. I'll raise an issue over there.

DimitriPapadopoulos:
My bad, I am actually the one who (temporarily) ignored `ISC001` and `ISC002` in [skeleton](https://github.com/jaraco/skeleton/blame/main/ruff.toml#L20-L21). However, this is only a temporary counter-measure to work around potential [Conflicting lint rules](https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules). At least the `ISC001` rule should be reinstated, when https://github.com/astral-sh/ruff/issues/8272 is addressed.


In the meantime, I think it makes sense to apply `ISC001` and `ISC002`, even if the rules are not yet enabled in the linter configuration.

PR #4436, opened 2024-06-26
Rename `_InstallerTypeT` and mark type aliases with `TypeAlias`

Avasam:
## Summary of changes
Noticed in https://github.com/python/typeshed/pull/12215
According to https://github.com/PyCQA/flake8-pyi/blob/main/ERRORCODES.md#Y043 , that was a bad name to have as it implies a TypeVar. I've renamed the alias to a more fitting `_StrictInstallerType` (alternatively: `_GenericInstallerType`)

This would've been caught by https://docs.astral.sh/ruff/rules/t-suffixed-type-alias/ if it was explicitly marked as a type alias.

Explicitly marked all type aliases I could easily find with `TypeAlias` (available in Python 3.10, backported to `typing_extensions`).

### Pull Request Checklist
- [x] Changes have tests (existing Ruff and type-checking tests)

[PR docs]:


setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -33,11 +33,12 @@
     from distutils.dist import _OptionsList
     from setuptools._importlib import metadata
     from setuptools.dist import Distribution
+    from typing_extensions import TypeAlias
 
-EMPTY: Mapping = MappingProxyType({})  # Immutable dict-like
-_ProjectReadmeValue = Union[str, Dict[str, str]]
-_CorrespFn = Callable[["Distribution", Any, StrPath], None]
-_Correspondence = Union[str, _CorrespFn]
+EMPTY: MappingProxyType = MappingProxyType({})  # Immutable dict-like
```
abravalheri, line 38: I think it makes more sense to preserve `Mapping` here... If Python had a immutable dict-object I would use that... `MappingProxyType` is just a implementation detail.

Avasam:
```suggestion
EMPTY: Mapping = MappingProxyType({})  # Immutable dict-like
```

abravalheri:
Thank you!

PR #4438, opened 2024-06-26
Exclude coverage for older Python version checks

Avasam:
## Summary of changes

Something I noticed from coverage failing for 1 line in https://github.com/pypa/setuptools/pull/4436 . The rationale is explained in config comment:
> Coverage is only run on the latest Python version, these will never be considered covered and we aim for 100% coverage

image

### Pull Request Checklist
- [x] Changes have tests (diffcov should pass for this PR)
- [x] News fragment added in [`newsfragments/`]. (not public facing)

[PR docs]:

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -20,3 +20,6 @@ exclude_also =
 	# jaraco/skeleton#97
 	@overload
 	if TYPE_CHECKING:
+	# Coverage is only run on the latest Python version
+	# these will never be considered covered and we aim for 100% coverage
+	if sys\.version_info <
--- a/setuptools/_path.py
+++ b/setuptools/_path.py
@@ -2,10 +2,10 @@ import os
 import sys
 from typing import Union
 
-if sys.version_info >= (3, 9):
-    StrPath = Union[str, os.PathLike[str]]  #  Same as _typeshed.StrPath
-else:
+if sys.version_info < (3, 9):
     StrPath = Union[str, os.PathLike]
+else:
+    StrPath = Union[str, os.PathLike[str]]  #  Same as _typeshed.StrPath
 
 
 def ensure_directory(path):
--- a/setuptools/compat/py310.py
+++ b/setuptools/compat/py310.py
@@ -4,7 +4,7 @@ import sys
 __all__ = ['tomllib']
 
 
-if sys.version_info >= (3, 11):
-    import tomllib
-else:  # pragma: no cover
+if sys.version_info < (3, 11):
     from setuptools.extern import tomli as tomllib
+else:
+    import tomllib


.coveragerc
```diff
@@ -20,3 +20,6 @@ exclude_also =
 	# jaraco/skeleton#97
 	@overload
 	if TYPE_CHECKING:
+	# Coverage is only run on the latest Python version
+	# these will never be considered covered and we aim for 100% coverage
+	if sys\.version_info <
```
Avasam, line 25: Technically this matches `<= (3, 12)` but I think that's enough of an edge case, would require more complex regex and extra book-keeping when bumping the Python version, that it's not worth accounting for.

jaraco: This file is managed primarily in [skeleton](/jaraco/skeleton) and there's already an effort underway to [implement coverage aggregation](https://github.com/jaraco/skeleton/issues/130), which would obviate the need for this change.

setuptools/_path.py
```diff
@@ -2,10 +2,10 @@
 import sys
 from typing import Union
 
-if sys.version_info >= (3, 9):
-    StrPath = Union[str, os.PathLike[str]]  #  Same as _typeshed.StrPath
-else:
+if sys.version_info < (3, 9):
     StrPath = Union[str, os.PathLike]
+else:
+    StrPath = Union[str, os.PathLike[str]]  #  Same as _typeshed.StrPath
```
jaraco, line 8: I've historically written `> (3, 9)` (or the equivalent `>= (3, 9)`) because I like to put the canonical, preferred code ahead of the legacy code. I don't feel strongly about it and I feel less strongly about it when the compatibility logic is in a `compat` submodule.

My thinking is that if someone is skimming the code, if the legacy code is first, they'll need to read the whole block to get a modern understanding of what's going on, whereas if the preferred code is first, they can stop after reading the first block and just know that the rest is for older Python versions.

This concern illustrates is a case where an `unless` syntax would be so nice to have:

```python
do:
    # preferred behavior
unless sys.version_info < (3, 9):
    # legacy behavior
```

Unfortunately, that's not a thing in Python, so the reader is forced to choose between the more elegant if clause or the more elegant preferred-first ordering.

jaraco (commented):
Given that we have someone working on a more permanent solution that avoids adding special exclusions per-project and making the code conform to a sub-optimal convention, I'm inclined to say let's just wait for the more permanent solution.

.coveragerc
Avasam: That would be better since then coverage would be correct in telling you that an environment-specific path was never taken.

jaraco: I'm glad you agree. I'll assume we should close this request, then, and await/pursue the upstream change. If you feel differently, feel free to advocate for including this change and we can re-open.

Issue #4439, opened 2024-06-26
AssertionError importing setuptools


jaraco:
I don't think that Python 3.11.2 is considered old.  However, I am unable to import setuptools without setting the environment variable that you want to remove.  Wouldn't it be better to fix setuptools first?

This is what I am seeing:

```
Python 3.11.2 (v3.11.2:878ead1ac1, Feb  7 2023, 10:02:41) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/culler/Library/Python/3.11/lib/python/site-packages/setuptools/__init__.py", line 8, in <module>
    import _distutils_hack.override  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/culler/Library/Python/3.11/lib/python/site-packages/_distutils_hack/override.py", line 1, in <module>
    __import__('_distutils_hack').do_override()
  File "/Users/culler/Library/Python/3.11/lib/python/site-packages/_distutils_hack/__init__.py", line 77, in do_override
    ensure_local_distutils()
  File "/Users/culler/Library/Python/3.11/lib/python/site-packages/_distutils_hack/__init__.py", line 64, in ensure_local_distutils
    assert '_distutils' in core.__file__, core.__file__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/distutils/core.py
>>> import os
>>> os.environ['SETUPTOOLS_USE_DISTUTILS'] = 'stdlib'
>>> import setuptools
>>> setuptools.__version__
'67.6.0'
```

_Originally posted by @culler in https://github.com/pypa/setuptools/issues/3625#issuecomment-1475353188_

culler:
My comment on #3625 was posted a long time ago.  In the meantime I have upgraded python on that system - including python 3.11, (even though I don't use it anymore).  So I no longer have a python 3.11.2 installation to test with.  My current python 3.11 is 3.11.6 with setuptools 68.1.0.  It seems to have no trouble importing setuptools.  I also have an installation of Python 3.11.0 on another machine, and it imports setuptools without issue.

[...]

PR #4441, opened 2024-06-26
Fallback to stdlib `distutils` if `setuptools._distutils` is missing

Avasam:
## Summary of changes

If `setuptools._distutils` is missing and `SETUPTOOLS_USE_DISTUTILS` environment variable isn't set to `stdlib`, importing `setuptools` will fail.
This change fallsback to stdlib despite `SETUPTOOLS_USE_DISTUTILS=local` if `setuptools._distutils` is missing and shows a warning for Python 3.11 and below.

To test this, simply delete or rename `setuptools._distutils` (either with an editable install or directly in your site-packages) and set `SETUPTOOLS_USE_DISTUTILS` to nothing or `local`. I'd like to write a test for this and may need some pointers to get started since this test would include deleting files (or at least making an import inaccessible)

Before:
```py
>>> import setuptools
  File "E:\Users\Avasam\Documents\Git\setuptools\setuptools\__init__.py", line 8, in <module>
    import _distutils_hack.override  # noqa: F401
  File "E:\Users\Avasam\Documents\Git\setuptools\_distutils_hack\override.py", line 1, in <module>
    __import__('_distutils_hack').do_override()
  File "E:\Users\Avasam\Documents\Git\setuptools\_distutils_hack\__init__.py", line 89, in do_override
    ensure_local_distutils()
  File "E:\Users\Avasam\Documents\Git\setuptools\_distutils_hack\__init__.py", line 76, in ensure_local_distutils
    assert '_distutils' in core.__file__, core.__file__
AssertionError: C:\Program Files\Python39\lib\distutils\core.py
```

After:
```py
>>> import setuptools
E:\Users\Avasam\Documents\Git\setuptools\_distutils_hack\__init__.py:66: UserWarning: environment variable SETUPTOOLS_USE_DISTUTILS is set to 'local' or None, but `setuptools._distutils` could not be imported. Falling back to stdlib distutils. This will lead to an error in Python 3.12 and above.
  warnings.warn(
```

Closes https://github.com/pypa/setuptools/issues/4439
The validity of this fix assumes that the user didn't themselves delete `setuptool._distutils` (messing up their own installation) and that distributing setuptools without `setuptools._distutils` and w/o setting an environment variable is a valid use-case. (sounds valid to me for vendor-stripped setuptools)

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/_distutils_hack/__init__.py
+++ b/_distutils_hack/__init__.py
@@ -33,12 +33,43 @@ def clear_distutils():
         del sys.modules[name]
 
 
+def get_local_distutils():
+    import importlib
+
+    try:
+        return importlib.import_module('setuptools._distutils')
+    except Exception:
+        # There are a couple of cases where setuptools._distutils
+        # may not be present:
+        # - An older Setuptools without a local distutils is
+        #   taking precedence. Ref #2957.
+        # - Path manipulation during sitecustomize removes
+        #   setuptools from the path but only after the hook
+        #   has been loaded. Ref #2980.
+        # - setuptools._distutils has been removed, either by the user
+        #   or by packaging / distribution.
+        # In either case, fall back to stdlib behavior.
+        return None
+
+
 def enabled():
     """
     Allow selection of distutils by environment variable.
     """
     which = os.environ.get('SETUPTOOLS_USE_DISTUTILS', 'local')
-    return which == 'local'
+    if which != 'local':
+        return False
+
+    if sys.version_info < (3, 12) and not get_local_distutils():
+        import warnings
+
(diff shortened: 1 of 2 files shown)


jaraco:
I really don't want to open up this code path. Since Setuptools adopted distutils, it's been primarily reliant on local distutils and designed only to fall back to stdlib if explicitly indicated by the user. This approach requires the user to have ownership over that fallback, and it's important for them to have ownership as we work to deprecate that path altogether (https://github.com/pypa/setuptools/issues/4137).

PR #4448, opened 2024-06-28
Apply assorted ruff/flake8-return rules (RET)

DimitriPapadopoulos:
## Summary of changes

```
RET502 Do not implicitly `return None` in function able to return non-`None` value
RET503 Missing explicit `return` at the end of function able to return non-`None` value
```

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -1087,6 +1087,7 @@ class SystemInfo:
             return 'v3.5', 'v2.0.50727'
         elif self.vs_ver == 8.0:
             return 'v3.0', 'v2.0.50727'
+        return None
 
     @staticmethod
     def _use_last_dir_name(path, prefix=''):
@@ -1648,6 +1649,7 @@ class EnvironmentInfo:
             path = join(prefix, arch_subdir, crt_dir, vcruntime)
             if isfile(path):
                 return path
+        return None
 
     def return_env(self, exists=True):
         """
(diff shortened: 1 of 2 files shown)


jaraco:
I'm not a fan of RET 503. It feels a lot to me like overspecifying the behavior, not dissimilar to:

```python
if bool(x):
  return True
else:
  return False
```

Instead of just:

```python
return bool(x)
```

Or writing:

```python
str("foo")
```

instead of just

```python
"foo"
```

Both of the former examples are more "explicit", but they only add noise. One of the things I loved about the Python language early on was that a blank `return` or an implied return at the end of a function body is indistinguishable from an explicit `return None`, so use the minimal form for consistency and conciseness.

I understand people's instinct to use meaningless statements as a signal about intent, but it just feels nitpicky to me.

My preference would be to stick with the more concise form until there's a documented language feature that enforces or honors a convention, such as to not have an implicit `None`.

I do believe I've seen other type checkers enforce a similar rule, though, so maybe I'm just a curmudgeon.

DimitriPapadopoulos:
My understanding of these rules is that an implicit `None` in itself is OK. What is not OK is a mixture of explicit and implicit:
```python
def f(i):
    if i > 0:
        return i
```
instead of:
```python
def f(i):
    if i > 0:
        return i
    return None
```
The rationale could be that newcomers to Python might get confused, and that a missing return might actually be an oversight and not an implicit `return None`.

DimitriPapadopoulos:
But I agree that's not the most useful of the ruff rules. To tell the truth, I haven't ever caught an actual oversight, as opposed to an implicit `return` or `None`.

PR #4454, opened 2024-07-02
Ignore ruff/tryceratops rule TRY003

DimitriPapadopoulos:
## Summary of changes

```
TRY003 Avoid specifying long messages outside the exception class
```

Applying this rule would mean creating lots of specialised exception classes. Not sure this would improve readability and maintainability in this context.

Addresses https://github.com/pypa/setuptools/pull/4450#issuecomment-2200583235.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -18,6 +18,7 @@ extend-select = [
 	"YTT", # flake8-2020
 ]
 ignore = [
+	"TRY003", # raise-vanilla-args, avoid multitude of exception classes
 	"TRY301", # raise-within-try, it's handy
 	"UP015", # redundant-open-modes, explicit is preferred
 	"UP030", # temporarily disabled


Avasam:
First case of a PR failing because of this: https://github.com/pypa/setuptools/pull/4464

Probably related to https://github.com/businho/pytest-ruff/releases/tag/v0.4.0 being released ~1h ago.

CC @jaraco @abravalheri to try and expedite this quick fix

jaraco:
I did confirm that downgrading to `pytest-ruff 0.3.2` suppresses the errors. I don't see anything in [the pytest-ruff changelog](https://github.com/businho/pytest-ruff/compare/v0.3.2...v0.4.0) that indicates that new rules are now enforced. Was it a mistake? Shouldn't we first see if pytest-ruff wants to yank the version? Is this issue going to affect other projects?

jaraco:
I see now there are probably several factors at play. Something about #4450 probably introduced broken config (or maybe there's broken config elsewhere) such that the failures weren't introduced, but now that pytest-ruff 0.4.0 fixes the handling of the brokenness in the config, the ruff checks are now running and failing. Maybe?

In any case, it looks like the issue is likely isolated to setuptools and its custom config so won't affect other projects right away.

Let's bring this fix in.

Since it seems we've already encountered scope creep on the TRY jobs, maybe it makes sense to opt-in to desired rules rather than opt out?

DimitriPapadopoulos:
I noticed `TRY003` appearing with ruff 0.5. I suspect the underlying version of ruff has been bumped to 0.5. I don't know how the version of ruff is chosen by [pytest-ruff](https://github.com/businho/pytest-ruff/).

Avasam:
Independent of pytest-ruff, `TRY003` is a code that is raised locall , as it should given current configs (that's how I noticed it, and should've been excluded in #4450).
Why it didn't raise on the CI previously is still an open question.

> I noticed `TRY003` appearing with ruff 0.5. I suspect the underlying version of ruff has been bumped to 0.5. I don't know how the version of ruff is chosen by [pytest-ruff](https://github.com/businho/pytest-ruff/).

This here PR used pytest-ruff==0.3.2 and ruff==0.5.0 when last run: https://github.com/pypa/setuptools/actions/runs/9756635721/job/26927344059?pr=4454#step:5:26

DimitriPapadopoulos:
My hope is that ruff will stabilise and such issues will get less and less frequent. The downside of opt-in is that the opt-in list might be much larger than the opt-out list, and would need to be updated from time to time to include new rules. Perhaps wait and see how often new rules break havoc?

Avasam:
Btw @jaraco TRY003 isn't new at all, it's been there since https://github.com/astral-sh/ruff/releases/tag/v0.0.236 , and I don't see any mention of it since 0.1 https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md

DimitriPapadopoulos:
Indeed. After downgrading to ruff 0.3, I still see `TRY003` errors:

```console
$ ruff --version
ruff 0.3.2
$ 
$ ruff check --extend-select TRY
[...]
Found 72 errors.
$ 
```

```console
$ ruff --version
ruff 0.5.1
$ 
$ ruff check --extend-select TRY
[...]
Found 72 errors.
$ 
```

Why I didn't experience `TRY003` errors in the CI jobs after #4450 remains a mystery...

DimitriPapadopoulos:
Yes, I seem to recall I had run pytest-ruff, not ruff. That's why!

Issue #4414, opened 2024-06-04
[BUG] pkg_resources.Requirement.parse raises different exception class on 70.0.0


pradyunsg:
### setuptools version

70.0.0

### Python version

N/A

### OS

N/A

### Additional environment information

N/A

### Description

In setuptools 70, `Requirement.parse` raises a different (private) exception class coming out of the `_vendor` path rather than the `extern` path. This means that there is no way for a caller of that method to catch the raised exception without reaching into the `_vendor` directory.

This was changed by https://github.com/pypa/setuptools/commit/e9995828311c5e0c843622ca2be85e7f09f1ff0d, if I understand correctly.

### Expected behavior

The class is not changed, or the class of the exception raised is publicly accessible.

### How to Reproduce

1. Create a file with the following contents:

```py
import warnings

warnings.simplefilter("ignore", Warning)

import setuptools
import pkg_resources  # noqa
from pkg_resources.extern.packaging.requirements import (  # noqa
    InvalidRequirement as ErrorFromExtern,
)
from pkg_resources._vendor.packaging.requirements import (  # noqa
    InvalidRequirement as ErrorFromVendor,
)

try:
    pkg_resources.Requirement.parse("a b")
except Exception as e:
    print(setuptools.__version__)
    print("  ", e.__class__)
    print("  ", isinstance(e, ErrorFromExtern))
    print("  ", isinstance(e, ErrorFromVendor))
```

2. Run this file in an environment with setuptools 70.0.0 and then in an environment with setuptools==69.5.1.
3. See the different class of the exception.

### Output

```console
69.5.1
   <class 'pkg_resources.extern.packaging.requirements.InvalidRequirement'>
   True
   False
70.0.0
   <class 'pkg_resources._vendor.packaging.requirements.InvalidRequirement'>
   False
   True
```

Avasam:
The only change that I can think of affecting this would be https://github.com/pypa/setuptools/pull/4348/files#diff-e3d3d454fa3a072c9f46f8affa27513892fbc2d245e87f57a5927a7be851de05

abravalheri:
This seems to be a bit of a quirk of Python's importlib/import machinery:

```python
# v69.5.1
from pkg_resources.extern.packaging import requirements as _packaging_requirements

print(f"{_packaging_requirements=}")
print(f"{_packaging_requirements.__spec__=}")
print(f"{_packaging_requirements.__loader__=}")

# --- ouput ---
[...]

PR #4457, opened 2024-07-02
Simplify vendoring and declare dependencies optionally

### LINKED ISSUE #2825: Stop vendoring packages

jaraco:
## Summary of changes

Although this PR doesn't fully remove vendored packages, it re-writes the logic to make the vendoring layer very thin and optional, removing lots of cruft that's accumulated around it.

In particular, it adds the vendored packages to `sys.path` instead of proxying them through a submodule. It adds them to the back of `sys.path` to allow naturally-installed packages to take precedence (vendored packages are only a fallback). The dependencies are no longer pinned to a specific version except that the packages in `_vendor` are for a specific version.

Benefits include:

- Gives precedence to naturally-installed packages if present (vendored packages are the fallback).
- Setuptools itself can import modules naturally, no more `extern` module redirection and associated double-accounting.
- Setuptools and pkg_resources share the same set of vendored packages (without creating a dependency).
- The `importlib_metadata` hack can be removed (as only one copy of `importlib_metadata` will ever be imported).
- Vendored packages no longer need to be patched.
- De-vendoring is as simple as removing the `_vendor` directory (and supplying the dependencies).
- It would be trivially easy to make the vendored packages an opt-in, such as requiring an environment variable.
- Mypy checks now catch real issues with dependencies.

In order to avoid cyclic dependencies, the declared dependencies are for now declared in an optional `core` extra so they now appear in the metadata.

Ref #2825

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
It's hard to see in GitHub, due to the changes to the vendored directories, but here's a stat summary excluding `_vendor`:

```
 setuptools debt/2825-devendor @ git diff --stat main.. -- ':(exclude)*/_vendor/*'
 .github/workflows/main.yml                |   1 -
 .gitignore                                |  23 ++++++++++
 conftest.py                               |   3 --
 mypy.ini                                  |  11 +++--
 newsfragments/2825.removal.rst            |   1 +
 pkg_resources/__init__.py                 |  33 ++++++++------
 pkg_resources/extern/__init__.py          | 104 -------------------------------------------
 pkg_resources/tests/test_resources.py     |   2 +-
 pyproject.toml                            |  14 +++++-
 setuptools/__init__.py                    |   3 ++
 setuptools/_core_metadata.py              |   8 ++--
 setuptools/_entry_points.py               |   6 +--
 setuptools/_importlib.py                  |  44 ++----------------
 setuptools/_itertools.py                  |   2 +-
 setuptools/_normalization.py              |   4 +-
 setuptools/_reqs.py                       |   4 +-
 setuptools/command/_requirestxt.py        |   4 +-
 setuptools/command/bdist_wheel.py         |  12 ++---
 setuptools/command/build_py.py            |   2 +-
 setuptools/command/easy_install.py        |   2 +-
 setuptools/command/editable_wheel.py      |   2 +-
 setuptools/command/egg_info.py            |   2 +-
 setuptools/command/test.py                |   4 +-
 setuptools/compat/py310.py                |   2 +-
 setuptools/config/_apply_pyprojecttoml.py |   4 +-
 setuptools/config/expand.py               |   4 +-
 setuptools/config/pyprojecttoml.py        |   2 +-
 setuptools/config/setupcfg.py             |   8 ++--
 setuptools/depends.py                     |   2 +-
 setuptools/dist.py                        |  10 ++---
 setuptools/extern/__init__.py             |  92 --------------------------------------
 setuptools/msvc.py                        |   3 +-
 setuptools/package_index.py               |   3 +-
 setuptools/tests/config/test_setupcfg.py  |   2 +-
 setuptools/tests/test_bdist_wheel.py      |   6 +--
 setuptools/tests/test_extern.py           |   2 +-
 setuptools/tests/test_setuptools.py       |   2 +-
 setuptools/tests/test_wheel.py            |   4 +-
 setuptools/wheel.py                       |   7 +--
 tools/vendored.py                         | 234 +++++++++++++-----------------------------------------------------------------------------------
 tox.ini                                   |   9 ++--
 41 files changed, 162 insertions(+), 525 deletions(-)
```

Net 363 lines removed.

abravalheri (commented):

This is a though one. In principle I have no problem with this and I think it would be fine. But it is hard to foresee the consequences of it.

The complicated part is that setuptools has no way of telling the installer/builder that the build-dependencies are "optional in a way", and that if they fail to install the dependencies (e.g. due to cycles) it is fine because we also have them bundled.

For example: if the user passes some flags to `pip` that forces `pip` to no use wheels for the installation, will this also mean that the build will fail with this new approach to vendoring?

I believe this is related to https://github.com/pypa/packaging-problems/issues/342. Ideally we should have a standard solution for that problem.

It would be interesting to hear @pfmoore and @pradyunsg thoughts on this matter.

pfmoore:
> It would be interesting to hear @pfmoore and @pradyunsg thoughts on this matter.

My main thought is that `--no-binary :all:` needs to be supported as an installation process. I don't personally have a need for this, but at the time we added in-tree build backend support, one of the key drivers was that there was a fairly vocal group of users who insisted that being able to build everything from source was essential. I assume that currently they are happy with the state of the ecosystem (at least, we haven't heard much from them recently), but we should ensure that a change like this doesn't trigger a new round of debate.

Unfortunately, I don't have any specific examples of people who should be consulted on this question. Maybe opening a discussion on the packaging Discourse would be worthwhile?

abravalheri:
Thank you @pfmoore.

This brings me to the question: if setuptools can build itself with no extra dependency listed on `pyproject.toml [build-system] requires` (thanks to bundling the vendored dependencies), but it does list dependencies in `pyproject.toml  [project] dependencies` does it count as creating a cycle in the same way as PEP 517 forbids.

pfmoore:
I'm not sure I follow your question. To build setuptools, the `pyproject.toml` says that there are no build requirements, the build backend path is ".", and you build using setuptools. If that works (i.e., all of the dependencies are picked up from vendored copies in the source tree) then building setuptools will work, yes.

However, if you want to install some project X that uses setuptools, the installer will set up a build environment that contains setuptools (building and then installing it as described above) and it will then install all of the dependencies listed in `project.depends` for setuptools. This will build each of those dependencies (likely using setuptools recursively to do so). It's possible that cached wheels will make this appear to work, but if you globally configure pip to not use a cache, you'll probably get an infinite recursion of build steps (which pip will stop after a certain limit, but I don't know if uv has considered this case, so be careful - it *does* crash your machine if left unchecked).

So I guess my answer is yes, this looks potentially dangerous to me.

jaraco:
I notice that hatchling is [able to declare dependencies](https://github.com/pypa/hatch/blob/332c407189a9e6001578eaa8baad95b466eea2f4/backend/pyproject.toml#L37-L43). Is that because none of the dependencies use hatchling as their build backend (trove-classifiers and pluggy use setuptools, tomli and packaging and pathspec use flit)? Does that mean that hatchling would break the world if pathspec or trove-classifiers were to switch to hatchling? I see hatchling does something clever - it relies on its own [ouroboros](https://github.com/pypa/hatch/blob/master/backend/src/hatchling/ouroboros.py) backend for getting build dependencies as part of `get_requires_for_build_*`.

I see that flit addresses this concern by implementing the backend with no dependencies (flit-core). Maybe setuptools should do something similar, and have a separate backend with no dependencies that can build setuptools.

> So I guess my answer is yes, this looks potentially dangerous to me.

Let me see if I can reproduce the danger. I'll get a Docker container that's resource-constrained and try to get the system to fall over building this branch from source.

Perhaps I should simply remove the declared dependencies for now, or move them to optional dependencies. That would separate this concern about cyclic (build) dependencies from simplifying the vendoring.

jaraco:
> Let me see if I can reproduce the danger. I'll get a Docker container that's resource-constrained and try to get the system to fall over building this branch from source.

I tried to reproduce the danger, but have thusfar been unsuccessful. I have this dockerfile:

```Dockerfile
FROM jaraco/multipy-tox
RUN git clone https://github.com/pypa/setuptools
WORKDIR setuptools
RUN git checkout debt/2825-devendor
ENV PIP_NO_CACHE=1
ENV PIP_NO_BINARY=:all:
# disable pip-run
ENV PYTHONPATH=
CMD py -3.12 -m pip install .
```

It appears to be succeeding in installing setuptools and its dependencies (though it fails on `autocommand` due to https://github.com/Lucretiel/autocommand/issues/28).

What I realized, though, is that when building a dependency and thus getting "setuptools" from source, it's getting it from PyPI, which doesn't yet have the declared dependencies. To simulate the danger, I'll need to upload an sdist of this PR somewhere and use that index.

> My main thought is that `--no-binary :all:` needs to be supported as an installation process. I don't personally have a need for this, but at the time we added in-tree build backend support, one of the key drivers was that there was a fairly vocal group of users who insisted that being able to build everything from source was essential. I assume that currently they are happy with the state of the ecosystem (at least, we haven't heard much from them recently), but we should ensure that a change like this doesn't trigger a new round of debate.

I took a moment to brush up on https://github.com/pypa/packaging-problems/issues/342, which attempts to capture the issue. I believe it's an outstanding issue and it's only been avoided by the following implicit constraint:

- build backends can only have (build or runtime) dependencies on projects that don't rely on that build backend.

I think I've got that right. A build backend can have dependencies if all of the dependencies rely on a different build backend (hatchling, for now). Or a build backend can have no dependencies (flit_core). I wonder if we should document this constraint somewhere. Is there a place where we could document this constraint?

jaraco:
In the latest commit, in order to avoid concerns about declaring dependencies (by default), I've moved them into an optional `core` extra. Now this PR is mainly about simplifying the vendoring.

I note that making this change revealed two bugs in the vendored packages that were masked by the naturally-installed dependencies.

pfmoore:
> I tried to reproduce the danger, but have thusfar been unsuccessful. I have this dockerfile

It's worth adding `-vvv` to the pip invocation and checking if everything really is being built from scratch. It's worth noting that the option is `--no-cache-dir`, so the environment variable should be `PIP_NO_CACHE_DIR`, not `PIP_NO_CACHE`, which is what you set).

> I think I've got that right. A build backend can have dependencies if all of the dependencies rely on a different build backend (hatchling, for now). Or a build backend can have no dependencies (flit_core). I wonder if we should document this constraint somewhere. Is there a place where we could document this constraint?

I don't think it's an explicit constraint, rather it's an implied consequence of the defined behaviour when combined with a requirement to build everything from source. In particular, it's not a constraint just on build backends, it affects projects that are dependencies of build backends as well - if setuptools depends on wheel, and wheel is built using setuptools, then *wheel* has a problem if you want to build it under a constraint of never using pre-built binaries. Conversely, if you are willing to use a multi-stage process where you build a setuptools wheel "by hand" and then use it to build everything else, there's no issue.

I'm not the best person to comment here, though, as I don't understand the whole "everything must be built from source" mindset - you have to start somewhere, and why can't that be "pre-build a setuptools wheel using some manual recipe, then use that to bootstrap everything else"? But from what I recall, there were some purists for whom that wasn't sufficient - and I never got a clear understanding of why.

jaraco:
Let's take the conversation about bootstrapping to the other issue for discussion.

jaraco:
In this dockerfile, I believe I've demonstrated the danger:

```Dockerfile
FROM jaraco/multipy-tox

# download the sdist and all of the dependencies
RUN pip download --no-binary :all: --dest index 'autocommand@git+https://github.com/jaraco/autocommand@debt/remove-duplicate-license' 'setuptools[core]@git+https://github.com/pypa/setuptools@242ef247e999bfe6e7a3a8eb4d984cb948ad0b4d'

# download flit-core for ordered-set
RUN pip download --no-binary :all: --dest index flit_core

# download setuptools-scm for jaraco.text
RUN pip download --no-binary :all: --dest index setuptools_scm

# download hatch-vcs for wheel
RUN pip download --no-binary :all: --dest index hatch-vcs

# download calver for trove-classifiers
RUN pip download --no-binary :all: --dest index calver

ENV PIP_NO_INDEX=1
ENV PIP_NO_CACHE_DIR=1

# disable pip-run
ENV PYTHONPATH=

CMD py -3.12 -m pip install --find-links=index setuptools
```

If I run that, it fails with `LookupError: file:///index/jaraco_text-3.12.1.tar.gz is already being built: jaraco.text>=3.7 from file:///index/jaraco_text-3.12.1.tar.gz (from setuptools)`.

If I change the setuptools tarball to the latest (`setuptools[core]@git+https://github.com/pypa/setuptools@debt/2825-devendor`), the routine completes (indicating that making the dependencies optional works around the danger).

jaraco:
As I've reduced the scope of this issue to merely simplifying the vendoring (and deferred considerations about declaring direct dependencies), would you take another look, @abravalheri ?

newsfragments/2825.removal.rst
```diff
@@ -0,0 +1 @@
+Now setuptools declares its own dependencies in the ``core`` extra. Dependencies are still vendored for bootstrapping purposes, but installed dependencies are preferred. Downstream packagers can de-vendor by simply removing the ``setuptools/_vendor`` directory.
```
abravalheri, line 1: Should we change this changelog entry?

Specifically the phrase "but installed dependencies are preferred" is a bit puzzling to me for the following reason:

* It seems to indicate that we want users to write:
   ```toml
   [build-system]
   requires = [
      "setuptools[core]"
   ]
   ...
   ```
   But we know that this can cause cycles and therefore clashes with PEP 517... So I don't think we should recommend that...

Due to the limitations that we have right now `core` can only be "informational" for the time being (but not intended for general use).

pkg_resources/__init__.py
```diff
@@ -74,6 +74,8 @@
 
 import _imp
 
+sys.path.append(os.path.dirname(__file__) + '/../setuptools/_vendor')
```
abravalheri, line 77: We might end up with 2 duplicated entries in `sys.path` right? (one for importing `pkg_resources` and another one for importing `setuptools`).

I suppose that this is not too bad...

setuptools/__init__.py
```diff
@@ -3,8 +3,11 @@
 import functools
 import os
 import re
+import sys
 from typing import TYPE_CHECKING
 
+sys.path.append(os.path.dirname(__file__) + '/_vendor')
```
abravalheri, line 9: I think this is worth a try. We might have to revert if it causes too much trouble, but it is good to experiment to see if it is a viable solution.

abravalheri (approved):
Thank you Jason, 

Great work on this! I've added a few minor comments, but overall, I think this is an interesting direction to explore. 👍

---

Just to clarify, this PR doesn't address the duplication of `packaging` in `pypa/distutils`, correct?

I'm of the opinion that `distutils` doesn't necessarily need to implement all the PEPs itself. For instance, we could use a simpler version of `canonicalize_name`, like `re.sub("[^A-Za-z0-9]+", "_", name)`, and let `setuptools` handle the more complex tasks.

Perhaps we could discuss ways to make `distutils` more extensible without the need for patching, especially considering our plans to deprecate `SETUPTOOLS_USE_DISTUTILS=stdlib`.

---

On another note, would you be open to revisiting our policy on adding vendored dependencies?

I understand that https://github.com/pypa/packaging-problems/issues/342 might take some time to reach a consensus, and in the meantime, it might be prudent to be more conservative with our dependencies.

While dependencies can certainly streamline our codebase and simplify maintenance, I'm starting to worry about the length of the chains of indirect/transient dependencies. Initially, we had `packaging`, `importlib_{resources,metadata}`, which are very reasonable. But now we're seeing unrelated dependencies like `autocommand` popping up, which might be a bit much.

There are also simpler implementations we could consider. For example, since `dict` has started to maintain insertion order in Python 3.6, it could potentially replace `orderedset`[^1].

This could be a good opportunity to establish some guidelines on when and why we decide to add a dependency.

Does this make sense? Looking forward to hearing your thoughts!

[^1]: We can also check to which extension it is safe to fallback to stdlib's `importlib.{metadata,resources}` when the backport is not available...

pkg_resources/__init__.py
jaraco: And it could be even more if those modules end up getting reloaded. My feeling was that such situations are probably rare and the inefficiencies caused are acceptable, compared with the cost of making this one-liner more complicated.

jaraco:
> Just to clarify, this PR doesn't address the duplication of `packaging` in `pypa/distutils`, correct?

That's right. I'm hoping to focus on deprecating distutils except as bundled. Once that's done, setuptools can fully adopt the distutils codebase and we can drop the dual-repo model (and share any vendored packages).

I've been tempted to _release_ distutils and then vendor it into setuptools, but I'm avoiding that approach as I'd like not to encourage more use of distutils except through setuptools.


> On another note, would you be open to revisiting our policy on adding vendored dependencies?

Sure. Feel free to start a discussion or issue about it.

> I understand that [pypa/packaging-problems#342](https://github.com/pypa/packaging-problems/issues/342) might take some time to reach a consensus, and in the meantime, it might be prudent to be more conservative with our dependencies.

I do aspire to invest time here, especially if I continue to have the luxury to pursue my open source work, but I agree, it's a potentially sticky problem.

> While dependencies can certainly streamline our codebase and simplify maintenance, I'm starting to worry about the length of the chains of indirect/transient dependencies. Initially, we had `packaging`, `importlib_{resources,metadata}`, which are very reasonable. But now we're seeing unrelated dependencies like `autocommand` popping up, which might be a bit much.

Yeah, I bemoan these transitive dependencies too. I would like to refactor some of these supporting libraries so that they have fewer unused dependencies (as inflect and autocommand are in many cases). This problem would be easier if packages could have "default extras" that could be deselected on install (or vendoring).

Maybe we should consider updating the vendoring routine to suppress/hide any dependencies that are actually not needed. That feels brittle and probably our time is better spent refactoring the libraries.

My instinct is that if there are more libraries in the dependency tree that are unused than those that are used, a reorganization is in order.

> There are also simpler implementations we could consider. For example, since `dict` has started to maintain insertion order in Python 3.6, it could potentially replace `orderedset`[1](#user-content-fn-1-c960100c048a5b3e4e9203dcdf1c5c1d).

I'm always a fan of lighter-weight dependencies or drop-in replacements from the stdlib.

> This could be a good opportunity to establish some guidelines on when and why we decide to add a dependency.

I'd be fine with establishing some guidelines. My default, and somewhat strongly-held, opinion is that dependencies should be readily adopted as needed. I can certainly write up my motivations and considerations in detail, but for here I'll summarize. I want to avoid repeating one's self (DRY), use the right tool for the job, and only make pragmatic compromises where necessary (or there's a strong case). I have a secondary motivation to publish, mature, and prove useful patterns that would be useful in the stdlib except for having a home to demonstrate and account for their value (things like [`compose'](https://jaracofunctools.readthedocs.io/en/latest/?badge=latest#jaraco.functools.compose) or [`Projection`](https://jaracocollections.readthedocs.io/en/latest/#jaraco.collections.Projection); I'm aiming to lay the egg of the chicken/egg problem (it's not used because it's not available; it's not made available because no one is using it).

Avasam:
Seems like this would be quite helpful for #2345 (and may fix #4414 ? )

pkg_resources/__init__.py
```diff
@@ -1118,7 +1119,7 @@ def markers_pass(self, req: Requirement, extras: tuple[str, ...] | None = None):
         evaluation. Otherwise, return True.
         """
         extra_evals = (
-            req.marker.evaluate({'extra': extra})
+            req.marker.evaluate({'extra': extra})  # type: ignore
             for extra in self.get(req, ()) + (extras or (None,))
         )
         return not req.marker or any(extra_evals)
```
Avasam, line 1125:
```suggestion
        if not req.marker:
            return True
        extra_evals = (
            req.marker.evaluate({'extra': extra})
            for extra in self.get(req, ()) + (extras or ("",))
        )
        return any(extra_evals)
```
or more compact:
```suggestion
        return not req.marker or any(
            req.marker.evaluate({'extra': extra})
            for extra in self.get(req, ()) + (extras or ("",))
        )
```


The early `req.marker` check solves [union-attr](https://mypy.readthedocs.io/en/latest/_refs.html#code-union-attr).

The `None`-->`""` change solves [dict-item](https://mypy.readthedocs.io/en/latest/_refs.html#code-dict-item) and avoids a deprecated code path (see  https://github.com/pypa/packaging/blob/793ee28b4c70886e8de5731e24e444388916b3ee/src/packaging/markers.py#L322-L323)
I thought about replacing `("",)` with an empty tuple `()` since `evaluate` defaults `extra` to an empty string, but that would change the result of `any` if `self.get(req, ()) + (extras or ())` results in an empty tuple.

pkg_resources/__init__.py
```diff
@@ -3454,7 +3458,7 @@ def __eq__(self, other: object):
     def __ne__(self, other):
         return not self == other
 
-    def __contains__(self, item: Distribution | str | tuple[str, ...]) -> bool:
+    def __contains__(self, item: Distribution | str) -> bool:
```
Avasam, line 3461:
```python
from packaging import specifiers as _packaging_specifiers
# ...
    def __contains__(
        self, item: Distribution | _packaging_specifiers.UnparsedVersion
    ) -> bool:
```
or
```suggestion
    def __contains__(
        self, item: Distribution | _packaging_version.Version | str
    ) -> bool:```

File: pkg_resources/__init__.py [hunk 6]
```diff
@@ -3464,7 +3468,10 @@ def __contains__(self, item: Distribution | str | tuple[str, ...]) -> bool:
         # Allow prereleases always in order to match the previous behavior of
         # this method. In the future this should be smarter and follow PEP 440
         # more accurately.
-        return self.specifier.contains(item, prereleases=True)
+        return self.specifier.contains(
+            item,  # type: ignore[arg-type]
```
Avasam, line 3472: GitHub doesn't allow me to add suggestions that far back in unchanged lines, but here's what the proper fix to ensure `item` isn't seen as potentially a `Distribution` by mypy.
```python
            _item = item.version
        else:
            _item = item

        # Allow prereleases always in order to match the previous behavior of
        # this method. In the future this should be smarter and follow PEP 440
        # more accurately.
        return self.specifier.contains(
            _item,
```

pkg_resources/__init__.py
jaraco: I very much prefer the more compact single expression and return value.

I wonder if there's a bug reported to mypy about the `union-attr` not being resolved ahead of the lazily-generated `extra_evals` (it theoretically could detect that the generator is not entered until after the `req.marker` is resolved not to be `None`).

I'm a little uneasy with the `""`, but since that appears to be the type-spec'd indicator for the empty extra, SGTM.

jaraco: Well heck, I tried to commit the suggestion, but GitHub failed, saying that I couldn't apply changes to deleted lines and I should refresh and try again. I refreshed and tried again, but the commit message I'd created was lost. Boo GitHub.

pkg_resources/__init__.py
Avasam: Isn't it a simple `if path not in sys.path` ?
```suggestion
_vendor_path = os.path.dirname(os.path.dirname(__file__)) + "/setuptools/_vendor"
if _vendor_path not in sys.path:
    sys.path.append(_vendor_path)
```

pkg_resources/__init__.py
jaraco: I tried again and got the same error:

image

jaraco: Indeed, when I go to the Files Changed tab, I can see that these suggestions are made against the original code, so although GitHub allows making the suggestion, it can't apply it. Also, the guidance to Refresh and try again is just broken, as anyone following that guidance would be caught in an infinite loop.

jaraco: Applied in 225f7fe32f5a07b99ff3fb403f517e6dbf208d68.

pkg_resources/__init__.py
jaraco: Applied in 0e1e7078560708b450ff5fdd59b9c1f425b15595.

setuptools/__init__.py
Avasam: To go along with this comment https://github.com/pypa/setuptools/pull/4457/files#r1670998269 :

```suggestion
_vendor_path = os.path.join(os.path.dirname(__file__), "_vendor")
if _vendor_path not in sys.path:
    sys.path.append(_vendor_path)
```

Sure it's "more complex" than a one-liner, but not really by much.

pkg_resources/__init__.py
Avasam: L GitHub indeed, broken because I manually wrote `suggestion` I guess. Glad you figured it out.

pkg_resources/__init__.py
jaraco: Yes, it's "as simple" as that, but as you've demonstrated, that changes this from a one-line expression into a multi-line, conditional code block. And since this logic needs to be copy/pasted, I'm loathe to make it more than one line. If it is more than one line, I'm tempted to create a separate module to contain the shared logic, but I'm also disinclined to do that because it would require creating a new top-level module.

The reason to have it in one place or in a single expression is it resists being tweaked. It resists, for example, the temptation to add behavior around the different lines, whereas a single expression stands on its own (it's self-encapsulating).

We could keep it a single expression with something like:

```suggestion
sys.path.extend(((vendor_path := os.path.join(os.path.dirname(os.path.dirname(__file__)), "setuptools", "_vendor")) not in sys.path) * [vendor_path])  # fmt: skip
```

It's convoluted, strains readability, but at least remains on one line and one expression.

Allowing ruff to reformat it results in:

```suggestion
sys.path.extend(
    (
        (
            vendor_path := os.path.join(
                os.path.dirname(os.path.dirname(__file__)), "setuptools", "_vendor"
            )
        )
        not in sys.path
    )
    * [vendor_path]
)
```

I'm a little unsure about whether to use `os.path.join` or not. I agree, it's often preferable (best practice) to use `join`, but in this case, it's conceivable that the path could be inside a zipfile, in which cases only `posixpath.sep` is supported, and since Windows has supported forward slashes since at least DOS 2.1, I'm comfortable using them for something like this hack (that we hope to retire). I'm not explicitly supporting vendoring for zip files, but it would be nice if it happens to work.

Indeed, I did some testing and confirmed that by using forward-slashes (and replacing `..` with another call to `dirname`), the technique works to load vendored packages even in a zip file:

```
 setuptools debt/2825-devendor [1] @ git diff
diff --git a/pkg_resources/__init__.py b/pkg_resources/__init__.py
index b8cf28cdb..21e070406 100644
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -74,7 +74,17 @@ from pkgutil import get_importer
 
 import _imp
 
-sys.path.append(os.path.dirname(__file__) + '/../setuptools/_vendor')
+sys.path.extend(
+    (
+        (
+            vendor_path := posixpath.join(
+                os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor'
+            )
+        )
+        not in sys.path
+    )
+    * [vendor_path]
+)
 
 # capture these to bypass sandboxing
 from os import utime
 setuptools debt/2825-devendor @ py -m zipfile -c pr.zip pkg_resources setuptools
 setuptools debt/2825-devendor @ env PYTHONPATH=pr.zip py -c "import sys; sys.path.remove(''); import pkg_resources" && echo done
<string>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  import sys; sys.path.remove(''); import pkg_resources
done
```

Replace `posixpath` with `ntpath` and the import fails. Interestingly, with `ntpath` the import only fails on Unix, so apparently the zip importer does support using backslashes inside the zipfile. Oh, yes, I reviewed the zipimporter logic recently and (to my surprise), it actually converts forward slashes to backslashes when reading zipfile entries on Windows.

Given those findings, I'm fine to use `os.path.join`, as much as I've come to dislike the pattern.

```suggestion
sys.path.extend(
    (
        (
            vendor_path := os.path.join(
                os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor'
            )
        )
        not in sys.path
    )
    * [vendor_path]
)
```

```suggestion
sys.path.extend(((vendor_path := os.path.join(os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor')) not in sys.path) * [vendor_path])  # fmt: skip
```

Let's go with that and use the same technique in both setuptools and pkg_resources.

jaraco: Done in 930ebe55b.

setuptools/__init__.py
jaraco: Done in 930ebe55b.

pkg_resources/__init__.py
Avasam: The `\` in zipfile is something I wouldn't even have considered, thanks for testing. My concern about consistent path separator was just so you wouldn't end up with the same path twice, differing only by slashes (inconsistent path separators would still protect against module reloading)

pkg_resources/__init__.py
jaraco: Done in 9eb89def7.

Avasam (approved):

Avasam (approved):

jaraco:
The test failures appear to be unrelated to the PR. The new failures also occur on main. I thought maybe it was the recent ruff release, but pinning to `ruff<0.5`, I get the same failures.

Avasam:
> The test failures appear to be unrelated to the PR. The new failures also occur on main. I thought maybe it was the recent ruff release, but pinning to `ruff<0.5`, I get the same failures.

With https://github.com/pypa/setuptools/pull/4454 merged it should be good now.

jaraco:
Now the tests are failing on Windows with:

```
---------------------------- Captured stderr call -----------------------------
ruff failed
  Cause: Failed to rename temporary cache file to D:\a\setuptools\setuptools\.ruff_cache\0.5.1\13567325068112760734
  Cause: failed to persist temporary file: Access is denied. (os error 5)
  Cause: Access is denied. (os error 5)

```

Likely unrelated to this change and maybe implicated in ruff 0.5.1 or pytest-ruff 0.4.

jaraco:
Issues with failing Windows builds were addressed in #4467. I plan to merge this, but only after 70.3.0 has been out and stable enough (a week or two without major regressions).

setuptools/_vendor/autocommand-2.2.2.dist-info/LICENSE
```diff
@@ -0,0 +1,166 @@
+GNU LESSER GENERAL PUBLIC LICENSE
```
carsongee, line 1: Sorry to ghost post on you, but why did autocommand get added here? The rest look they were modifications, but this seems like a new _vendor package entirely. I ask because as it is LGPL, we have some licensing concerns. Thanks!

Avasam: Licensing concerns aside, I can answer your first question: `autocommand` is a dependency of `jaraco.text`. Setuptools doesn't need it, it's just pulled in with dependencies. If you *really* need to I'm sure you can devendor it.

pkg_resources/__init__.py
```diff
@@ -74,6 +74,8 @@
 
 import _imp
 
+sys.path.extend(((vendor_path := os.path.join(os.path.dirname(os.path.dirname(__file__)), 'setuptools', '_vendor')) not in sys.path) * [vendor_path])  # fmt: skip
```
gaborbernat, line 77: This hack breaks when this package is within a zip file... https://github.com/pypa/virtualenv/pull/1727 cc @jaraco

jaraco: Sorry to hear that.

It's not fair to characterize this as a hack. It's the vendoring approach, which is required by the ecosystem to satisfy PEP 517's "no circular build dependencies" constraint. Setuptools is following best-available standards for supplying sophisticated behavior. There's a lot of work going on in https://github.com/pypa/packaging-problems/issues/342 to solve the issue more holistically so that build backends can have and declare proper dependencies.

When I first authored this change, I considered the zipimport case and thought that this approach might work for the zipimport case but I never verified.

Since the release of this change, the vendoring support is there as a fallback for environments that cannot include the dependencies naturally. The best approach would be for the app/environment builder to ensure that `setuptools[core]` is installed (such that vendoring is unnecessary).

I'd be open to expanding the vendoring support to support better the zipimport case. I'll open an issue about it.

PR #4464, opened 2024-07-09
Test coverage config update for protocols

Avasam:
Just a test to confirm whether
1. Single-line elipsis bodies are automatically excluded from coverage (making `@overload` redundant)
2. Excluding Protocols would resolve failing coverage for https://github.com/pypa/setuptools/pull/4192

If this works like I think it would, I'll PR to the skeleton.


Avasam:
Results of this test are satisfying and a PR has been opened upstream : https://github.com/jaraco/skeleton/pull/135

PR #4470, opened 2024-07-11
Rename arguments on _EditableFinder and _EditableNamespaceFinder

mtreglia-gpsw:
This commit rename the `find_spec` arguments in _EditableFinder and _EditableNamespaceFinder to be compliant of the standard `importlib.abc.MetaPathFinder` interface.

https://docs.python.org/3/library/importlib.html#importlib.machinery.PathFinder.find_spec https://github.com/abravalheri/typeshed/blob/eecac02ef7fdbe82b9dc57fa3ca4bb9d341116d7/stdlib/_typeshed/importlib.pyi

The `find_module` arguments were not renamed because the method `find_module` is deprecated starting from Pyhton3.10 and has been removed in Pyhton3.12

https://docs.python.org/3/reference/import.html#the-meta-path

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -805,7 +805,7 @@ PATH_PLACEHOLDER = {name!r} + ".__path_hook__"
 
 class _EditableFinder:  # MetaPathFinder
     @classmethod
-    def find_spec(cls, fullname: str, _path=None, _target=None) -> ModuleSpec | None:
+    def find_spec(cls, fullname: str, path=None, target=None) -> ModuleSpec | None:  # type: ignore
         # Top-level packages and modules (we know these exist in the FS)
         if fullname in MAPPING:
             pkg_path = MAPPING[fullname]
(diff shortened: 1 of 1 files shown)


abravalheri:
Thanks @mtreglia-gpsw .

Could you please add the specific type-ignore annotation?

abravalheri:
Thank you very much @mtreglia-gpsw .

I don't know how this is handled in other libraries or if there is a convention for preferring `del`, but I would normally add the specific type ignore comment instead...

Anyway, I guess that is a matter of preference.

mtreglia-gpsw:
> Thank you very much @mtreglia-gpsw .
> 
> I don't know how this is handled in other libraries or if there is a convention for preferring `del`, but I would normally add the specific type ignore comment instead...
> 
> Anyway, I guess that is a matter of preference.

Indeed with the `del` is not required the `#type-ignore`. 
The `del` will just immediately decrement the ref counter for the variable instead of doing it when out of scope. 

It's a matter of preference, If conventionally you prefer the type-ignore I will change it.

abravalheri:
Thank you.

PR #4471, opened 2024-07-16
Update distutils-legacy.rst

artificial-intelligence:
fix some small capitalization typos


lazka:
`Command` is a class and named that way and not `command`.

@artificial-intelligence Could you explain your change?

jaraco:
In addition to this change being mistaken, it's also found originally in [distutils](/pypa/distutils), so any changes to this doc should go there, but probably there should be no changes.

PR #4229, opened 2024-02-20
Fix "ModuleNotFoundError" for bdist_wheel on Python 3.12

christian-krieg:
## Summary of changes
When running ``python3 setup.py bdist_wheel`` on Python 3.12, setuptools exits with:

```
writing byte-compilation script '/tmp/<tmpfilename>.py'
[...]
Traceback (most recent call last):
  File "/tmp/<tmpfilename>.py", line 1, in <module>
    from distutils.util import byte_compile
ModuleNotFoundError: No module named 'distutils'
```
After some research, I learned that `distutils` moved to `setuptools` (https://github.com/pypa/setuptools/issues/3661).

Importing `setuptools` in the generated byte-compilation script solves the issue.

Closes 

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added in [`newsfragments/`].
  _(See [documentation][PR docs] for details)_

[`newsfragments/`]: https://github.com/pypa/setuptools/tree/master/newsfragments
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #4484, opened 2024-07-18
Fix ModuleNotFoundError for _distutils/util.py:byte_compile

kfollesdal:
## Summary of changes
Solve `ModuleNotfoundError 'No module named 'distutils'` for function `_distutils/util.py:byte_compile.` By adding `import setuptools` to the script in `byte_complie`, `distutils` will be available.

The python package `thrift` is broken for python 3.12 in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/a0812b380b9d17e9dd02b677e4052bf446b97cbe/pkgs/development/python-modules/thrift/default.nix#L12), since `distutils` was removed in python 3.12. The [thrift](https://github.com/apache/thrift/blob/master/lib/py/setup.py) python package use setuptools/distutils to build. 

The build process use the function ` _distutils/util.py:byte_compile`, that run a python script in a subprocess. Problem is that we get error:

```
       >   File "/build/tmpgob9mq6_.py", line 1, in <module>
       >     from distutils.util import byte_compile
       > ModuleNotFoundError: No module named 'distutils'
```
 when it try to build. Problem is that in the subprocess the `setuptool` is not imported and therefore are `distutils` not available.  
 
 This PR solve the problem by import setuptools first in the script.

### DIFF
--- a/setuptools/_distutils/util.py
+++ b/setuptools/_distutils/util.py
@@ -414,6 +414,7 @@ def byte_compile(  # noqa: C901
             with script:
                 script.write(
                     """\
+import setuptools
 from distutils.util import byte_compile
 files = [
 """


kfollesdal:
Discovered that it is a duplicate of https://github.com/pypa/setuptools/pull/4229/

kfollesdal:
Maybe relevant? https://github.com/pypa/distutils/issues/224

kfollesdal:
Close since duplicate of https://github.com/pypa/setuptools/pull/4229

Found a solutions by modify `PYTHONPATH` before build: 
```
DISTUTILS=$(mktemp -d)
ln -s ${setuptools}/${python.sitePackages}/setuptools/_distutils "$DISTUTILS/distutils"
PYTHONPATH="$PYTHONPATH:$DISTUTILS"
```

PR #4492, opened 2024-07-18
Add Python version specifiers to [core] dependencies

mgorny:
## Summary of changes

Add Python version specifiers to importlib_metadata, importlib_resources and tomli dependencies, to require them only on Python versions on which they are actually used.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/newsfragments/4492.misc.rst
+++ b/newsfragments/4492.misc.rst
@@ -1,0 +1,1 @@
+Now backports in ``core`` dependencies are installed only on Python versions requiring them.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,9 +107,9 @@ core = [
 	"ordered-set>=3.1.1",
 	"more_itertools>=8.8",
 	"jaraco.text>=3.7",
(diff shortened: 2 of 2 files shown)


jaraco:
Unfortunately, applying this change will almost certainly break the [update routine for the vendored packages](https://github.com/pypa/setuptools/blob/3fca2492380dfe2820529829c5b9acb072c7fecb/tools/vendored.py#L51-L72) due to [pip not respecting environment markers](https://github.com/pypa/pip/issues/12770). If we add these exclusions, we'll also at the very least need to add them to the workaround in that routine. Or perhaps something similar to the `upgrade_core` could be done to strip the environment markers as a workaround for direct dependencies.

Even better would be to find a solution to pypa/pip#12770; perhaps `uv` doesn't have that issue and could be used instead.

jaraco:
In 022cedbf2, I switched to uv for vendoring and it works great, so no longer have any concerns about the markers.

mgorny:
Thanks!

Issue #4473, opened 2024-07-17
Emergent ruff error with 71.0 release


jaraco:
Although tests passed last week, after merging #4457 and cutting a release, tests started failing with this lint error:

```
>           raise RuffError(stdout.decode())
E           pytest_ruff.RuffError: setuptools/command/editable_wheel.py:302:39: E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
E               |
E           300 |         for name in build.get_sub_commands():
E           301 |             cmd = self.get_finalized_command(name)
E           302 |             if name == "build_py" and type(cmd) != build_py_cls:
E               |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^ E721
E           303 |                 self._safely_run(name)
E           304 |             else:
E               |

.tox/py/lib/python3.11/site-packages/pytest_ruff/__init__.py:101: RuffError
```

jaraco:
It seems it was bugfix release of ruff that enabled the new error:

```
 setuptools main 🐚 ruff check setuptools/command/editable_wheel.py
All checks passed!
 setuptools main 🐚 pipx upgrade ruff
⚠️ Found a space in the home path. We heavily discourage this, due to multiple incompatibilities. Please check our docs for more information on
    this, as well as some pointers on how to migrate to a different home path.
upgraded package ruff from 0.5.1 to 0.5.2 (location: /Users/jaraco/Library/Application Support/pipx/venvs/ruff)
 setuptools main 🐚 ruff check setuptools/command/editable_wheel.py
setuptools/command/editable_wheel.py:302:39: E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
    |
300 |         for name in build.get_sub_commands():
301 |             cmd = self.get_finalized_command(name)
302 |             if name == "build_py" and type(cmd) != build_py_cls:
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^ E721
303 |                 self._safely_run(name)
304 |             else:
    |

Found 1 error.
```

jaraco:
Looks like the change was due to an unintended false negative (https://github.com/astral-sh/ruff/pull/12300).

jaraco:
Oh, gosh. There's more than one.

```
>           raise RuffError(stdout.decode())
[...]

PR #4500, opened 2024-07-20
Update mypy to 1.11

Avasam:
## Summary of changes
- Update mypy to 1.11
- Slightly relax version lock to allow automatic point releases
- Updated `follow_imports = silent` comment to be more generalized / future-proof
- As per https://github.com/pypa/setuptools/issues/4473#issuecomment-2234155911 , indicate myself as owner of the mypy pin, until the removal of `types-setuptools` from typeshed. Right now a CI failure is expected almost every version, especially as more annotations are added to setuptools, and fixes upstreamed to typeshed. Which would require a lock, then unlock PRs everytime (more churn than just locking and updating when ready), or non-obvious fixes (mainly due to distutils subtyping). Once `types-setuptools` is obsoleted, things should be more stable. @jaraco 

Goes towards unblocking #4352
Closes #4432

### Pull Request Checklist
- [x] Changes have tests (these are test updates)
- [x] News fragment added in [`newsfragments/`]. (no user facing changes)

[PR docs]:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,10 @@ test = [
 	# for tools/finalize.py
 	'jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"',
 	"pytest-home >= 0.5",
-	"mypy==1.10.0", # pin mypy version so a new version doesn't suddenly cause the CI to fail
+	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
+	# until types-setuptools is removed from typeshed.
+	# For help with static-typing issues, or mypy update, ping @Avasam 
+	"mypy==1.11",
 	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
 	"tomli",
 	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ test = [
 	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
 	# until types-setuptools is removed from typeshed.
 	# For help with static-typing issues, or mypy update, ping @Avasam 
-	"mypy==1.11",
+	"mypy==1.11.*",
 	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
 	"tomli",
 	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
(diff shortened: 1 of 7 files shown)


setuptools/command/install.py
```diff
@@ -29,7 +32,8 @@ class install(orig.install):
         'old-and-unmanageable',
         'single-version-externally-managed',
     ]
-    new_commands = [
+    # Type the same as distutils.command.install.install.sub_commands
+    new_commands: ClassVar[list[tuple[str, Callable[[Any], bool] | None]]] = [
```
Avasam, line 36: This was causing invariance issues when trying to append `install.new_commands` to `install.sub_commands` at the end of this module

setuptools/tests/test_wheel.py
```diff
@@ -547,7 +553,7 @@ def __repr__(self):
 @pytest.mark.parametrize(
     'params',
     WHEEL_INSTALL_TESTS,
-    ids=list(params['id'] for params in WHEEL_INSTALL_TESTS),
+    ids=[params['id'] for params in WHEEL_INSTALL_TESTS],
```
Avasam, line 556: Not a necessary change, but I noticed this improvement whilst I was looking at this line. Would've been caught by https://docs.astral.sh/ruff/rules/unnecessary-generator-list/

pyproject.toml
```diff
@@ -59,7 +59,10 @@ test = [
 	# for tools/finalize.py
 	'jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"',
 	"pytest-home >= 0.5",
-	"mypy==1.10.0", # pin mypy version so a new version doesn't suddenly cause the CI to fail
+	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
+	# until types-setuptools is removed from typeshed.
+	# For help with static-typing issues, or mypy update, ping @Avasam 
+	"mypy==1.11",
```
jaraco, line 65: This pin is the same as ==1.11.0. Is that what you intended, or did you want something like `~=1.11` or `==1.11.*`?

setuptools/tests/test_wheel.py
jaraco: I'm good with it. Thanks for highlighting!

jaraco (approved):

pyproject.toml
Avasam: Yes. Seems I mixed up with npm
```suggestion
	"mypy==1.11.*",
```

PR #4503, opened 2024-07-21
Mark abstract base classes and methods

Avasam:
## Summary of changes

Works towards #2345

Merged `@abstractmethod` from typeshed and searched source code for `abstract`

### Pull Request Checklist
- [x] Changes have tests (existing static and unit tests should pass)

[PR docs]:


setuptools/__init__.py
```diff
@@ -226,6 +227,33 @@ def reinitialize_command(self, command, reinit_subcommands=False, **kw):
...
+
+    @abstractmethod
+    def run(self) -> None:
+        """
+        Execute the actions intended by the command.
+        (Side effects **SHOULD** only take place when ``run`` is executed,
+        for example, creating new files or writing to the terminal output).
+        """
+        raise NotImplementedError
+
```
Avasam, line 256: Is it possible to for the class' doc to reference its own methods in rst ? As to not duplicate the descriptions ?
```
    .. method:: initialize_options(self)
        Set or (reset) all options/attributes/caches used by the command
        to their default values. Note that these values may be overwritten during
        the build.
    .. method:: finalize_options(self)
        Set final values for all options/attributes used by the command.
        Most of the time, each option/attribute/cache should only be set if it does not
        have any value yet (e.g. ``if self.attr is None: self.attr = val``).
    .. method:: run(self)
        Execute the actions intended by the command.
        (Side effects **SHOULD** only take place when ``run`` is executed,
        for example, creating new files or writing to the terminal output).
```

newsfragments/4503.feature.rst
```diff
@@ -0,0 +1 @@
+Mark abstract base classes and methods with `abc.ABC` and `abc.abstract_method` -- by :user:`Avasam`
```
Avasam, line 1:
```suggestion
Mark abstract base classes and methods with `abc.ABC` and `abc.abstractmethod` -- by :user:`Avasam`
```

setuptools/__init__.py
abravalheri: Probably yes. We are using `autocommand` magic in: https://github.com/pypa/setuptools/blob/main/docs/userguide/extension.rst?plain=1#L78-L89.

Might require some adjustments, but even if we have to mention them without a link just with double backticks, it should be fine.

abravalheri: See https://setuptools--4503.org.readthedocs.build/en/4503/userguide/extension.html#setuptools.Command

abravalheri: Can we remove this part of the docstring and replace with a simple mention in this PR?

> ```
>     .. method:: initialize_options(self)
>         Set or (reset) all options/attributes/caches used by the command
>         to their default values. Note that these values may be overwritten during
>         the build.
>     .. method:: finalize_options(self)
>         Set final values for all options/attributes used by the command.
>         Most of the time, each option/attribute/cache should only be set if it does not
>         have any value yet (e.g. ``if self.attr is None: self.attr = val``).
>     .. method:: run(self)
>         Execute the actions intended by the command.
>         (Side effects **SHOULD** only take place when ``run`` is executed,
>         for example, creating new files or writing to the terminal output).
> ```

(Just to avoid the duplication in the docs).

Avasam: Oh nice, I *can* easily self-reference without having to include module and class names.

abravalheri:
Thank you very much!

PR #4504, opened 2024-07-22
Merge simple type annotations from typeshed

Avasam:
## Summary of changes

~~Please review and merge first: https://github.com/pypa/setuptools/pull/4505~~  Done
~~Merging #4581 and #4584 first will reduce changes in this PR~~ Done
It's important to validate this PR with mypy, let's re-enable it first and fix existing failures: https://github.com/pypa/setuptools/pull/4604

Not done in this PR:
- Abstract methods and classes: https://github.com/pypa/setuptools/pull/4503
- `@overload`: https://github.com/pypa/setuptools/pull/4506
- return types
- `AbstractSandbox`'s dynamically created methods
- expanded *args & **kwargs, `@type_check_only`, and other TypedDict from typeshed

Step 6.1 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

### Pull Request Checklist
- [x] Changes have tests (no new tests, existing static tests should pass)
- [x] News fragment added in [`newsfragments/`]. (nothing user-facing as static typing users would still be using `types-setuptools`)

[PR docs]:


setuptools/config/pyprojecttoml.py
```diff
@@ -94,7 +95,7 @@ def read_configuration(
         If not given a dummy object will be created and discarded after the
         configuration is read. This is used for auto-discovery of packages and in the
         case a dynamic configuration (e.g. ``attr`` or ``cmdclass``) is expanded.
-        When ``expand=False`` this object is simply ignored.
+        When ``expand:bool=False`` this object is simply ignored.
```
Avasam, line 98:
```suggestion
        When ``expand=False`` this object is simply ignored.
```

setuptools/config/pyprojecttoml.py
```diff
@@ -117,9 +118,9 @@ def read_configuration(
     if "distutils" in tool_table:
         _ExperimentalConfiguration.emit(subject="[tool.distutils]")
 
-    # There is an overall sense in the community that making include_package_data=True
+    # There is an overall sense in the community that making include_package_data:bool=True
```
Avasam, line 121:
```suggestion
    # There is an overall sense in the community that making include_package_data=True
```

setuptools/config/pyprojecttoml.py
```diff
@@ -117,9 +118,9 @@ def read_configuration(
...
     # the default would be an improvement.
-    # `ini2toml` backfills include_package_data=False when nothing is explicitly given,
+    # `ini2toml` backfills include_package_data:bool=False when nothing is explicitly given,
```
Avasam, line 123:
```suggestion
    # `ini2toml` backfills include_package_data=False when nothing is explicitly given,
```

setuptools/extension.py
```diff
@@ -126,10 +127,12 @@ class Extension(_Extension):
       specified on Windows. (since v63)
     """
 
-    def __init__(self, name, sources, *args, **kw):
+    def __init__(
+        self, name: str, sources: list[str], *args, py_limited_api: bool = False, **kw
+    ):
```
Avasam, line 132: From @abravalheri  https://github.com/pypa/setuptools/pull/4505/files#r1709465928
> Jason might have added support for `PathLike`sources recently in [pypa/distutils#237](https://github.com/pypa/distutils/pull/237) ([`v72.2.0`/setuptools/_distutils/extension.py#L29](https://github.com/pypa/setuptools/blob/v72.2.0/setuptools/_distutils/extension.py?rgh-link-date=2024-08-08T13%3A10%3A07Z#L29)).
> 
> (PS.: granted they are incompatible with `SETUPTOOLS_USE_DISTUTILS=stdlib` but that is a deprecated code path).

_distutils_hack/__init__.py
```diff
@@ -1,6 +1,8 @@
 # don't import any costly modules
+from __future__ import annotations
 import sys
 import os
+from types import TracebackType
```
Avasam, line 5: > # don't import any costly modules

Here I don't have the choice to import either `from types import TracebackType` or `from typing import TYPE_CHECKING`. Dunno which one is "costlier"

setuptools/command/bdist_egg.py
```diff
@@ -10,13 +12,17 @@
...
+from typing import TYPE_CHECKING
 
 from setuptools.extension import Library
 from setuptools import Command
 from .._path import ensure_directory
 
 from sysconfig import get_path, get_python_version
 
+if TYPE_CHECKING:
+    from zipfile import _ZipFileMode
```
abravalheri, line 24: I think we should avoid using private members...

setuptools/command/build_ext.py
```diff
@@ -402,7 +402,7 @@ def link_shared_object(
         library_dirs=None,
         runtime_library_dirs=None,
         export_symbols=None,
-        debug=False,
+        debug: bool = False,
```
abravalheri, line 405: This is a funny one...

Does this mean that when we define a class attribute like `inplace = False` the inferred type is already `bool`, but if we have a function argument with `debug=False` the inferred type is `Literal[False]`?

That sounds like the type-checker is very inconsistent...

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -30,14 +30,14 @@
...
-    from distutils.dist import _OptionsList
+    from distutils.dist import _OptionsList  # Comes from typeshed
     from setuptools._importlib import metadata
     from setuptools.dist import Distribution
     from typing_extensions import TypeAlias
 
 EMPTY: Mapping = MappingProxyType({})  # Immutable dict-like
 _ProjectReadmeValue: TypeAlias = Union[str, Dict[str, str]]
-_CorrespFn: TypeAlias = Callable[["Distribution", Any, StrPath], None]
+_CorrespFn: TypeAlias = Callable[["Distribution", Any, Union[StrPath, None]], None]
```
abravalheri, line 40: Do we need the optional StrPath here? I think it is better to keep `StrPath` instead of `Union[StrPath, None]`, no?

In the `apply` function the very first thing we do is:

`root_dir = os.path.diname(filename) or "."`, so we always have a `StrPath` value that we use to call al the functions, not?

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -150,7 +150,9 @@ def _guess_content_type(file: str) -> str | None:
     raise ValueError(f"Undefined content type for {file}, {msg}")
 
 
-def _long_description(dist: Distribution, val: _ProjectReadmeValue, root_dir: StrPath):
+def _long_description(
+    dist: Distribution, val: _ProjectReadmeValue, root_dir: StrPath | None
```
abravalheri, line 154: Similar comment to this and all the other functions: I think `root_dir` is always StrPath.

setuptools/config/expand.py
```diff
@@ -51,7 +51,7 @@
     from setuptools.dist import Distribution
 
 _K = TypeVar("_K")
-_V = TypeVar("_V", covariant=True)
+_VCo = TypeVar("_VCo", covariant=True)
```
abravalheri, line 54: It seems that the convention in `typeshed` is `_V_co` (example: https://github.com/python/typeshed/blob/e9b3b03764d0557e103e1b2c4d95ce02965db22b/stdlib/builtins.pyi#L79, https://github.com/python/typeshed/blob/e9b3b03764d0557e103e1b2c4d95ce02965db22b/stubs/aiofiles/aiofiles/base.pyi#L8)... Not that I care much about it, but if we are not following a convention that I prefer the shorter `_V`...

setuptools/config/setupcfg.py
```diff
@@ -43,19 +43,25 @@
...
+    tuple. The first element of the tuple indicates the origin of the option value
+    (e.g. the name of the configuration file where it was read from),
+    while the second element of the tuple is the option value itself
+    """
+    AllCommandOptions: TypeAlias = Dict[str, SingleCommandOptions]
+    """cmd name => its options"""
+
+else:
+    SingleCommandOptions = Dict[str, Tuple[str, Any]]
+    AllCommandOptions = Dict[str, SingleCommandOptions]
```
abravalheri, line 59: What is the big advantage here of keeping two copies in the `if/else`? Isn't it better to just keep everything outside of the if as in the original?

setuptools/discovery.py
```diff
@@ -57,12 +57,15 @@
...
-
-chain_iter = itertools.chain.from_iterable
-
 if TYPE_CHECKING:
     from setuptools import Distribution
+    from typing_extensions import TypeAlias
+
+    StrIter: TypeAlias = Iterator[str]
+else:
+    StrIter = Iterator[str]
```
abravalheri, line 66: Same comment about 2 copies inside and outside the `if`.

setuptools/dist.py
```diff
@@ -194,8 +198,10 @@ def check_packages(dist, attr, value):
 
 
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Distribution = distutils.core.Distribution
+    _Distribution: TypeAlias = distutils.core.Distribution
```
abravalheri, line 204: Should we just do `from distutils.core import Distribution as _Distribution` then?

setuptools/dist.py
```diff
@@ -607,7 +618,7 @@ def parse_config_files(self, filenames=None, ignore_option_errors=False):
         self._finalize_requires()
         self._finalize_license_files()
 
-    def fetch_build_eggs(self, requires):
+    def fetch_build_eggs(self, requires: _StrOrIter):
```
abravalheri, line 621: By the time we reach this function the argument passed to `requires` is always `list[str]`.

The trouble happens at the `__init__` function, although the user can pass `_StrOrIter` for any of the requirement-related arguments, by the time the `__init__` function finishes all the requirements are processed into a `list[str]`.

setuptools/extension.py
```diff
@@ -25,8 +28,10 @@ def _have_cython():
 # for compatibility
 have_pyrex = _have_cython
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Extension = distutils.core.Extension
+    _Extension: TypeAlias = distutils.core.Extension
```
abravalheri, line 34: Should we then just do `from distutils.core import Extension as _Extension`?

setuptools/package_index.py
```diff
@@ -396,7 +396,7 @@ def process_filename(self, fn, nested=False):
             self.debug("Found: %s", fn)
             list(map(self.add, dists))
 
-    def url_ok(self, url, fatal=False):
+    def url_ok(self, url, fatal: bool = False):
```
abravalheri, line 399:
```suggestion
    def url_ok(self, url, fatal: bool = False) -> bool:
```

setuptools/dist.py
```diff
@@ -753,7 +764,7 @@ def exclude_package(self, package):
                 if p.name != package and not p.name.startswith(pfx)
             ]
 
-    def has_contents_for(self, package):
+    def has_contents_for(self, package: str):
```
abravalheri, line 767:
```suggestion
    def has_contents_for(self, package: str) -> bool:
```

setuptools/__init__.py
```diff
@@ -114,8 +114,10 @@ def setup(**attrs):
 setup.__doc__ = distutils.core.setup.__doc__
 
 if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
     # Work around a mypy issue where type[T] can't be used as a base: https://github.com/python/mypy/issues/10962
-    _Command = distutils.core.Command
+    _Command: TypeAlias = distutils.core.Command
```
abravalheri, line 120: Should we then just do `from distutils.core import Command as _Command`?

Avasam: That works to not use `TypeAlias`, which is more concise, so I think that's an improvement. But you're still stuck with `if TYPE_CHECKING` check.

abravalheri: That is OK, just trying to streamline whenever possible...

Avasam:
@abravalheri I'll answer a couple comments at once:
- I have a handful of TypeAliases unnecessarily duplicated behind `TYPE_CHECKING` check whilst a `from __future__ import annotations` is sufficient to use an annotation that doesn't exist at runtime. So those should be simplified now.
- As quickly mentioned in the PR description, and described in my step-by-step plan in https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833, I purposefully avoided adding return type annotations here. I plan on merging return annotations from typeshed as a direct follow-up. Then re-enabling `ANN2` checks for setuptools.

setuptools/config/expand.py
Avasam: You are correct. Ironically I copied this from the typeshed stub... where I made the same mistake and it wasn't, caught there!
image

setuptools/dist.py
Avasam: I typed the function by what it supports. Which is the same as `setuptools._reqs.parse`. A `list[str]`is a valid type to pass to a `_StrOrIter`, so it doesn't create any typing issue here.

setuptools/command/build_ext.py
Avasam: Both are inferred as `bool`.

Mypy only checks the body of typed functions, and a function isn't considered typed until all its params are annotated (*or* a return type is explicitly added for functions with no parameters).

Pyright has the `reportMissingParameterType` check: https://github.com/microsoft/pyright/blob/main/docs/configuration.md

https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/ won't trigger for a public function with no annotation even if all params have default values.

Until all `link_shared_object` params are annotated, it won't change anything for mypy. So if it feels inconsistent to you, I can remove that annotation for now.

abravalheri: Ah, ok I see, it is not about the inferred type, it is about when a function is considered typed.

I don't have any problems in leaving this as you suggested in the PR changes.

setuptools/sandbox.py
```diff
@@ -411,7 +422,7 @@ def _remap_pair(self, operation, src, dst, *args, **kw):
 class DirectorySandbox(AbstractSandbox):
     """Restrict operations to a single subdirectory - pseudo-chroot"""
 
-    write_ops = dict.fromkeys([
+    write_ops: dict[str, None] = dict.fromkeys([
```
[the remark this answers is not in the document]
Avasam: The inferred type is `dict[str, Any | None]` to be exact (by both pyright and mypy). Because *we* know the value should never be used (this really ought to be a `set`, not a `dict`) we can be more precise on the type. Preventing some misuse like `self.write_ops["open"] = 1` and communicating more clearly that there's no use for the values of this dict.

The `| None` in `Any | None` did already prevent `test: str = self.write_ops["open"]`

setuptools/build_meta.py
```diff
@@ -289,7 +297,7 @@ def _arbitrary_args(self, config_settings: _ConfigSettings) -> Iterator[str]:
 
 
 class _BuildMetaBackend(_ConfigSettingsTranslator):
-    def _get_build_requires(self, config_settings, requirements):
+    def _get_build_requires(self, config_settings: _ConfigSettings, requirements):
```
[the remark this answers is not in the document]
Avasam: Doesn't cause any issue. I didn't add annotations to `_get_build_requires` because it's not typed in typeshed. `config_settings` param being typed is incidental to adding all `_ConfigSettings`.

But since you're mentioning it here and it's not an issue, I've added the type annotation to `requirements` param

setuptools/__init__.py
Avasam: Certainly! And when I have my head deep in it sometimes I'll miss some stuff (like not needing `TYPE_CHECKING` to *use* `TypeAlias`)

setuptools/config/_apply_pyprojecttoml.py
Avasam: We don't *need* to, it would simply be arbitrary restrictions since all correspondances in `PYPROJECT_CORRESPONDENCE` *do* support a `None` `root_dir` (they all end up defaulting to `os.getcwd()` or `os.curdir`)

setuptools/__init__.py
```diff
@@ -180,7 +180,7 @@ def _ensure_stringlike(self, option, what, default=None):
             )
         return val
 
-    def ensure_string_list(self, option):
+    def ensure_string_list(self, option: str):  # type: ignore[override] # Fixed in typeshed for next mypy release
```
Avasam, line 183: Let's be explicit about the expected fixed version, so we don't just "wait forever" if this isn't actually fixed.

```suggestion
    def ensure_string_list(self, option: str):  # type: ignore[override] # Fixed in typeshed for mypy 1.12
```

setuptools/build_meta.py
```diff
@@ -145,18 +151,18 @@ def suppress_known_deprecation():
...
+_ConfigSettings: TypeAlias = Union[Mapping[str, Union[str, List[str], None]], None]
 """
 Currently the user can run::
 
     pip install -e . --config-settings key=value
     python -m build -C--key=value -C key=value
 
 - pip will pass both key and value as strings and overwriting repeated keys
-  (pypa/pip#11059).
+(pypa/pip#11059).
```
Avasam, line 162:
```suggestion
  (pypa/pip#11059).
```

setuptools/build_meta.py
```diff
@@ -145,18 +151,18 @@ def suppress_known_deprecation():
...
 - build will accumulate values associated with repeated keys in a list.
-  It will also accept keys with no associated value.
-  This means that an option passed by build can be ``str | list[str] | None``.
+It will also accept keys with no associated value.
+This means that an option passed by build can be ``str | list[str] | None``.
```
Avasam, line 165:
```suggestion
  It will also accept keys with no associated value.
  This means that an option passed by build can be ``str | list[str] | None``.
```

setuptools/discovery.py
```diff
@@ -57,12 +57,12 @@
...
 
-StrIter = Iterator[str]
-
-chain_iter = itertools.chain.from_iterable
-
 if TYPE_CHECKING:
     from setuptools import Distribution
+    from typing_extensions import TypeAlias
+
+StrIter: TypeAlias = Iterator[str]
```
Avasam, line 64: I don't feel like this alias brings anything of value:
- It doesn't work around any runtime issue
- It doesn't abstract away any complexity
- It's not clearer than the type it's aliasing, or doesn't help explain it
- It's barely any more concise (because it's using a shorthand instead of being clear and explicit)

**But** it's not private. Could we maybe deprecate it ?

setuptools/command/sdist.py
```diff
@@ -45,7 +47,7 @@ class sdist(orig.sdist):
     ]
 
     distribution: Distribution  # override distutils.dist.Distribution with setuptools.dist.Distribution
-    negative_opt = {}
+    negative_opt: ClassVar[dict[str, str]] = {}  # type: ignore[misc] # Fixed upstream in typeshed to be a ClassVar
```
Avasam, line 50:
```suggestion
    negative_opt: ClassVar[dict[str, str]] = {}  # type: ignore[misc] # Fixed upstream in typeshed to be a ClassVar. Should be included in mypy 1.12
```

setuptools/command/sdist.py
```diff
@@ -45,7 +47,7 @@ class sdist(orig.sdist):
     ]
 
     distribution: Distribution  # override distutils.dist.Distribution with setuptools.dist.Distribution
-    negative_opt = {}
+    negative_opt: ClassVar[dict[str, str]] = {}  # type: ignore[misc] # Fixed upstream in typeshed to be a ClassVar. Should be included in mypy 1.12
```
Avasam, line 50: https://github.com/python/typeshed/pull/12403/files#diff-7708dfacd21509625134fb42599740dcb16ef7721b6065e9bb97b716dbafc5d7R15

setuptools/command/easy_install.py
```diff
@@ -711,9 +711,9 @@ def easy_install(self, spec, deps=False):
             else:
                 return self.install_item(spec, dist.location, tmpdir, deps)
 
-    def install_item(self, spec, download, tmpdir, deps, install_needed=False):
+    def install_item(self, spec, download, tmpdir, deps, install_needed: bool = False):
         # Installation is also needed if file in tmpdir or is not an egg
-        install_needed = install_needed or self.always_copy
+        install_needed = install_needed or bool(self.always_copy)
```
Avasam, line 716: Either this, or:
- default `always_copy` to `False` (I wasn't sure if `None` was used as a special meaning "not set by user");
- allow `install_needed` param to be `None`;
- use a differently named variable in `install_item` (like `_install_needed`);

setuptools/command/editable_wheel.py
```diff
@@ -138,7 +137,7 @@ def run(self):
             self._create_wheel_file(bdist_wheel)
         except Exception:
             traceback.print_exc()
-            # TODO: Fix false-positive [attr-defined] in typeshed
+            # TODO: Pending fix for false-positive [attr-defined] in typeshed https://github.com/python/typeshed/pull/12467
```
Avasam, line 140: This has been merged and mypy 1.12 is out
```suggestion
```

Avasam:
I've added more distutils workaround so that this can be merged w/o #4691, but I'd ask to consider making distutils stubs available for all Python versions so that it stops being an issue.

setuptools/command/build_py.py
```diff
@@ -48,14 +50,14 @@ def finalize_options(self):
...
-        preserve_mode=True,
-        preserve_times=True,
-        link=None,
-        level=1,
+        infile: StrPath,
+        outfile: StrPath,
+        preserve_mode: bool = True,
+        preserve_times: bool = True,
+        link: str | None = None,
+        level: object = 1,
```
abravalheri, line 60:
```python
        level: object = 1,
```

This one looks odd, but I suppose that is because it assumes either `int` or `boolean` values elsewhere.

abravalheri (approved):
Thank you very much @Avasam.

I added 2 more questions, but I think we can merge the PR as it is, after we apply the rebase and test. Worst case scenario we can have follow ups.

setuptools/command/build_py.py
Avasam: `object` can be used to mean "any parameter type" for unused params, (Typeshed has the alias `_typeshed.Unused`), but that might lead to LSP violations since subclasses should now support any `object`.

I should change it to `int` (Unless you tell me that subclasses of these classes should *not* support an "optimization level")
image

abravalheri:
So there was a problem in the tests with the latest version of `ruff`, and clash with the version of ruff in the `.pre-commit-hooks.yaml`.

I bumped the hook version just to keep the ball rolling. Hopefully it will not clash in the next merge with skeleton.

Avasam:
Ah yes, the formatter just added merging of short multiline implicit string concatenation in preview mode (along with a few other small changes).

Might be worth a standalone run on main.

abravalheri:
> Might be worth a standalone run on main, as to not "pollute" this PR with unrelated changes.

WIP in https://github.com/pypa/setuptools/pull/4705.

Avasam:
Btw you can squash merge this, otherwise I'll just squash and force-push.

setuptools/command/build_py.py
Avasam: I'll need to go fix it in typeshed first (or #4689) I'll leave it "as-is" for this PR.

abravalheri: Thank you @Avasam, once the CI finishes running I will squash it.

Avasam:
Wooh! A big one done!

PR #4505, opened 2024-07-22
Runtime changes for typeshed parameter annotations merge

Avasam:
## Summary of changes
Extracted from https://github.com/pypa/setuptools/pull/4504
These are the runtime changes that should be reviewed more carefully.

Works towards #2345 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -12,6 +12,7 @@ __ https://setuptools.pypa.io/en/latest/deprecated/easy_install.html
 
 from __future__ import annotations
 
+from collections.abc import Iterable
 from glob import glob
 from distutils.util import get_platform
 from distutils.util import convert_path, subst_vars
@@ -26,6 +27,7 @@ from distutils.command.build_scripts import first_line_re
 from distutils.command import install
 import sys
 import os
+from typing import TYPE_CHECKING
 import zipimport
 import shutil
 import tempfile
@@ -78,6 +80,8 @@ from ..compat import py39, py311
 from .._path import ensure_directory
 from jaraco.text import yield_lines
 
+if TYPE_CHECKING:
+    from typing_extensions import Self
 
 # Turn on PEP440Warnings
 warnings.filterwarnings("default", category=pkg_resources.PEP440Warning)
@@ -2055,19 +2059,21 @@ class CommandSpec(list):
         return os.environ.get('__PYVENV_LAUNCHER__', _default)
 
     @classmethod
-    def from_param(cls, param):
+    def from_param(cls, param: Self | str | Iterable[str] | None) -> Self:
         """
         Construct a CommandSpec from a parameter to build_scripts, which may
         be None.
         """
         if isinstance(param, cls):
             return param
-        if isinstance(param, list):
+        if isinstance(param, str):
+            return cls.from_string(param)
+        if isinstance(param, Iterable):
             return cls(param)
         if param is None:
             return cls.from_environment()
-        # otherwise, assume it's a string.
-        return cls.from_string(param)
+        # AttributeError to keep backwards compatibility, this should really be a TypeError though
+        raise AttributeError(f"Argument has an unsupported type {type(param)}")
 
     @classmethod
     def from_environment(cls):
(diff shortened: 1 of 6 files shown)


setuptools/command/easy_install.py
```diff
@@ -2055,19 +2059,18 @@ def _sys_executable(cls):
...
-        if isinstance(param, list):
+        if isinstance(param, str):
+            return cls.from_string(param)
+        if isinstance(param, Iterable):
             return cls(param)
-        if param is None:
-            return cls.from_environment()
-        # otherwise, assume it's a string.
-        return cls.from_string(param)
+        return cls.from_environment()
```
Avasam, line 2073: If this gracious fallback is undesired, I can explicitly raise instead.

abravalheri: Let's raise it explicitly because before `cls.from_environment()` was only used in the case of `None`. I suspect that the previous fallback to `cls.from_string` would cause a `TypeError` if the user passed non-sensical args.

We don't want to change that behaviour

setuptools/command/editable_wheel.py
```diff
@@ -466,8 +466,14 @@ def _create_file(self, relative_output: str, src_file: str, link=None):
...
-        mappings = {self._normalize_output(k): v for k, v in output_mapping.items()}
-        mappings.pop(None, None)  # remove files that are not relative to build_lib
+        mappings = {
+            k: v
+            for k, v in (
+                (self._normalize_output(k), v) for k, v in output_mapping.items()
+            )
+            # remove files that are not relative to build_lib
+            if k is not None
+        }
```
abravalheri, line 476: Probably a bit more readable... The name of the intermediary variable may be different, just used the first thing that came to my mind.

```suggestion
        normalised = ((self._normalize_output(k), v) for k, v in output_mapping.items())
        # remove files that are not relative to build_lib
        mappings = {k: v for k, v in normalised if k is not None}
```

setuptools/extension.py
```diff
@@ -126,10 +127,12 @@ class Extension(_Extension):
       specified on Windows. (since v63)
     """
 
-    def __init__(self, name, sources, *args, **kw):
+    def __init__(
+        self, name: str, sources: list[str], *args, py_limited_api: bool = False, **kw
```
abravalheri, line 131: Jason might have added support for `PathLike`sources recently in https://github.com/pypa/distutils/pull/237 (https://github.com/pypa/setuptools/blob/v72.2.0/setuptools/_distutils/extension.py#L29).

(PS.: granted they are incompatible with `SETUPTOOLS_USE_DISTUTILS=stdlib` but that is a deprecated code path).

setuptools/command/editable_wheel.py
Avasam: Looks great thanks

setuptools/extension.py
Avasam: Thanks for the heads up, that added annotation isn't necessary for this PR. I'd rather avoid possible type conflicts with sdtlib distutils for now. I'll tackle that later.

setuptools/command/easy_install.py
Avasam: Raising an error instead of a fallback makes sense since the param unexpectedly goes unused.

It was an `AttributeError` because non-string first argument to `shlex` is assumed to be a stream.
```py
  File "C:\Program Files\Python39\lib\shlex.py", line 140, in read_token
    nextchar = self.instream.read(1)
AttributeError: 'int' object has no attribute 'read'
```
At the `from_param` level it should really be a TypeError, but that's technically a non-backwards compatible change.

I raise an `AttributeError`, maybe can be changed for a `TypeError` next major version (or now if you don't mind the error type change)

abravalheri: Thanks @Avasam, I am happy with either.

I don't think that changing to `TypeError` should be a big impact because it looks like a very edge-y case.
Probably the important part is to raise the exception, not the exact exception class. So I am OK with risking it.

But equally happy to take the conservative approach, given how susceptible `setuptools` is to Hyrum's Law...

Avasam: How about this: Keep it to `AttributeError` for this PR, I'll open a follow-up that changes to `TypeError` that you can merge before the next major update. That should cover our bases and improve the communicated error.

abravalheri: Thanks @Avasam.

abravalheri:
Thank you!

PR #4524, opened 2024-07-29
Add guidance on pinning the upper bound in build-system.requires.

### LINKED ISSUE #4521: [Docs] Missing recommendation on an upper bound version pin for build backends
### Summary

Related to #4519 and #4520, but independent of them, there seems to be, at the very least, an important documentation bug:

Both on [Configuring setuptools using `pyproject.toml` files](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) and on [Building and Distributing Packages with Setuptools: `setup.cfg`-only projects](https://setuptools.pypa.io/en/latest/setuptools.html#setup-cfg-only-projects), the Setuptools documentation instructs developers to request the Setuptools build backend without an upper bound version pin, e.g:

```toml
[build-system]
requires = [
  "setuptools >= 40.9.0",
]
```

```toml
[build-system]
requires = ["setuptools", "setuptools-scm"]
```

However this is not the first or the second time that a package fails to build with newer versions of setuptools that used to work with previous versions.

Pip's documentation has a slightly safer recommendation on [Build Process: Build time dependencies](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-time-dependencies):

```toml
[build-system]
requires = ["setuptools ~= 58.0", "cython ~= 0.29.0"]
```

(please ignore the `cython` pin, it's just for illustration purposes).

Considering the automatic behavior on build isolation to install the latest setuptools allowed by a package's build backend, and of Setuptools upgrade breaking builds based on earlier Setuptools releases, then Setuptools documentation should explicitly document adding an upper bound version pin for Setuptools.

Note that this is not a complaint on the fact that this kind of breakage can happen, as long as it happens between major versions, with an appropriately visible deprecation, but we should teach users of Setuptools how not to be unnecessarily exposed to it.

Notice that this will make it more likely that old packages might stop being installable if old versions of Setuptools itself become uninstallable by, say, newer versions of pip (as has happened), but I believe that older versions of setuptools becoming uninstallable happens more rarely than newer versions of Setuptools breaking older packages.

### OS / Environment

all

### Additional Information

When the documentation is improved, packages will hopefully start declaring an upper bound for versions of setuptools as a build-backend, which means build isolation by frontends (like pip or uv) will not install an incompatible version of setuptools for building a setuptools-build-backed package.

### Code of Conduct

jaraco:
## Summary of changes

Closes #4521

### Pull Request Checklist
- [ ] ~~Changes have tests~~
- [ ] ~~News fragment added in [`newsfragments/`].~~

[PR docs]:

### DIFF
--- a/docs/userguide/dependency_management.rst
+++ b/docs/userguide/dependency_management.rst
@@ -30,9 +30,40 @@ requisite library is ``setuptools``. Specify the build backend in a
     requires = ["setuptools"]
     #...
 
-Please note also to include any other ``setuptools`` plugins
+Also include any other ``setuptools`` plugins
 (e.g., :pypi:`setuptools_scm`, :pypi:`setuptools-golang`, :pypi:`setuptools-rust`)
 or build-time dependency (e.g., :pypi:`Cython`, :pypi:`cppy`, :pypi:`pybind11`).
+
+.. code-block:: toml
+
+    [build-system]
+    requires = ["setuptools", "cython", "setuptools_scm"]
+
+
+If the project depends on a feature introduced in a specific version of Setuptools,
+it is good practice to specify it as a lower bound:
+
+.. code-block:: toml
+
+    [build-system]
+    requires = ["setuptools >= 61.2"]
+
+Some may be tempted to also include an upper-bound for yet unreleased major
+versions (e.g. ``setuptools <= 70``) or pin to a specific version (e.g.
+``setuptools == 70.0.4``) in order to avoid the project being uninstallable
+should those backward-incompatible changes affect this release of the project.
+Setuptools maintainers recommend strongly against this precautionary approach.
+The team primarily maintains one release, the latest monotonically-increasing
+release, and encourages users to use that latest release (work at HEAD). As a
+result, the team is cognizant of and takes responsibility for making
+backward-incompatible changes and aims to mitigate the impact of any breaking
+changes prior to releasing that change. By pinning against an unreleased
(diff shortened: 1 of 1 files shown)


docs/userguide/dependency_management.rst
```diff
@@ -30,10 +30,41 @@ requisite library is ``setuptools``. Specify the build backend in a
...
+.. code-block:: toml
+
+    [build-system]
+    requires = ["setuptools >= 61.2"]
+
+Some may be tempted to also include an upper-bound for yet unreleased major
+versions (e.g. ``setuptools <= 70``) or pin to a specific version (e.g.
+``setuptools == 70.0.4``) in order to avoid the project being uninstallable
+should those backward-incompatible changes affect this release of the project.
+Setuptools maintainers recommend strongly against this precautionary approach.
```
jaraco, line 55: @abravalheri I'd like your consent, especially on this statement, but also would appreciate any other feedback.

abravalheri: No problems, from my side Jason, please go ahead. Thank you very much for having a look at this.

eli-schwartz (approved):
I broadly agree with this advice.

The numpy.distutils situation is actually a good example of a case where upper bounds are the correct thing to do -- but that was both a known, released version, and a long-running (multi-year!) event. It was a special enough case that I don't know it's worth devoting any digital ink to. :)

abravalheri (approved):

PR #4531, opened 2024-07-31
Remove post-release tags on setuptools' own build.

jaraco:
Ref #4530

## Summary of changes

Remove the post release tags for a cleaner build with the intention of later adopting setuptools_scm for version inference.

No news fragment is needed as it doesn't affect official releases.

### Pull Request Checklist
- [ ] ~~Changes have tests~~
- [ ] ~~News fragment added in [`newsfragments/`].
  _(See [documentation][PR docs] for details)_~~

[PR docs]:


jaraco:
The docs build is failing on main.

jaraco:
On further consideration, maybe this change does deserve a news fragment. Integrators probably have to enact the same workaround to convert the in-dev version to a released version, and that step will no longer be necessary after this change.

Issue #4529, opened 2024-07-31
[Question] AttributeError: 'Extension' object has no attribute '_needs_stub'


alexhutman:
### setuptools version

72.1.0

### Python version

Python 3.11.1

### OS

Ubuntu

### Additional environment information

I am running Ubuntu via a [Docker image](https://github.com/alexhutman/ZeroForcingNumber/blob/a503769a4020b2a189703ce7e070062153ae1a69/Dockerfile) that is based upon [Sage 10.3](https://hub.docker.com/layers/sagemath/sagemath/10.3/images/sha256-7643cc6f7caec104c6d9aad28dbbaff45c0d8d13da95318c161da50269867a29?context=explore).

### Description

I saw issue #3339 -- I believe my scenario is different. I subclass `build_ext`, and therefore `finalize_options()` should be called normally. The build works fine during a normal build (`pip install .`), however using an editable build (`pip install --editable .`) produces the aforementioned error.

I have a fix for this lined up [here](https://github.com/pypa/setuptools/compare/main...alexhutman:setuptools:main). Please let me know if I am good to send in that PR. The install worked for me after changing those lines.

### Expected behavior

I expected it to succeed just as it does while not building in `--editable` mode.

### How to Reproduce

1. Try this Minimum Example subclass (remembering to install `--editable`):
```python
class zf_cythonize(build_ext):
    base_directives = dict(
         binding=False,
         language_level=3,
    )
    def run(self):
        dist = self.distribution
        ext_modules = dist.ext_modules
        if ext_modules:
            dist.ext_modules[:] = self.cythonize(ext_modules)
        super().run()

    def cythonize(self, extensions):
        # Run Cython with -Werror on continuous integration services
        # with Python 3.6 or later
        from Cython.Compiler import Options
        Options.warning_errors = False

        compiler_directives = dict(**self.base_directives)
        from Cython.Build.Dependencies import cythonize
        return cythonize(extensions,
                         compiler_directives=compiler_directives)
```

---

OR:

1. Clone [this repo](https://github.com/alexhutman/ZeroForcingNumber) and checkout `a503769a4020b2a189703ce7e070062153ae1a69`
2. Build normally (`docker build -t zeroforcing .`). This should succeed with no problems.
3. Change the Dockerfile to add `--editable` after `pip install`: `sage -pip install --editable zeroforcing/`
4. Rerun and observe the error

### Output

```console
docker build -t zeroforcing .

[...]

PR #4535, opened 2024-08-01
Fix most mypy `attr-defined` & pyright `reportAttributeAccessIssue` by explicitly defining attributes on classes

Avasam:
## Summary of changes

Fix most mypy `attr-defined` & pyright `reportAttributeAccessIssue` for #2345 by explicitly defining attributes on classes

I have some ideas on how to address the last `attr-defined` cause by `DistributionMetadatalicense_files` `DistributionMetadata.license_file`, but that's out of scope for this PR.

`get_finalized_command`, `get_command_obj` and `get_command_class` still require casting, but that should be fixed in typeshed.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/editable_wheel.py
+++ b/setuptools/command/editable_wheel.py
@@ -153,6 +153,7 @@ class editable_wheel(Command):
             self._create_wheel_file(bdist_wheel)
         except Exception:
             traceback.print_exc()
+            # TODO: Fix false-positive [attr-defined] in typeshed
             project = self.distribution.name or self.distribution.get_name()
             _DebuggingTips.emit(project=project)
             raise
--- a/setuptools/extension.py
+++ b/setuptools/extension.py
@@ -128,7 +128,8 @@ class Extension(_Extension):
 
     # These 4 are set and used in setuptools/command/build_ext.py
     # The lack of a default value and risk of `AttributeError` is purposeful
-    # to avoid people forgetting to call finalize_options if they modify the extension list
+    # to avoid people forgetting to call finalize_options if they modify the extension list.
+    # See example/rationale in https://github.com/pypa/setuptools/issues/4529.
     _full_name: str  #: Private API, internal use only.
     _links_to_dynamic: bool  #: Private API, internal use only.
     _needs_stub: bool  #: Private API, internal use only.
--- a/setuptools/extension.py
+++ b/setuptools/extension.py
@@ -126,6 +126,14 @@ class Extension(_Extension):
       specified on Windows. (since v63)
     """
 
+    # These 4 are set and used in setuptools/command/build_ext.py
+    # The lack of a default value and risk of `AttributeError` is purposeful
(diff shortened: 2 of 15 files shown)


setuptools/command/editable_wheel.py
```diff
@@ -153,6 +153,7 @@ def run(self):
             self._create_wheel_file(bdist_wheel)
         except Exception:
             traceback.print_exc()
+            # TODO: Fix false-positive [attr-defined] in typeshed
```
Avasam, line 156: https://github.com/python/typeshed/pull/12467

setuptools/extension.py
```diff
@@ -126,6 +126,12 @@ class Extension(_Extension):
       specified on Windows. (since v63)
     """
 
+    # These 4 are set and used in setuptools/command/build_ext.py
+    _full_name = ""  #: Private API, internal use only.
+    _links_to_dynamic = False  #: Private API, internal use only.
+    _needs_stub = False  #: Private API, internal use only.
+    _file_name = ""  #: Private API, internal use only.
```
abravalheri, line 133: There is one very edge-y case where it is useful to not define default values before hand: to avoid people forgetting to call `finalize_options` if they modify the extension list like in https://github.com/pypa/setuptools/issues/4529.

Tricky... In this case I feel like would like to have the cake (static type annotations) but eat it too (have the exception happening if someone forgets to call `finalize_options`).

abravalheri: Maybe wrap in a `if TYPE_CHECKING`?

Avasam: One could also declare the attribute type without a default value:
```py
>>> class A:
...   foo: str
...   bar: str
...
>>> A().foo
AttributeError: 'A' object has no attribute 'foo'
```

If it's an issue with static checkers, then it can be guarded by `if TYPE_CHECKING`. Either way I'll add a comment that a potential `AttributeError` is expected and wanted. You can have your cake and eat it too, here 😄

setuptools/extension.py
```diff
@@ -126,6 +126,14 @@ class Extension(_Extension):
       specified on Windows. (since v63)
     """
 
+    # These 4 are set and used in setuptools/command/build_ext.py
+    # The lack of a default value and risk of `AttributeError` is purposeful
+    # to avoid people forgetting to call finalize_options if they modify the extension list
```
abravalheri, line 131:
```suggestion
    # to avoid people forgetting to call finalize_options if they modify the extension list.
    # See example/rationale in https://github.com/pypa/setuptools/issues/4529.
```

abravalheri (approved):

Avasam:
New uncovered line https://github.com/pypa/setuptools/actions/runs/10306778328/job/28530606282?pr=4535#step:5:108
> setuptools/tests/__init__.py (66.7%): Missing lines 12

Is better resolved by https://github.com/jaraco/skeleton/issues/130

abravalheri:
Thank you!

PR #4538, opened 2024-08-02
Merge with distutils@d7ffdb9c7

jaraco:
- **Distutils C++ support**
- **👹 Feed the hobgoblins (delint).**
- **Adjust tests further to accommodate C++ support.**
- **👹 Feed the hobgoblins (delint).**
- **Ignore sysconfig variables for LDCXXSHARED as it appears not to exist on PyPy.**
- **ENH: Extension should be able to accept PathLike sources objects**
- **Adjust expectation to match behavior.**
- **For consistency, ensure Extension.sources is always a pathlib.Path object and adjust expectations in tests.**
- **When computing input_opt, ensure src is a string (when it could be a pathlib Path).**
- **In filelist, allow for self.files to be pathlib objects.**
- **Revert "Disable cygwin tests for now. Ref pypa/setuptools#3921"**
- **find_library_file: port from doctest to normal test and fix for cygwin**
- **test_customize_compiler_with_msvc_python: skip for cygwin**
- **Prefer f-strings**
- **Remove r string now that docstring is gone.**
- **👹 Feed the hobgoblins (delint).**
- **👹 Feed the hobgoblins (delint).**
- **Allow path objects**
- **Need to include 'three' in the input.**
- **Consolidate str and Path handling.**
- **Expand convert_path to also accept pathlib.Path objects.**
- **Prefer simply 'pathlib' for import.**
- **Extract a singledispatchmethod _copy for handling the copy of each data file.**
- **Use explicit registration for compatibility with older Pythons.**
- **Prefer os.PathLike in convert_path**
- **Simply accept pathlib objects and convert them early for compatibilty with upstream.**
- **Convert needs to accept None for Setuptools' sake.**
- **In test_convert_path, utilize posixpath.join and ntpath.join for maximum compatibility with other libraries.**
- **Wrap paths in PurePosixPath to ensure that any WindowsPaths don't get backslashes.**
- **convert_path no longer fails if passed a path with a trailing slash. Instead, trailing slashes are stripped just as they are with pathlib.Path.**
- **convert_path now converts to a platform-native path.Path, but then calls `.as_posix()` on it.**
- **Separate test_convert_path into two tests to avoid interactions in monkeypatching os.path.**
- **Remove expectation that a ValueError is raised for data_files being drive-relative absolute on Windows.**
- **Simplify convert_path by simply relying on the logic in PurePath.**
- **Harmonize convert_path tests across Unix and Windows.**
- **Consolidate convert_path tests and just generate the expected value in a platform-sensitive way. Should fix failures on mingw.**
- **Use simple import**
- **venv_install_options: add missing clear_argv fixture**
- **CI: add a job for running tests under MSVC CPython with GCC as the default compiler**
- **mingw: make get_msvcr() a noop**
- **Remove unused RangeMap**
- **Harmonize '-shared' parameter for C and C++ in Mingw32CCompiler**
- **In sysconfig.customize_compiler, initialize ldcxxshared from config vars.**
- **Refactored customize_compiler to reduce logical branches and extract _add_flags logic.**
- **Don't add flags when the value is None, such as when on PyPy, LDCXXSHARED is unset.**
- **Rely on monkeysession to monkeypatch.**
- **Apply isort. Ref pypa/distutils#240**

## Summary of changes

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
After pypa/distutils#275, I'm going to open a new PR.

PR #4549, opened 2024-08-08
Redundant abstractmethod run in `setuptools.command.setopt.option_base`

Avasam:
## Summary of changes

@abravalheri I noticed this abstractmethod was redundant (doesn't break anything) by marking the class as `ABC` after https://github.com/pypa/setuptools/pull/4503 was merged

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/command/setopt.py
+++ b/setuptools/command/setopt.py
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from abc import ABC
 from distutils.util import convert_path
 from distutils import log
 from distutils.errors import DistutilsOptionError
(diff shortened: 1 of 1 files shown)


abravalheri:
Should we also remove the `ABC` from the definition `class option_base(Command, ABC)`, as command is now already an ABC? Or is it left there intentionally (to make it more explicit)?

Avasam:
> Should we also remove the `ABC` from the definition `class option_base(Command, ABC)`, as command is now already an ABC? Or is it left there intentionally (to make it more explicit)?

It's there explicitly.
When subclassing a class with abstract methods, type-checkers will remind you to implement all its abstract methods. Unless you mark it with `ABC` (or `meta=ABCMeta`, same thing) where it'll inherit the "Abstract" state of the parent class.

In other words, either `option_base` adds all 3 abstract methods, or it submetaclasses ABCMeta.

abravalheri:
Thank you!

PR #4556, opened 2024-08-11
Ruff: enable all pyflakes and perf rules

Avasam:
## Summary of changes

Setuptools conforms to all the Ruff-implemented Pyflakes and Perflint rules except `PERF203` and a single instance of `PERF102`. So we can be less picky about their selection.

`PERF203` is worth looking at, but is often a micro-optimisation with false-positives, let's not block the CI on that. I applied it on some tests.

### Pull Request Checklist
- [x] Changes have tests (these are new static tests)
- [x] News fragment added in [`newsfragments/`]. (no public-facing change)

[PR docs]:

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -2187,7 +2187,7 @@ class ScriptWriter:
         spec = str(dist.as_requirement())
         for type_ in 'console', 'gui':
             group = type_ + '_scripts'
-            for name, ep in dist.get_entry_map(group).items():
+            for name in dist.get_entry_map(group).keys():
                 cls._ensure_safe_name(name)
                 script_text = cls.template % locals()
                 args = cls._get_script_args(type_, name, header, script_text)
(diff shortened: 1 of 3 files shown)


Avasam:
It's been a couple PRs where the PyPy test gets randomly cancelled, that's weid

setuptools/command/easy_install.py
```diff
@@ -2187,7 +2187,7 @@ def get_args(cls, dist, header=None):
         spec = str(dist.as_requirement())
         for type_ in 'console', 'gui':
             group = type_ + '_scripts'
-            for name, ep in dist.get_entry_map(group).items():
+            for name in dist.get_entry_map(group).keys():
```
abravalheri, line 2190: If we are not using the values, then we can just iterate over the original object no? There would be no reason to call `.keys()`, right?

abravalheri (approved):
Thank you very much.

setuptools/command/easy_install.py
Avasam: yeah iterating over a dict should be the same as iterating over its keys since we don't do anything special with the view-object

abravalheri:
Thank you @Avasam.

PR #4575, opened 2024-08-17
Using `dict` as an `OrderedDict` and allowed using `dict` as an ordered type in `setuptools.dist.check_requirements`

Avasam:
## Summary of changes

Allows using `dict` as an ordered type in ``setuptools.dist.check_requirements``
https://github.com/pypa/setuptools/pull/4574 Gave me the idea for this PR

### Pull Request Checklist
- [x] Changes have tests (tests updated with new messages and dict)

[PR docs]:

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -11,12 +11,12 @@ from pathlib import Path
 from typing import (
     TYPE_CHECKING,
     Any,
+    Dict,
     List,
     MutableMapping,
-    NoReturn,
+    Sequence,
     Tuple,
     Union,
-    overload,
 )
 
 from more_itertools import partition, unique_everseen
@@ -30,7 +30,6 @@ from . import (
     command as _,  # noqa: F401 # imported for side-effects
 )
 from ._importlib import metadata
-from ._reqs import _StrOrIter
 from .config import pyprojecttoml, setupcfg
 from .discovery import ConfigDiscovery
 from .monkey import get_unpatched
@@ -63,7 +62,13 @@ for use with `isinstance`.
 """
 _Sequence: TypeAlias = Union[Tuple[str, ...], List[str]]
 # This is how stringifying _Sequence would look in Python 3.10
-_requence_type_repr = "tuple[str, ...] | list[str]"
+_sequence_type_repr = "tuple[str, ...] | list[str]"
+_OrderedStrSequence: TypeAlias = Union[str, Dict[str, Any], Sequence[str]]
+"""
+:meta private:
+Avoid single-use iterable. Disallow sets.
+A poor approximation of an OrderedSequence (dict doesn't match a Sequence).
+"""
 
 
 def __getattr__(name: str) -> Any:  # pragma: no cover
@@ -97,7 +102,7 @@ def assert_string_list(dist, attr: str, value: _Sequence) -> None:
         assert ''.join(value) != value
     except (TypeError, ValueError, AttributeError, AssertionError) as e:
         raise DistutilsSetupError(
-            f"{attr!r} must be of type <{_requence_type_repr}> (got {value!r})"
+            f"{attr!r} must be of type <{_sequence_type_repr}> (got {value!r})"
         ) from e
 
 
@@ -173,15 +178,11 @@ def invalid_unless_false(dist, attr, value):
     raise DistutilsSetupError(f"{attr} is invalid.")
 
 
-@overload
-def check_requirements(dist, attr: str, value: set | dict) -> NoReturn: ...
-@overload
-def check_requirements(dist, attr: str, value: _StrOrIter) -> None: ...
-def check_requirements(dist, attr: str, value: _StrOrIter) -> None:
+def check_requirements(dist, attr: str, value: _OrderedStrSequence) -> None:
     """Verify that install_requires is a valid requirements list"""
     try:
         list(_reqs.parse(value))
-        if isinstance(value, (dict, set)):
+        if isinstance(value, set):
             raise TypeError("Unordered types are not allowed")
     except (TypeError, ValueError) as error:
         msg = (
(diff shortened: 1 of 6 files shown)


setuptools/dist.py
```diff
@@ -52,17 +70,17 @@ def check_importable(dist, attr, value):
         ) from e
 
 
-def assert_string_list(dist, attr, value):
+def assert_string_list(dist, attr: str, value: _OrderedIterable):
```
Avasam, line 73: This is more like `assert_ordered_iterable` now. I also realize that whilst this is all technically type-safe, we may not want to allow `dict` for a user-facing reason:
If a user is told they can use a dict here, they're likely going to expect that it means they can use a nested value of some sort, not a "dict with unused values". I also think that forcing a user to coerce their Iterable into a tuple or list in their config file so niche it should be fine.

Avasam: I ended up reverting a lot of this and updating messages instead.

setuptools/dist.py
```diff
@@ -799,7 +820,7 @@ def _include_misc(self, name, value):
             )
         else:
             new = [item for item in value if item not in old]
-            setattr(self, name, old + new)
+            setattr(self, name, list(old) + new)
```
Avasam, line 823: By adding annotations to the method, mypy now raises this issue. This fixes a `TypeError` when a `Distribution`'s old included attribute was a `tuple`. I can create a small separate PR for this if you'd like.

setuptools/dist.py
```diff
@@ -138,18 +153,22 @@ def invalid_unless_false(dist, attr, value):
...
+@overload
+def check_requirements(dist, attr: str, value: set) -> NoReturn: ...
+@overload
+def check_requirements(dist, attr: str, value: _StrOrIter) -> None: ...
+def check_requirements(dist, attr: str, value: _StrOrIter) -> None:
     """Verify that install_requires is a valid requirements list"""
     try:
         list(_reqs.parse(value))
-        if isinstance(value, (dict, set)):
+        if isinstance(value, set):
```
Avasam, line 164: I find it curious that this is the only function in this file checking for a "not set" rather than a "tuple or list". Is there any expectation to allow working with any iterable other than tuple or list? Bbasically the same question I had in https://github.com/pypa/setuptools/pull/4575#discussion_r1720847729

> I also realize that whilst this is all technically type-safe, we may not want to allow `dict` for a user-facing reason: If a user is told they can use a dict here, they're likely going to expect that it means they can use a nested value of some sort, not a "dict with unused values". I also think that forcing a user to coerce their Iterable into a tuple or list in their config file so niche it should be fine.

jaraco: In general, I try not to be overly-prescriptive about input types. For example, accept `Iterable` or `Sequence` so as not to unnecessarily constrain the user. Probably only `tuple` or `list` are used commonly, but I'd be open to allowing users to pass in any ordered object with an `__iter__`.

In this case, we want to assert that whatever value was passed is ordered. The `isinstance(, set)` check is a poor approximation of that, and it's there mainly as a rough check that the user didn't make a silly mistake.

setuptools/dist.py
```diff
@@ -138,11 +139,15 @@ def invalid_unless_false(dist, attr, value):
     raise DistutilsSetupError(f"{attr} is invalid.")
 
 
-def check_requirements(dist, attr, value):
+@overload
+def check_requirements(dist, attr: str, value: set) -> NoReturn: ...
```
jaraco, line 143: I don't like this at all. We don't want to be declaring that `set` is valid, because this function is specifically designed to prevent `set`. I would normally say let's just drop the imperative check and let users rely on mypy to check their types, but I'm guessing it's unlikely that anyone is type-checking their `setup.py`. Also, I'm unsure if there are other code paths. For example, can a user supply a set using `pyproject.toml`? It seems not.

So maybe the best thing would be to drop this override and suppress mypy errors/warnings about its usage.

setuptools/dist.py
```diff
@@ -138,11 +139,15 @@ def invalid_unless_false(dist, attr, value):
...
+@overload
+def check_requirements(dist, attr: str, value: _StrOrIter) -> None: ...
```
jaraco, line 145: I think `_StrOrIter` is wrong here. An `Iterable` could be an `Iterator` that can be exhausted after the first `_reqs.parse()`. What we really want here is an `OrderedSequence`, which might be defined as every `Sequence` that's not unordered. I'm not sure if it's possible to define that in mypy, so an approximation is fine (maybe even `_StrOrIter`).

setuptools/dist.py
Avasam: > We don't want to be declaring that set is valid

That's what this overload is doing, *preventing* the use of `set` in `check_requirements` by indicating it'll never return (ie it'll always raise or exit the program).

There's some cases where it doesn't help (like if `check_requirement` is the last statement of a block), but without it `set` was already always valid anyway because it matches `_StrOrIter`.

One could also use the static `@deprecated` decorator on that overload to further re-enforce and message the user, but that's a bit cheesy (and wouldn't be clean as it'd require a runtime `typing_extension` import or alias)

Finally as per https://github.com/pypa/setuptools/pull/4575/files#r1733535055 , we can instead try to restrict the type of `value` further. For instance `Sequence[str], which already matches a type of nested strings incidentally, and would prevent passing `dict` or `set`.

image

jaraco: Thanks for the explanation. As long as IDEs aren't tempted to _suggest_ passing a `set` because of this annotation, I'm fine with it.

Avasam: > Thanks for the explanation. As long as IDEs aren't tempted to _suggest_ passing a `set` because of this annotation, I'm fine with it.

I just left home, but later I can take a look and show you what it looks like in VSCode/Pylance (especially when it comes to intellisense suggestions). As to not make assumptions here.

Avasam: Because of the necessary overload order, it still suggested sets first. Whilst it also showed it would never return `NoReturn` (return type), that didn't feel right. So I went with your other suggestion. I think that's good enough.

abravalheri:
Thank you very much.

PR #4578, opened 2024-08-20
Type sequence checks in setuptools/dist.py

Avasam:
## Summary of changes

Split from https://github.com/pypa/setuptools/pull/4575

Fixes a `TypeError` when a ``Distribution``'s old included attribute was a `tuple`.

Note that I'm purposefully avoiding using `Iterable`, `Sequence` or `Container` because `str` matches all those. And I didn't wanna complicate the checks to allow "any iterable except str" (maybe in a future PR)

### Pull Request Checklist
- [x] Changes have tests (tests updated with new messages and dict)

[PR docs]:

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -140,8 +140,7 @@ def _check_marker(marker):
 def assert_bool(dist, attr, value):
     """Verify that value is True, False, 0, or 1"""
     if bool(value) != value:
-        tmpl = "{attr!r} must be a boolean value (got {value!r})"
-        raise DistutilsSetupError(tmpl.format(attr=attr, value=value))
+        raise DistutilsSetupError(f"{attr!r} must be a boolean value (got {value!r})")
 
 
 def invalid_unless_false(dist, attr, value):
@@ -163,11 +162,11 @@ def check_requirements(dist, attr: str, value: _StrOrIter) -> None:
         if isinstance(value, (dict, set)):
             raise TypeError("Unordered types are not allowed")
     except (TypeError, ValueError) as error:
-        tmpl = (
+        msg = (
             f"{attr!r} must be a string or iterable of strings "
             f"containing valid project/version requirement specifiers; {error}"
         )
-        raise DistutilsSetupError(tmpl) from error
+        raise DistutilsSetupError(msg) from error
 
 
 def check_specifier(dist, attr, value):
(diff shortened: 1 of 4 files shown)


setuptools/dist.py
```diff
@@ -799,7 +819,7 @@ def _include_misc(self, name, value):
             )
         else:
             new = [item for item in value if item not in old]
-            setattr(self, name, old + new)
+            setattr(self, name, list(old) + new)
```
Avasam, line 822: By adding annotations to the method, mypy now raises this issue. This fixes a `TypeError` when a `Distribution`'s old included attribute was a `tuple`.

setuptools/dist.py
```diff
@@ -138,18 +152,22 @@ def invalid_unless_false(dist, attr, value):
...
             raise TypeError("Unordered types are not allowed")
     except (TypeError, ValueError) as error:
         tmpl = (
-            "{attr!r} must be a string or list of strings "
-            "containing valid project/version requirement specifiers; {error}"
+            f"{attr!r} must be a string or iterable of strings "
+            f"containing valid project/version requirement specifiers; {error}"
         )
-        raise DistutilsSetupError(tmpl.format(attr=attr, error=error)) from error
+        raise DistutilsSetupError(tmpl) from error
```
abravalheri, line 170: Now the string is no longer a template, we could rename the variable from `tmpl` to `msg`, just for improving readability.

setuptools/dist.py
```diff
@@ -36,9 +46,35 @@
...
+
+def __getattr__(name: str) -> Any:  # pragma: no cover
+    if name == "sequence":
+        SetuptoolsDeprecationWarning.emit(
+            "`setuptools.dist.sequence` is an internal implementation detail.",
+            "Please define your own `sequence = tuple, list` instead.",
+            due_date=(2025, 8, 28),  # Originally added on 2024-08-27
+        )
+        return _sequence
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
abravalheri, line 77: @Avasam, how about this for the deprecation of `sequence`?

After a [quick search on github](https://github.com/search?q=%2F%28from+setuptools.*sequence%29%7C%28import+setuptools.*sequence%29%2F+language%3APython&type=code), I do see people accessing it, so we cannot simply remove it...

Avasam: I had opted out of wanting to replace`sequence` by `typing.Sequence` in https://github.com/pypa/setuptools/pull/4575#discussion_r1720850125 because there's a lot of implications when it comes to new checks and supported types (like having to explicitely check for `str` first everytime, because a `str` is a valid `Sequence[str]`, both at runtime and in type-checking).
Not to say it wouldn't be an improvement, just that's not something I wanna go into for the scope of those PRs.

Is making `sequence` private something you want to do independent of those changes ?

abravalheri: > Is making sequence private something you want to do independent of those changes ?

I think so. It seems to be such a trivial implementation detail, I don't know why people are importing it from setuptools.

It is probably good to deprecate it now just in case at some point we decide to remove it.

PR #4585, opened 2024-08-20
Made `setuptools.package_index.Credential` a `NamedTuple`

Avasam:
## Summary of changes

Docstring said "like a namedtuple", but I don't see why it can't be simplified to a `NamedTuple`

Contributes to #2345 and my batches of dunder typing (this was the only untyped `__iter__` outside tests)

One potential difference is the repr:
`'<setuptools.package_index.Credential object at 0x000001D0C827AD90>'`
`Credential(username='a', password='b')`
(but `__str__` already printed the values)

### Pull Request Checklist
- [x] Changes have tests (added doctest for __str__, direct access is already tested)

[PR docs]:

### DIFF
--- a/setuptools/package_index.py
+++ b/setuptools/package_index.py
@@ -18,6 +18,7 @@ import urllib.parse
 import urllib.request
 from fnmatch import translate
 from functools import wraps
+from typing import NamedTuple
 
 from more_itertools import unique_everseen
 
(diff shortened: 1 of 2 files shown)


setuptools/package_index.py
```diff
@@ -1001,21 +1002,20 @@ def _encode_auth(auth):
...
-    def __iter__(self):
-        yield self.username
-        yield self.password
+    username: str
+    password: str
 
-    def __str__(self):
-        return '%(username)s:%(password)s' % vars(self)
+    def __str__(self) -> str:
+        return f'{self.username}:{self.password}'
```
Avasam, line 1018: I considered `return ":".join(self)` but felt explicit was better.

setuptools/package_index.py
```diff
@@ -1001,21 +1002,20 @@ def _encode_auth(auth):
...
-    """
-    A username/password pair. Use like a namedtuple.
+class Credential(NamedTuple):
     """
+    A username/password pair.
 
-    def __init__(self, username, password):
-        self.username = username
-        self.password = password
+    Displayed separated by `:`.
```
Avasam, line 1009: Unsure if this line is redundant
```suggestion
```

abravalheri:
I like this one. We can revert if we find problems in the future.

PR #4562, opened 2024-08-12
Pin Ruff to a lower bound rather than pinning pytest-ruff to an upper bound

Avasam:
## Summary of changes

Closes #4528

### Pull Request Checklist
- [x] Changes have tests (nothing new to test)
- [x] News fragment added in [`newsfragments/`]. (not user facing)
  _(See [documentation][PR docs] for details)_

[`newsfragments/`]: https://github.com/pypa/setuptools/tree/master/newsfragments
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #4586, opened 2024-08-21
Move static-checkers-only dependencies into their dedicated extras

Avasam:
## Summary of changes

https://github.com/pypa/setuptools/commit/0588af036a2b38318c99b39a3549d587b6eb1b4b added extras specific for type-checking and "checkers" (lint+format). I think other dependencies that are only specified for these tests should also be moved there ?

### Pull Request Checklist
- [x] Changes have tests (config change only, tests should pass as-is)

[PR docs]:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,14 +54,6 @@ test = [
 	# for tools/finalize.py
 	'jaraco.develop >= 7.21; python_version >= "3.9" and sys_platform != "cygwin"',
 	"pytest-home >= 0.5",
-	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
-	# until types-setuptools is removed from typeshed.
-	# For help with static-typing issues, or mypy update, ping @Avasam 
-	"mypy==1.11.*",
-	# No Python 3.11 dependencies require tomli, but needed for type-checking since we import it directly
-	"tomli",
-	# No Python 3.12 dependencies require importlib_metadata, but needed for type-checking since we import it directly
-	"importlib_metadata",
 	"pytest-subprocess",
 
 	# workaround for pypa/setuptools#4333
@@ -134,6 +126,11 @@ type = [
 	"pytest-mypy",
 
 	# local
+
+	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
+	# until types-setuptools is removed from typeshed.
+	# For help with static-typing issues, or mypy update, ping @Avasam 
+	"mypy==1.11.*",
 ]
 
 
--- a/setuptools/tests/integration/test_pip_install_sdist.py
+++ b/setuptools/tests/integration/test_pip_install_sdist.py
@@ -202,13 +202,12 @@ def build_deps(package, sdist_file):
     "Manually" install them, since pip will not install build
     deps with `--no-build-isolation`.
     """
-    import tomli as toml
-
     # delay importing, since pytest discovery phase may hit this file from a
     # testenv without tomli
+    from setuptools.compat.py310 import tomllib
 
     archive = Archive(sdist_file)
-    info = toml.loads(_read_pyproject(archive))
+    info = tomllib.loads(_read_pyproject(archive))
     deps = info.get("build-system", {}).get("requires", [])
     deps += EXTRA_BUILD_DEPS.get(package, [])
     # Remove setuptools from requirements (and deduplicate)


pyproject.toml
```diff
@@ -131,7 +125,21 @@ enabler = [
...
+	# pin mypy version so a new version doesn't suddenly cause the CI to fail,
+	# until types-setuptools is removed from typeshed.
+	# For help with static-typing issues, or mypy update, ping @Avasam 
+	"mypy==1.11.*",
+	# Conditionally imported vendored dependencies.
+	# They're not needed at runtime for Python version outside the [core] extra,
+	# but they're still needed for type-checking since we import them directly
+	"tomli",
+	"importlib_resources",
+	"importlib_metadata",
```
abravalheri, line 142: I wonder, can we have these conditional on the Python version?

I think it would be interesting from the testing perspective to see how the test suite behaves when these packages are not installed.

In the `_importlib.py` file we have `if sys.version_info ...` protecting the imports, do you think the typechecking would be smart enough to deal with that?

abravalheri (approved):

pyproject.toml
Avasam: Yes, type checkers should understand simple version and platform checks.

abravalheri: That is an interesting point.

What happens if we run `mypy` (via `pytest-mypy`) with the Python 3.12 executable, but in `mypy.ini` we have `python_version = 3.8`? Will `mypy` simulate that the file is running under `3.8` even for things like `sys.version_info`?

Avasam: Exactly. And it'll take into account modules that are present (or not) in stdlib as specified by typeshed. You could run all your version and platform checks from the same Python executable (as long as all your packages are present in your environment and you don't use new syntax that required AST changes)

abravalheri: Thank you for the clarification @Avasam. 

Now that #4352 is merged should we remove `tomli`, `importlib_*` so that the test suite can check against the stdlib's version? Or are these dependencies still needed?

Avasam: We should be able to remove them. I'm on it :)

pyproject.toml
```diff
@@ -118,8 +103,17 @@ core = [
...
 check = [
+	# upstream
 	"pytest-checkdocs >= 2.4",
 	"pytest-ruff >= 0.2.1; sys_platform != 'cygwin'",
+
+	# local
+
+	# require newer pytest-ruff than upstream for pypa/setuptools#4368
+	# also exclude cygwin for pypa/setuptools#3921
+	'pytest-ruff >= 0.3.2; sys_platform != "cygwin"',
```
Avasam, line 114: This was actually for https://github.com/businho/pytest-ruff/issues/17 , which was re-opened because the output wasn't actually cleaned up. So this extra requirement is unneeded.

Should I remove it now ?
```suggestion
```

Otherwise I've also done it in https://github.com/pypa/setuptools/pull/4562

setuptools/tests/integration/test_pip_install_sdist.py
```diff
@@ -202,13 +202,15 @@ def build_deps(package, sdist_file):
...
     # delay importing, since pytest discovery phase may hit this file from a
     # testenv without tomli
+    if sys.version_info >= (3, 11):
+        import tomllib
+    else:  # pragma: no cover
+        import tomli as tomllib
 
     archive = Archive(sdist_file)
-    info = toml.loads(_read_pyproject(archive))
+    info = tomllib.loads(_read_pyproject(archive))
```
Avasam, line 213: I think technically this is not currently needed to pass CI because our mypy configs cause this method to not be type-checked *yet*.

But it would be needed eventually or with #4192

pyproject.toml
abravalheri: Sounds like a fair assessment. Thank you very much @Avasam.

We can do it in this PR on in any other following one.

setuptools/tests/integration/test_pip_install_sdist.py
```diff
@@ -202,13 +202,15 @@ def build_deps(package, sdist_file):
...
     deps with `--no-build-isolation`.
     """
-    import tomli as toml
-
     # delay importing, since pytest discovery phase may hit this file from a
     # testenv without tomli
+    if sys.version_info >= (3, 11):
+        import tomllib
+    else:  # pragma: no cover
+        import tomli as tomllib
```
abravalheri, line 210: The following may simplify:

```suggestion
    from setuptools.compat.py310 import tomllib
```

PR #4593, opened 2024-08-24
Reraise sensible errors from auto_chmod

Avasam:
## Summary of changes

My best understanding is that this is old Python2 code that was trying to re-raise an error with additional information in the message. (back when raise could take a tuple of strings)

This PR changes the nonsensical ``TypeError: 'Exception' object is not subscriptable`` that would always happen now for for sensible and complete messages with the old error as part of the trace. Optionally we could just reraise.

### Pull Request Checklist
- [ ] Changes have tests (TODO)

[PR docs]:


setuptools/command/easy_install.py
```diff
@@ -1786,13 +1788,18 @@ def _first_line_re():
...
+    """shutils onexc callback to automatically call chmod for certain functions."""
+    if os.name != 'nt':
+        raise OSError(f"Can't call auto_chmod on non-nt os {os.name!r}") from exc
+    supported_methods = {os.unlink, os.remove}
+    if func not in supported_methods:
+        raise ValueError(
+            f"Argument func is not one of {[method.__name__ for method in supported_methods]} (got: {func.__name__!r})"
+        ) from exc
+    chmod(arg, stat.S_IWRITE)
+    return func(arg)
```
Avasam, line 1802: Alternatively, less detailed, but much more concise:
```suggestion
    if func in [os.unlink, os.remove] and os.name == 'nt':
        chmod(arg, stat.S_IWRITE)
        return func(arg)
    raise exc
```

setuptools/command/easy_install.py
```diff
@@ -1786,13 +1788,18 @@ def _first_line_re():
...
-        chmod(arg, stat.S_IWRITE)
-        return func(arg)
-    et, ev, _ = sys.exc_info()
-    # TODO: This code doesn't make sense. What is it trying to do?
-    raise (ev[0], ev[1] + (" %s %s" % (func, arg)))
+# Must match shutil._OnExcCallback
+def auto_chmod(func: Callable[..., _T], arg: str, exc: BaseException) -> _T:
+    """shutils onexc callback to automatically call chmod for certain functions."""
+    if os.name != 'nt':
+        raise OSError(f"Can't call auto_chmod on non-nt os {os.name!r}") from exc
```
abravalheri, line 1795: This is an interesting aspect of the previous implementation (not related to your changes): why can we not call `chmod(path, stat.S_IWRITE)` on non Windows OSs?

I had a look on the `git blame` history and found that this check was introduced in https://github.com/pypa/setuptools/commit/d675527eae964f3b69d5867f4df8800d10767d5c, but I cannot guess from the commit message why it is restricted to Windows only.

This thread https://bugs.python.org/issue22040 seems to suggest that rather than the workaround not being applicable to other OSs, it is the problem that seems to only appear on Windows...

I wonder if we actually need this extra check...

I can see other projects in the community just going for it, without checking for windows:

https://github.com/python-poetry/poetry/blob/5d19ab0d8e9b734e3452c6e1210a7f2501709397/src/poetry/utils/helpers.py#L93-L98

https://github.com/scientific-python/cookie/blob/f170dd542dfc133a0de3a36ac3c5c36470f066d7/noxfile.py#L43-L45


The most complex handling I saw were from pre-commit and pip, but still there is no OS-specific check:

https://github.com/pre-commit/pre-commit/blob/504149d2ca57f3115ccd4fa9c6ae692929b2282a/pre_commit/util.py#L205-L219

https://github.com/pypa/pip/blob/main/src/pip/_internal/utils/misc.py#L153-L191

Avasam: I don't see why either, just ran `os.chmod(".", stat.S_IWRITE)` in WSL and it doesn't immediately error out or whatever.

> seems to suggest that rather than the workaround not being applicable to other OSs, it is the problem that seems to only appear on Windows...

I'd make more sense to just pass-through if not Windows, or call `os.chmod` indescriminately.

Avasam: Here's another way of seeing it: if we know the error only happens on Windows, there's no point in "fixing and retrying", so it was just immediately rearising to keep the traceback to a minimum.

abravalheri:
Thank you very much!

PR #4555, opened 2024-08-10
Check MSVC arm64 variant on arm64 host

saschanaz:
## Summary of changes

For now this selects Hostx86/arm64 because of old distutils decision:

https://github.com/pypa/setuptools/blob/a6f7484b25628ae351d4b96fa6254ada387ece10/setuptools/_distutils/_msvccompiler.py#L181-L189

This works but just runs slower. Perhaps we can safely require arm64 MSVC for arm64 python installations?

Closes #4553

### Pull Request Checklist
- [ ] Changes have tests - [`test_get_vc2017()` exists but never fails](https://github.com/pypa/setuptools/blob/a6f7484b25628ae351d4b96fa6254ada387ece10/setuptools/tests/test_msvc14.py#L54-L66)
- [x] News fragment added in [`newsfragments/`].
  _(See [documentation][PR docs] for details)_

[`newsfragments/`]: https://github.com/pypa/setuptools/tree/master/newsfragments
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #2334, opened 2020-08-16
Resolving error text issue (#2302)

arencambre:
## Summary of changes

Changed two error strings to be up to date.

Closes 2302

### Pull Request Checklist
- [x] Changes have tests
- [x] News fragment added in changelog.d. See [documentation](http://setuptools.readthedocs.io/en/latest/developer-guide.html#making-a-pull-request) for details

PR #4600, opened 2024-08-27
Remove monkeypatching of _msvccompiler.

jaraco:
The monkeypatch indicates that the functionality was ported from Python 3.8, so the version in distutils should be adequate now.

I note that there have been some [changes to this module](https://github.com/pypa/setuptools/commits/4c990b9eab4d33faf517fcc2080662ebde0d2841/setuptools/msvc.py) since it was ported from Python 3.8, including:

- [x] Arm support #4555 (addressed by pypa/distutils#285)
- [x] requiresAny support #3950 (addressed by pypa/distutils#289)
- [x] VS/VC 2017 support #2663 (addressed by pypa/distutils#289) 
- [x] #2334 (addressed by pypa/distutils#289)
- [x] ~~https://github.com/pypa/setuptools/pull/1904~~ (this was the port from Python 3.8.0)

We'll want to have those issues supported by pypa/distutils ~~if not fully dropping support for stdlib distutils~~ before enacting this change.

~~Merge after #4606.~~

## Summary of changes

Closes 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


abravalheri:
There are probably some changes in setuptools/msvc.py that need to be portrait to distutils.

jaraco:
I updated the description to highlight the relevant changes.



> We'll want to have those issues supported by pypa/distutils if not fully dropping support for stdlib distutils before enacting this change.

On further consideration, I think it's safe to force Windows users who need improved VC/VS support not to use distutils from the stdlib, so support in pypa/distutils should be adequate.

jaraco:
I've added #1904 to the list of changes to review. There are a lot of changes in there that don't match with what's in distutils, presumably because distutils has evolved since Python 3.8. I'm hoping it's not the case that these additional behaviors (like those to `find_vcvarsall`) weren't subtly added in #1904.

jaraco:
Oh, interesting. So it looks like the contents in msvc were indeed drawn from [3.8.0](https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Lib/distutils/_msvccompiler.py#L99-L136) but that implementation changed dramatically throughout the 3.8 series (especially in https://github.com/python/cpython/commit/0d20364b132014eec609b900997c34779a4d548c). This means that although bugs were fixed in stdlib distutils, those fixes have been largely unavailable to Setuptools' users.

We do want to adopt the fixes that are applied in distutils (and not go backward on those changes).

That means that there's liable to be more churn when adopting this change than I'd like, as distutils can't be made a drop-in replacement for what setuptools users expect. Depending on the amount of disruption, we may need to yank releases or back out changes until the differences can be assessed.

saschanaz:
> Arm support https://github.com/pypa/setuptools/pull/4555 (addressed by https://github.com/pypa/distutils/pull/285)

But those are independent, no? The former is to find vcvarsall and the latter is to get the right tool from that vcvarsal.

saschanaz:
Oh, folded into https://github.com/pypa/setuptools/commit/196d44b8a1bd626764e9679c1b4b77ccc1d7ed3f. Never mind then.

jaraco:
Thanks for reviewing. Please let me know how Setuptools 74 works for you. I'm happy to address any unexpected failures promptly.

saschanaz:
The linked issues are unhappy with setuptools.msvc.EnvironmentInfo being removed, is there an equivalent now?

jaraco:
There isn't. I did not realize they were in use. I was assuming and did not verify that those classes were only previously used for internal implementation details for functionality that had been removed.

Since users are relying on it, we should bring them back, and add some tests.

PR #3067, opened 2022-02-01
[`pyproject.toml`: part 3] Add means to apply configuration from `pyproject.toml`

abravalheri:
> *This is part of #2970, split into several stacked PRs as requested in https://github.com/pypa/setuptools/pull/2970#pullrequestreview-868151658*
> *Follows #3066*

This change targets adding a standardised project configuration format as initially proposed by PEP 621 (with the possibility of modification/evolution according to follow-up PEPs).

For the configuration fields not covered by the standards, the approach proposed in [`ini2toml`](https://ini2toml.readthedocs.io/en/latest/setuptools_pep621.html) is adopted.

This operation relies on vendorised external libraries `tomli` and `validate-pyproject`.

## Summary of changes

- Implement `read_configuration` from `pyproject.toml` - 3eb62ac2912f8c0a8bfbdc5dbff809f983a14e2a
- Expand dynamic `entry_points` from `pyproject.toml` - 112e8a930314eddf9f35159ea890aaec750843c8
- Add means of applying config read from `pyproject.toml` to dist - 303791af3ed6be54d0a3663b1738537be6b30f32
- Add the `apply_configuration` API to `setuptools.config.setupcfg` - 43e34361e072880841654412c61d9c508e28a5f9
- Test `pyproject.toml` config has the same effect as `setup.cfg` -  c6907cb008fd647f9180e29ec4ee780e148b832c

Please see the commit message of each individual commit for more information about the changes.

### Pull Request Checklist
- [x] Changes have tests
- [ ] News fragment added in [`changelog.d/`].
  _(See [documentation][PR docs] for details)_

[`changelog.d/`]: https://github.com/pypa/setuptools/tree/master/changelog.d
[PR docs]:
https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

PR #4604, opened 2024-08-27
Re-enable mypy and sync mypy.ini from skeleton

Avasam:
## Summary of changes

Ref https://github.com/jaraco/skeleton/issues/143

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -64,6 +64,6 @@ ignore_missing_imports = True
 
 # Even when excluding a module, import issues can show up due to following import
 # https://github.com/python/mypy/issues/11936#issuecomment-1466764006
-[mypy-setuptools.config._validate_pyproject.*,setuptools._distutils.*]
+[mypy-setuptools.config._validate_pyproject.*,setuptools._vendor.*,setuptools._distutils.*]
 follow_imports = silent
 # silent => ignore errors when following imports
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,16 +1,26 @@
 [mypy]
-# CI should test for all versions, local development gets hints for oldest supported
-# But our testing setup doesn't allow passing CLI arguments, so local devs have to set this manually.
-# python_version = 3.8
+## upstream
+
+# Is the project well-typed?
 strict = False
+
+# Early opt-in even when strict = False
 warn_unused_ignores = True
 warn_redundant_casts = True
-# required to support namespace packages: https://github.com/python/mypy/issues/14057
+enable_error_code = ignore-without-code
+
+# Support namespace packages per https://github.com/python/mypy/issues/14057
 explicit_package_bases = True
 
 disable_error_code =
 	# Disable due to many false positives
 	overload-overlap,
+
+## local
+
+# CI should test for all versions, local development gets hints for oldest supported
+# But our testing setup doesn't allow passing CLI arguments, so local devs have to set this manually.
+# python_version = 3.8
 
 exclude = (?x)(
 	# Avoid scanning Python files in generated folders
--- a/setuptools/config/expand.py
+++ b/setuptools/config/expand.py
@@ -203,7 +203,8 @@ def _load_spec(spec: ModuleSpec, module_name: str) -> ModuleType:
         return sys.modules[name]
     module = importlib.util.module_from_spec(spec)
     sys.modules[name] = module  # cache (it also ensures `==` works on loaded items)
-    spec.loader.exec_module(module)  # type: ignore
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
     return module
 
 
@@ -285,10 +286,11 @@ def find_packages(
 
     from setuptools.discovery import construct_package_dir
 
-    if namespaces:
-        from setuptools.discovery import PEP420PackageFinder as PackageFinder
+    # check "not namespaces" first due to python/mypy#6232
+    if not namespaces:
+        from setuptools.discovery import PackageFinder
     else:
-        from setuptools.discovery import PackageFinder  # type: ignore
+        from setuptools.discovery import PEP420PackageFinder as PackageFinder
 
     root_dir = root_dir or os.curdir
     where = kwargs.pop('where', ['.'])
@@ -359,7 +361,8 @@ def entry_points(text: str, text_source="entry-points") -> dict[str, dict]:
     entry-point names, and the second level values are references to objects
     (that correspond to the entry-point value).
     """
-    parser = ConfigParser(default_section=None, delimiters=("=",))  # type: ignore
+    # Using undocumented behaviour, see python/typeshed#12700
+    parser = ConfigParser(default_section=None, delimiters=("=",))  # type: ignore[call-overload]
     parser.optionxform = str  # case sensitive
     parser.read_string(text, text_source)
     groups = {k: dict(v.items()) for k, v in parser.items()}
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -1418,7 +1418,7 @@ class EnvironmentInfo:
             os.path.join(prefix, arch_subdir, crt_dir, vcruntime)
             for (prefix, crt_dir) in itertools.product(prefixes, crt_dirs)
         )
-        return next(filter(os.path.isfile, candidate_paths), None)
+        return next(filter(os.path.isfile, candidate_paths), None)  # type: ignore[arg-type] #python/mypy#12682
 
     def return_env(self, exists=True):
         """
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2777,7 +2777,7 @@ class EntryPoint:
         if require:
             # We could pass `env` and `installer` directly,
             # but keeping `*args` and `**kwargs` for backwards compatibility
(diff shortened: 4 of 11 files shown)


mypy.ini
```diff
@@ -51,6 +51,6 @@ ignore_missing_imports = True
 
 # Even when excluding a module, import issues can show up due to following import
 # https://github.com/python/mypy/issues/11936#issuecomment-1466764006
-[mypy-setuptools.config._validate_pyproject.*,setuptools._distutils.*]
+[mypy-setuptools.config._validate_pyproject.*,setuptools._vendor.*,setuptools._distutils.*]
```
jaraco, line 54: Do we know why the exclusions had to change? Presumably the mypy tests were passing before it was disabled.

Avasam: Whilst the CI setup makes it pass, locally on Windows:
```
setuptools\_vendor\packaging\_manylinux.py:187: error: Cannot find implementation or library stub for module named "_manylinux"  [import-not-found]
setuptools\_vendor\packaging\_manylinux.py:187: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

Avasam:
The "Fix ConfigHandler generic" commit is related to https://github.com/python/mypy/issues/17631#issuecomment-2277491389 (same symptoms). Not sure why it wasn't failing before, didn't fail locally (granted I'm running from Windows on 3.9, so it could be environment-specific, despite version/platform flags), and only failed on the CI on (MacOS and/or 3.12)[^1].

But it still revealed an underlying misuse of the `TypeVar`

[^1]: No other MacOS or Python 3.12 action had time to complete, so idk which one was the specific issue.

jaraco:
I still see some test failures.

```
_________________________ setuptools/config/expand.py __________________________
[gw3] linux -- Python 3.10.14 /home/runner/work/setuptools/setuptools/.tox/py/bin/python
206: error: "type: ignore" comment without error code (consider "type: ignore[union-attr]" instead)  [ignore-without-code]
291: error: "type: ignore" comment without error code (consider "type: ignore[assignment]" instead)  [ignore-without-code]
362: error: "type: ignore" comment without error code (consider "type: ignore[call-overload]" instead)  [ignore-without-code]
______________________ setuptools/config/pyprojecttoml.py ______________________
[gw3] linux -- Python 3.10.14 /home/runner/work/setuptools/setuptools/.tox/py/bin/python
48: error: "type: ignore" comment without error code (consider "type: ignore[union-attr]" instead)  [ignore-without-code]
__________________________ pkg_resources/__init__.py ___________________________
[gw0] linux -- Python 3.10.14 /home/runner/work/setuptools/setuptools/.tox/py/bin/python
2780: error: "type: ignore" comment without error code (consider "type: ignore[arg-type]" instead)  [ignore-without-code]
__________________ setuptools/tests/test_editable_install.py ___________________
[gw2] linux -- Python 3.10.14 /home/runner/work/setuptools/setuptools/.tox/py/bin/python
881: error: "type: ignore" comment without error code (consider "type: ignore[index]" instead)  [ignore-without-code]
883: error: "type: ignore" comment without error code (consider "type: ignore[index]" instead)  [ignore-without-code]
891: error: "type: ignore" comment without error code (consider "type: ignore[index]" instead)  [ignore-without-code]
892: error: "type: ignore" comment without error code (consider "type: ignore[index]" instead)  [ignore-without-code]
```

What's the plan for those?

Also, thanks for reconciling the differences between setuptools' and skeleton's mypy.ini.

Avasam:
> I still see some test failures.
> What's the plan for those?

I rebase and fix them :)

setuptools/config/expand.py
```diff
@@ -203,7 +203,9 @@ def _load_spec(spec: ModuleSpec, module_name: str) -> ModuleType:
         return sys.modules[name]
     module = importlib.util.module_from_spec(spec)
     sys.modules[name] = module  # cache (it also ensures `==` works on loaded items)
-    spec.loader.exec_module(module)  # type: ignore
+    if spec.loader is None:
+        raise AttributeError(f"spec {spec} is missing a loader")
```
Avasam, line 207: or this if you think it should really never happen and want to remove the check in optimized mode

```suggestion
    assert spec.loader is not None
```

setuptools/config/expand.py
```diff
@@ -285,10 +287,10 @@ def find_packages(
 
     from setuptools.discovery import construct_package_dir
 
-    if namespaces:
-        from setuptools.discovery import PEP420PackageFinder as PackageFinder
+    if not namespaces:
+        from setuptools.discovery import PackageFinder
     else:
-        from setuptools.discovery import PackageFinder  # type: ignore
+        from setuptools.discovery import PEP420PackageFinder as PackageFinder
```
Avasam, line 293: Avoids python/mypy#6232

setuptools/config/expand.py
```diff
@@ -359,7 +361,8 @@ def entry_points(text: str, text_source="entry-points") -> dict[str, dict]:
     entry-point names, and the second level values are references to objects
     (that correspond to the entry-point value).
     """
-    parser = ConfigParser(default_section=None, delimiters=("=",))  # type: ignore
+    # TODO: Explain why passing None is fine here when ConfigParser.default_section expects to be str
+    parser = ConfigParser(default_section=None, delimiters=("=",))  # type: ignore[call-overload]
```
Avasam, line 365: Could this be ?
```suggestion
    parser = ConfigParser(default_section="", delimiters=("=",))
```

setuptools/msvc.py
```diff
@@ -1418,7 +1418,7 @@ def VCRuntimeRedist(self) -> str | None:
             os.path.join(prefix, arch_subdir, crt_dir, vcruntime)
             for (prefix, crt_dir) in itertools.product(prefixes, crt_dirs)
         )
-        return next(filter(os.path.isfile, candidate_paths), None)
+        return next(filter(os.path.isfile, candidate_paths), None)  # type: ignore[arg-type] #python/mypy#12682
```
Avasam, line 1421: Ref https://github.com/python/mypy/issues/12682

Avasam:
@jaraco Bump on this since currently static typing validation from mypy is disabled (at least pyright's running). And some of my pending PRs should be validated by it.

setuptools/config/expand.py
jaraco: I prefer the suggested line. It's more succinct and seems to carry the same expectation.

setuptools/config/expand.py
jaraco: The meaning of `default_section` is [documented here](https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour). It suggests that the default section could be overridden from the default of `"DEFAULT"`.

I don't know why the original author passed `default_section=None`. Reading briefly through the [configparser code](https://github.com/python/cpython/blob/main/Lib/configparser.py), it looks like the `default_section` is often compared for equality, so `None` is effectively treated as "no default". Presumably that was the intention of this usage.

Probably what should happen here is configparser should be explicit about what types are allowed here. I scanned through the configparser tests, but didn't see anything that explicitly allowed or excluded `None` as a valid type, and since `None` appears to be valid per this usage, it's probably more accurately an `Any` or `Hashable` type. 

My recommendation - file a ticket with python/cpython (or python/typeshed) indicating that the type spec for this parameter is ambiguous and should be refined, and then link that issue here instead of a TODO (since there's nothing "to do" in setuptools until such a refinement is present).

jaraco (approved):
LGTM. Just a couple lingering concerns.

setuptools/config/expand.py
Avasam: The author being @abravalheri in https://github.com/pypa/setuptools/pull/3067/files#diff-2e8e2332e00f145022b07f1a5de895e643f38570a1765bd2da26e27fe29b64cdR305 I'll ask even if that was a couple years ago :)

Do you remember if there was any reason to set `ConfigParser`'s `default_section` to `None` instead of using the default of `"DEFAULT"` ? (re-exported as `configparser.DEFAULTSECT`)

Avasam: In the mean time, I marked setuptool's usage as undocumented behaviour, linking to his issue/question I created: https://github.com/python/typeshed/issues/12700

Further actions can come in a follow-up PR once the question is resolved and an action has been taken.

Avasam:
Other PRs started failing too with a strange error (not the same though), so CI failure may or may not be related to these changes (although I don't see which of my changes could cause it)

Avasam:
Now that the weird CI issue seems resolved. I'd like to merge this before bumping mypy to 1.12. And to unblock #4504

setuptools/config/expand.py
abravalheri: > Do you remember if there was any reason to set `ConfigParser`'s `default_section` to `None` instead of using the default of `"DEFAULT"` ? (re-exported as `configparser.DEFAULTSECT`)

I believe I was trying to achieve the behaviour described by Jason, i.e. "no default".

Avasam: @abravalheri Thanks. This confirms what we thought. I think the new comment of calling this out as undocumented behaviour but not changing it is probably fine.

Issue #4638, opened 2024-09-07
[BUG] Unclear selection of package data in sdist and wheels


DanielYang59:
### setuptools version

setuptools==74.1.2

### Python version

Python 3.12.5 (MacOS) Python 3.10.12 (Ubuntu-WSL2)

### OS

MacOS Sonoma 14.6.1 and Ubuntu-WSL2 22.04 LTS

### Additional environment information

Well I haven't installed `setuptools` on my fridge (yet), maybe later

### Description

I was trying to [replace explicit declare of package data with `include-package-data = true` for `pymatgen`](https://github.com/DanielYang59/pymatgen/tree/package-data), but noticed `python3 -m build` could generate source distribution with package data missing (even after I removed `dist` and `src/pymatgen.egg-info`). I believe no other files were controlling this process (no `MANIFEST.in`, [no related args in `setup.py`](https://github.com/DanielYang59/pymatgen/blob/package-data/setup.py)).

Here's the related section in `pyproject.toml` file ([this is the complete file](https://github.com/DanielYang59/pymatgen/blob/package-data/pyproject.toml)):

```
[build-system]
build-backend = "setuptools.build_meta"

[tool.setuptools]
include-package-data = true

[tool.setuptools.packages.find]
where = ["src"]
include = ["pymatgen", "pymatgen.*"]
```

### Expected behavior

I would expect `python3 -m build` to behave consistently and generate wheels and sdist exactly as specified in `pyproject.toml` file.

The generated wheel ([the wheel on pypi](https://pypi.org/project/pymatgen/#files) for example) and sdist should include multiple package data files (json and more):
```
wget https://files.pythonhosted.org/packages/7f/9e/b8b1014e4a81ffb048a905e71efde31a4512ea4ae00421356ea62c52ad2d/pymatgen-2024.8.9-cp312-cp312-macosx_11_0_arm64.whl

>>> unzip -l pymatgen-2024.8.9-cp312-cp312-macosx_11_0_arm64.whl  | grep json
    16284  2024-08-09 17:16   pymatgen/entries/calc_compounds.json.gz
    34835  2024-08-09 17:16   pymatgen/entries/exp_compounds.json.gz
    17909  2024-08-09 17:16   pymatgen/entries/data/nist_gas_gf.json
    40327  2024-08-09 17:16   pymatgen/entries/data/g_els.json
...  more 
```

### How to Reproduce

```
[...]

PR #4643, opened 2024-09-10
[Docs] Clarify usage of `include_package_data/package_data/exclude_package_data` on package data files

DanielYang59:
## Summary of changes

- Clarify usage of `include_package_data/package_data/exclude_package_data` in `pyproject.toml` on inclusion of package data files, as a follow up of #4638
- Clarify priority level https://github.com/pypa/setuptools/issues/4638#issuecomment-2338508890


docs/userguide/datafiles.rst
```diff
@@ -35,16 +37,34 @@ For example, if the package tree looks like this::
             ├── data1.txt
             └── data2.txt
 
-and you supply this configuration:
+When at least one of the following conditions are met:
```
DanielYang59, line 40: The `include_package_data` keyword would only have an effect when condition is met, perhaps it's good to state these upfront?

docs/userguide/datafiles.rst
```diff
@@ -35,16 +37,34 @@ For example, if the package tree looks like this::
...
-and you supply this configuration:
+When **at least one** of the following conditions are met:
+
+1. These files are included via the :ref:`MANIFEST.in <Using MANIFEST.in>` file,
+   like so::
+
+        include src/mypkg/*.txt
+        include src/mypkg/*.rst
+
+2. They are being tracked by a version control system such as Git, Mercurial
```
DanielYang59, line 48: I guess "version control system" might be more appropriate in this context than ["revision control system"](https://en.wikipedia.org/wiki/Revision_Control_System):

> Revision Control System (RCS) is an early implementation of a [version control system](https://en.wikipedia.org/wiki/Version_control_system) (VCS).

From [the Wikipedia of Git](https://en.wikipedia.org/wiki/Git):
> Git ([/ɡɪt/](https://en.wikipedia.org/wiki/Help:IPA/English))[[8]](https://en.wikipedia.org/wiki/Git#cite_note-:0-9) is a [distributed version control](https://en.wikipedia.org/wiki/Distributed_version_control) [system](https://en.wikipedia.org/wiki/Software_system)[[9]](https://en.wikipedia.org/wiki/Git#cite_note-FOOTNOTEChaconStraub201429%E2%80%9331-10) that tracks versions of [files](https://en.wikipedia.org/wiki/Computer_file).

And [SVN](https://en.wikipedia.org/wiki/Apache_Subversion):

> Apache Subversion (often abbreviated SVN, after its command name svn) is a [version control](https://en.wikipedia.org/wiki/Version_control) system distributed as [open source](https://en.wikipedia.org/wiki/Open-source_software) under the [Apache License](https://en.wikipedia.org/wiki/Apache_License).[[2]](https://en.wikipedia.org/wiki/Apache_Subversion#cite_note-3)

docs/userguide/datafiles.rst
```diff
@@ -172,7 +177,7 @@ file, nor require to be added by a revision control system plugin.
 
 .. note::
         If your glob patterns use paths, you *must* use a forward slash (``/``) as
-        the path separator, even if you are on Windows.  Setuptools automatically
+        the path separator, even if you are on Windows. ``Setuptools`` automatically
```
abravalheri, line 180: As far as I remember in the docs we use capital Setuptools without monospaced font when referring to the project name and ``setuptools`` with monospaced font when referring to the package. Either one is fine, but ``Setuptools`` looks a bit off. I think it would be better if we use either the capital letter or the monospace font but not both.

```suggestion
        the path separator, even if you are on Windows. Setuptools automatically
```

docs/userguide/datafiles.rst
abravalheri: I am happy with both.

I believe that the entry in Wikipedia that you cited is referring to a specific "object" in the "category" of Revision Control Systems, also called "Revision Control System" :P, but not the category itself.

For our purposes "Revision Control System" or "Revision Control Software" are synonyms with "Version Control System". Based in your comment, Mercurial seems to consider the same.

It would be interesting however to keep the usage consistent and use the same term across the whole documentation. So if we are changing this paragraph, we should also change elsewhere. Alternatively we can also add a `(also known as Version Control System)` note somewhere...

abravalheri (commented):
Thank you very much for your contribution!

docs/userguide/datafiles.rst
DanielYang59: Okay, I would revert this change because these two are to some extent interchangeable, and is not the focus of this PR (plus I know little about terminology used elsewhere across the docs).

docs/userguide/datafiles.rst
DanielYang59: Thanks for the info, it looks like in this scenario we're more referring to the package (and therefore `setuptools` being more suitable)?

docs/userguide/datafiles.rst
```diff
@@ -2,13 +2,14 @@
...
-
-Setuptools focuses on this most common type of data files and offers three ways
+Old packaging installation methods in the Python ecosystem have
+traditionally allowed the inclusion of "data files" (files beyond
+:ref:`the default set <manifest>` ), which are placed in a platform-specific
+location. However, the most common use case for data files distributed
+with a package is for use *by* the package, usually by including the
+data files **inside the package directory**.
+
+``Setuptools`` focuses on this most common type of data files and offers three ways
```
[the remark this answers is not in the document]
DanielYang59: > Thank you for your suggestion. I understand that your main objective was to clarify what “data files” are.

Yes exactly. Because this is the starting sentence of the `Data Files Support` section, and we have to make sure users understand what is considered "data file" in the first place, such that they could decide whether they even need these keywords at all.

For example C source code/READMEs are not "Python" files, but we don't need to declare them to include them into the package. And therefore we might need to clarify the following definition of "data files":
> we define "data files" as non-Python files


> the proposed parenthesis seem to suggest: data file = file not included in the "default set". This could lead to confusion, as what defines a data file is not fundamentally related to whether a file is included in the default set or not. 

Thanks a ton for the clarification! Yes my sketch indeed would need more discussion and check.

> They may also be referred to as "resource files."

Personally I think the following terminologies are already very confusing: "package data", "data file", "source distribution", "files inside the wheel", and their difference is pretty unclear, and can we somehow define/clarify them, and perhaps avoid introducing more synonyms ("resource files" in this case)?

DanielYang59:
I just noticed [this](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/) from official Python docs:

> MANIFEST.in does not affect binary distributions such as wheels.

docs/userguide/datafiles.rst
abravalheri: Great, thank you!

docs/userguide/datafiles.rst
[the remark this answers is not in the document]
abravalheri: I have no problems with omitting `resource files`, I just mentioned before because other packages started calling them that way (e.g. `importlib_resources`). Please feel free to use and modify my suggestion to align with your vision in the contribution.

----

> For example C source code/READMEs are not "Python" files, but we don't need to declare them to include them into the package.

Yes, in this case C/README files are not data files, because they don't fit in the remaining part of the suggested definition:

```rst
... that are installed alongside Python modules and packages on the user's machine
when they install a :term:`distribution <Distribution Package>` from PyPI
or via a ``.whl`` file.
These files are typically intended for use at runtime by the package itself or
to influence the behavior of other packages or systems.
```

C-files are not installed along side Python files in the user's machine and also are not intended to be used in runtime, right?

----

> I just noticed [this](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/) from official Python docs:
> 
> > MANIFEST.in does not affect binary distributions such as wheels.


That is not 100% precise is it? `MANIFEST.in` determines what goes into the `sdist`, and then the contents of the `sdist` influence what goes into the wheel (specially when `include-package-data = true` which is the default)... So there is a potential indirect effect there. That is because the build process works more or less like in the mermaidjs diagram below[^1]:

```mermaid
graph LR
    src[fa:fa-laptop-code source code] --> b1(fa:fa-hammer)
    
    subgraph build process
      b1 --> sdist(fa:fa-file-code sdist) --> b2(fa:fa-hammer)
      b2 --> wheel(fa:fa-box wheel)
       
      build([build dependencies]) --> b1
      build([build dependencies]) --> b2
    end

    subgraph installation process
        wheel --> pkg
        pkg(fa:fa-box-open) --> inst[fa:fa-python installed packages]
        runtime([runtime dependencies]) --> inst
    end

    classDef deps fill:#2aa198,stroke:#333
    classDef env fill:#6c71c4,stroke:#333
    build:::deps
    runtime:::deps
```

[^1]: The "build process" creates distribution artifacts/packages. `sdist` distributions are meant to be platform independent (but may contain varying levels of optimisation - e.g. compiling Python to C via Cython). `wheel` distributions contain the final files meant to be directly copied to the user's machine, and therefore may be platform specific.

docs/userguide/datafiles.rst
DanielYang59: Thanks I would fold this

docs/userguide/datafiles.rst
```diff
@@ -2,12 +2,20 @@
...
-:ref:`the default set <manifest>` ), which are placed in a platform-specific
-location. However, the most common use case for data files distributed
-with a package is for use *by* the package, usually by including the
-data files **inside the package directory**.
+In the Python ecosystem, the term "data files" is used in various complex scenarios
+and can have nuanced meanings.
+For the purposes of this documentation, we define "data files" as non-Python files
+that are installed alongside Python modules and packages on the user's machine
+when they install a :term:`distribution <Distribution Package>` from PyPI
+or via a ``.whl`` file.
```
DanielYang59, line 10: I would tag and explain something I want to change (I find myself lose track of what I was originally intended if I overthink), please free feel to comment.

I believe the source (PyPI vs a `.whl` file) is not important here, but whether we're installing from a binary distribution (`.whl`) or source distribution that matters.

docs/userguide/datafiles.rst
```diff
@@ -351,6 +351,38 @@ Any files that match these patterns will be *excluded* from installation,
...
+
+TODO:
+- better formatted logic expression
+- check logic correctness (especially regarding parentheses)
+
+Meanwhile, to further clarify the interplay between these three keywords,
+to include certain data file into the source distribution, the following
+logic condition has two be met::
+
+    (m or p) and not (i or e)  # TODO: why "not i"?
```
DanielYang59, line 367: Is this a typo? Why `not i` to include data file into the sdist?

docs/userguide/datafiles.rst
```diff
@@ -364,10 +364,9 @@ Meanwhile, to further clarify the interplay between these three keywords,
 to include certain data file into the source distribution, the following
 logic condition has two be met::
 
-    (m or p) and not (i or e)  # TODO: why "not i"?
+    m or (p and not (i or e))  # TODO: why not i?
```
DanielYang59, line 367: @abravalheri Can you help me have a look here? why `not i` for source distribution here?

docs/userguide/datafiles.rst
```diff
@@ -356,24 +356,21 @@ even if they were listed in ``package_data`` or were included as a result of usi
...
-TODO:
-- better formatted logic expression
-- check logic correctness (especially regarding parentheses)
-
 Meanwhile, to further clarify the interplay between these three keywords,
 to include certain data file into the source distribution, the following
 logic condition has two be met::
 
-    m or (p and not (i or e))  # TODO: why not i?
+    m or (p and not e)
```
DanielYang59, line 363: I think for documentation, using plain text (or, and, not) might be more readable than symbols.

Also perhaps we should link to [the original experiment repo](https://github.com/abravalheri/experiment-setuptools-package-data?tab=readme-ov-file#results-for-setuptools6060) in case someone want to know different behaviour in different versions of `setuptools`

docs/userguide/datafiles.rst
abravalheri: We are probably better off if we don't define data files in terms of source distributions or source trees...

docs/userguide/datafiles.rst
abravalheri: > why `not i`?

Sorry, I copied over the wrong expression. This is the most updated https://github.com/abravalheri/experiment-setuptools-package-data?tab=readme-ov-file#results-for-setuptools6060.

```
s = m + e'.p
w = i.e'.m + e'.p  -- Which can be simplified => e' . (i.m + p)
                   -- and is also equivalent to => e' . (i.s + p)
```

`setuptools==58.3.0` had a but that made it operate in a non intuitive way.

docs/userguide/datafiles.rst
abravalheri: If we are going for plain text, then it is probably better to use the full keywords:

```
MANIFEST.in OR (package-data AND NOT exclude-package-data)
```

Or something like that.

docs/userguide/datafiles.rst
```diff
@@ -337,6 +351,40 @@ Any files that match these patterns will be *excluded* from installation,
...
+    i - "include-package-data = true" is set
+    e - file selected by "exclude-package-data"
+    p - file selected by "package-data"
+    m - file included in "MANIFEST.in"
+
+.. note::
+    Different versions of ``setuptools`` might behave differently. The above
+    description applies to versions after ``58.5.3`` (exclusive). For information
+    on the behavior of earlier versions and more details, please refer to the
+    `GitHub repository <https://github.com/abravalheri/experiment-setuptools-package-data>`_.
```
abravalheri, line 387: Not sure if we need this note, because that was a bug in prior versions (so we don't really need to document bug behaviour).

We could instead use something like:

```suggestion
   Please make sure to use a recent version of setuptools, to avoid potential problems and bugs.
```

Or even omit it completely.

docs/userguide/datafiles.rst
DanielYang59: No problem, I noticed that! Thanks for the confirmation!

docs/userguide/datafiles.rst
DanielYang59: > If we are going for plain text, then it is probably better to use the full keywords

This one looks great! It's much better to avoid defining such custom notations for simplicity.

Meanwhile it looks like all capital keywords (AND/OR/NOT) would no be highlighted while and/or/not will, it is just my browser?

abravalheri: Well, please fell free to highlight or use any kind of emphasis (e.g. bold, italics, uppercase, etc) the way you think makes more sense 😅

docs/userguide/datafiles.rst
DanielYang59: Thanks for the input, let's remove it then. I was not aware of the background and thought we have to maintain backwards awareness (didn't know it was a buggy behaviour).

docs/userguide/datafiles.rst
DanielYang59: Of course, thanks for the advice!

docs/userguide/datafiles.rst
DanielYang59: > We are probably better off if we don't define data files in terms of source distributions or source trees

Sorry it looks like I might have caused confusion here. I was suggesting:
```diff
- When they install a distribution from PyPI or via a `.whl` file
+ When they install a distribution from source distribution or via a `.whl` file
```

Because `install from PyPI` is not important here, because by `install from PyPI`, we could install both from `.whl` file (if wheel is built for that particular platform) or source distribution. I hope I didn't misunderstand anything?

DanielYang59:
@abravalheri  I believe I have included everything I want to include, perhaps it's time to polish the language and check the correctness?

Some questions unrelated to this PR:
1. The following file name has spaces in it, is it intended or maybe rename it to a proper Unix name (escaping spaces looks quite messy)?: https://github.com/pypa/setuptools/blob/main/docs/python%202%20sunset.rst
2. I got the following notifications while building docs locally, perhaps update `conf.py` accordingly?
```
intersphinx inventory has moved: https://twine.readthedocs.io/en/stable//objects.inv -> https://twine.readthedocs.io/en/stable/objects.inv
inventory <https://build.pypa.io/en/latest> contains multiple definitions for std:label:python--m-build--v
intersphinx inventory has moved: https://packaging.python.org/en/latest//objects.inv -> https://packaging.python.org/en/latest/objects.inv
intersphinx inventory has moved: https://packaging.pypa.io/en/latest//objects.inv -> https://packaging.pypa.io/en/latest/objects.inv
```

docs/userguide/datafiles.rst
abravalheri: I think the former phrase is better for the definition than the second. Alternatively only the following is also fine:

```diff
- When they install a distribution from source distribution or via a `.whl` file
+ when they install a distribution via a `.whl` file
```

I believe We should not define define data files as per `source distribution`. A Python package always need to be transformed into a wheel, so the files inside of the wheel is what actually matters.

In the previous version `install from PyPI` is being used as an equivalent of `pip install package`, which in the end of the day downloads (preferentially) a `.whl` file.

docs/userguide/datafiles.rst
```diff
@@ -2,6 +2,16 @@
 Data Files Support
 ====================
 
+In the Python ecosystem, the term "data files" is used in various complex scenarios
+and can have nuanced meanings. For the purposes of this documentation,
+we define "data files" as non-Python files that are installed alongside Python
+modules and packages on the user's machine when they install a
+:term:`distribution <Distribution Package>` from either source distribution
+or via a binary distribution (``.whl`` file for example).
```
abravalheri, line 10:
```suggestion
:term:`distribution <Distribution Package>` via :term:`wheel <Wheel>`.
```

abravalheri:
Thank you very much

DanielYang59:
No problem, thanks a ton for guiding me through all these changes :) I learned a lot!

DanielYang59:
Cool! Thanks for letting me know. You're an amazing reviewer, thank you!

PR #4666, opened 2024-10-01
Use minimum requirement for jaraco.functools

### LINKED ISSUE #4665: [BUG] Minimum requirement for jaraco.functools
### setuptools version

75.1.0

### Python version

3.10

### OS

MacOS

_No response_

### Description

setuptools uses a function from `jaraco.functools` called `splat`:
https://github.com/pypa/setuptools/blob/v75.1.0/setuptools/_distutils/_modified.py#L6

That function is only available from version 4.x.x. So `jaraco.functools` should have a minimum requirement:
https://github.com/pypa/setuptools/blob/v75.1.0/pyproject.toml#L106

### Expected behavior

Should have a version conflict when installing setuptools in combination with other packages that require jaraco.functools < 4.

### How to Reproduce

Install setuptools with another package that require jaraco.functools < 4.

### Output

```console
File "/app/xxx.py", line 9, in <module>
  from distutils.util import strtobool
File "/poetry-env/lib/python3.10/site-packages/setuptools/_distutils/util.py", line 23, in <module>
  from ._modified import newer
File "/poetry-env/lib/python3.10/site-packages/setuptools/_distutils/_modified.py", line 6, in <module>
  from jaraco.functools import splat
ImportError: cannot import name 'splat' from 'jaraco.functools' (/poetry-env/lib/python3.10/site-packages/jaraco/functools.py)
```

anderstheet:
## Summary of changes

Adds a minimum requirement to jaraco.functools

Closes #4665

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ core = [
 
 	# for distutils
 	"jaraco.collections",
-	"jaraco.functools",
+	"jaraco.functools>=4",
 	"packaging",
 	"more_itertools",
 ]


pyproject.toml
```diff
@@ -103,7 +103,7 @@ core = [
 
 	# for distutils
 	"jaraco.collections",
-	"jaraco.functools",
+	"jaraco.functools>=4",
```
Avasam, line 106: I guess this should be changed here too, since that repo is where the vendored `distutils` (and hence, these deps) is coming from: https://github.com/pypa/distutils/blob/378984e02edae91d5f49425da8436f8dd9152b8a/pyproject.toml#L23

anderstheet: Isn't that what this changes? 🤔

Avasam: Normally, any change done in the `_distutils` folder should be done in the distutils repo *instead*.

I'm just not sure if pyproject dependencies follow the same rule (ie: are they refreshed from distutils or are they manually updated). So you could do the same change over there.

Anyway, maintainers will let you if that's the case.

anderstheet: Ahh makes sense.

I have created a pull request in the `distutils` repo 👍 
https://github.com/pypa/distutils/pull/301

jaraco (approved):

abravalheri:
Thank you very much!

PR #4691, opened 2024-10-17
Vendor distutils stubs

### LINKED ISSUE #4689: [FR] Vendor distutils stubs

### Alternative Solutions

### Additional context

### Code of Conduct

Avasam:
## Summary of changes
Note: Please consider #4704 as an alternative !

See commit history. Basically I copied `stdlib/distutils` from typeshed, then applied `stubs/setuptools/_distutils` from typeshed on top.
The following "fixups" have been brought back into typeshed and pypa/distutils, if you want we could wait for them to be merged and then I can rebase my changes here so reduce the amount of changes in my "fixups" commit:
- https://github.com/pypa/distutils/pull/303
- https://github.com/python/typeshed/pull/12839
- https://github.com/python/typeshed/pull/12841
- https://github.com/python/typeshed/pull/12842

I think these could go directly in https://github.com/pypa/distutils, vendoring them back into setuptools using the already existing process, to make it possibly easier to keep stubs and implementation in sync. But increasing the charge of pypa/distutils contributors. Do note it **has** to exist in a folder named `distutils` or `distutils-stubs`, preferably by itself.

Closes #4689

### Pull Request Checklist
- [x] Changes have tests (these fix tests going forward :D )
- [ ] News fragment added in [`newsfragments/`]. (no public facing changes as long as they're not published)

[PR docs]:


pyproject.toml
```diff
@@ -203,6 +203,8 @@ include-package-data = true
 include = [
 	"setuptools*",
 	"pkg_resources*",
+	# TODO: Include distutils stubs with package once we're confident in them
+	# "typings/distutils-stubs",
```
Avasam, line 207: We don't need to ship anything as long as typeshed provides stubs for `setuptools/_distutils`. We can synchronize once ready.
As I mentioned in the issue, there's also always the option to ship it as a different package 9in which case it could live completely separately from setuptools and pypa/distutils)

mypy.ini
```diff
@@ -18,6 +18,9 @@ disable_error_code =
 
 ## local
 
+# Use our custom stubs for distutils
+mypy_path = $MYPY_CONFIG_FILE_DIR/typings
```
Avasam, line 22: I used pyright's default of `typings` so that I didn't have to configure it there. But that folder name can be arbitrary.

Avasam:
Between #4704, #4861 and https://github.com/pypa/distutils/pull/329 , I don't see how this option is desirable anymore. The maintenance cost is just too high compared to the alternatives.

Avasam:
Between #4704, #4861 and https://github.com/pypa/distutils/pull/329 , I don't see how this option is desirable anymore. The maintenance cost is just too high compared to the alternatives.

PR #4704, opened 2024-10-25
Statically alias distutils-stubs to setuptools._distutils

### LINKED ISSUE #4689: [FR] Vendor distutils stubs

Avasam:
## Summary of changes

Closes https://github.com/pypa/setuptools/issues/4689
This is an alternative to https://github.com/pypa/setuptools/pull/4691. This makes type-checkers and editors see `distutils` as basically an alias to `setuptools.`

Advantage:
- Less maintenance work, these stubs can be autogenerated when updating the vendored stubs (if a file is removed or added)
- No need to keep in sync with typeshed
- Static typing follows directly pypa/distutils without concerns with stdlib/distutils compatibility

Disadvantage:
- Typing isn't as good or as complete as typeshed's (solved by https://github.com/pypa/distutils/pull/329)
- Typing improvements and fixes need to be done "in source" in pypa/distutils. But this is somewhat already the long-term plan (https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833 but directly in pypa/distutils)
- pypa/distutils could accidentally introduce a typing breaking change for setuptools, which probably needs to be ignored with a TODO when updating the vendor.

### Pull Request Checklist
- [x] Changes have tests (existing type-checking tests)
- [ ] News fragment added in [`newsfragments/`]. (no need until this is shipped with setuptools)

[PR docs]:


setuptools/__init__.py
```diff
@@ -228,7 +223,7 @@ def reinitialize_command(
     ) -> _Command:
         cmd = _Command.reinitialize_command(self, command, reinit_subcommands)
         vars(cmd).update(kw)
-        return cmd
+        return cmd  # pyright: ignore[reportReturnType] # pypa/distutils#307
```
Avasam, line 226: Ref.: https://github.com/pypa/distutils/pull/307

Avasam:
I've added a comment at the top of geenrated files to make it clear that they are generated and how to regenerate them.

setuptools/command/install.py
```diff
@@ -180,4 +180,5 @@ def do_egg_install(self) -> None:
 # XXX Python 3.1 doesn't see _nc if this is inside the class
 install.sub_commands = [
     cmd for cmd in orig.install.sub_commands if cmd[0] not in install._nc
-] + install.new_commands
+] + install.new_commands  # type: ignore[operator]
+# TODO: Type sub_commands/new_commands to avoid variance issues in pypa/distutils (like python/typeshed#11951)
```
Avasam, line 184: https://github.com/pypa/distutils/pull/318

setuptools/command/sdist.py
```diff
@@ -50,10 +50,12 @@ class sdist(orig.sdist):
...
 
     distribution: Distribution  # override distutils.dist.Distribution with setuptools.dist.Distribution
-    negative_opt: ClassVar[dict[str, str]] = {}
+    # TODO: Mark class-level mutables as ClassVars in pypa/distutils (like python/typeshed#12403)
+    negative_opt: ClassVar[dict[str, str]] = {}  # type: ignore[misc]
 
     README_EXTENSIONS = ['', '.rst', '.txt', '.md']
-    READMES = tuple('README{0}'.format(ext) for ext in README_EXTENSIONS)
+    # TODO: Mark some class-level tuples as as not fixed-length pypa/distutils (like python/typeshed#12403)
+    READMES = tuple('README{0}'.format(ext) for ext in README_EXTENSIONS)  # type: ignore[assignment]
```
Avasam, line 58: https://github.com/pypa/distutils/pull/319

setuptools/tests/test_build_ext.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
Avasam, line 2:
```suggestion
```

Avasam:
I'm currently putting my efforts towards https://github.com/pypa/setuptools/pull/4861 as a solution. Which does the same thing, but autogenerates the stubs instead of having them being checked in.

Issue #4629, opened 2024-09-03
[FR] Implement PEP 639


befeleme:
### What's the problem this feature will solve?

The chaotic land of expressing license metadata in Python packaging will become more unified.

### Describe the solution you'd like

See: https://peps.python.org/pep-0639
This will update the Core Metadata version to 2.4, so the previous features, updating them to 2.2 and 2.3, should land in first.

### Additional context

A previous draft of the PEP has been already implemented in setuptools.
There's a feature request for another bit of the draft: https://github.com/pypa/setuptools/issues/3596
The full implementation, including the Core Metadata bump, will require also the support on the side of PyPI first: https://github.com/pypi/warehouse/issues/16620

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

abravalheri:
Thank you very much @befeleme.

Also possibly pending on https://github.com/pypa/packaging/pull/799 (unless we skip the validation for now).

Any member of the community interested in giving it a go is very welcome!

abravalheri:
@befeleme one thing that I was also discussing in the discourse (see https://discuss.python.org/t/enforcing-consistent-metadata-for-packages/50008/101 and https://discuss.python.org/t/enforcing-consistent-metadata-for-packages/50008/102) is that I don't believe setuptools is in a condition to adopt version 2.2 of the core metadata, specifically `Dynamic`.

As discussed in that thread, I don't think we can right now ensure 100% that `Dynamic` will work as specified in [PEP 643](https://peps.python.org/pep-0643/). And I also believe that changing setuptools such that we become able to implement that is not trivial, likely causing lots of disruption for existing packages.

So that is why we cannot "bump" into 2.2, and therefore we cannot bump into 2.4 either.

If that is OK, we can adopt PEP 639 without bumping the metadata to 2.4. Or bump the metadata to 2.4 and use `Dynamic` only as "best-effort" indicator, without providing the guarantees that the PEP 643 wants us to enforce.

/cc @pfmoore @jaraco

---

One example in which setuptools would have trouble to provide guarantees is when a package has `pyproject.toml` without `[project]` and uses a plugin that overwrites metadata. Another situation is if an user adds custom command classes that write `PKG-INFO` or `METADATA` (like `egg-info`, `sdist`, `bdist_wheel`, `editable_wheel`) - this way they have complete control over the contents of the file.

In general `setuptools` can provide a good indicator if a field is filled dynamically or not, but I don't think we can "guarantee" it.

[...]

Issue #3596, opened 2022-09-19
[FR] Support License-File root license directory (PEP 639)


hroncok:
### What's the problem this feature will solve?

The License-File core metadata tag definition in https://peps.python.org/pep-0639/ was updated. Specifically:

> Root License Directory (Short: License Directory)
> The directory under which license files are stored in a project/distribution
> and the root directory that their paths, as recorded under the License-File
> core metadata fields, are relative to. Defined here to be the project root
> directory for source trees and source distributions, and a subdirectory
> named licenses of the directory containing the core metadata
> (i.e., the .dist-info/licenses directory) for built distributions and
> installed projects.

And:

> As specified by this PEP, [License-File] value is also that file’s path
> relative to the root license directory in [...] installed projects [...]

Translated to human language, when the `License-File` looks like this:

```
License-File: LICENSE
```

The file should be located in `...dist-info/licenses/LICENSE`. Currently, it is located in `...dist-info/LICENSE` directly, based on a previous iteration of this PEP (which is til a draft).

### Describe the solution you'd like

When including license files in dist-info, move them to the `licenses` directory.

### Alternative Solutions

An alternate solution is that a tool that reads the License-File core metadata tag should look for the files in both locations.

### Additional context

hatchling already follows this and includes the license files in the licenses directory.

Thank you :heart: 

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

abravalheri:
Thank you for the ticket @hroncok.

My doubt here is if we should wait until PEP 639 is finalised instead of risking implementing it again and having to change later.

This issue also might require coordination with the `wheel` package.

befeleme:
> My doubt here is if we should wait until PEP 639 is finalised instead of risking implementing it again and having to change later.

PEP 639 has been provisionally accepted (with one of the conditions being implementation in at least two build backends :pleading_face:), so this issue now describes the solution agreed upon.

abravalheri:
Thank you very much @befeleme.

If any member of the community is interested in giving it a go, PRs are very welcome.

webknjaz:
UPD: PEP 639 has been accepted in full.

PR #4706, opened 2024-10-28
Add initial support for License-Expression (PEP 639)

cdce8p:
## Summary of changes
- Update `validate_pyproject` to `0.22`
https://github.com/abravalheri/validate-pyproject/blob/v0.22/CHANGELOG.rst
- Initial support for `License-Expression` core metadata

```diff
 [project]
 name = "spam"
 ...
-license = {text = "MIT"}
+license = "MIT"
```

**Missing**: The license expression isn't yet validated. The next version of `validate_pyproject` will likely add that once a new version of packaging has been released. From the PEP:
> Build and publishing tools SHOULD check that the License-Expression field contains a valid SPDX expression, including the validity of the particular license identifiers (as [defined above](https://peps.python.org/pep-0639/#spdx)).

It should be fine to omit that part for the moment.

Refs: #4629


abravalheri:
Hi @cdce8p thank you very much for having a look on this.

We probably have to organise so that PEP 639 lands after PEP 685 implementation (ideally we want a couple of days/weeks between the two so that users have time to report potential bugs).

setuptools/_distutils/dist.py
```diff
@@ -1221,6 +1223,11 @@ def get_license(self):
 
     get_licence = get_license
 
+    def get_license_expression(self):
+        return self.license_expression
+
+    get_license_expression = get_license_expression
+
```
abravalheri, line 1230: Would it be possible to implement this feature without changing the `_distutils` folder?

In general, we don't modify the contents to that folder directly. Instead, the code is maintained in the `pypa/distutils` repository and merged back in setuptools periodically,  ([see also our development guidelines](https://setuptools.pypa.io/en/latest/development/developer-guide.html#code-conventions-and-other-practices)).

cdce8p: > Would it be possible to implement this feature without changing the `_distutils` folder?

Yes. Removed those changes as they weren't necessary to begin with. I added them initially but the better solution was to add `license_expression` to `_DISTUTILS_UNSUPPORTED_METADATA` in `setuptools/dist.py`. This is also fully compatible with `SETUPTOOLS_USE_DISTUTILS=stdlib`. https://github.com/pypa/setuptools/pull/4706/commits/390b95f2b904bd75df48fa64e6370bedbfb63457

cdce8p:
> We probably have to organise so that PEP 639 lands after PEP 685 implementation (ideally we want a couple of days/weeks between the two so that users have time to report potential bugs).

Maybe I'm missing something obvious but I'm not quite sure how [PEP 685](https://peps.python.org/pep-0685/) is related. Would you mind explaining it a bit more?

The way I've designed the PR, it shouldn't depend on anything. It just allows users to specify `license = MIT` in `pyproject.toml` which will be used for the `License-Expression` field. Note, the metadata version is still `2.2` (not `2.4` which is the first to "require" support for PEP 639).

cdce8p:
Opened #4734 with just the `validate-pyproject` update. This will make the changes here easier to review.

abravalheri:
> Maybe I'm missing something obvious but I'm not quite sure how [PEP 685](https://peps.python.org/pep-0685/) is related. Would you mind explaining it a bit more?

The problem is that the metadata version of the core metadata is incremental:
- PEP 685 stablishes 2.2
- PEP 639 stablishes 2.3

So we cannot implement PEP 639 before 685, otherwise metadata validation may fail.

Even if the PR is independent we cannot merge it now because of the [nature of the releases of setuptools](https://setuptools.pypa.io/en/latest/development/releases.html) (we cannot risk PEP 639 support being released before PEP 685 support).

cdce8p:
> So we cannot implement PEP 639 before 685, otherwise metadata validation may fail.

The validation will only happen once we update the metadata version.
https://github.com/pypa/packaging/blob/24.2/src/packaging/metadata.py#L826-L831

If not, a lot of wheel uploads would probably fail as we currently include `License-File` metadata that's not up to spec. https://github.com/pypa/setuptools/issues/3596 (Fix: https://github.com/pypa/setuptools/pull/4728).

What I'm proposing here is simply to add the basic support to setuptools. It won't get validated by PyPI until we're ready to move to `2.4` but then all projects using the "new" syntax for `pyproject.license` will already be ready to go.

abravalheri:
Sorry, I have been using the wrong PEP number, PEP 685 is important, but the really big one for us to implement before 639 is [PEP 643](https://peps.python.org/pep-0643/) (the one with `Dynamic`).

cdce8p:
> So for the sake of reducing the unknown risks, I propose we pipeline the changes you kindly contributed after the implementation of 685. Unfortunately, it means that people cannot start modifying their `pyproject.toml` immediately, but it feels safer.

I understand your reasoning. One last suggestion. The last days I've been looking into how `hatchling` does it as I noticed it was one of the earliest adopters of `License-Expression` in the metadata. Turns out the packaging update caused upload failures https://github.com/pypa/hatch/issues/1786. I can't reproduce it with PyPI at the moment. However, `packaging==24.2` rejects any `License-Expression` and `License-File` keys with metadata `<2.4`. That would be relevant for us too. After the issue report, they removed the PEP 639 keys from the metadata and now include the `project.license` as part of the `License` key.

**My suggestion**
- Remove the `License-File` keys from the generate metadata as this might cause upload failures once the new packaging version is rolled out to PyPI.
- Keep the license files in the `.dist-info` folder but move them to the `licenses` subfolder in according to the spec. Essentially #4728
- Add support for `project.license` by back-populating the `License` key, similar to https://github.com/pypa/hatch/pull/1805 Together with #4734, this would allow users to use the new syntax while keeping the old metadata format.

cdce8p:
@abravalheri With the last updates, I removed the `License-Expression` field and instead back-populate the `License` field with the validated license expression. That's the approach currently taken by [hatchling](https://github.com/pypa/hatch/releases/tag/hatchling-v1.26.2). Since it doesn't add any new fields, would it be ok to do now?

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -181,10 +182,12 @@ def _long_description(
...
 
 
-def _license(dist: Distribution, val: dict, root_dir: StrPath | None):
+def _license(dist: Distribution, val: str | dict, root_dir: StrPath | None):
     from setuptools.config import expand
 
-    if "file" in val:
+    if isinstance(val, str):
+        _set_config(dist, "license_expression", _static.Str(val))
+    elif "file" in val:
```
abravalheri, line 190: @cdce8p I updated this part of the code to solve the merge conflict. I hope that is OK.

cdce8p: Thanks. Feel free to ping me, especially if one of my PRs need attention. I'm usually fairly quick to respond.

_Small side note. If I review PRs from other people, I only update them with merge commits. The rebase one makes it unnecessarily difficult for the OP to follow and understand what's changed on their own PR. I myself only do force pushes if my PR hasn't been reviewed yet, afterwards only merge commits._

If that's ok with you, I'll force push the rebase once #4796 is done. Just so I'm aware what has changed.

abravalheri: > If that's ok with you, I'll force push the rebase once https://github.com/pypa/setuptools/pull/4796 is done. Just so I'm aware what has changed.

Thanks @cdce8p. That would be helpful.
I usually try to avoid merging the full main branch back to a PR, because it messes up the graph visualisation (which becomes difficult to follow), that is why I went with a rebase, sorry if that disrupted your workflow.

cdce8p: > I usually try to avoid merging the full main branch back to a PR, because it messes up the graph visualisation (which becomes difficult to follow), that is why I went with a rebase, sorry if that disrupted your workflow.

Don't worry about it. Guess everyone has their own preferred workflow. Just wrote that as it took me a bit to actually see what you've changed during the rebase.

> Thanks @cdce8p. That would be helpful.

👍🏻

pyproject.toml
```diff
@@ -11,10 +11,10 @@ authors = [
 ]
 description = "Easily download, build, install, upgrade, and uninstall Python packages"
 readme = "README.rst"
+license = "MIT"
```
abravalheri, line 14: This is a bit complicated.

Setuptools has many vendored dependencies (which may vary with time), so the correct license expression is not that simple. Additionally, I don't know if it is going to be trivial to derive a correct expression automatically from the code base at each release (e.g. what happens if a maintainer change a vendored dependency? how can we guarantee the license expression is up-to-date?)...

Would it make more sense to simply omit the license expression and rely on `license-files` or an automatic scan of the relevant directories?

setuptools/_core_metadata.py
```diff
@@ -88,6 +88,7 @@ def read_pkg_file(self, file):
     self.url = _read_field_from_msg(msg, 'home-page')
     self.download_url = _read_field_from_msg(msg, 'download-url')
     self.license = _read_field_unescaped_from_msg(msg, 'license')
+    self.license_expression = _read_field_unescaped_from_msg(msg, 'license_expression')
```
abravalheri, line 91: Should this use an `_` (underscore) or `-` (dash) for the message field string?

setuptools/config/_apply_pyprojecttoml.py
```diff
@@ -181,16 +182,19 @@ def _long_description(
...
+        _set_config(dist, "license_expression", _static.Str(val))
     else:
-        _set_config(dist, "license", _static.Str(val["text"]))
+        if "file" in val:
+            # XXX: Is it completely safe to assume static?
+            value = expand.read_files([val["file"]], root_dir)
+            _set_config(dist, "license", _static.Str(value))
+            dist._referenced_files.add(val["file"])
+        else:
+            _set_config(dist, "license", _static.Str(val["text"]))
```
abravalheri, line 197: I would prefer to flatten this nested `else{if, else}` into a `elif,else` as in the [suggested change](https://github.com/pypa/setuptools/compare/97e8f94e24a57a446dee965f71532acacf7354d9..f3a1fdddf62fc61195b9a5e2ed8c66f4b4613238#diff-3c50c435ad8328e83caa459b99929ae4273860772b44c14f49039927636c5c04L182):

image

setuptools/config/_validate_pyproject/extra_validations.py
```diff
@@ -24,6 +24,13 @@ class RedefiningStaticFieldAsDynamic(ValidationError):
     )
 
 
+class IncludedDependencyGroupMustExist(ValidationError):
+    _DESC = """An included dependency group must exist and must not be cyclic.
+    """
+    __doc__ = _DESC
+    _URL = "https://peps.python.org/pep-0735/"
+
+
```
abravalheri, line 33: I am supposing the reason for these files to be changed is that this PR is cascaded on top of https://github.com/pypa/setuptools/pull/4734. So ideally we should merge that other one first and possibly rebase. I have changed the PR to target the [`feature/pep639`](https://github.com/pypa/setuptools/tree/feature/pep639) branch so that we can do that without risking an intermediary/incomplete release.

setuptools/dist.py
```diff
@@ -402,6 +405,23 @@ def _normalize_requires(self):
...
+                InformationOnly.emit(f"Normalizing '{license_expr}' to '{normalized}'")
+                self.metadata.license_expression = normalized
+
+            for cl in self.metadata.get_classifiers():
+                if not cl.startswith("License :: "):
+                    continue
+                raise InvalidConfigError(
+                    "License classifier are deprecated in favor of the license expression. "
+                    f"Remove the '{cl}' classifier."
+                )
```
abravalheri, line 423: Because it is a deprecation, should we give a `warning` first before evolving to an error? (we can skip the deprecated classifier in the meantime).

setuptools/_core_metadata.py
```diff
@@ -175,7 +176,7 @@ def write_field(key, value):
         if attr_val is not None:
             write_field(field, attr_val)
 
-    license = self.get_license()
+    license = self.license_expression or self.get_license()
```
abravalheri, line 179: So this mean that we need a follow-up PR for the `License-Expression`, right?

pyproject.toml
cdce8p: Let me omit it for now. I mainly used `MIT` here as that matches the existing classifier `License :: OSI Approved :: MIT License` but I'm aware of the "vendor" dilemma. This is something which can be discussed at a later point.

**IMNAL**: During the discussion on the Python packaging docs I made the argument that vendoring, if done correctly, usually doesn't change the project license as long as all vendored dependencies are included and licenses correctly.

https://github.com/pypa/packaging.python.org/pull/1662#discussion_r1880319370
https://discuss.python.org/t/pep-639-improving-license-clarity-with-better-package-metadata/2154/39

setuptools/config/_apply_pyprojecttoml.py
cdce8p: I thought about this as well. However I decided to use to nesting to clearly separate the before / after PEP 639 cases. In a flattened list it just looks like they are equal which isn't the case.

At some point, it might also make sense to add a DeprecationWarning for the legacy case which will be easier with the nested style.

setuptools/_core_metadata.py
cdce8p: Yes, it should be `-`. Will update it.

setuptools/config/_validate_pyproject/extra_validations.py
cdce8p: > I am supposing the reason for these files to be changed is that this PR is cascaded on top of #4734.

Correct. We need https://github.com/abravalheri/validate-pyproject/pull/210 and https://github.com/abravalheri/validate-pyproject/pull/217 to be able to parse and validate `project.license` correctly. They are only available in `v0.23.0`.

> So ideally we should merge that other one first and possibly rebase.

I created the other PR as I thought it might be easier to review. "Just a simple vendored dependency bump". It's identical to the first commit here. (A rebase will resolve it).

> I have changed the PR to target the [`feature/pep639`](https://github.com/pypa/setuptools/tree/feature/pep639) branch so that we can do that without risking an intermediary/incomplete release.

Not sure that's necessary. Will leave a comment about it one the issue to keep it in a central place: https://github.com/pypa/setuptools/issues/4759#issuecomment-2659761258

setuptools/dist.py
cdce8p: I'd be open to that as well. My reasoning for the error was that the users isareupdating the metadata anyway. Might as well error then. Deprecations tend to get unnoticed unfortunately, especially with CD pipelines.

setuptools/_core_metadata.py
cdce8p: Correct. This PR was designed so that it itself didn't require a metadata_version bump. That could / should happen at a later point once all pieces are in place. I.e. setuptools has adopted version `2.3`, the license files are stored in the correct folder and license expression parsing is handled correctly.

abravalheri:
Thank you very much @cdce8p. How do you feel about rebasing this on the [`feature/pep639 `](https://github.com/pypa/setuptools/tree/feature/pep639) branch?

abravalheri:
Problem with cygwin in the CI possibly solved by https://github.com/pypa/setuptools/pull/4832

setuptools/dist.py
```diff
@@ -402,6 +404,31 @@ def _normalize_requires(self):
...
+                    continue
+                license_classifiers_found = True
+                SetuptoolsDeprecationWarning.emit(
+                    "License classifier are deprecated in favor of the license expression.",
+                    f"Please remove the '{cl}' classifier.",
+                    see_url="https://peps.python.org/pep-0639/",
+                    due_date=(2027, 2, 17),  # Introduced 2025-02-17
+                )
+            if license_classifiers_found:
+                self.metadata.set_classifiers(classifiers)
```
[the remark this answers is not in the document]
cdce8p: Tbh I don't think setuptools should modify the classifiers in the first place. We should either emit an error or a warning to the user to do so themselves. Everything else would just be surprising in the end. Pushed a new commit to revert that part.

> (It might also be the case your intention is to introduce a more timid version of the skip/warning in this PR, and later un-nest it and make it more prominent in a follow up. Please let me know if that is the case).

At some point it might make sense to add another warning to nudge users to use license expressions. However, I do not plan to add it at this time.

PR #4707, opened 2024-10-28
Merge TypedDicts from typeshed

Avasam:
## Summary of changes

Follow-up to #4504, works towards #2345
I didn't include TypeDicts for `**kwargs` params as part of this PR. keyword args will be its own thing.

### Pull Request Checklist
- [x] Changes have tests (existing type-checking and runtime tests)
- [x] News fragment added in [`newsfragments/`]. (not public facing yet)

[PR docs]:

### DIFF
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -34,7 +34,7 @@ import zipimport
 from collections.abc import Iterable
 from glob import glob
 from sysconfig import get_path
-from typing import TYPE_CHECKING, Callable, NoReturn, TypeVar
+from typing import TYPE_CHECKING, Callable, NoReturn, TypedDict, TypeVar
 
 from jaraco.text import yield_lines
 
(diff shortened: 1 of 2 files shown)


setuptools/command/easy_install.py
```diff
@@ -2038,14 +2038,19 @@ def chmod(path, mode):
         log.debug("chmod failed: %s", e)
 
 
+class SplitArgs(TypedDict, total=False):
```
abravalheri, line 2041: Should we prefer introducing private terms?

abravalheri (approved):
Thank you very much!

setuptools/command/easy_install.py
Avasam: Is it intended for devs to assign to `split_args` at any point (and I mean re-assigning to the attribute directly, not mutating the dictionary)? If not, it could be made private. (the same reasoning can be applied to TypeDict used solely in args vs return types)

abravalheri: Ideally users should not be importing more symbols in general 😅, and more so from deprecated modules.

Hopefully devs will not have to assign new values, but if they do I think they should simply use a literal dict with the correct keys. Mypy should be able to cope with that right?

Avasam: Yes using a dict literal to assign to a typeddict should be fine. The module being deprecated is even more reason. I'll make them both private.
If a user really needs to they can still import the private symbol for reference.

PR #4709, opened 2024-10-28
Re-enable ANN2 for setuptools

Avasam:
## Summary of changes

Follow-up to #4504, works towards #2345
I didn't include long/complex overloads in this PR, that may come later

This has some overlap with #4711, but they can be merged in either order.

### Pull Request Checklist
- [x] Changes have tests (existing type-checking and runtime tests)
- [x] News fragment added in [`newsfragments/`]. (no public facing changes)

[PR docs]:

### DIFF
--- a/setuptools/__init__.py
+++ b/setuptools/__init__.py
@@ -186,7 +186,7 @@ class Command(_Command):
             )
         return val
 
-    def ensure_string_list(self, option: str):
+    def ensure_string_list(self, option: str) -> None:
         r"""Ensure that 'option' is a list of strings.  If 'option' is
         currently a string, we split it either on /,\s*/ or /\s+/, so
         "foo bar baz", "foo,bar,baz", and "foo,   bar baz" all become
@@ -226,7 +226,7 @@ class Command(_Command):
     ) -> _Command:
         cmd = _Command.reinitialize_command(self, command, reinit_subcommands)
         vars(cmd).update(kw)
-        return cmd
+        return cmd  # pyright: ignore[reportReturnType] # pypa/distutils#307
 
     @abstractmethod
     def initialize_options(self) -> None:
--- a/setuptools/command/bdist_egg.py
+++ b/setuptools/command/bdist_egg.py
@@ -16,7 +16,7 @@ from typing import TYPE_CHECKING, Literal
 from setuptools import Command
 from setuptools.extension import Library
 
-from .._path import ensure_directory
+from .._path import StrPathT, ensure_directory
 
 from distutils import log
 from distutils.dir_util import mkpath, remove_tree
@@ -440,13 +440,13 @@ INSTALL_DIRECTORY_ATTRS = ['install_lib', 'install_dir', 'install_data', 'instal
 
 
 def make_zipfile(
-    zip_filename,
+    zip_filename: StrPathT,
     base_dir,
     verbose: bool = False,
     dry_run: bool = False,
     compress=True,
     mode: _ZipFileMode = 'w',
-):
+) -> StrPathT:
     """Create a zip file from all the files under 'base_dir'.  The output
     zip file will be named 'base_dir' + ".zip".  Uses either the "zipfile"
     Python module (if available) or the InfoZIP "zip" utility (if installed
@@ -455,7 +455,7 @@ def make_zipfile(
     """
     import zipfile
 
-    mkpath(os.path.dirname(zip_filename), dry_run=dry_run)
+    mkpath(os.path.dirname(zip_filename), dry_run=dry_run)  # type: ignore[arg-type] # python/mypy#18075
     log.info("creating '%s' and adding '%s' to it", zip_filename, base_dir)
 
     def visit(z, dirname, names):
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -12,8 +12,7 @@ from typing import Iterable, Iterator
 
 from more_itertools import unique_everseen
 
-from setuptools._path import StrPath
-
+from .._path import StrPath, StrPathT
 from ..dist import Distribution
 from ..warnings import SetuptoolsDeprecationWarning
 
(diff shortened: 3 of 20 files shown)


setuptools/__init__.py
```diff
@@ -226,7 +226,7 @@ def reinitialize_command(
     ) -> _Command:
         cmd = _Command.reinitialize_command(self, command, reinit_subcommands)
         vars(cmd).update(kw)
-        return cmd
+        return cmd  # pyright: ignore[reportReturnType] # pypa/distutils#307
```
Avasam, line 229: Ref.: https://github.com/pypa/distutils/issues/307

setuptools/command/build_py.py
```diff
@@ -50,20 +49,20 @@ def finalize_options(self):
...
         level: object = 1,
-    ):
+    ) -> tuple[StrPathT | str, bool]:
         # Overwrite base class to allow using links
         if link:
             infile = str(Path(infile).resolve())
-            outfile = str(Path(outfile).resolve())
-        return super().copy_file(
+            outfile = str(Path(outfile).resolve())  # type: ignore[assignment] # Re-assigning a str when outfile is StrPath is ok
+        return super().copy_file(  # pyright: ignore[reportReturnType] # pypa/distutils#309
```
Avasam, line 65: Ref.: https://github.com/pypa/distutils/issues/309

setuptools/glob.py
```diff
@@ -129,24 +149,24 @@ def _rlistdir(dirname):
...
 
-def has_magic(s):
+def has_magic(s: str | bytes) -> bool:
     if isinstance(s, bytes):
-        match = magic_check_bytes.search(s)
+        return magic_check_bytes.search(s) is not None
     else:
-        match = magic_check.search(s)
-    return match is not None
+        return magic_check.search(s) is not None
```
Avasam, line 166: mypy will declare the first `match` as typed `Math[bytes] | None`, so the second assignement doesn't respect that typing, this is the same issue as https://github.com/python/mypy/issues/10736 AFAICT

Either the variable needs to be explicitely typed as `Match[bytes] | Match[str] | None`, or we simply don't assign.

setuptools/depends.py
```diff
@@ -51,7 +52,9 @@ def version_ok(self, version):
             and self.format(version) >= self.requested_version
         )
 
-    def get_version(self, paths=None, default: str = "unknown"):
+    def get_version(
+        self, paths=None, default: str | int = "unknown"
+    ) -> str | int | None:
```
Avasam, line 57: `get_version`, `get_module_constant`, and `extract_constant` *can* return `Any`, but I'm not sure that's within intended usage ?

setuptools/command/bdist_egg.py
```diff
@@ -455,7 +455,7 @@ def make_zipfile(
     """
     import zipfile
 
-    mkpath(os.path.dirname(zip_filename), dry_run=dry_run)
+    mkpath(os.path.dirname(zip_filename), dry_run=dry_run)  # type: ignore[arg-type] # python/mypy#18075
```
Avasam, line 458: Ref: https://github.com/python/mypy/issues/18075

setuptools/command/build_py.py
```diff
@@ -50,20 +49,20 @@ def finalize_options(self):
...
         link: str | None = None,
         level: object = 1,
-    ):
+    ) -> tuple[StrPathT | str, bool]:
         # Overwrite base class to allow using links
         if link:
             infile = str(Path(infile).resolve())
-            outfile = str(Path(outfile).resolve())
-        return super().copy_file(
+            outfile = str(Path(outfile).resolve())  # type: ignore[assignment] # Re-assigning a str when outfile is StrPath is ok
```
Avasam, line 64: This may or may not also be related to https://github.com/python/mypy/issues/18075

setuptools/depends.py
abravalheri: Yeah, this module is a bit annoying...

It looks very "vestigial", not very used in the setuptools code-base itself. But is exported as part of the API.

`get_module_constant` and `extract_constant` seem to be very likely returning something compatible with the return values of `ast.parse_literal` and possibly other objects that can be disassembled directly from Python bytecode. But I guess `Any` is the best thing we can do about that.

`get_version` is returning `None | Literal["default"] | T` if `self.format` is defined as `Callable[..., T]` or `Any` otherwise. But I don't think it is worth to add any form of validation here, we want to avoid breaking this API (potentially it is a good candidate for deprecation or moving to a different 3rd-party package).

Avasam: Ok so essentially `Any`, but I changed it to at least give some hints.

setuptools/dist.py
```diff
@@ -52,6 +51,9 @@
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
 
+    from pkg_resources import Distribution as _pkg_resources_Distribution
```
abravalheri, line 54: It is good this is inside a `if TYPE_CHECKING` block. Thank you!

Ideally we want to avoid importing `pkg_resources` in runtime because of the performance hit.

It is only acceptable to "lazily" import `pkg_resources` on deprecated code paths (but never eagerly).

abravalheri:
I don't know if it was a false negative but apparently after merging this in top of the previous merges we have a `mypy` error:

image
https://github.com/pypa/setuptools/actions/runs/11617414532/job/32352587471#step:9:71

Maybe there is a new version of `mypy`?

@Avasam, do you have any info?

Avasam:
> I don't know if it was a false negative but apparently after merging this in top of the previous merges we have a `mypy` error:
>
> […]
> 
> @Avasam, do you have any info? 

I can take a look at it tomorrow. I'm omw to a Halloween party 🎃

If needed for other PRs you can suppress the error temporarily (or suppress for entire file if it only happens on one side of the 3.12 distutils boundary)

PR #4711, opened 2024-10-28
setuptools ANN201 autofixes for fully untyped functions

Avasam:
## Summary of changes

I commented `[lint.per-file-ignores] "setuptools/**" = ["ANN2"]` and `[lint.flake8-annotations] ignore-fully-untyped = true` in `ruff.toml` then ran `ruff check setuptools --select=ANN201 --fix --unsafe-fixes`.
I removed the added return types that were revealing hidden mypy issues (often `initialize_options`) to keep the changes here completely automated.

This has some overlap with #4709, but they can be merged in either order.

### Pull Request Checklist
- [x] Changes have tests (this increased the amount of type-checked code)
- [x] News fragment added in [`newsfragments/`]. (no public facing changes)

[PR docs]:

### DIFF
--- a/setuptools/config/setupcfg.py
+++ b/setuptools/config/setupcfg.py
@@ -477,7 +477,7 @@ class ConfigHandler(Generic[Target]):
         parser = (lambda _, v: values_parser(v)) if values_parser else (lambda _, v: v)
         return cls._parse_section_to_dict_with_key(section_options, parser)
 
-    def parse_section(self, section_options):
+    def parse_section(self, section_options) -> None:
         """Parses configuration file section.
 
         :param dict section_options:
@@ -713,7 +713,7 @@ class ConfigOptionsHandler(ConfigHandler["Distribution"]):
 
         return find_kwargs
 
-    def parse_section_entry_points(self, section_options):
+    def parse_section_entry_points(self, section_options) -> None:
         """Parses `entry_points` configuration file section.
 
         :param dict section_options:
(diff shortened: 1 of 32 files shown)


setuptools/config/setupcfg.py
```diff
@@ -280,7 +281,7 @@ def _section_options(
             yield name.lstrip('.'), value
 
     @property
-    def parsers(self):
+    def parsers(self) -> NoReturn:
```
abravalheri, line 284: Ok, isn't this going to make more complicated inheritance and require some form of overload annotation later?

It is kind of an abstract method that needs to be implemented.
The expected signature is something like `() -> dict[str, Callable[[str], Any]]` or more specific[^1].
I suppose that for the sake of documentation/understanding that return value as a `dict` is better than `NoReturn`.

[^1]: The return values for the callable are probably something like all the types that can be produced by `ast.parse_literal` (or subtypes) + `packaging.specifiers.SpecifierSet`

abravalheri (approved):

setuptools/config/setupcfg.py
Avasam: `NoReturn` is a bottom type, so as far as subclassing it shouldn't be an issue.

For consumers it means that if something returns a `ConfigHandler`, they'd see accessing the property `parsers` as never returning (unless they use `isinstance` to ensure a subtype or something).

So you are correct that 
> I suppose that for the sake of documentation/understanding that return value as a `dict` is better than `NoReturn`.

Even better, this property should be marked abstract, so type-checkers tells the user to never instanciate `ConfigHandler`, which implies that any `ConfigHandler` the user gets should be a concrete subclass, so returning `NoReturn` for a `NotImplementedError` is no longer relevant.

But that's for a separate PR ! (I'll just revert this line here)

abravalheri:
Thank you very much.

PR #4728, opened 2024-11-03
Store license-files in licenses subfolder

### LINKED ISSUE #3596: [FR] Support License-File root license directory (PEP 639)

cdce8p:
## Summary of changes
The initial draft of [PEP 639](https://peps.python.org/pep-0639/) specified that license files should be written directly to the `.dist-info` folder. This was changed in later iterations. The files should now be written to `.dist-info/licenses` and the root folder structure should be recreated to avoid name conflicts.

Also added support for recursive `license_files` specifier, e.g. `**/LICENSE`. Those are [allowed](https://peps.python.org/pep-0639/#add-license-files-key) according to the PEP.

Closes #3596
Refs #4629


setuptools/dist.py
```diff
@@ -418,7 +418,10 @@ def _finalize_license_files(self) -> None:
             patterns = ['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']
 
         self.metadata.license_files = list(
-            unique_everseen(self._expand_patterns(patterns))
+            map(
+                lambda path: path.replace("\\", "/"),
```
abravalheri, line 422: Should we use `os.sep` here instead of `\\` (e.g. if somehow `\\` ends showing up on a linux build, we want the build to crash right, because it would mean incorrect input from the user)?

setuptools/tests/test_bdist_wheel.py
```diff
@@ -311,7 +331,7 @@ def test_licenses_deprecated(dummy_dist, monkeypatch, tmp_path):
     bdist_wheel_cmd(bdist_dir=str(tmp_path)).run()
 
     with ZipFile("dist/dummy_dist-1.0-py3-none-any.whl") as wf:
-        license_files = {"dummy_dist-1.0.dist-info/DUMMYFILE"}
+        license_files = {"dummy_dist-1.0.dist-info/licenses/licenses/DUMMYFILE"}
```
abravalheri, line 334: May be worth to add a comment with a note for future maintainers along the lines of:

> The file structure already has a `licenses` folder. PEP 639 adds an extra `licenses` folder inside the metadata directory.

Or change the example to use a different directory name all together when creating the test project directory (clearer approach, but more work).

setuptools/dist.py
cdce8p: I don't believe that's necessary here. `path` isn't a user input but rather the glob matched license path from `self._expand_patterns`. The idea with this line is just to normalize the separator since the core metadata field only allows forward `/`.

cdce8p: Not sure I've ever seen that in the wild but you're right it's a possibility. Pushed https://github.com/pypa/setuptools/pull/4728/commits/3e9b9c7fbe4b753bbc80ee8be01c9a8a9233970d to use `os.sep`.

abravalheri:
Thank you!

PR #4729, opened 2024-11-04
Fix formatting and imports for Python 3.9

cdce8p:
## Summary of changes
Small PR to fix CI tests.

### DIFF
--- a/setuptools/_importlib.py
+++ b/setuptools/_importlib.py
@@ -1,12 +1,7 @@
+import importlib.resources as resources  # noqa: F401
 import sys
 
 if sys.version_info < (3, 10):
     import importlib_metadata as metadata  # pragma: no cover
 else:
     import importlib.metadata as metadata  # noqa: F401
-
-
-if sys.version_info < (3, 9):
-    import importlib_resources as resources  # pragma: no cover
-else:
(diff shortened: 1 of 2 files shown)


abravalheri:
Thanks, @cdce8p. This has some overlaps with https://github.com/pypa/setuptools/pull/4718, but I think that trying to keep all the PRs strictly separated by topic is showing up to be tricky.

setuptools/_importlib.py
```diff
@@ -1,12 +1,7 @@
...
 if sys.version_info < (3, 10):
     import importlib_metadata as metadata  # pragma: no cover
 else:
     import importlib.metadata as metadata  # noqa: F401
-
-
-if sys.version_info < (3, 9):
-    import importlib_resources as resources  # pragma: no cover
-else:
-    import importlib.resources as resources  # noqa: F401
```
abravalheri, line 12: I think that if we don't do this change, we might be able to keep the PRs thematically separated... let me see if we can do that.

abravalheri:
Thank you, I have manually merged this one with some changes: https://github.com/pypa/setuptools/commit/7a298d9abd05d02d89021f08cd2bb36bcca9952b.

PR #4742, opened 2024-11-09
Add support for windmc compiling routine

raedrizqie:
## Summary of changes

Needed for compiling `pywin32` modules


abravalheri:
Hi @raedrizqie thank you very much for the contribution.

I am afraid that in the `pypa/setuptools` repository we don't modify the files `setuptools/_distutils` directory directly. Instead, the directory is merged from the https://github.com/pypa/distutils repository periodically.

It is probably best to propose and discuss your change in the `pypa/distutils` repo.

raedrizqie:
I see.. Thanks for the reply.

PR #4744, opened 2024-11-10
ANN204 (missing return type for special methods) autofixes

Avasam:
## Summary of changes

Based on https://github.com/pypa/setuptools/pull/4736, so merge that one first.

Step 6.2 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833

This will produce many type issues that need to be resolved. I'll do that in other PRs and keep this as draft in the mean time.

### Pull Request Checklist
- [x] Changes have tests (type checkers, and this increases static type-checking coverage)
- [x] News fragment added in [`newsfragments/`]. (not public facing yet)

[PR docs]:


setuptools/command/bdist_egg.py
```diff
@@ -262,6 +266,8 @@ def zap_pyfiles(self):
 
                     pattern = r'(?P<name>.+)\.(?P<magic>[^.]+)\.pyc'
                     m = re.match(pattern, name)
+                    # We shouldn't ind any non-pyc files in __pycache__
+                    assert m is not None
```
Avasam, line 270: Not strictly an annotation change, but I included it here because it didn't feel worth doing a PR just for this

setuptools/command/bdist_egg.py
```diff
@@ -262,6 +266,8 @@ def zap_pyfiles(self):
 
                     pattern = r'(?P<name>.+)\.(?P<magic>[^.]+)\.pyc'
                     m = re.match(pattern, name)
+                    # We shouldn't find any non-pyc files in __pycache__
+                    assert m is not None
```
Avasam, line 270: Not strictly an annotation change, but I included it here because it didn't feel worth doing a PR just for this

setuptools/command/bdist_wheel.py
```diff
@@ -369,7 +369,7 @@ def get_tag(self) -> tuple[str, str, str]:
             )
         return tag
 
-    def run(self) -> None:
+    def run(self):
```
Avasam: I opted out of annotating this specific `run` in this PR for the same reason I didn't do `initialize_options`.

(not I'm commenting on a specific commit, this PR doesn't remove any return annotation)

setuptools/config/expand.py
```diff
@@ -414,7 +414,7 @@ def _get_package_dir(self) -> Mapping[str, str]:
         return {} if pkg_dir is None else pkg_dir
 
     @property
-    def package_dir(self) -> Mapping[str, str]:
+    def package_dir(self) -> LazyMappingProxy[str, str]:
```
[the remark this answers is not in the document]
Avasam:
```suggestion
    def package_dir(self) -> Mapping[str, str]:
```

setuptools/__init__.py
```diff
@@ -108,11 +108,12 @@ def _fetch_build_eggs(dist: Distribution):
         raise
 
 
-def setup(**attrs):
+def setup(**attrs) -> Distribution:
     logging.configure()
     # Make sure we have any requirements needed to interpret 'attrs'.
     _install_setup_requires(attrs)
-    return distutils.core.setup(**attrs)
+    # Override return type of distutils.core.Distribution with setuptools.dist.Distribution
```
abravalheri, line 115:
```suggestion
    # Override return type of distutils.core.Distribution with setuptools.dist.Distribution
    # (implicitly implemented via `setuptools.monkey.patch_all`).
```

This is just an optional clarification note. Not necessarily you need to act on that, only if you are planning to change the comments/ignores in `develop` (I can add it locally later, ).

abravalheri: In principle, we could modify those comments for clarity but possibly not in this PR?

I think there is also a subtle difference between the definition inside of a `Command` class and inside the `setup()` function:

For the command classes, the `Distribution` object is passed in as an [argument to the constructor](https://github.com/pypa/setuptools/blob/main/setuptools/_distutils/cmd.py#L89), so there is no direct patching of type of the given object `Distribution` itself[^1]. It is a "dependency injection" kind of situation, the object simply preserves the same type it had when it was passed in. Possibly that could be co-variant but I guess it would be impossible to satisfy static analysis given that `distutils.core.Distribution` is patched somewehere else in runtime?

On the other hand, inside `setup(...)` we are directly forwarding the call to `distutils.core.setup(...)`, so I think that the comment helps to understand the "misterious runtime behaviour", when it is expected that `distutils.core.setup(...)` should return the `distutils` version of `Distribution` (but in runtime it does not).

[^1]: Despite `distutils.cmd.Distribution` being patched, note how that [module](https://github.com/pypa/setuptools/blob/main/setuptools/_distutils/cmd.py) only uses `Distribution` for dynamic checks but never instantiate an object.

PR #4753, opened 2024-11-19
Runtime typing fixes for typeshed return type merge

Avasam:
## Summary of changes

Runtime typing fixes for #4744 that *shouldn't* incur any functional change for the end-users.

### Pull Request Checklist
- [x] Changes have tests (type-checking and existing tests)
- [x] News fragment added in [`newsfragments/`]. (not public-facing)

[PR docs]:


setuptools/command/bdist_wheel.py
```diff
@@ -220,7 +220,7 @@ class bdist_wheel(Command):
 
     def initialize_options(self) -> None:
         self.bdist_dir: str | None = None
-        self.data_dir: str | None = None
+        self.data_dir = ""
```
Avasam, line 223: `data_dir` is *always* set in `finalize_options`, so if I understand correctly this value doesn't matter and it'll always be a string.

setuptools/command/build_py.py
```diff
@@ -90,10 +89,7 @@ def __getattr__(self, attr: str):
         return orig.build_py.__getattr__(self, attr)
 
     def build_module(self, module, module_file, package):
-        outfile, copied = orig.build_py.build_module(self, module, module_file, package)
-        if copied:
-            self.__updated_files.append(outfile)
```
Avasam, line 95: Looks like `__updated_files` was completely unused, and its mangled

abravalheri: I see, apparently it was introduced for the `2to3` transformation: https://github.com/pypa/setuptools/commit/5b568886e649fad8811f81484dc88fa5111faa94.

In this case, does it make sense to keep this method? (It will only call `super()` right?)

setuptools/unicode_utils.py
```diff
@@ -65,10 +65,10 @@ def _read_utf8_with_fallback(file: str, fallback_encoding=py39.LOCALE_ENCODING)
 
 
 def _cfg_read_utf8_with_fallback(
-    cfg: ConfigParser, file: str, fallback_encoding=py39.LOCALE_ENCODING
+    cfg: RawConfigParser, file: str, fallback_encoding=py39.LOCALE_ENCODING
 ) -> None:
     """Same idea as :func:`_read_utf8_with_fallback`, but for the
-    :meth:`ConfigParser.read` method.
+    :meth:`RawConfigParser.read` method.
```
abravalheri, line 71: This also allows any subclass, right?

abravalheri (approved):
Thank you.

setuptools/unicode_utils.py
Avasam: Right. Setuptools passes a `RawConfigParser`, but `ConfigParser`, which is a subclass, will be statically allowed.

setuptools/command/build_py.py
Avasam: You're right! I didn't even notice that. I'd remove the method entirely at this point.

abravalheri: I agree, it is better to remove it.

PR #4755, opened 2024-11-20
Type setuptools/msvc.py dir methods and properties

Avasam:
## Summary of changes

Extracted from #4744 with additional fixes

### Pull Request Checklist
- [x] Changes have tests (type-checking tests)

[PR docs]:

### DIFF
--- a/setuptools/msvc.py
+++ b/setuptools/msvc.py
@@ -17,6 +17,10 @@ from typing import TYPE_CHECKING, TypedDict
 
 from more_itertools import unique_everseen
 
+from setuptools.compat import py310
+
+from ._path import StrPath
+
 import distutils.errors
 
 if TYPE_CHECKING:
@@ -135,7 +139,7 @@ class PlatformInfo:
             else rf'\{self.target_cpu}'
         )
 
-    def cross_dir(self, forcex86=False):
+    def cross_dir(self, forcex86=False) -> str:
         r"""
         Cross platform specific subfolder.
 
@@ -306,7 +310,7 @@ class RegistryInfo:
         node64 = '' if self.pi.current_is_x86() or x86 else 'Wow6432Node'
         return os.path.join('Software', node64, 'Microsoft', key)
 
-    def lookup(self, key, name):
+    def lookup(self, key: str, name: str) -> str | None:
         """
         Look for values in registry in Microsoft software registry.
 
@@ -319,7 +323,7 @@ class RegistryInfo:
 
         Return
         ------
-        str
+        str | None
             value
         """
         key_read = winreg.KEY_READ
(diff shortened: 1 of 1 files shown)


setuptools/msvc.py
```diff
@@ -716,9 +718,9 @@ def UniversalCRTSdkDir(self):
         for ver in vers:
             sdkdir = self.ri.lookup(self.ri.windows_kits_roots, 'kitsroot%s' % ver)
             if sdkdir:
-                return sdkdir or ''
+                return sdkdir
 
-        return None
+        return ''
```
Avasam, line 723: This one I'm not entirely certain of, and does change the possible return types.

dir-related properties of this class are a mixed bag on whether they return `None` or `''` as a fallthrough case. And are almost never documented as returning None (which is fixed in this PR).

However all other properties that *can* return `None` are checked for truthiness when used, this one isn't. (hence it was causing issues for https://github.com/pypa/setuptools/pull/4744)

The redundant `or ''` makes me think this may have been intended to not return `None`. But the true intention is unclear to me.

abravalheri: Thank you very much for having a look on this.

It is a complicated decision. For the sake of backwards compatibility I think a safe approach would be to stick with the observed behaviour instead of the original intention?

Even if `return None` was a late automatic addition, the original function could still returning `None` under certain circumstances isn't it? So it may be safer for us to accept it can return `None`?

Avasam: [`0788a65` (#4755)](https://github.com/pypa/setuptools/pull/4755/commits/0788a65e3efab6c6d34a012b577eac4b43fd30bd) Shows the diff letting this return `None`.

Same resolved `TypeError` in `UniversalCRTSdkLastVersion`, `UCRTLibraries`, `UCRTIncludes` and their caller when `UniversalCRTSdkDir` returns `None`.
If you want these to also keep the exact same exception raising behaviour, I can raise an explicit `TypeError` in them.

setuptools/msvc.py
```diff
@@ -751,16 +755,16 @@ def NetFxSdkVersion(self):
...
         ------
-        str
+        str | None
             path
         """
-        sdkdir = ''
+        sdkdir: str | None = ''
         for ver in self.NetFxSdkVersion:
             loc = os.path.join(self.ri.netfx_sdk, ver)
             sdkdir = self.ri.lookup(loc, 'kitsinstallationfolder')
```
Avasam, line 770: Because `RegistryInfo.lookup` can return `None`, and this method returns the last falsy result if all falsy. Then the return type is potentially `None`.

setuptools/msvc.py
abravalheri: I am more inclined towards the 2nd option for being the balance between leaving things as they are and providing debugging information if necessary.

setuptools/msvc.py
```diff
@@ -17,6 +17,10 @@
 
 from more_itertools import unique_everseen
 
+from setuptools.compat import py310
+
+from ._path import StrPath
```
abravalheri, line 22: Let's use relative imports to align with `from ._path import StrPath`

```suggestion
from ._path import StrPath
from .compat import py310
```

abravalheri (approved):
Thank you very much!

Issue #4483, opened 2024-07-18
[BUG] TypeError: canonicalize_version() got an unexpected keyword argument 'strip_trailing_zero'


tim-s-ccs:
### setuptools version

setuptools>=71.0.0

### Python version

Python 3.12

### OS

Ubuntu

### Additional environment information

I found this bug when running the test for the project in GitHub actions

### Description

When I install the requirements in my Python package project (`pip-sync requirements-dev.txt`) I get an error within setuptools which says:
```
Preparing metadata (setup.py) ... error
error: subprocess-exited-with-error

× python setup.py egg_info did not run successfully.
...
TypeError: canonicalize_version() got an unexpected keyword argument 'strip_trailing_zero'
```
This error was caused by using setuptools >= 71.0.0

Not to speculate but it looks like the issue was introduced by this commit in setuptools: https://github.com/pypa/setuptools/commit/00384a5f4fd22c653172b99feefe13b0009eb870

And this commit is including new changes from the packaging project and its this commit here where we can see the issue:
https://github.com/pypa/packaging/commit/cc938f984bbbe43c5734b9656c9837ab3a28191f
https://github.com/pypa/packaging/blame/4493dfcd95a893f676a7aa4bd17c547bea676371/src/packaging/utils.py#L58

### Expected behavior

I would expect `pip-sync requirements-dev.txt` to run successfully

### How to Reproduce

Clone https://github.com/tim-s-ccs/example-python-package.git
Make sure you are using Python 3.12 and you have pip-tools installed (`pip install --upgrade pip wheel pip-tools`)
Make sure you are using the correct version of setuptools `pip install --upgrade setuptools==71.0.0`
Run `pip-compile requirements-dev.in`

### Output

```console
    error: subprocess-exited-with-error
    
    × python setup.py egg_info did not run successfully.
    │ exit code: 1
    ╰─> [42 lines of output]
        running egg_info
        creating /private/var/folders/kt/n16_msf56_z3kfyb65l37wxhb_m4s_/T/pip-pip-egg-info-3sca_19y/example_python_package.egg-info
        writing /private/var/folders/kt/n16_msf56_z3kfyb65l37wxhb_m4s_/T/pip-pip-egg-info-3sca_19y/example_python_package.egg-info/PKG-INFO
[...]

PR #4758, opened 2024-11-23
Unvendor safe dependencies

jaraco:
- **Move dependencies with no build- or run-time dependency on Setuptools to natural dependencies.**
- **Rename 'vendor' to 'target'**
- **Inline the update function.**
- **Only vendor the dependencies defined in core.**
- **Re-vendor dependencies.**
- **Also declare the dependencies as build requirements.**
- **Remove test_no_missing_dependencies, as that guarantee no longer holds.**
- **Update reference to wheel, no longer vendored.**

## Summary of changes

Progress toward #2825.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
This change will incidentally fix #4483, as now packaging becomes a proper dependency instead of an implicit/hidden one.

abravalheri:
Hi Jason, thank you very much for having a look on this.

I not 100% sure about the approach. I like the fact that setuptools can bootstrap itself right now without the help of any external tools other than Python stdlib. One of the bootstrap stories in the ecosystem involves installing `flit-core`, `build`, `installer`... I have the impression that an equally viable alternative story with `setuptools` + `pip` (with the difference that you are going to have a fully featured environment at that end of the process). That is something nice to have.

One possible alternative is documenting that setuptools should be installed as `setuptools[core]` whenever outside of the context of `[build-system] requires = [...]` + raising more informative import errors in the well known scenarios (like `importlib-metadata` and `packaging`).

jaraco:
Thanks for the feedback. I'm not yet sure how I'm going to proceed with this, but given that the tests aren't yet passing, I'm converting this to a draft.

PR #4762, opened 2024-12-01
correct 'namespace' term to 'namespace-package' in quickstart doc

deronnax:
## Summary of changes

The original reference pointed to python generic [namespace](https://docs.python.org/3.11/glossary.html#term-namespace),  in the python glossary, which is wrong. What it want to point to is [package namespace](https://docs.python.org/3.11/glossary.html#term-namespace-package).
I don't think it needs an entry in the changelog.

### DIFF
--- a/docs/userguide/quickstart.rst
+++ b/docs/userguide/quickstart.rst
@@ -199,7 +199,7 @@ Package discovery
 -----------------
 For projects that follow a simple directory structure, ``setuptools`` should be
 able to automatically detect all :term:`packages <package>` and
-:term:`namespaces <namespace>`. However, complex projects might include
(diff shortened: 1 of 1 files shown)


abravalheri:
@deronnax do you know why he have a warn for "term not in glossary" in https://github.com/pypa/setuptools/actions/runs/12298183072/job/34325797816#step:5:153?

Do we have to remove the hyphen maybe?

deronnax:
Hum. Not at all, sorry.  Let's revert if it is a problem.

jaraco:
It is a problem: https://github.com/pypa/setuptools/issues/4779.

PR #4766, opened 2024-12-10
Fix "Wheel naming is not following PEP 491 convention"

### LINKED ISSUE #3777: [BUG] Wheel naming is not following PEP 491 convention
### setuptools version

setuptools==66.0.0

### Python version

python 3.8

### OS

macOS

_No response_

### Description

I am building a wheel for a python package using setuptools and it seems that the naming of my wheel file is not respecting the [PEP 491](https://peps.python.org/pep-0491/#escaping-and-unicode) convention.

For external reasons, I need to name my package with the structure `{namespace}.{package-name}`. If I follow the convention, I would expect that my wheel file is named: `namespace_package_name-0.1.0-py2.py3-none-any.whl`.

However, I get this name for my wheel: `namespace.package_name-0.1.0-py2.py3-none-any.whl`, which is not respecting the convention.

### Expected behavior

I expect the "." in the package name to be replaced with a "_" in the wheel name.

### How to Reproduce

1. create a hello_world package structure:
   ```
   hello_world
   |- my_package
   |    |- __init__.py
   |    |- hello_world.py
   |- pyproject.toml
   |- setup.cfg
   ```
2. in `setup.cfg`, write:
    ```
    [metadata]
    name = namespace.my-package
    version = 0.1.0
    
    [options]
    zip_safe = False
    include_package_data = True
    packages = find:
    ```
4. in  `project.toml`, write:
    ```
    [build-system]
    requires = ["setuptools==66.0.0", "wheel"]
    ```
6. run `pipx run build` at the root of your hello_world package
7. inspect the `dist` directory which was created

### Output

in the dist folder, you find:

- `namespace.my_package-0.1.0-py3-none-any.whl`

di:
## Summary of changes

Fixes #3777.

~Marking this as draft as there is one outstanding failing test that I have been unable to resolve (`setuptools/tests/test_easy_install.py::TestSetupRequires::test_setup_requires_override_nspkg`).~

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


abravalheri:
Thank you!


The failing test seems to be happening now regardless of this change: https://github.com/pypa/setuptools/actions/runs/12298183072/job/34321035714#step:3:49.


I don't know if anything changed in the latest versions of Python being distributed to macOS, so that `__file__` is no longer available on the scope.

di:
Aha, that's a different failure actually, `setuptools/tests/test_easy_install.py::TestSetupRequires::test_setup_requires_override_nspkg` is the test that is legitimately failing and would need to be updated here.

abravalheri:
> `setuptools/tests/test_easy_install.py::TestSetupRequires::test_setup_requires_override_nspkg` is the test that is legitimately failing and would need to be updated here.

This looks like it is checking deprecated functionality? Not sure we have to account such an edge case since we are moving away from `easy_install`. Maybe we can do a major bump to show that the change may be considered a breaking change.

di:
I'm not sure! I think the test is just broken in a way I haven't been able to figure out (it mixes the distribution name and the namespace name as one thing in the helper functions) but I think the underlying functionality probably still works?

di:
All tests are passing, this is now ready for review.

setuptools/command/bdist_wheel.py
```diff
@@ -134,7 +134,13 @@ def get_abi_tag() -> str | None:
...
 
 def safer_name(name: str) -> str:
-    return safe_name(name).replace("-", "_")
+    return (
+        # Per https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization
+        re.sub(r"[-_.]+", "-", safe_name(name))
+        .lower()
+        # Per https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode
+        .replace("-", "_")
+    )
```
jaraco, line 143: I'm not a fan of this duplicated code being further expanded. I'd like to see this logic consolidated (in setuptools if not in something like packaging), especially since now it's more than one line that's being duplicated.

setuptools/tests/test_easy_install.py
```diff
@@ -1120,8 +1120,10 @@ def make_nspkg_sdist(dist_path, distname, version):
     package with the same name as distname.  The top-level package is
     designated a namespace package).
     """
-
-    parts = distname.split('.')
+    # The project name might not contain periods. Replace dashes and
+    # underscores with periods before constructing the namespace.
+    namespace = distname.replace('-', '.').replace('_', '.')
```
jaraco, line 1125: This seems slightly terrible. Is this helper called with a normalized distname, in which case it's not possible to know what the separators are? Is it always called with a normalized distname, or is it called with the specified distribution name? This function, if called with `more-itertools` will produce a namespace package for `more`. I guess I can see from the previous implementation that it always assumes there's a single namespace component at the front.

Still, the previous implementation would have accepted `j_o.foo` and produced a namespace package `j_o`, but this new implementation will get `j-o-foo` and produce a namespace packgae `j`.

More importantly, however, I notice that `setuptools.setup` is being called with `name={distname}`. In that case, `{distname}` should not be normalized and should contain the periods.

I think this change is a mistake, and instead, this function should require that `distname` contain at least one period.

setuptools/tests/test_easy_install.py
```diff
@@ -670,11 +670,11 @@ def test_setup_requires_override_nspkg(self, use_setup_cfg):
 
         with contexts.save_pkg_resources_state():
             with contexts.tempdir() as temp_dir:
-                foobar_1_archive = os.path.join(temp_dir, 'foo.bar-0.1.tar.gz')
-                make_nspkg_sdist(foobar_1_archive, 'foo.bar', '0.1')
+                foobar_1_archive = os.path.join(temp_dir, 'foo_bar-0.1.tar.gz')
+                make_nspkg_sdist(foobar_1_archive, 'foo_bar', '0.1')
```
jaraco, line 674: Per the comment in make_nspkg_sdist, distname should be `foo.bar`.

jaraco (approved):
It doesn't make me happy to approve this change, but I've conceded the fight. Let's align with the spec.

setuptools/command/bdist_wheel.py
di: Resolved in 9f764e8.

setuptools/tests/test_easy_install.py
di: Resolved in 8795306, ae5df97 and 1ec0c91.

setuptools/tests/test_easy_install.py
di: Resolved in 8795306.

di:
This is ready for re-review.

jaraco (approved):

di:
@abravalheri @jaraco Would it be possible to get a release out with this change? This is blocking https://github.com/pypi/warehouse/pull/17378 and https://github.com/pypi/warehouse/issues/17377. Happy to prep/propose the release if that would help!

abravalheri:
Ok, let's get this out before we merge more stuff so that is easier to track breaking changes if any (or gauge any push back from the community).

abravalheri:
@di, sorry for the delay and thank you for working on this. CI is running for the release https://github.com/pypa/setuptools/tree/refs/tags/v75.8.1.

If we want to be in the safe side, it might make sense to wait some days before to check if we have any pushbacks or bugs reported by the community.

abravalheri:
But, we had some unexplained CI errors: https://github.com/pypa/setuptools/actions/runs/13523090359/job/37786779855.

I will try to re-run the failed jobs shortly.

Issue #4759, opened 2024-11-26
[BUG] `tool.setuptools.license-files` results in invalid metadata


dnicolodi:
### setuptools version

setuptools==74.1.2

### Python version

Python 3.13

### OS

any

### Description

If any of the the glob patterns specified in`tool.setuptools.license-files` matches a file in the package, setuptools generates invalid metadata: it includes a `License-File` field while specifying `Metadata-Version` to be `2.1`. This is invalid and `packaging` raises an exception while parsing the metadata. This likely results in the resulting distributions to not be accepted by PyPI.

Because `tool.setuptools.license-files`  has a default value of `['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']` the problem can be encountered also in packages that do not explicitly set this field in `pyproject.toml` but happen to have a file matching the default glob pattern.

### Expected behavior

Do not emit the `License-File` field or do it and specify `Metadata-Version: 2.4` as per PEP 639.

### How to Reproduce

Here is a short reproducer:

```console
$ cat >pyproject.toml <<EOF
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"
[project]
name = "foo"
version = "1.2.3"
[tool.setuptools]
license-files = ["LICENSE"]
EOF
$ touch LICENSE
$ uv build .
$ tar xf dist/foo-1.2.3.tar.gz
$ python
>>> import pathlib
>>> import packaging.metadata
>>> packaging.metadata.Metadata.from_email(pathlib.Path('dist/foo-1.2.3/PKG-INFO').read_text())
```

### Output

```python
  + Exception Group Traceback (most recent call last):
  |   File "<python-input-3>", line 1, in <module>
  |     packaging.metadata.Metadata.from_email(pathlib.Path('dist/foo-1.2.3/PKG-INFO').read_text())
  |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/Users/daniele/src/twine/.venv/lib/python3.13/site-packages/packaging/metadata.py", line 781, in from_email
  |     raise ExceptionGroup(
  |         "invalid or unparsed metadata", exc_group.exceptions
  |     ) from None
  | ExceptionGroup: invalid or unparsed metadata (1 sub-exception)
  +-+---------------- 1 ----------------
[...]

PR #4771, opened 2024-12-17
Stop writing License-File

hauntsaninja:
https://github.com/pypa/setuptools/issues/4759


jaraco:
I think this is the wrong approach. We don't want to degrade the behavior, but instead declare the proper metadata version.

PR #4783, opened 2024-12-31
Enforce ruff/flake8-pytest-style rules (PT)

DimitriPapadopoulos:
I haven't addressed `PT012`.

As for `PT007`, I have used the default style because it minimises changes to the current codebase:
```toml
[lint.flake8-pytest-style]
parametrize-values-type = "list"
parametrize-values-row-type = "tuple"
```

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


Avasam:
I think you'll want to enable [allow-dict-calls-with-keyword-arguments](https://docs.astral.sh/ruff/settings/#lint_flake8-comprehensions_allow-dict-calls-with-keyword-arguments)

PR #4786, opened 2025-01-03
setuptools.msvc: set host_dir correctly on ARM64 hosts

swt2c:
## Summary of changes

This corrects the setting of the host directory name on ARM64 hosts so that the `EnvironmentInfo.VCTools` property and `EnvironmentInfo.return_env()['path']` includes the correct path to CL.EXE.  Currently, it returns the X64 directory on ARM64 hosts.  See below; note `HostX64` in first listed path.

```
$ python
Python 3.13.1 (tags/v3.13.1:0671451, Dec  3 2024, 20:04:37) [MSC v.1942 64 bit (ARM64)] on win32
>>> import setuptools.msvc
>>> ei = setuptools.msvc.EnvironmentInfo('arm64')
>>> ei.return_env()['path']
'C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.42.34433\\bin\\HostX64\\arm64;C:\\Program Files (x86)\\Windows Kits\\10\\Bin\\arm64;C:\\Program Files (x86)\\Windows Kits\\10\\Bin\\10.0.22621.0\\arm64;C:\\Users\\talbert\\.blah_env\\Scripts;C:\\Program Files\\Git\\clangarm64\\bin;C:\\Program Files\\Git\\usr\\bin;C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0;C:\\Windows\\System32\\OpenSSH;C:\\Program Files\\Git\\cmd;C:\\Users\\talbert\\AppData\\Local\\Programs\\Python\\Python313-arm64\\Scripts;C:\\Users\\talbert\\AppData\\Local\\Programs\\Python\\Python313-arm64;C:\\Users\\talbert\\AppData\\Local\\Programs\\Python\\Launcher;C:\\Users\\talbert\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Program Files\\Git\\usr\\bin\\vendor_perl;C:\\Program Files\\Git\\usr\\bin\\core_perl'

```

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


setuptools/msvc.py
```diff
@@ -1039,7 +1039,13 @@ def VCTools(self):
...
             host_dir = (
-                r'bin\HostX86%s' if self.pi.current_is_x86() else r'bin\HostX64%s'
+                r'bin\HostX86%s'
+                if self.pi.current_is_x86()
+                else (
+                    r'bin\HostARM64%s'
+                    if self.pi.current_cpu == 'arm64'
+                    else r'bin\HostX64%s'
+                )
             )
```
Avasam, line 1049: Not a maintainer here, but personally I find if-else chains to be much more readable and concise than nested ternaries:
```py
            if self.pi.current_is_x86():
                host_dir = r'bin\HostX86%s'
            elif self.pi.current_cpu == 'arm64':
                host_dir = r'bin\HostARM64%s'
            else:
                host_dir = r'bin\HostX64%s'
```

swt2c: Updated.

setuptools/msvc.py
```diff
@@ -1038,9 +1038,13 @@ def VCTools(self):
...
         elif self.vs_ver >= 15.0:
-            host_dir = (
-                r'bin\HostX86%s' if self.pi.current_is_x86() else r'bin\HostX64%s'
-            )
+            if self.pi.current_is_x86():
+                host_dir = r'bin\HostX86%s'
+            elif self.pi.current_cpu == 'arm64':
+                host_dir = r'bin\HostARM64%s'
+            else:
+                host_dir = r'bin\HostX64%s'
```
jaraco, line 1046: Even better than an if/else statement would be something like:

```suggestion
            host_dir = os.path.join('bin', f'Host{self.host_id.upper()}%s')
```

And then of course implement the property `host_id` to reflect the correct string to insert. In general, it's good practice to DRY - avoid repeating the assignment and prefix and suffix of the value.

setuptools/msvc.py
```diff
@@ -1038,9 +1038,13 @@ def VCTools(self):
             tools += [os.path.join(si.VCInstallDir, path)]
 
         elif self.vs_ver >= 15.0:
-            host_dir = (
-                r'bin\HostX86%s' if self.pi.current_is_x86() else r'bin\HostX64%s'
-            )
+            if self.pi.current_is_x86():
+                host_dir = r'bin\HostX86%s'
+            elif self.pi.current_cpu == 'arm64':
```
jaraco, line 1043: Note that this comparison is no longer symmetrical with the check for 'x86'. Better would be to be consistent. Either use `self.pi.current_cpu == 'x86'` inline or implement a new `self.pi.current_is_arm64()` method.

jaraco (changes requested):
Since this PR isn't linked to any issue, can you update the PR description to elaborate on the issue? In particular, answer questions like: Why does it matter? How did you encounter it? Under what conditions is it relevant (e.g. does it affect all users of Setuptools on ARM, or some subset)?

I see lots of other places in this module where the assumption is that there is x86 and x64. I'm not sure it makes sense to add ARM support here, but not elsewhere. What's the value in fixing just this issue and not other parts in the module that deal with processor architecture?

swt2c:
> Since this PR isn't linked to any issue, can you update the PR description to elaborate on the issue? In particular, answer questions like: Why does it matter? How did you encounter it? Under what conditions is it relevant (e.g. does it affect all users of Setuptools on ARM, or some subset)?
> 
> I see lots of other places in this module where the assumption is that there is x86 and x64. I'm not sure it makes sense to add ARM support here, but not elsewhere. What's the value in fixing just this issue and not other parts in the module that deal with processor architecture?

I thought the description was pretty clear about what the problem was (it really is just about the `path` item as returned by `EnvironmentInfo.return_env()`, but I added a Python interpreter run that shows the exact output that you currently get.  This issue affects anyone who is using EnvironmentInfo.return_env() on an ARM64 Windows system and relying on the `path` that is returned.

This is the only problem that I'm actually aware of with setuptools.msvc on Windows ARM64.  I just did another look through the file and while yes, there are a lot of cases where it assumes there is only x86 and x64, I didn't see any cases where it is actually doing the wrong thing on ARM64.  The rest of the output of EnvironmentInfo.return_env() looks fine.

setuptools/msvc.py
swt2c: Adding a property for something that's only used once seemed a bit of overkill.  Let me know if what I did works.

setuptools/msvc.py
swt2c: Addressed.

jaraco (approved):

setuptools/msvc.py
jaraco: I took it one step further in 3218d666e49245b0878db8b19ba20d8cd72d6424.

swt2c:
Thanks @jaraco for the fixups and adding a test.  I wasn't very creative in thinking about how to test this without a Windows ARM64 host.

PR #4789, opened 2025-01-04
Don't emit warning for explicitly included data files

cdce8p:
## Summary of changes

From what I can tell based on the discussion in #3340, it would make sense to hide the `Package would be ignored` warning for explicitly including data files (via `tool.setuptools.package-data` or similar).

### DIFF
--- a/newsfragments/4789.misc.rst
+++ b/newsfragments/4789.misc.rst
@@ -1,0 +1,1 @@
+Don't emit ``Package would be ignored`` warning for explicitly included data files.
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -194,6 +194,10 @@ class build_py(orig.build_py):
             files = ei_cmd.filelist.files
 
         check = _IncludePackageDataAbuse()
+        data_files = {
+            package: filenames
+            for package, _, _, filenames in self.get_data_files_without_manifest()
+        }
         for path in self._filter_build_files(files, egg_info_dir):
             d, f = os.path.split(assert_relative(path))
             prev = None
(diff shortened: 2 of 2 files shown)


abravalheri:
I have a different understanding of the previous discussion: if the package is not included, then the data files contained by it cannot be included (adding the package itself is a pre-requisite for adding its data files)...

So the warning is a call to action for the users to add the package name in the package list (this would be along the lines of "explicit is better than implicit").

cdce8p:
> I have a different understanding of the previous discussion: if the package is not included, then the data files contained by it cannot be included (adding the package itself is a pre-requisite for adding its data files)...

My specific example which motivated the PR is Mypy. It uses `packages.find.include` with a wildcard `["mypy*", ...]` and additionally `package-data`. The paths used `package-data` are subfolders as it would clutter the main folder too much otherwise. The warning in this case is pointless IMO.

https://github.com/python/mypy/blob/v1.14.1/pyproject.toml#L79-L92

> So the warning is a call to action for the users to add the package name in the package list (this would be along the lines of "explicit is better than implicit" and raising awareness of the users that for Python purposes the directories are packages during runtime and can even be imported).

My goal here would be to think of a better user story, ie. what can you do if you see the warning, are aware of it but still want to go ahead with it? _Namespace packages aren't really an option._ Some alternative ideas
- If explicit is better them implicit, would it make sense to add a new option, maybe `tool.setuptools.data-folder` where users can explicitly name folder (globs) which should opt-out of this check?
- As the warning is quite annoying, I even thought about overwriting `build_py. _filter_build_files` for mypy just to get rid of it. Maybe we could instead refactor `analyze_manifest` so it wouldn't be necessary to overwrite a private method.
https://github.com/pypa/setuptools/blob/fb7f3d3cab98f2c00cbf18fe887ce73007f49e18/setuptools/command/build_py.py#L196-L197

PR #4796, opened 2025-01-09
Fix ruff issues

cdce8p:


cdce8p:
/CC @abravalheri This fixes the CI issues seen in the recent runs. Not sure it needs a news entry. Feel free to let me know if I should add one though.

abravalheri:
Thank you very much.

PR #4802, opened 2025-01-14
Ran autofixes for `missing-return-type-special-method (ANN204)` with `ignore-fully-untyped = false`

Avasam:
I set `[lint.flake8-annotations] ignore-fully-untyped = false` then ran `ruff check --select=ANN204 --fix --unsafe-fixes`

## Summary of changes

Step 6.3 of https://github.com/pypa/setuptools/issues/2345#issuecomment-1627561833
Some slight overlaps with https://github.com/pypa/setuptools/pull/4744

### Pull Request Checklist
- [x] Changes have tests (this adds static tests)

[PR docs]:

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -66,13 +66,12 @@ ignore = [
 	"UP038", # Using `X | Y` in `isinstance` call is slower and more verbose https://github.com/astral-sh/ruff/issues/7871
 	# Only enforcing return type annotations for public functions
 	"ANN202", # missing-return-type-private-function
-	"ANN204", # missing-return-type-special-method
 ]
 
 [lint.per-file-ignores]
 # Suppress nuisance warnings about module-import-not-at-top-of-file (E402) due to workaround for #4476
 "setuptools/__init__.py" = ["E402"]
-"pkg_resources/__init__.py" = ["E402"]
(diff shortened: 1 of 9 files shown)


abravalheri (commented):
Thank you very much @Avasam.

I am not sure what will be the policy regarding modifying the `pkg_resource` files now that it has been scheduled for removal and that Jason has spent some time exporting the commit history to the `pypa/pkg_resources` repo. I also don't know if there are plans of keep updating that other repo or if it was a simple experiment.

Maybe @jaraco can clarify what is his preference.

Avasam:
Oh yeah this PR was created before #5007

Avasam:
I've excluded `pkg_resources` from this PR.

jaraco:
For future reference, avoid changing pkg_resources. I want to avoid committing anything here that we'd miss there, assuming we decide to use it. But if it's convenient or it happens that we change something here, it's no big deal. Don't sweat it either way.

Avasam:
When it comes to checkers (Ruff, mypy, pyright), the `pkg_resources` folder can just be ignored, even on a per-rule level. Makes it easy to keep automatically enforcing project practices without touching `pkg_resources`.

PR #4818, opened 2025-02-01
Drop test dependency on `path` package

Avasam:
## Summary of changes

As far as I can tell, `pathlib`+`jaraco.path` should be sufficient, and it's only used in tests. There may be issues with vendored deps, let's see what the CI says.

Closes 

### Pull Request Checklist
- [x] Changes have tests (these are the test changes)
- [x] News fragment added in [`newsfragments/`]. (no public changes, tests only)

[PR docs]:

### DIFF
--- a/setuptools/tests/config/test_pyprojecttoml.py
+++ b/setuptools/tests/config/test_pyprojecttoml.py
@@ -5,7 +5,6 @@ from inspect import cleandoc
 import jaraco.path
 import pytest
 import tomli_w
-from path import Path
 
 import setuptools  # noqa: F401 # force distutils.core to be patched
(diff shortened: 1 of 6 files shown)


Avasam:
After making this PR for something that looked like an unnecessary extra dependency. I went and looked at what exactly is `path` (my first assumption being that it was some compat module from before `pathlib`). Especially https://pypi.org/project/path#advantages .
It makes me curious about the vague separation between `jaraco/path` and `jaraco/jaraco.path`.

I also noticed that `path.Path.parts` and `pathlib.Path.parts` are incompatible (one's a method, the other's a property)

PR #4841, opened 2025-02-18
Validate license-files glob patterns

cdce8p:
Alternative approach to #4838.

Validate each pattern and raise an exception if it is invalid or a pattern doesn't match any license files.
This will also apply to existing configs with `license-files` in `setup.cfg` and `tools.setuptools.license-files`, however I don't think that's an issue as these are almost always real problems and easy to fix.

Ref: #4829

### DIFF
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -70,6 +70,8 @@ Avoid single-use iterable. Disallow sets.
 A poor approximation of an OrderedSequence (dict doesn't match a Sequence).
 """
 
+_license_files_allowed_chars = re.compile(r'^[\w\-\.\/\*\?\[\]]+$')
+
 
 def __getattr__(name: str) -> Any:  # pragma: no cover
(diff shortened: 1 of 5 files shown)


setuptools/dist.py
```diff
@@ -438,35 +440,73 @@ def _finalize_license_files(self) -> None:
...
+            raise InvalidConfigError(
+                f"License file pattern '{pattern}' should be relative and "
+                "must not start with '/'"
+            )
+        if _license_files_allowed_chars.match(pattern) is None:
+            raise InvalidConfigError(
+                f"License file pattern '{pattern}' contains invalid "
+                "characters. "
+                "https://packaging.python.org/en/latest/specifications/pyproject-toml/#license-files"
+            )
```
cdce8p, line 490: This is similar to the validation pattern I added in `flit` a few weeks ago. https://github.com/pypa/flit/pull/705

PR #4849, opened 2025-02-22
Fix typos

shenxianpeng:
## Summary of changes

I reviewed the TODO comments and attempted to address some of them.

Additionally, I found some typos during this process. Perhaps the setuptools project could introduce spellcheck hooks in .pre-commit-config.py to catch and fix more typos automatically.

### Pull Request Checklist
~- [ ] Changes have tests~

[PR docs]:


Avasam:
> introduce spellcheck hooks in .pre-commit-config.py to catch and fix more typos automatically

I doubt the signal-to-noise ratio would be worth it. Having random spellcheck passes once in a while like you did is probably enough. But that's just my opinion.

setuptools/_vendor/autocommand-2.2.2.dist-info/METADATA
```diff
@@ -320,7 +320,7 @@ optional arguments:
 STOP and STEP default to 1
 ```
 
-Even though autocommand is being applied to the `wrapper` returned by `print_yielded`, it still retreives the signature of the underlying `seq` function to create the argument parsing.
+Even though autocommand is being applied to the `wrapper` returned by `print_yielded`, it still retrieves the signature of the underlying `seq` function to create the argument parsing.
```
abravalheri, line 323: We should not modify files in the `_vendor` and `_distutils` directly via PRs to setuptools. Instead it is better to send the PRs to the upstream repositories.

abravalheri (commented):
Thank you very much @shenxianpeng.

Some of these are not strictly typos (just a different orthography due to differences between English speaking countries), but it should be fine to convert between them.

(In terms of potential follow ups for actions/pre-commit hooks, I would avoid being too restrictive in terms of what we accept from contributors, e.g. people should feel welcome to contribute using the way they right English in their countries regardless if they come from Europe/Oceania/Africa/America etc...).

Could you please have avoid submitting changes to the `_vendor` and or `_distutils` repository?

PR #4855, opened 2025-02-26
pkg_resources: Correctly handle normalized project names

### LINKED ISSUE #4853: [BUG] packages of the form package.something renamed to package_something in 75.8.1
### setuptools version

setuptools==75.8.1

### Python version

python3.10

### OS

windows

### Description

We have packages using pyproject.toml of the form:

`something.module` that use implicit namespaces for `something`

The latest 75.8.1 seems to rename the `something.module-1.2.3.dist-info` to be instead: `something_module-1.2.3.dist-info`

rolling back to 75.8.0 this problem does not exist.

This breaks folks using something of the form `pkg_resources.require('something.module')` as it now only works if they have `pkg_resources.require('something-module')`

### Expected behavior

packages should continue to have the `something.module` form

### How to Reproduce

```
git clone https://github.com/krpatter-intc/setuptools-example.git
cd setuptools-example
python -m build
```
And you'll note the naming difference in the wheel (and the .dist-info file)

You can then run:

```
import pkg_resources
pkg_resources.require("something.module") # this now throws an exception
```

### Output

```console

```

I'm aware pkg_resources is on its way out, but this still feels bad that the package has a new name. I'm not sure what the new equivalent/replacement for `pkg_resources.require` is.

di:
## Summary of changes

This fixes an issue where `pkg_resources.require` treats project names with periods differently than other project names. 

Before:
```
>>> import pkg_resources
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
>>> pkg_resources.require('zip-po')
[zip-po 0.0.0 (/home/di/.pyenv/versions/3.12.4/lib/python3.12/site-packages)]
>>> pkg_resources.require('zip_po')
[zip-po 0.0.0 (/home/di/.pyenv/versions/3.12.4/lib/python3.12/site-packages)]
>>> pkg_resources.require('zip.po')
  File "/home/di/git/pypa/setuptools/pkg_resources/__init__.py", line 1061, in require
    needed = self.resolve(parse_requirements(requirements))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/git/pypa/setuptools/pkg_resources/__init__.py", line 888, in resolve
    dist = self._resolve_dist(
           ^^^^^^^^^^^^^^^^^^^
  File "/home/di/git/pypa/setuptools/pkg_resources/__init__.py", line 929, in _resolve_dist
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'zip.po' distribution was not found and is required by the application
```

After:
```
>>> import pkg_resources
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
>>> pkg_resources.require('zip-po')
[zip-po 0.0.0 (/home/di/.pyenv/versions/3.12.4/lib/python3.12/site-packages)]
>>> pkg_resources.require('zip_po')
[zip-po 0.0.0 (/home/di/.pyenv/versions/3.12.4/lib/python3.12/site-packages)]
>>> pkg_resources.require('zip.po')
[zip-po 0.0.0 (/home/di/.pyenv/versions/3.12.4/lib/python3.12/site-packages)]
```

The simplest fix would be a one-character update to `safe_name`, which currently does not normalize periods when producing a "safe" project name:

https://github.com/pypa/setuptools/blob/56c055b653f080544f490e198605974c71b6fe35/pkg_resources/__init__.py#L1551-L1556 

However, since this is a public API, changes here might be more disruptive, so I opted to change `pkg_resources` usage of `safe_name` internally instead.

Closes #4853.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2911,7 +2911,7 @@ class Distribution:
         platform: str | None = None,
         precedence: int = EGG_DIST,
     ) -> None:
-        self.project_name = safe_name(project_name or 'Unknown')
+        self.project_name = safe_name(project_name or 'Unknown').replace('.', '-')
         if version is not None:
             self._version = safe_version(version)
         self.py_version = py_version
@@ -3459,7 +3459,7 @@ class Requirement(packaging.requirements.Requirement):
         """DO NOT CALL THIS UNDOCUMENTED METHOD; use Requirement.parse()!"""
         super().__init__(requirement_string)
         self.unsafe_name = self.name
-        project_name = safe_name(self.name)
+        project_name = safe_name(self.name).replace('.', '-')
         self.project_name, self.key = project_name, project_name.lower()
         self.specs = [(spec.operator, spec.version) for spec in self.specifier]
         self.extras = tuple(map(safe_extra, self.extras))
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -3459,8 +3459,8 @@ class Requirement(packaging.requirements.Requirement):
         """DO NOT CALL THIS UNDOCUMENTED METHOD; use Requirement.parse()!"""
         super().__init__(requirement_string)
         self.unsafe_name = self.name
-        project_name = safe_name(self.name).replace('.', '-')
-        self.project_name, self.key = project_name, project_name.lower()
+        self.project_name = safe_name(self.name)
+        self.key = self.project_name.lower().replace('.', '-')
         self.specs = [(spec.operator, spec.version) for spec in self.specifier]
         self.extras = tuple(map(safe_extra, self.extras))
         self.hashCmp = (
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -2911,7 +2911,7 @@ class Distribution:
         platform: str | None = None,
         precedence: int = EGG_DIST,
     ) -> None:
-        self.project_name = safe_name(project_name or 'Unknown').replace('.', '-')
+        self.project_name = safe_name(project_name or 'Unknown')
         if version is not None:
             self._version = safe_version(version)
         self.py_version = py_version
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -453,11 +453,11 @@ class TestWorkdirRequire:
             """,
         "pkg3_mod.egg-info/PKG-INFO": """
             Name: pkg3.mod
-            Version: 1.2.3
+            Version: 1.2.3.4
             """,
         "pkg4.mod.egg-info/PKG-INFO": """
             Name: pkg4.mod
-            Version: 0.42
+            Version: 0.42.1
             """,
     }
 
@@ -466,8 +466,8 @@ class TestWorkdirRequire:
         [
             ("pkg1.mod", "1.2.3", "pkg1.mod>=1"),
             ("pkg2.mod", "0.42", "pkg2.mod>=0.4"),
-            ("pkg3.mod", "1.2.3", "pkg3.mod<=2"),
-            ("pkg4.mod", "0.42", "pkg4.mod>0.2,<1"),
+            ("pkg3.mod", "1.2.3.4", "pkg3.mod<=2"),
+            ("pkg4.mod", "0.42.1", "pkg4.mod>0.2,<1"),
         ],
     )
     def test_require_normalised_name(self, tmp_path, monkeypatch, name, version, req):
--- a/pkg_resources/tests/test_pkg_resources.py
+++ b/pkg_resources/tests/test_pkg_resources.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import builtins
 import datetime
+import inspect
 import os
 import plistlib
 import stat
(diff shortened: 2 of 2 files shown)


pkg_resources/__init__.py
```diff
@@ -2911,7 +2911,7 @@ def __init__(
         platform: str | None = None,
         precedence: int = EGG_DIST,
     ) -> None:
-        self.project_name = safe_name(project_name or 'Unknown')
+        self.project_name = safe_name(project_name or 'Unknown').replace('.', '-')
```
abravalheri, line 2914: I don't think we should change `project_name`.
If I am not mistaken in all the discussions we reached the conclusion that the "publicly facing" name of projects should stay as given by the users (e.g. with `.`), the "canonicalised/normalised" name would be only a aspect of "machine driven" process, e.g. dependency resolution, file naming etc. That is also why PyPA does not require normalisation of metadata.

abravalheri (commented):
Thank you very much @di for having a look on this.


I was working on a regression test.
Would you consider adding the following to the `test_pkg_resources`?

```python
import inspect

# ...

class TestWorkdirRequire:
    def fake_site_packages(self, tmp_path, monkeypatch, dist_files):
    def fake_site_packages(self, tmp_path, monkeypatch, dist_files):
        site_packages = tmp_path / "site-packages"
        site_packages.mkdir()
        for file, content in self.FILES.items():
            path = site_packages / file
            path.parent.mkdir(exist_ok=True, parents=True)
            path.write_text(inspect.cleandoc(content), encoding="utf-8")

        monkeypatch.setattr(sys, "path", [site_packages])
        return os.fspath(site_packages)

    FILES = {
        "pkg1_mod-1.2.3.dist-info/METADATA": """
            Metadata-Version: 2.4
            Name: pkg1.mod
            Version: 1.2.3
            """,
        "pkg2.mod-0.42.dist-info/METADATA": """
            Metadata-Version: 2.1
            Name: pkg2.mod
            Version: 0.42
            """,
        "pkg3_mod.egg-info/PKG-INFO": """
            Name: pkg3.mod
            Version: 1.2.3
            """,
        "pkg4.mod.egg-info/PKG-INFO": """
            Name: pkg4.mod
            Version: 0.42
            """,
    }

    @pytest.mark.parametrize(
        ("name", "version", "req"),
        [
            ("pkg1.mod", "1.2.3", "pkg1.mod>=1"),
            ("pkg2.mod", "0.42", "pkg2.mod>=0.4"),
            ("pkg3.mod", "1.2.3", "pkg3.mod<=2"),
            ("pkg4.mod", "0.42", "pkg4.mod>0.2,<1"),
        ],
    )
    def test_require_normalised_name(self, tmp_path, monkeypatch, name, version, req):
        # https://github.com/pypa/setuptools/issues/4853
        site_packages = self.fake_site_packages(tmp_path, monkeypatch, self.FILES)
        ws = pkg_resources.WorkingSet([site_packages])

        [dist] = ws.require(req)
        assert dist.version == version
        assert dist.project_name == name
        assert os.path.commonpath([dist.location, site_packages]) == site_packages
```

pkg_resources/tests/test_pkg_resources.py
```diff
@@ -425,3 +426,56 @@ def test_normalize_path_backslash_sep(self, unnormalized, expected):
...
+        ],
+    )
+    def test_require_normalised_name(self, tmp_path, monkeypatch, name, version, req):
+        # https://github.com/pypa/setuptools/issues/4853
+        site_packages = self.fake_site_packages(tmp_path, monkeypatch, self.FILES)
+        ws = pkg_resources.WorkingSet([site_packages])
+
+        [dist] = ws.require(req)
+        assert dist.version == version
+        assert dist.project_name == name
```
abravalheri, line 480:
```python
assert dist.project_name == name
```

Sorry apparently this expectation was misleading from my part.

di: Yeah, looks like underscores are already being converted to dashes for `project_name` here vis `safe_name` prior to this change. Do you want to assert on that instead?

abravalheri: I think we can simply remove this assert. We possibly can have slightly different versions for each mock distribution so that they don't overlap, but not fundamental.

di: Dropped it in 65e078f like you did in 2c242238f536c4b942812632ba9dd0b1c48b4b85.

abravalheri:
@krpatter-intc

Would you be able to test if this PR solves your problem?

I believe that can use `"setuptools @ https://github.com/di/setuptools/archive/refs/heads/fix/4853.zip"` (or some variation of this) in your `pyproject.toml` to download the version in this PR.

If it does not work, could you then have a look on https://github.com/pypa/setuptools/pull/4856 (`"setuptools @ https://github.com/abravalheri/setuptools/archive/refs/heads/issue-4853.zip"`)?

krpatter-intc:
> @krpatter-intc, would you be able to test if this PR solves your problem?

Yep, I can confirm this fixes `pkg_resources.require` to take any format from the test package I built (and confirmed it was built/installed using the 75.8.1 release):

```
>>> pkg_resources.require("sample.module_under")
[sample-module-under 1 (c:\temp\example\.venv\lib\site-packages)]
>>> pkg_resources.require("sample-module_under") 
[sample-module-under 1 (c:\temp\example\.venv\lib\site-packages)]
>>> pkg_resources.require("sample-module-under") 
[sample-module-under 1 (c:\temp\example\.venv\lib\site-packages)]
>>>
```

Thanks for the quick turnaround!

abravalheri:
Thank you very much @krpatter-intc.

My initial idea was to go with this PR, but there are some tests failing, so we might go instead with #4856 (I think it will also solve your problem).

PR #4857, opened 2025-02-27
Reduce Ruff configs that duplicates upstream after skeleton merge

Avasam:
## Summary of changes

These configs are now repeated with skeleton upstream and can be removed.

### Pull Request Checklist
- [N/A] Changes have tests
- [N/A] News fragment added in [`newsfragments/`].

[PR docs]:

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -10,13 +10,13 @@ exclude = [
 [lint]
 extend-select = [
 	# upstream
- 
+
 	"C901", # complex-structure
 	"I", # isort
 	"PERF401", # manual-list-comprehension
 	"W", # pycodestyle Warning
- 
- 	# Ensure modern type annotation syntax and best practices
+
+	# Ensure modern type annotation syntax and best practices
 	# Not including those covered by type-checkers or exclusive to Python 3.11+
 	"FA", # flake8-future-annotations
 	"F404", # late-future-import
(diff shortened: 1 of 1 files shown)


ruff.toml
```diff
@@ -30,21 +30,17 @@ extend-select = [
 
 	# local
 	"ANN2", # missing-return-type-*
-	"F", # Pyflakes
```
abravalheri, line 33: The `"F"` rule is not present in skeleton.
Should I assume that it is automatically enabled by `ruff`?

ruff.toml
```diff
@@ -10,13 +10,13 @@ exclude = [
...
- 
+
 	"C901", # complex-structure
 	"I", # isort
 	"PERF401", # manual-list-comprehension
 	"W", # pycodestyle Warning
- 
- 	# Ensure modern type annotation syntax and best practices
+
+	# Ensure modern type annotation syntax and best practices
```
abravalheri, line 19: These cleanups of whitespace ideally should be done in `skeleton`, otherwise there might be some merge conflicts in the future, isn't it?

abravalheri (approved):
Thank you very much @Avasam

ruff.toml
Avasam: No need to assume, here's the reference! 
- https://docs.astral.sh/ruff/linter/#rule-selection
- https://docs.astral.sh/ruff/settings/#lint_extend-select

ruff.toml
Avasam: Those whitespaces are odd. (btw that was autoformatted). Not sure if I accidentally added them in the skeleton or if Jason accidentally introduced them when merging. I'll take a look, but yeah.

Avasam: https://github.com/jaraco/skeleton/pull/164

abravalheri:
Thank you very much.

PR #4861, opened 2025-03-04
Generate `distutils-stubs` on install

### LINKED ISSUE #4689: [FR] Vendor distutils stubs

Avasam:
## Summary of changes

Dependent on https://github.com/pypa/distutils/pull/382

3rd iteration to close #4689 (doesn't solve *everything* mentioned in that issue, but it would greatly reduce the scope so I'd prefer re-opening as a new issue)

This is built off of https://github.com/pypa/setuptools/pull/4704 where the stubs are automatically generated, but now they're done on install instead of existing in the source code.

Further Advantages over #4704:
- Doesn't pollute the source or history

Further Disadvantages over #4704:
- Contributors must install the local setuptools repo in their environment (`pip install .` or `pip install -e .`) to get proper type-checking in their editor on Python 3.12+ when it comes to pypa/distutils derived types. (they already had to install `types-setuptools`, this replaces that)

### Pull Request Checklist
- [x] Changes have tests (existing build+install and type-checking tests)

[PR docs]:


Avasam:
Aaaah, right, I forgot that `distutils-stubs` will never override the stdlib types from typeshed. So this only affects Python 3.12+
For end users: This is already the case with typeshed's own setuptools stubs, so that's fine.
For setuptools maintainers: This means we can't use this to get correct types in our own code.

Overall still an improvement for the end users (and typeshed won't have to special case this stub anymore) and a necessary step for https://github.com/pypa/setuptools/issues/2345

setup.py
```diff
@@ -37,6 +38,23 @@ def pypi_link(pkg_filename):
...
+    for path in _vendored_distutils_path.rglob("*.py"):
+        relative_path = path.relative_to(_vendored_distutils_path)
+        if relative_path.parts[0] == "tests":
+            continue
+        stub_path = (destination / relative_path).with_suffix(".pyi")
+        stub_path.parent.mkdir(exist_ok=True)
+        module = (
+            "setuptools._distutils."
+            + str(relative_path.with_suffix("")).replace(os.sep, ".")
+        ).removesuffix(".__init__")
```
Avasam, line 54: Alternative code for readability (but less "correct"):
```suggestion
        module = (
            ("setuptools._distutils." + str(relative_path.with_suffix("")))
            .replace(os.sep, ".")
            .removesuffix(".__init__")
        )
```

setup.py
```diff
@@ -37,6 +38,22 @@ def pypi_link(pkg_filename):
...
+def generate_distutils_stubs(destination: Path) -> None:
+    for path in _vendored_distutils_path.rglob("*.py"):
+        relative_path = path.relative_to(_vendored_distutils_path)
+        if relative_path.parts[0] == "tests":
+            continue
+        stub_path = (destination / relative_path).with_suffix(".pyi")
+        stub_path.parent.mkdir(exist_ok=True)
+        module = "setuptools._distutils." + str(relative_path.with_suffix("")).replace(
+            os.sep, "."
+        ).removesuffix(".__init__")
```
Avasam, line 53: Alternative code for readability (but less "correct"):
```suggestion
        module = (
            ("setuptools._distutils." + str(relative_path.with_suffix("")))
            .replace(os.sep, ".")
            .removesuffix(".__init__")
        )
```

Playing with it some more:
```suggestion
        module = (
            "setuptools._distutils."  # (don't unwrap)
            + str(relative_path)
            .removesuffix(".py")
            .replace(os.sep, ".")
            .removesuffix(".__init__")
        )
```

Avasam:
@jaraco I think that this version has minimal (if any) downsides, solves real issues, is literally as good as typeshed's and works towards https://github.com/pypa/setuptools/issues/2345

Avasam:
I think this doesn't work with tox because of PEP660 editable installs? I'll need help to get that working.

Avasam:
I very nearly got it working with a custom build backend to support PEP 660 editable installs, which also happens to leave the stubs in the source (which I can just `.gitignore`). Bringing all the benefits of #4704 .
But I'm currently blocked on this mypy issue: https://github.com/python/mypy/issues/18775

build_with_distutils_stubs.py
```diff
@@ -0,0 +1,74 @@
...
+        module = "setuptools._distutils." + str(relative_path.with_suffix("")).replace(
+            os.sep, "."
+        ).removesuffix(".__init__")
+        if str(relative_path) == "__init__.py":
+            # Work around python/mypy#18775
+            stub_path.write_text("""\
+from typing import Final
+
+__version__: Final[str]
+""")
```
Avasam, line 46: Turns out even typeshed had to do this, I just missed it: https://github.com/python/typeshed/blob/main/stubs/setuptools/distutils/__init__.pyi

setuptools/command/build_py.py
```diff
@@ -48,21 +49,50 @@ def finalize_options(self):
...
+    def copy_file(
         self,
         infile: StrPath,
         outfile: StrPathT,
         preserve_mode: bool = True,
         preserve_times: bool = True,
         link: str | None = None,
-        level: object = 1,
-    ) -> tuple[StrPathT | str, bool]:
+        level: int = 1,
```
Avasam, line 60: This was typed as `object` because it's unused. However, distutils types it as `int`.

So either distutils should change it to `object` (or `Any`) to indicate that anything can be passed to `level` and that subclasses are *not* intended to implement it.
Or, if some subclasses are meant to implement it, then this should indeed be `level: int` even if unused.

build_with_distutils_stubs.py
```diff
@@ -0,0 +1,74 @@
+"""Generate distutils stub files inside the source directory before packaging.
+We have to do this as a custom build backend for PEP 660 editable installs.
+Doing it this way also allows us to point local type-checkers to types/distutils,
+overriding the stdlib types even on Python < 3.12."""
```
Avasam, line 4: Looks like I got this working without leaving an unchecked `types/` folder in the source.
```suggestion
"""
```

Avasam:
Necessary distutils changes have been PRd in https://github.com/pypa/distutils/pull/357

setuptools/command/build_ext.py
```diff
@@ -109,7 +112,7 @@ def _get_inplace_equivalent(self, build_py, ext: Extension) -> tuple[str, str]:
 
     def copy_extensions_to_source(self) -> None:
         build_py = self.get_finalized_command('build_py')
-        for ext in self.extensions:
+        for ext in self.extensions or ():
```
Avasam: This would previously raise if `self.extensions` was `None`

setuptools/config/setupcfg.py
```diff
@@ -183,7 +182,7 @@ def parse_configuration(
             command_options,
             ignore_option_errors,
             ensure_discovered,
-            distribution.package_dir,
+            dict(distribution.package_dir),
```
Avasam: distutils' `Distribution.package_dir` is declared as `Mapping[str, str]`

setuptools/discovery.py
```diff
@@ -331,7 +331,7 @@ def _root_dir(self) -> StrPath:
     def _package_dir(self) -> dict[str, str]:
         if self.dist.package_dir is None:
             return {}
-        return self.dist.package_dir
+        return dict(self.dist.package_dir)
```
Avasam: distutils' `Distribution.package_dir` is declared as `Mapping[str, str]`, but a non-dict would fail later on `_package_dir.copy()`.

setuptools/_distutils/dist.py
```diff
@@ -251,7 +251,9 @@ def __init__(self, attrs: MutableMapping[str, Any] | None = None) -> None:  # no
         # Distribution as a convenience to the developer.
         self.packages: list[str] | None = None
         self.package_data: dict[str, list[str]] = {}
-        self.package_dir: Mapping[str, str] | None = None
+        # Mutated in place by setuptools' auto-discovery,
+        # so keep it a `dict` rather than a read-only `Mapping`.
+        self.package_dir: dict[str, str] | None = None
```
Avasam, line 256: Must be changed upstream in distutils

Done in https://github.com/pypa/distutils/pull/431

Issue #4921, opened 2025-03-25
Consider accepting `-` and `_` in extras name


jamesliu4c:
Here is the changelog message for 78.0.2:

> Postponed removals of deprecated dash-separated and uppercase fields in setup.cfg. All packages with deprecated configurations are advised to move before 2026.

I advocate that the better option is to accept `-` and `_`, and remove the deprecation warning. Anything else would risk major breaking changes throughout the ecosystem again.

I traced down the issue where making the deprecation into a breaking change was discussed. https://github.com/pypa/setuptools/issues/4864#issuecomment-2708805236.

Quoting rather liberally,

> Now we have a couple of options:

> 1. Transform the deprecation warning in a proper error, to force the end-users to implement the necessary changes in their code-base.
> 2. Postpone it if we feel that the community is not ready for the change yet.
> 3. Give-up entirely on the deprecation on the basis that there might be unmaintained packages out there only published to PyPI as sdists important to the environment.
> a. We can just settle for forever warning the deprecation without removing backwards compatibility.
> b. Change the parsing of setup.cfg accept both - or _ without warnings.

And @jaraco said in [response](https://github.com/pypa/setuptools/issues/4864#issuecomment-2709033980):

> That said, I'm leaning toward option 1, get some signal out there to non-compliant projects, and then if the disruption is large, rollback and fallback to option 2 (now with the affected projects having visibility to the issue).

The ticket was closed when the tests were marked as xfail [here](https://github.com/pypa/setuptools/commit/be965883cfa5b210dd009e9d9f56d785b6b51d6a).

@zanieb warned that the disruption would be quite large [here](https://github.com/pypa/setuptools/issues/4864#issuecomment-2749855688):

> > That said, I'm leaning toward option 1, get some signal out there to non-compliant projects,
>
> I think rolling out changes like this is quite disruptive to the ecosystem and is to aggressive for a foundational package like this. I would appreciate more effort determining if changes like this would be disruptive ahead of time.
> ...
> > Change the parsing of setup.cfg accept both - or _ without warnings.
>
> Is there a downside to this approach? It looks like this was originally deprecated because the extra names were not installed correctly by pip (https://github.com/pypa/setuptools/issues/1608) but that doesn't seem like a strong motivation? (I tried pip install '.[foo-b]' with the extra name foo_b and it seems to work fine now)

[...]

Issue #4913, opened 2025-03-24
[BUG] Setuptools allow arbitrarily wrong configs in `setup.cfg`


abravalheri:
The issue in https://github.com/pypa/setuptools/issues/4910 unveiled something much deeper.
It seems that setuptools has been allowing arbitrary fields (like `description_file` described in #4910) in `setup.cfg` (possibly without a visible warning -- needs investigation)[^1].

***Expected outcome***

Setuptools should issue a deprecation warning when a invalid field in used in `setup.cfg`.

[^1]: Funny thing is that issue #4910 always involves `description-file`. I find it hard to believe that it is a coincidence, possibly many packages CTRC+C, CTRL+V the same config all over the place.

zr40:
`description-file` in `setup.cfg` might have originated from PBR in 2013 (then called oslo.packaging): https://github.com/openstack/pbr/blob/aa4641f0e6e668d4eb3a0c1f70a3e15e83097c6a/README.rst and they still support this field today, albeit in the underscore form: https://github.com/openstack/pbr/blob/master/pbr/util.py#L321

fungi:
It goes back even further, PBR (nee oslo.packaging) was a fork of Distutils2, which implemented the same `description-file` option.

aman2454:
I have seen a different one than description-file, 
` setuptools.errors.InvalidConfigError: Invalid dash-separated key 'force-manifest' in 'sdist' (setup.cfg), please use the underscore name 'force_manifest' instead`

fungi:
The `force-manifest` entries look like they're for distutils, which allows passing default command-line options through setup.cfg files, and `--force-manifest` is one such distutils option.

I suppose these examples raise a question: should other packaging tools and SetupTools plugins be allowed to extend setup.cfg with their own directives alongside SetupTools use of the same file?

Apparently some do, and have for years. The heritage of setup.cfg files goes back beyond PBR/d2to1/Distutils2 to the original distutils which relied on it for declarative configuration decades before SetupTools (re)copied that design (you'll find references to setup.cfg files for distutils in documentation from 2001, possibly earlier). Allowing and ignoring unrecognized entries has provided backward-compatibility with existing packages already relying on this functionality for a very long time.

abravalheri:
> I suppose these examples raise a question: should other packaging tools and SetupTools plugins be allowed to extend setup.cfg with their own directives alongside SetupTools use of the same file?

[...]

PR #4923, opened 2025-03-26
Downgrade hyphen and uppercase deprecation warning to info

jamesliu4c:
## Summary of changes
This changes `enforce_underscore` and `enforce_option_lowercase` from deprecation warnings to information only.

Closes 
https://github.com/pypa/setuptools/issues/4921

### Pull Request Checklist
~~- [] Changes have tests~~

[PR docs]:

### DIFF
--- a/newsfragments/4923.misc.rst
+++ b/newsfragments/4923.misc.rst
@@ -1,0 +1,1 @@
+This changes ``enforce_underscore`` and ``enforce_option_lowercase`` from deprecation warnings to information only.
--- a/setuptools/dist.py
+++ b/setuptools/dist.py
@@ -628,17 +628,16 @@ class Distribution(_Distribution):
 
         underscore_opt = opt.replace('-', '_')
         affected = f"(Affected: {self.metadata.name})." if self.metadata.name else ""
-        SetuptoolsDeprecationWarning.emit(
+        InformationOnly.emit(
             f"Invalid dash-separated key {opt!r} in {section!r} (setup.cfg), "
             f"please use the underscore name {underscore_opt!r} instead.",
             f"""
-            Usage of dash-separated {opt!r} will not be supported in future
-            versions. Please use the underscore name {underscore_opt!r} instead.
+            Usage of dash-separated {opt!r} is not supported by Python packaging
+            specifications. Please use the underscore name {underscore_opt!r} instead.
             {affected}
             """,
             see_docs="userguide/declarative_config.html",
-            due_date=(2026, 3, 3),
-            # Warning initially introduced in 3 Mar 2021
+            # Changed from warning to info after https://github.com/pypa/setuptools/issues/4910
         )
         return underscore_opt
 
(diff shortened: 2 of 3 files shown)


newsfragments/4923.misc.rst
```diff
@@ -0,0 +1 @@
+This changes ``enforce_underscore`` and ``enforce_option_lowercase`` from deprecation warnings to information only.
```
webknjaz, line 1: This is supposed to read coherently in a change log.

webknjaz:
Could you link a post where maintainers decided that this is acceptable?

newsfragments/4923.misc.rst
jamesliu4c: Ok, so how do you suggest I word this?

jamesliu4c:
> Could you link a post where maintainers decided that this is acceptable?

I am not sure this has been decided it is acceptable. It seems that there is not consensus based on the conversation in [the issue](https://github.com/pypa/setuptools/issues/4921), so I made this PR to clarify my meaning. If it is better to mark as draft, we can do that.

abravalheri (changes requested):
Thank you very much @jamesliu4c for working on this but unfortunately this is a -1 from my side (other maintainers may have different opinions).

Before describing the reasons for the -1, I would like to clarify that with the benefit of the hindsight, it seems requiring `_` for setuptools config fields is not really the core problem here. Based on the coarse feedback we received, if setuptools was dealing only with fields defined by itself, the impact of requiring `_` do not seem very strong. The impression is that the community has sorted that one out. The crux of the problem is apparently #4913, which is something completely unforeseen when we studied what to do with the deprecation warning introduced many years ago.

(I am not saying that setuptools should necessarily go around failing for fields it does not recognise, but that is a discussion in the context of #4913).

Now I apologise if I was not clear enough in my comment in https://github.com/pypa/setuptools/issues/4921#issuecomment-2752837435. I am (and I believe other maintainers would also be) are very happy to consider other alternatives that would solve the problem in a better way. But the alternative would need to be a **better** way of doing things, specifically:

- Can we rework the code in a way that it is simplified and at the same time can handle configurations with dashes?
- Can we change the code in a way that it seamlessly handle the dash x underscore dilemma and reduce the bloat in `setuptools.dist`?

(I also added a couple of suggestions in https://github.com/pypa/setuptools/issues/4921#issuecomment-2752837435).

Particularly it is also dear to me that we avoid having to read entry-points metadata with `importlib` (and without having to hardcode or CTRL+C,CTRL+V the information of the existing commands that exists in `pyproject.toml`). I think that operation is too complex/costly if the intention is to simply accept dashes[^1].

While we are open to discuss and consider accepting `-`, I think that it should come with a counterpart that also improves the codebase and/or performance (or at least a very compelling study of the state of the community and a convincing why the deprecation can not be handled by the packages holding down the change).

[^1]: I do need to clarify here that PR 4870 itself did not eliminate loop-ing through entry-points, but it was a step towards that direction. The initial idea was to give the community a very specific error easy to interpret and fix for a while before completely removing the convoluted mechanisms that identify the non compliant fields/sections (I believe that PR 4870 tided-up the previous implementation and potentially fixed some bugs, but alone it was not the final destination).

newsfragments/4923.misc.rst
webknjaz: Well, that looks like a TODO-list item because it uses infinitive. It's usually present simple or the past tense, explaining what the impact is for the end-user.

jaraco:
I don't have a strong opinion on it, so I defer to Anderson's opinion. Therefore, I'm closing this as unmerged. We can revisit if a compelling argument can be made that convinces me or (preferably) Anderson.

PR #4945, opened 2025-04-13
easy_install.get_win_launcher: Default to pythonw.exe in GUI mode

hmartinez82:
## Summary of changes

When creating a launcher script where no explicit python binary is defined, `pythonw.exe` should be used in Windows when in GUI mode.

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/launcher.c
+++ b/launcher.c
@@ -291,8 +291,8 @@ int run(int argc, char **argv, int is_gui) {
     *ptr-- = '\0';
 
     if (strncmp(python, "#!", 2)) {
-        /* default to python.exe if no #! header */
-        strcpy(python, "#!python.exe");
+        /* default to python[w].exe if no #! header */
+        strcpy(python, is_gui ? "#!pythonw.exe" : "#!python.exe");
     }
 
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for the contrib. It does seem like a reasonable change.

Nevertheless, I don't think there's any life left in easy_install or the script launchers. These installers are deprecated and unsupported except for critical security updates. Moreover, any changes to the launcher will require a rebuild of the script launchers, and I'm not even sure if that's possible these days.

I don't plan to merge this, but if you could tell us what motivated this change, it would be good to document, in case it affects others with the same use-case.

hmartinez82:
@jaraco The last file in this PR shows why https://github.com/msys2/MINGW-packages/pull/23966/files

jaraco:
Thanks for the clarification. Since easy_install is being actively removed, I believe this issue is moot. Glad there's a suitable workaround. A better workaround would be to use build/pip/uv to build/install packages and rely on the scripts they provide.

PR #4949, opened 2025-04-17
Fix error message string formatting

ngoldbaum:
I hit this error in the wild and noticed it had a typo.


abravalheri:
Thank you very much for catching this!

PR #4955, opened 2025-04-20
Rely on pip for setup.py develop

jaraco:
Removes reliance on easy_install while retaining a develop command, providing some extended support for the develop command while unblocking removal of easy_install.

Ref #917

- **Replace develop command with redirection to pip.**
- **Ensure the deprecation warning is retained.**
- **Bring back support for no-deps.**

## Summary of changes

Closes 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
There are almost certainly some incompatibilities with this change. For example, it increases the reliance on pip. It removes uninstall support (although uninstall is still available using pip). The output almost certainly changes. Also, because the installation goes through PEP 517, the installation layout will be different.

That said, this feature is intended to provide a soft landing for the users and workflows still reliant on `setup.py develop`. Users truly reliant on the prior behavior can pin to the prior version of Setuptools.

PR #4962, opened 2025-04-27
Update test to honor new behavior in importlib_metadata 8.7

### LINKED ISSUE #4961: TestInvalidEntryPoint failing following release of importlib_metadata 8.7
In https://github.com/python/importlib_metadata/pull/518, `importlib_metadata` started adding some validation for invalid entry points. Now they're no longer allowed to be constructed and will raise a ValueError if invalid input is presented.

This new behavior clashes with Setuptools' own expectation that the entry points will fail lazily when relevant attributes are accessed and what exception is thrown.

This change seems to have blocked the Setuptools 80 release because somehow `importlib_metadata` is getting used at head, presumably because some other package is pulling it in.

jaraco:
Fixes #4961

## Summary of changes

Closes 

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/newsfragments/4961.bugfix.rst
+++ b/newsfragments/4961.bugfix.rst
@@ -1,0 +1,1 @@
+Update test to honor new behavior in importlib_metadata 8.7.
--- a/setuptools/_entry_points.py
+++ b/setuptools/_entry_points.py
@@ -15,6 +15,10 @@ def ensure_valid(ep):
     """
     Exercise one of the dynamic properties to trigger
     the pattern match.
+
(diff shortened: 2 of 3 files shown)


jaraco:
I started looking into specifically trapping the error and replacing it with an OptionError, but ultimately I decided that would add too much complication, and the whole point of validating the error in importlib metadata is so that other libraries don't have to do so, so I've instead marked the validation routine as deprecated in favor of eventual retirement when importlib_metadata 8.7 or later (and its stdlib equivalent) are available.

PR #4973, opened 2025-05-02
Short-circuit get_requires hooks when there are no extra build dependencies

ichard26:
## Summary of changes

The PEP 517 `get_requires_for_build_(wheel|sdist)` hooks function by running the `egg_info` command with a patched Distribution that raises a special exception when build dependencies (from `setup_requires`) are about to be installed.

This works great at minimizing the overhead of the requires hooks... except when there are no dynamic build dependencies. In such a situation, the injected `install_build_eggs()` shim that raises `SetupRequirementsError` is never invoked, thus the entire `egg_info` command will execute. This often results in substantial overhead for modern projects that use `[build-system].requires` and not `setup_requires`.

With this simple script at the root of the pip project, this change brings the execution time from ~300ms to ~100ms.

```python
import setuptools.build_meta
setuptools.build_meta.get_requires_for_build_wheel()
```

This helps a fair bit with reducing the overhead with PEP 517 processing.


ichard26:
I have no idea why this is causing the prepare metadata hooks to blow up. If anyone else has any ideas, I'm all ears. I'm utterly stumped.

ichard26:
Oh, it looks like the `dist-info` command depends on the `egg-info` command having ran before it...? That's going to kill this idea if true.

jaraco:
> Oh, it looks like the `dist-info` command depends on the `egg-info` command having ran before it...? That's going to kill this idea if true.

I do believe that's the case. Soon, however, the biggest dependencies on egg_info (easy_install) should be removed, which gets us much closer to being able to consolidate the metadata handling behavior in dist_info instead of egg_info, so maybe there's hope.

abravalheri:
> Oh, it looks like the dist-info command depends on the egg-info command having ran before it...?

It is a bit more nuanced, I believe, `dist-info` will try to to run `egg-info` if it has not run yet. But there might be some caching happening.

---

> I have no idea why this is causing the prepare metadata hooks to blow up. If anyone else has any ideas, I'm all ears. I'm utterly stumped.

I think that this part of the source code is very fragile... `build_meta` is patching something that has already being patched by `setuptools/__init__.py`. It is way too much patching and things are not really obvious.

Ultimately this happens because it is not very straight forward to align PEP 517-style build hooks with existing `setup.py` scripts. The expectations of who detains the control flow do not line up. So the existing `build_meta` implementation hacks a `goto-like` feature to circumvent these limitations.


In https://github.com/pypa/setuptools/compare/main...abravalheri:setuptools:issue-4973?expand=1, I am studying if we can reduce the patching... Let's see if it find the same errors happening...

ichard26:
That branch looks quite promising! Let me know if you turn it into a PR and I'll be happy to verify whether it improves `get_requires_for_build_(wheel|sdist)` performance like this PR.

ichard26:
Taking a closer look at @abravalheri's branch, it seems like https://github.com/pypa/setuptools/commit/70d4fc75bb431011546dee6cc9e46a48b20a75d1 is the ticket to fixing the dist-info issues during bootstrapping, thus no major refactoring is necessary to achieve the intended performance improvement. PTAL!

ichard26:
@abravalheri while #5008 undergoes discussion and review, it would be nice to consider landing this small (but meaningful!) optimization. It improves the performance of the get requires hooks for the vast majority of projects (which do not declare any extra build dependencies). The main risk is that projects relying on their `setup.py` executing fully while setting up the build environment. This is already precarious though as build isolation may be disabled (thus the get requires hooks are skipped entirely) and setup.py is already skipped if there **are** extra build dependencies.

TIA! Let me know if there is anything else I need to do to move this forward (other than wait, of course). Edit: I'm aware CI is red, but it was red on main when I last pushed. I'll try rebasing this on main.

PR #4992, opened 2025-05-11
skeleton sync

### LINKED ISSUE #4981: sdist is missing license file
[Recently](https://github.com/pypa/setuptools/commit/9a81db3c77bc106017dcd4b0853a5a94f43ae33c) the license file got removed from git and so it is no longer in the sdist package too.  I think this is wrong because the MIT license explicitly states this:

```
The above copyright notice and this permission notice shall be included in
all copies or substantial portions of the Software.
```

and so I believe the full license wording should be included in the sdist package.

Disclaimer: IANAL

jaraco:
- **Python 3 is the default nowadays (jaraco/skeleton#173)**
- **Add coherent.licensed plugin to inject license texts into the build.**
- **Revert "Python 3 is the default nowadays (jaraco/skeleton#173)" (jaraco/skeleton#175)**

## Summary of changes

The main change here is the introduction of the 'coherent.licensed' plugin. This plugin takes the declared license expression and uses it to inject the corresponding license into the sdist and wheel artifacts.

`coherent.licensed` is developed using the `coherent.build` backend (and system), but that means the sdist is distributed using flit-core, so doesn't introduce any circular dependencies at build time.

Closes #4981

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 80.4.0
+current_version = 80.5.0
 commit = True
 tag = True
 
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,3 +1,18 @@
+v80.5.0
+=======
+
+Features
+--------
+
+- Replaced more references to pkg_resources with importlib equivalents. (#3085)
+
+
+Misc
+----
+
+- #4982
+
+
 v80.4.0
 =======
 
--- a/docs/userguide/interfaces.rst
+++ b/docs/userguide/interfaces.rst
@@ -7,7 +7,7 @@ In addition to operating as a library, because newer versions of Setuptools are 
 
 In order to have the ability to make sensible changes to the project, downstream developers and consumers should avoid depending on internal implementation details of the library and should rely only on the supported interfaces:
 
-- *Tier 1*: APIs required by modern PyPA packaging standards (:pep:`517`, :pep:`660`) and Documented APIs for customising build behavior or creating plugins (:doc:`/userguide/extension`, :doc:`references/keywords`):
+- *Tier 1*: APIs required by modern PyPA packaging standards (:pep:`517`, :pep:`660`) and Documented APIs for customising build behavior or creating plugins (:doc:`/userguide/extension`, :doc:`/references/keywords`):
 
    These APIs are expected to be extremely stable and have deprecation notices and periods prior to backward incompatible changes or removals.
 
--- a/newsfragments/3085.feature.rst
+++ b/newsfragments/3085.feature.rst
@@ -1,0 +1,1 @@
+Replaced more references to pkg_resources with importlib equivalents.
--- a/newsfragments/3085.feature.rst
+++ b/newsfragments/3085.feature.rst
@@ -1,1 +1,0 @@
-Replaced more references to pkg_resources with importlib equivalents.
--- a/newsfragments/4981.feature.rst
+++ b/newsfragments/4981.feature.rst
@@ -1,0 +1,1 @@
+Added a build dependency on coherent.licensed to inject the declared license text at build time.
--- a/newsfragments/4982.misc.rst
+++ b/newsfragments/4982.misc.rst
@@ -1,1 +1,0 @@
-Mark more tests failing without Internet access as ``uses_network``.
--- a/newsfragments/4995.misc.rst
+++ b/newsfragments/4995.misc.rst
@@ -1,0 +1,1 @@
+Fixed ``no attribute 'join_path'`` error in Windows test
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -440,11 +440,7 @@ def _macos_arch(machine):
 
 
 def get_build_platform():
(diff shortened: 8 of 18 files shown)


abravalheri:
This possibly brings the same dependency cycle limitations of PEP 517, described in https://github.com/coherent-oss/coherent.build/issues/32.

abravalheri:
Other than the discussion on the `coherent.build` cycles,
it seems that the plugin also brings dependencies:

https://inspector.pypi.io/project/coherent-licensed/0.1.1/packages/7c/99/6ce7f20634d19e4a37b597e14d26ae9e9cdebcc534b468e465dad7ae6d34/coherent_licensed-0.1.1.tar.gz/coherent_licensed-0.1.1/pyproject.toml#line.11

That will introduce a dependency cycle: https://github.com/psf/requests

jaraco:
> So my understanding is that `coherent.build` replaces the `pyproject.toml` file in the `sdist`?

The original design was that `coherent.build` wouldn't have a pyproject.toml at all in the repo source code (and in fact still doesn't), and it would inject a minimal pyproject.toml (with the build-system table) to any sdist.

The more recent design based on https://github.com/coherent-oss/system/issues/12 instead "compiles" the repo source into a flit-core-based sdist. It's still opt-in, but I'm inclined to make that the default behavior, because it seems to be the _only_ solution I've found to the [bootstrapping problem](https://github.com/pypa/packaging-problems/issues/342).

> Or maybe that is the idea for future releases, but not present in 0.33.1 yet?
> 
> https://inspector.pypi.io/project/coherent-build/0.33.1/packages/26/7d/39d4e1aaf4b25034845756372d610f9a414e3f72ab45a36075380813dd5c/coherent_build-0.33.1.tar.gz/coherent_build-0.33.1/pyproject.toml

Whoops. I had to do a manual release of 0.33.1 due to breakage in the 0.33.0 version and I forgot to opt-in to the flit-core backend. See [0.33.0](https://inspector.pypi.io/project/coherent-build/0.33.0/packages/35/43/5e4001ee6c9486e1e8cbc45abdd0c1a13902cbded3d333600f3bce541088/coherent_build-0.33.0.tar.gz/coherent_build-0.33.0/pyproject.toml) for an example of how the code gets compiled to flit. I can simply cut another release 0.33.2 and it will have the same behavior ([done](https://gist.github.com/jaraco/08532d5d3d3324156a98dfb1277fcd33)).



> it seems that the plugin also brings dependencies:

Yes, indeed. It looks like requests is using setuptools. Thanks for catching that. That's annoying, and somewhat out of my control. I wish there were a proper solution to the bootstrapping problem that didn't involve migrating every package to flit-core or re-writing software to use suboptimal dependencies. I could probably rewrite coherent.licensed not to use requests, but that is only a bandaid and an existential risk (should another dependency creep in).

Fact of the matter it's the same people insisting that we need a LICENSE injected that also are imposing this bootstrapping problem. Nobody's solving the problem, just insisting on it.

\</rant>

I'll update coherent.licensed not to depend on requests. Then let's proceed with this change, no?

abravalheri:
> I've [updated](https://inspector.pypi.io/project/coherent-licensed/0.2.0/packages/87/6e/339f698e3de95a2019377c43c2525db2001d0534e9f9d5b217eb74863c90/coherent_licensed-0.2.0.tar.gz/coherent_licensed-0.2.0/pyproject.toml) coherent.licensed not to depend on requests. Then let's proceed with this change, no?

Thank you very much Jason for working on this.
From my point of view this should be fine.

jaraco:
Merged as https://github.com/pypa/setuptools/commit/b74789e2aa3227e85d61b40708959b35d7f666cc

rgommers:
> Fact of the matter it's the same people insisting that we need a LICENSE injected that also are imposing this bootstrapping problem. Nobody's solving the problem, just insisting on it.

Let's be very clear here. *You* created this problem all by yourself by deleting the license file. Which you could simply put back, but that would be admitting you made a mistake, which you seem unable to do. Hence *you* are also the one responsible for the bootstrapping problem here.

In addition, you broke build hermeticity when building directly from a release tag (some distros do this) with this PR, because the new build dependency downloads content from the internet. See https://packaging.python.org/en/latest/discussions/downstream-packaging/#do-not-use-the-internet-during-the-build-process for why that's a problem.

tiran:
Breaking hermetic builds and introducing cyclic dependencies is not acceptable.

The removal of the license file from git and sdists is in violation of my contributions to setuptools and setuptools' license. You have until 2025-05-16 AOE to restore the license file.

tiran:
To be absolute clear here: I am not asking you politely to please, please consider restoring the license file.

I'm telling you that you **MUST** restore the license file ASAP. There is zero room for debate or backtalk. Your actions are in violation of setuptools' license and my rights as a contributor to setuptools and distutils. I'm taking OSS licenses very serious and will **NOT** accept any objection.

A bunch of people including me have tried to convince you on multiple occasions over the past few weeks. You have been ignoring everybody. I hope that a direct, point-blank demand will make you understand how serious your violation is.

CC: CoC working group members @di @gaborbernat @dstufft

PR #5006, opened 2025-05-20
Removed pkg_resources from test_wheel.

jaraco:
- **Remove test_wheel_install_pep_503.**
- **Remove reliance on pkg_resources in test_wheel.**
- **Add news fragment.**

## Summary of changes

Ref #3085

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/newsfragments/3085.feature.rst
+++ b/newsfragments/3085.feature.rst
@@ -1,0 +1,1 @@
+Removed reliance on pkg_resources in test_wheel.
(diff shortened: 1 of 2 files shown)


jaraco:
Oh, but the pyright stuff should probably be addressed.

jaraco:
The pyright tests are failing now that I'm using `importlib.metadata` for the metadata and `__getitem__` in older versions of the package could possibly return None. I don't care. I want them to crash or fail if the keys aren't present. I'll just try to suppress the pyright errors.

Issue #4073, opened 2023-10-07
[Docs] Clarity on transition for `pkg_resources.declare_namespace`


tonybaloney:
### Summary

I'm trying to work out how to remove uses of:

```python
import pkg_resources
pkg_resources.declare_namespace(__name__)
```

Now that `pkg_resources` is deprecated.

There is a docs section on this:

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#pkg-resource-style-namespace-package

Which suggests the following snippet:

```python
__import__("pkg_resources").declare_namespace(__name__)
```

Won't this also not work because `pkg_resources` is deprecated?

### Additional Information

na

### Code of Conduct

- [X] I agree to follow the PSF Code of Conduct

abravalheri:
Please note that the docs don't actually suggest replacing it with `__import__("pkg_resources").declare_namespace(__name__)`.

In the [documentation's page](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#legacy-namespace-packages) that you mentioned, there is a clear reference that `pkg_resources` is deprecated:

> Historically, there were two methods to create namespace packages. One is the `pkg_resources` style supported by `setuptools` and the other one being `pkgutils` style offered by `pkgutils` module in Python. Both are now considered ***deprecated*** despite the fact they still linger in many existing packages. These two differ in many subtle yet significant aspects and you can find out more on [Python packaging user guide](https://packaging.python.org/guides/packaging-namespace-packages/).

Instead the docs contain an [example](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages) of namespace package implemented using the technique described in [PEP 420](https://peps.python.org/pep-0420/).

The link in the Python Packaging User Guide also contains a warning about legacy namespaces and recommends PEP 420 approach (a.k.a. "native namespace packages").

I am not sure which kind of clarity you would be interested... Could you please provide a PR with your suggestion on how to improve it[^1]?

[^1]: Please just have in mind that it is important to keep some form of the documentation of the legacy/deprecated approach, so people supporting old packages can still find helpful documentation. That is why we cannot simply exclude all the references completely.

Issue #2463, opened 2020-11-25
Tests fail in my environment and I do not know why


MeggyCal:
Hi, I am trying to run tests for setuptools 50.3.2 for openSUSE Tumbleweed (so most packages are in their newest version) and I am getting the following error:

```
[  228s] ==================================== ERRORS ====================================
[  228s] ___________________ ERROR at setup of test_distutils_stdlib ____________________
[  228s] 
[  228s] tmpdir = local('/tmp/pytest-of-abuild/pytest-0/test_distutils_stdlib0')
[  228s] 
[  228s]     @pytest.fixture
[  228s]     def venv(tmpdir):
[  228s]         env = VirtualEnv()
[  228s]         env.root = path.Path(tmpdir)
[  228s]         env.req = os.getcwd()
[  228s] >       return env.create()
[  228s] 
[  228s] setuptools/tests/test_distutils_adoption.py:28: 
[  228s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  228s] /usr/lib/python3.8/site-packages/jaraco/envs.py:24: in create
[  228s]     self.install()
[  228s] /usr/lib/python3.8/site-packages/jaraco/envs.py:39: in install
[  228s]     subprocess.check_call(cmd, env=env)
[  228s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[  228s] 
[  228s] popenargs = ([Path('/tmp/pytest-of-abuild/pytest-0/test_distutils_stdlib0/.env/bin/python'), '-m', 'pip', 'install', '/home/abuild/rpmbuild/BUILD/setuptools-50.3.2'],)
[  228s] kwargs = {'env': {'COLORTERM': '1', 'CPU': 'i686', 'CSHEDIT': 'emacs', 'FORCE_SOURCE_DATE': '1', ...}}
[  228s] retcode = 1
[  228s] cmd = [Path('/tmp/pytest-of-abuild/pytest-0/test_distutils_stdlib0/.env/bin/python'), '-m', 'pip', 'install', '/home/abuild/rpmbuild/BUILD/setuptools-50.3.2']
[  228s] 
[  228s]     def check_call(*popenargs, **kwargs):
[...]

Issue #2262, opened 2020-07-15
Incompatibility between setuptools and pyInstaller


thetheo2:
We have a specific testcase as below.  
The testcase passes at setuptools==49.0.1, PyInstaller==3.6, and fails at newer releases of setuptools as below--
 ./hello_world 
Traceback (most recent call last):
  File "hello_world.py", line 4, in <module>
    import pkg_resources.py2_warn
ModuleNotFoundError: No module named 'pkg_resources'
[7155] Failed to execute script hello_world
~~~~
The hello_world.py script is as below--
import pkg_resources.py2_warn
print('hello world')
try :
    import numpy
    print(numpy.sqrt(2))
except (ImportError,AttributeError) :
    pass
try :
    import requests
    r = requests.get('https://api.github.com/repos/psf/requests')
    print(r.json()['description'])
except (ImportError,AttributeError) :
    pass
~~~~
PyInstaller is run as below--
../../pp_src/invio/bin/pyinstaller --clean -d all -F hello_world.py

jaraco:
pkg_resources is deprecated and being removed. This issue is obsolete.

Issue #2007, opened 2020-02-26
`pkg_resources.evaluate_marker` ignores the `extra` argument


ender-wieczorek:
**Problem:**

Trying to evaluate markers like `extra == "test" and python_version == "2.7"` gives an error:
```
File "/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 1378, in evaluate_marker
 return marker.evaluate()
 File "/usr/local/lib/python2.7/dist-packages/pkg_resources/_vendor/packaging/markers.py", line 301, in evaluate
 return _evaluate_markers(self._markers, current_environment)
 File "/usr/local/lib/python2.7/dist-packages/pkg_resources/_vendor/packaging/markers.py", line 226, in _evaluate_markers
 lhs_value = _get_env(environment, lhs.value)
 File "/usr/local/lib/python2.7/dist-packages/pkg_resources/_vendor/packaging/markers.py", line 208, in _get_env
 "{0!r} does not exist in evaluation environment.".format(name)
pkg_resources._vendor.packaging.markers.UndefinedEnvironmentName: 'extra' does not exist in evaluation environment.
```

An example of such marker is in the `testpath` distribution that `jupyter` depends on.

**Fix:**

```
diff --git a/pkg_resources/__init__.py b/pkg_resources/__init__.py
index 75563f95..db089550 100644
--- a/pkg_resources/__init__.py
+++ b/pkg_resources/__init__.py
@@ -1376,7 +1376,7 @@ def evaluate_marker(text, extra=None):
     """
     try:
         marker = packaging.markers.Marker(text)
-        return marker.evaluate()
+        return marker.evaluate({'extra': extra})
     except packaging.markers.InvalidMarker as e:
         raise SyntaxError(e)
```

sksDonni:
@ender-wieczorek I am encountering the same problem while working on another project. I tried the fix that was suggested to no avail. The log is something like this-

T```
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/workspace/oppia/oppia/scripts/start.py", line 32, in <module>
    from . import install_third_party_libs
  File "scripts/install_third_party_libs.py", line 54, in <module>
[...]

PR #5007, opened 2025-05-21
Remove pkg_resources

### LINKED ISSUE #3085: Setuptools should not rely on pkg_resources

jaraco:
- **Remove pkg_resources**
- **Remove assertion that 'import pkg_resources' works; it's no longer an invariant.**
- **Mark pbr integration test as xfail.**
- **Mark namespace tests as xfail. Support for pkg_resources-style namespaces is dropped.**
- **Add news fragment.**

## Summary of changes

Closes #3085

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


jaraco:
To be sure, this change is a big deal. It's been decades in the making to provide replacements for this functionality. There are known open issues and efforts to remove the dependence on pkg_resources and probably other use-cases that haven't migrated. My hope is that these cases can pin to Setuptools<81.

abravalheri (commented):
Thank you very much Jason for the effort in untangling the 2 code bases.
I agree that it is time to start thinking about the removal of `pkg_resources`.

I propose we do it in 2 stages:

1. Explicitly mention a "due date" to the [warning message](https://github.com/pypa/setuptools/blob/v80.8.0/pkg_resources/__init__.py#L98-L103)
2. Removal once the "due date" arrives.

---

I have 2 doubts:

- Does the `namespace_packages` (deprecated) keyword for `setup.py` works without `pkg_resources.declare_namespace`?
  
  If I understood correctly previously these worked in tandem, but I don't know what is the situation now after the latest changes in https://github.com/pypa/setuptools/blob/v80.8.0/setuptools/namespaces.py#L47-L63.

- Is the idea of https://github.com/pypa/pkg_resources to also publish something to PyPI? Even if the package is born deprecated it may be a "last-resort" alternative for users that rely on packages that have not received updates for a long time, or that are still working on the migration...

  I know at least that folks from `zope`/`plone` have been slowly trying to replace `pkg_resources.declare_namespace` for a while, but are not there quite yet (refs: https://github.com/pypa/setuptools/pull/3434#issuecomment-1459868662, https://github.com/zopefoundation/meta/issues/194), so maybe that would be useful if they take longer than an eventual "due date" to do so.

  Update: I can see in https://github.com/pypa/setuptools/issues/3085#issuecomment-2899334790 that this is not straight forward.

jaraco:
> * Does the `namespace_packages` (deprecated) keyword for `setup.py` works without `pkg_resources.declare_namespace`?

It's a mixed bag.

If I recall correctly, when pip-installed, a declared namespace gets its `__init__.py` omitted, so ends up acting like a PEP 420 namespace. I also seem to recall that there's a `.pth` file that gets installed under certain circumstances (although maybe not any more after the removal of easy_install).

Looking at zope.interface as an example:

```
 🐚 py -3.13 -m pip-run git+https://github.com/zopefoundation/zope.interface -- -c "import zope, pathlib; print('files in zope:', list(pathlib.Path(zope.__path__[0]).iterdir())); print('nspkg.pth file:', next(pathlib.Path(zope.__path__[0]).parent.glob('*-nspkg.pth')).read_text())"
files in zope: [PosixPath('/var/folders/f2/2plv6q2n7l932m2x004jlw340000gn/T/pip-run-r43zfapw/zope/interface')]
nspkg.pth file: import sys, types, os;p = os.path.join(sys._getframe(1).f_locals['sitedir'], *('zope',));importlib = __import__('importlib.util');__import__('importlib.machinery');m = sys.modules.setdefault('zope', importlib.util.module_from_spec(importlib.machinery.PathFinder.find_spec('zope', [os.path.dirname(p)])));m = m or sys.modules.setdefault('zope', types.ModuleType('zope'));mp = (m or []) and m.__dict__.setdefault('__path__',[]);(p not in mp) and mp.append(p)
```

The same is [true for the wheel](https://inspector.pypi.io/project/zope-interface/7.2/packages/76/71/e6177f390e8daa7e75378505c5ab974e0bf59c1d3b19155638c7afbf4b2d/zope.interface-7.2-cp310-cp310-macosx_10_9_x86_64.whl/).

These results confirm that, at least for the pip-installed packages, the use of pkg_resources is completely absent.

This confirmation gives me higher confidence that this removal is at least safer than I'd previously thought.

In fact, I wonder if `declare_namespace` is in use anywhere at all.

>       If I understood correctly previously these worked in tandem, but I don't know what is the situation now after the latest changes in https://github.com/pypa/setuptools/blob/v80.8.0/setuptools/namespaces.py#L47-L63.

There was a time where even the -nspkg.pth file did nothing and fell back to PEP 420 on Python 3.3+. In 27b7d4e47a4d80f5f377cefb87f91aa0e0110246, however, I limited the PEP 420 fallback to Python 3.3 and 3.4. I didn't leave any good justification for making that change (no linked issue, no mention in changelog). Maybe it had something to do with #1214. I wonder if that change was a mistake and could have been allowed to simply fall back to PEP 420.

> In fact, I wonder if `declare_namespace` is in use anywhere at all.

I guess it has to still be used for the editable case... at least where `setup.py develop` was used.

Confirmed - it's still used even for the modern editable case:

```
 zope.interface master 🐚 py -3.13 -m pip-run -e .
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
>>> import zope
>>> zope.__file__
'/Users/jaraco/draft/zope.interface/src/zope/__init__.py'
>>> zope.__path__
['/Users/jaraco/draft/zope.interface/src/zope']
```

I'm less worried about the editable case, because it's outside the typical integration and production workflows and the developer has a lot more control to implement workarounds, such as to pin setuptools.

> * Is the idea of https://github.com/pypa/pkg_resources to also publish something to PyPI? Even if the package is born deprecated it may be a "last-resort" alternative for users that rely on packages that have not received updates for a long time, or that are still working on the migration...

As you've observed, I was considering this option, but I'm still not confident it's the right thing to do. It would require getting the name opened up for use, which I'm fairly confident the maintainers would do for this purpose.

But it would still require the users to install it. It wouldn't be installed implicitly as Setuptools is in some environments. And if they have the ability to install a custom `pkg_resources` package, they probably also have the ability to pin to an older Setuptools... so maybe "pin setuptools" is the universal workaround.

> I propose we do it in 2 stages:
> 
>     1. Explicitly mention a "due date" to the [warning message](https://github.com/pypa/setuptools/blob/v80.8.0/pkg_resources/__init__.py#L98-L103)
> 
>     2. Removal once the "due date" arrives.

I'm good with that plan. I'll set the due date for 6 months and mark this PR as draft in the meantime.

abravalheri:
Thank you very much Jason for adding the due date and providing the review about the namespaces.

> But it would still require the users to install it. It wouldn't be installed implicitly as Setuptools is in some environments. And if they have the ability to install a custom pkg_resources package, they probably also have the ability to pin to an older Setuptools... so maybe "pin setuptools" is the universal workaround.

As a brainstorm: there is one possibility to make it "installed implicitly" in build environments.

I mentioned in https://github.com/pypa/setuptools/issues/3085#issuecomment-2904033881 that we could return `pkg_resources` in `setuptools.build_meta:__legacy__.get_requires_for_build_sdist/wheel`.

But I am not sure this direction is worth pursuing specially because `pip` (and posibly `uv`?) don't fully[^1] fallback by default yet to `setuptools.build_meta:__legacy__` when `pyproject.toml` is absent (instead they run `python setup.py ...`).

[^1]: This behaviour is meant to be the default, but the `pip` team has-postponed/never-manage-to-implement this change to avoid backwards incompatibility. So what we have right now is an heuristic that sometimes uses `python setup.py` and sometimes `setuptools.build_meta:__legacy__`.

Avasam:
Incidentally, this would close most `pkg_resources`-related issues:
- Mentioned in title https://github.com/pypa/setuptools/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20pkg_resources%20in%3Atitle (except https://github.com/pypa/setuptools/issues/863)

As well as:
- https://github.com/pypa/setuptools/issues/271
- https://github.com/pypa/setuptools/issues/362
- https://github.com/pypa/setuptools/issues/369
- https://github.com/pypa/setuptools/issues/402
- https://github.com/pypa/setuptools/issues/413
- https://github.com/pypa/setuptools/issues/591
- https://github.com/pypa/setuptools/issues/759
- https://github.com/pypa/setuptools/issues/1095
- https://github.com/pypa/setuptools/issues/1383
- https://github.com/pypa/setuptools/issues/1677
- https://github.com/pypa/setuptools/issues/1708
- https://github.com/pypa/setuptools/issues/1722
- https://github.com/pypa/setuptools/issues/1759
- https://github.com/pypa/setuptools/issues/2262
- https://github.com/pypa/setuptools/issues/2463

There are a lot more issues referencing `pkg_resources`, but I only listed those that seemed to come from obvious direct `pkg_resources` usage, or were raising issues/feature requests specifically about `pkg_resources`'s API.

(as a sidenote: while looking at these, I found a few issues that may be closable already, I commented on those, so https://github.com/pypa/setuptools/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20commenter%3AAvasam%20pkg_resources should have a handful of issues you can close already)

setuptools/tests/test_namespaces.py
```diff
@@ -49,6 +52,7 @@ def test_mixed_site_and_non_site(self, tmpdir):
         with paths_on_pythonpath(map(str, targets)):
             subprocess.check_call(try_import)
 
+    @pytest.mark.xfail(reason="pkg_resources has been removed")
     def test_pkg_resources_import(self, tmpdir):
```
Avasam: Is there a point to keeping such tests ?

jaraco:
The PyPy tests are repeatedly failing (https://github.com/pypa/setuptools/actions/runs/21794694800/job/62891257402?pr=5007). When I run the PyPy tests locally, I get a different failure.

```
FAILED setuptools/tests/test_egg_info.py::TestEggInfo::test_egg_info_includes_setup_py - FileNotFoundError: [Errno 2] No such file or directory: 'foo.egg-info/SOURCES.txt'
```

Re-running with `-k test_egg_info`, the test passes, so it's another flaky test.

Oh, right. PyPy tests are known to be flaky and aren't required to pass checks.

jaraco:
I'm merging and releasing this now. We can roll it back if it causes unexpected or unmitigatable harm.

Avasam:
> Incidentally, this would close most `pkg_resources`-related issues:
> 
> * Mentioned in title [pypa/setuptools/issues (sort:updated-desc is:open pkg_resources in:title)](https://github.com/pypa/setuptools/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20pkg_resources%20in%3Atitle)

There's still many open and I'm pretty sure all of the issues from that query can be closed. (and even if it's an issue that can be similarly replicated, they should be rescoped to the new libraries, meaning it's probably not setuptool's problem anyway) 

As well as:
- https://github.com/pypa/setuptools/issues/4073
- https://github.com/pypa/setuptools/issues/2007
- https://github.com/pypa/setuptools/issues/1931
- https://github.com/pypa/setuptools/issues/406
- https://github.com/pypa/setuptools/issues/5051

Avasam:
I'm curious about a handful of leftovers.

Documentation is currently being handled in: https://github.com/pypa/setuptools/pull/5173

Dependencies and configs in: https://github.com/pypa/setuptools/pull/5175

---

So these are the leftovers I'm curious about:

``zip-safe``, ``eager-resources`` and ``namespace-packages`` mentioned in https://github.com/pypa/setuptools/blob/main/docs/userguide/pyproject_config.rst#setuptools-specific-configuration are they still used by legacy ``easy_install`` and ``setup.py install`` or can they be cleaned up ? (I vaguely remember this burden being shifted to pip)

IDEM for ``namespace_packages`` has something replaced the `pkg_resources` implementation or is this now unusable ?

https://github.com/pypa/setuptools/blob/main/setuptools/script%20(dev).tmpl & https://github.com/pypa/setuptools/blob/main/setuptools/script.tmpl have a hard requirement on `pkg_resources`

https://github.com/pypa/setuptools/blob/main/setuptools/wheel.py still uses `__import__('pkg_resources').declare_namespace(__name__)` in `_fix_namespace_packages`

https://github.com/pypa/setuptools/blob/main/setuptools/_scripts.py still fallbacks to `pkg_resources`

There's still a `pkg_resources` references in tests:
- https://github.com/pypa/setuptools/blob/main/setuptools/tests/fixtures.py
- https://github.com/pypa/setuptools/blob/main/setuptools/tests/namespaces.py
- https://github.com/pypa/setuptools/blob/main/setuptools/tests/test_develop.py
- https://github.com/pypa/setuptools/blob/main/setuptools/tests/test_editable_install.py
- https://github.com/pypa/setuptools/blob/main/setuptools/tests/test_wheel.py

PR #5033, opened 2025-06-14
Add ruff rules PIE

cclauss:
## Summary of changes

https://docs.astral.sh/ruff/rules/#flake8-pie-pie

Fix eight instances of:
% `ruff rule PIE810`
> The `startswith` and `endswith` methods accept tuples of prefixes or suffixes respectively.
Passing a tuple of prefixes or suffixes is more efficient and readable than calling the method multiple times.

> References
- [Python documentation: `str.startswith`](https://docs.python.org/3/library/stdtypes.html#str.startswith)
- [Python documentation: `str.endswith`](https://docs.python.org/3/library/stdtypes.html#str.endswith)

These transformations to reduce unnecessary repetitive function calls have been recommended since 2006 when [they were introduced](https://docs.python.org/2/whatsnew/2.5.html#other-language-changes) in Python 2.5

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


newsfragments/5033.misc.rst
```diff
@@ -0,0 +1 @@
+Add ruff rules PIE for minor optimizations to avoid repeated calls to str.startswith and str.endswith.
```
Avasam, line 1: I don't think usage of Ruff or the PIE rule group concerns users of setuptools.

```suggestion
Avoid repeated calls to ``str.startswith`` and ``str.endswith``.
```

Avasam (commented):
PIE is indeed a pretty good rule group to enable unconditionally.

abravalheri:
There is probably some recent update in Ruff causing the CI to fail in an unchanged portion of the code:


```python
>           raise RuffError(stdout.decode())
E           pytest_ruff.RuffError: setuptools/tests/test_build_py.py:168:5: PT031 `pytest.warns()` block should contain a single simple statement
E               |
E           167 |       msg = r"Python recognizes 'mypkg\.tests' as an importable package"
E           168 | /     with pytest.warns(SetuptoolsDeprecationWarning, match=msg):
E           169 | |         # TODO: To fix #3260 we need some transition period to deprecate the
E           170 | |         # existing behavior of `include_package_data`. After the transition, we
E           171 | |         # should remove the warning and fix the behavior.
E           172 | |
E           173 | |         if os.getenv("SETUPTOOLS_USE_DISTUTILS") == "stdlib":
E           174 | |             # pytest.warns reset the warning filter temporarily
E           175 | |             # https://github.com/pytest-dev/pytest/issues/4011#issuecomment-423494810
E           176 | |             warnings.filterwarnings(
E           177 | |                 "ignore",
E           178 | |                 "'encoding' argument not specified",
E           179 | |                 module="distutils.text_file",
E           180 | |                 # This warning is already fixed in pypa/distutils but not in stdlib
E           181 | |             )
E           182 | |
E           183 | |         build_py.finalize_options()
E           184 | |         build_py.run()
E               | |______________________^ PT031
E           185 |
E           186 |       build_dir = Path(dist.get_command_obj("build_py").build_lib)
```

abravalheri:
We can probably (temporarily) pin Ruff < [0.12.0](https://github.com/astral-sh/ruff/releases/tag/0.12.0) to avoid problems with new rules being enforced.

cclauss:
Could `build_py.finalize_options()` go before the `with` block?

Or `# noqa: PT031`?

abravalheri:
I don't know in this case because of the warnings filter bespoke modification inside the context, may be fragile...

Let's see if the CI fails in https://github.com/pypa/setuptools/pull/5042.

Avasam:
You can pin ruff temporarily as to not expand this PR's scope if you want.
I'm going through my active projects to update to 0.12 so I'll likely to it for setuptools+distutils as well today if no one else does it before me.

abravalheri:
@Avasam I created something in https://github.com/pypa/setuptools/pull/5042 just as a stopgap. We probably can merge and rebase this PR.

@cclauss feel free to rebase/merge on upstream/main.

cclauss:
https://github.com/pypa/setuptools/pull/5042#issuecomment-2981093777

cclauss:
@abravalheri Your review, please.

abravalheri:
Thank you very much @cclauss.

sins921 (approved):

PR #5047, opened 2025-07-11
Fix load_launcher_manifest() not finding the manifest file

lazka:
In #4963 load_launcher_manifest() was moved to a different directory, but the "launcher manifest.xml" file it references wasn't, leading to FileNotFoundError being raised when used.

Move the manifest file to the new location.


lazka:
Thanks!

Issue #5075, opened 2025-08-26
Bump vendored package `platformdirs`


daeho-ro:
Most recent version of `platformdirs` contains more robust homebrew location detection, so hope this can be updated.

jaraco:
I'm pretty sure this was addressed in a recent revendoring. Please comment and we can re-open if not.

PR #5076, opened 2025-08-26
feat: update platformdirs

daeho-ro:
## Summary of changes

Bump `platformdirs` package

Closes 
- https://github.com/pypa/setuptools/issues/5075

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:


jaraco:
I updated all vendored packages, so this change is obsolete. Thanks for the contrib.

(side note, as a general rule, we only let maintainers run the vendor update logic, so the best you can do is request an update in an issue)

PR #5083, opened 2025-09-14
BUG: fix a bug where calling `distutils.build_ext.finalize_options` more than once would raise an exception

neutrinoceros:
## Summary of changes

This allows `build_ext.finalize_options` to be called more than once without crashing. This is helpful in h5py and yt where we cannot really finalize options of this class before needing side effects (which live in the `run` method, as recommended).

I'm opening as a draft and without a test for now because I'm running out of time but I'll make sure to add one before opening for review.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:

### DIFF
--- a/newsfragments/5083.bugfix.rst
+++ b/newsfragments/5083.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where calling ``build_ext.finalize_options`` more than once would
+raise an exception.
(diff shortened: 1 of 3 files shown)


neutrinoceros:
this evolved into 2 separate bug fixes, so I split the second one out into #5084

abravalheri:
Hi @neutrinoceros thank you very much for working on this.

Would it be possible to submit changes to the `_distutils` repository directly to the https://github.com/pypa/distutils repo? The code in that folder is not maintained directly here. Setuptools every now and then pulls all the changes from `pypa/distutils`, so any change we do here gets overwritten.

neutrinoceros:
thanks, I wasn't aware ! here's the new PR: https://github.com/pypa/distutils/pull/385

PR #5098, opened 2025-10-21
"Lorem ipsum.txt" -> "Lorem_ipsum.txt"

x1sc0:
Using spaces in file names is not ideal


jaraco:
Vendored packages aren't managed in this repo. See the jaraco.text project, and search for past issues around this concern.

PR #5103, opened 2025-10-28
`_Finder.find` - fix docstring markdown rendering

Andrej730:
`*` was rendered as cursive in tooltips. Used double backticks - the way it's done in other docstrings of the same file.

Before:
image

After:
image


abravalheri:
Thank you very much.

The CI errors are unrelated.

PR #5132, opened 2025-12-20
Prevent object file collisions in parallel extension builds

e-luks:
## Summary of changes

Parallel builds of extensions that share source files may write to the same object file paths under a common build directory, resulting in race conditions and non-deterministic build outputs.

Use a per-extension subdirectory within build_temp to isolate object files and ensure deterministic, parallel-safe builds.

### Pull Request Checklist
- [X] Changes have tests

[PR docs]:

### Tests
I tested this change downstream at Gentoo after encountering https://bugs.gentoo.org/967476 and https://bugs.gentoo.org/945376 (packages https://pypi.org/project/librt and https://pypi.org/project/pymongo/). Since the build failures are encountered non-determinstically, i compiled both librt and pymongo as often as needed, capping the compilation attempts to 100.
- Before the change I encountered the build failures after at most 10 compilations
- After the change I didn't encounter the issue, compiling both librt and pymongo successfully 100 times.

### DIFF
--- a/newsfragments/3942.bugfix.rst
+++ b/newsfragments/3942.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed non-deterministic parallel builds of extensions that share source files.
--- a/setuptools/_distutils/command/build_ext.py
+++ b/setuptools/_distutils/command/build_ext.py
@@ -562,9 +562,12 @@ class build_ext(Command):
         for undef in ext.undef_macros:
             macros.append((undef,))
 
+        # Per-extension build dir to avoid conflicts in parallel builds.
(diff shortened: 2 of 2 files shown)


Avasam:
Please not that https://setuptools.pypa.io/en/latest/development/developer-guide.html#code-conventions-and-other-practices
> the `setuptools/_distutils` directory should not be modified directly when contributing to the `setuptools` project. Instead, this directory is maintained as a separated project in https://github.com/pypa/distutils, and periodically merged into `setuptools`.

e-luks:
@Avasam thanks, opened https://github.com/pypa/distutils/pull/393

PR #5141, opened 2026-01-15
Bump version 80.9.0 -> 80.9.1

SamhithGardas:


SamhithGardas:
@jaraco @abravalheri, Created patch for the setuptools with jaraco.context==6.1.0 fix. Could you please take a look? 

Reason I submit this is because of the artifatories were unable to recognize the change that @jaraco  made yesterday on the same version.

SamhithGardas:
The release was already pushed as 75.3.3 but the transitive dependency still exists.

Issue #4946, opened 2025-04-14
Undisclosed security vulnerability


jaraco:
This issue is to track a yet-undisclosed security vulnerability.

WilliamRoyNelson:
@jaraco Would it be possible to backport this fix to 75.3.x for Python 3.8 support?
I know Python 3.8 reached End of Life in 2024, but virtualenv still supports it:
https://github.com/pypa/virtualenv/blob/2ec4db5d60d9374a0869703c757ead04a7b67c84/src/virtualenv/seed/wheels/embed/__init__.py#L8-L13

It looks like their policy is to support Python versions for 18 months after EOL:
https://github.com/pypa/virtualenv/discussions/2526#discussioncomment-5347173

And because setuptools 75.3.2 is bundled with virtualenv, it causes problems with vulnerability scanning tools.

jaraco:
> And because setuptools 75.3.2 is bundled with virtualenv, it causes problems with vulnerability scanning tools.

Unfortunately, no backport will fix versions pinned to the vulnerable versions.

See the referenced issue where we're exploring options for a backport.

jaraco:
> [@jaraco](https://github.com/jaraco) Would it be possible to backport this fix to 75.3.x for Python 3.8 support? I know Python 3.8 reached End of Life in 2024, but virtualenv still supports it: https://github.com/pypa/virtualenv/blob/2ec4db5d60d9374a0869703c757ead04a7b67c84/src/virtualenv/seed/wheels/embed/__init__.py#L8-L13
> 
> It looks like their policy is to support Python versions for 18 months after EOL: [pypa/virtualenv#2526 (comment)](https://github.com/pypa/virtualenv/discussions/2526#discussioncomment-5347173)
> 
> And because setuptools 75.3.2 is bundled with virtualenv, it causes problems with vulnerability scanning tools.

I did this for #5009 and 75.3.3 and it has caused quite a bit of confusion and consternation (see #5142), so I'm reluctant to do it again. Setuptools' policy is to drop support for EOL Pythons shortly after they go EOL. Supporting backports should be reserved for critical issues. These vulnerable code paths are unlikely to be hit, as they are deprecated code paths and the preferred code paths are to install first with pip or Setuptools will invoke pip directly, bypassing this vulnerable code path.

[...]

PR #5100, opened 2025-10-25
Backport CVE-47273 fix from 78.1.1 to 75.3.2

UladzimirTrehubenka:
Fix #4946, allow to use upcoming (I hope) 75.3.3 against Python 3.8 (which is still widely used) and prevent issues with vulnerability scanning tools.

PR #5150, opened 2026-01-20
Backport CVE-47273 fix from 78.1.1 to 75.3.3

UladzimirTrehubenka:
New version of https://github.com/pypa/setuptools/pull/5100 - fix https://github.com/pypa/setuptools/issues/4946 (commit https://github.com/pypa/setuptools/commit/250a6d1), allow to use 75.3.X against Python 3.8 (which is still widely used) and prevent issues with vulnerability scanning tools.


jaraco:
Published as v75.3.4.

PR #5160, opened 2026-01-23
Fix misleading `Multiple directories found` assertion in `_find_info_directory`

ooooo-create:
## Summary of changes

If the search failed to find any candidates (len(candidates) == 0) in a directory that had no subdirectories or multiple subdirectories (len(dirs) != 1), the condition evaluated to True. This triggered the assertion checks for candidates == 1, resulting in a confusing AssertionError: Multiple {suffix} directories found, even though zero candidates were actually found.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/build_meta.py
+++ b/setuptools/build_meta.py
@@ -356,12 +356,11 @@ class _BuildMetaBackend(_ConfigSettingsTranslator):
         for parent, dirs, _ in os.walk(metadata_directory):
             candidates = [f for f in dirs if f.endswith(suffix)]
 
-            if len(candidates) != 0:
-                assert len(candidates) == 1, f"Multiple {suffix} directories found"
+            if len(candidates) != 0 or len(dirs) != 1:
+                assert len(candidates) == 1, (
+                    f"Exactly one {suffix} should have been produced, but found {len(candidates)}: {candidates}"
+                )
                 return Path(parent, candidates[0])
-
-            if len(dirs) != 1:
-                break
 
         msg = f"No {suffix} directory found in {metadata_directory}"
         raise errors.InternalError(msg)


abravalheri:
Hi @ooooo-create, thank you very much for the contribution.

Could you please fill in an issue report containing the description of the problem you are trying to solve first, with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)?

These checks are somewhat old in the codebase, despite being refactored a bunch of times, this check has always been present in one form or other since the introduction of PEP 517. So I would like to understand a bit better which kinds of problems you may have had.


In a different iteration, this piece of code had a different message: `f"Exactly one {suffix} should have been produced"`. Would that be a better error message?

ooooo-create:
> Could you please fill in an issue report containing the description of the problem you are trying to solve first, with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)?

Thanks for your suggestion. Here is a bash script to reproduce it
```
mkdir /tmp/proj1
cd /tmp/proj1

touch setup.py

python -m venv .venv
.venv/bin/python -m pip install -U pip
.venv/bin/python -m pip install -e .
```

However, I ran into this issue because I had multiple setup() calls in my setup.py that were triggered by different conditions. I accidentally caused the execution flow to bypass all of them, and realized that the resulting error—AssertionError: Multiple .egg-info directories found—was actually misleading.

> In a different iteration, this piece of code had a different message: `f"Exactly one {suffix} should have been produced"`. Would that be a better error message?

yes~ it's better, i will take it.

abravalheri:
I see... Thank you very much.

Yeah, it is an `assert` statement, which probably means that the original author never expected this statement to be reached in a real scenario (and also they were more concerned with the case where multiple dirs are found) ... and I see why it is triggering in your build.

I think we should go with a better error message and avoid changing too much the code itself.

abravalheri:
Thank you!

PR #5181, opened 2026-02-10
Restore `pkg_resources`

### LINKED ISSUE #5174: [BUG] Restore `pkg_resources`
### setuptools version

setuptools==82.0.0

### Python version

Python 3.10

### OS

n/a

_No response_

### Description

(I am reporting this a bug, though it's likely to be more of a discussion, at least initially. I don't see this kind of technical discussion happening in GitHub Discussion though).

setuptools 82.0.0 recently removed the `pkg_resources` library. This breaks all packages that rely on this for either build-time and runtime. While users can work around this at runtime by pre-installing an older setuptools version in the environment, doing so for build-time dependency on `pkg_resources` is far more difficult, given the rise of of build isolation in pip and other tools in recent years (I believe this was the key reason for closing https://github.com/pypa/setuptools/issues/863). While workaround do exist for pip (which has the `--no-build-isolation` hammer), I don't believe other tools like tox have equivalent options. The removal of `pkg_resources` thus effectively breaks installation of older packages, which is very unfortunate given these are still quite common in the wild.

### Expected behavior

The entire reason for `build_meta.__legacy__` existence is to allow us to continue using packages that haven't migrated to newer packaging standards, whether that is because the packages are no longer maintained or we're intentionally testing older versions (for e.g. stable or historical release branches). Removing `pkg_resources` breaks this contract owing to `pkg_resources`'s use as a build-time dependency for things like namespace packages.

I believe `pkg_resources` should continue to be installed in the environment legacy configuration. Whether this "installation" is handled by pip or setuptools is likely an implementation detail, but in either case the backing code to support it is small, and it could easily be filed away somewhere and never looked at again.

Am I missing something? Is there an alternative approach users can use and that we could document? As always, I am happy to work on a solution once one has been agreed on.

### How to Reproduce

n/a

### Output

n/a

stephenfin:
## Summary of changes

As discussed in #5174, restore the `pkg_resources` package in setuptools. The removal warning is not restored and documentation for same is left in `doc/deprecated`. Additional documentation is provided to clarify that this module is frozen and no new PRs of any kind will be accepted. To this effect, ruff checks are also disabled for this module to prevent fixes being needed for future checks.

Closes #5174

### Pull Request Checklist

- [x] Changes have tests

[PR docs]: https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request


abravalheri:
From my pov the approach discussed in https://github.com/pypa/setuptools/issues/863#issuecomment-3872808142 is more promising, but it would be nice to hear from the other maintaienrs.

stephenfin:
> From my pov the approach discussed in [#863 (comment)](https://github.com/pypa/setuptools/issues/863#issuecomment-3872808142) is more promising, but it would be nice to hear from the other maintaienrs.

Thanks, @abravalheri :pray: Probably best to keep the discussion on the two issues so I've responded there.

stephenfin:
The two `collateral` jobs appears to be failing due to an access issue and an existing issue from `main` (I get the same issue building docs there using `tox -e docs`).

firmwarecostum:
Can this be accepted? There are still many who need this.

PR #5183, opened 2026-02-11
Fix broken docs build

stephenfin:
In rST, double backticks are literals while single backticks are default role. The default role is setuptools is (unfortunately) configured to be `any`. This is now causing a docs failures since the `pkg_resources` doc was moved in commit 75072c43, preventing Sphinx from finding it.


abravalheri:
Thank you!

PR #5193, opened 2026-03-03
Add initial support for abi3.abi3t tag from PEP 803

### LINKED ISSUE #5205: [FR] Support building extensions and wheels with the abi3t ABI (PEP 803)

The steering council approved PEP 803 and Python 3.15 will support a new abi3t ABI. Extensions built using this ABI can be imported on both the free-threaded and GIL-enabled builds of CPython.

Adding support for building these extensions and wheels will allow projects using setuptools to reduce their maintenance burden and build one wheel for both Python 3.15 builds and automatically support 3.16 and newer when those builds are released.

See PEP 803 for more details: https://peps.python.org/pep-0803/

There are two possible ways of getting this done.

### Require a free-threaded interpreter to build abi3t wheels

The main benefit of this approach is it doesn't require any new UI in setuptools. We would repurpose the `py_limited_api` keyword argument for `Extension` and `bdist_wheel` option so that if they are set and the build is happening with a free-threaded interpreter, setuptools will produce `.abi3t.so` extension files on posix platforms and `abi3t.abi3` wheels on all platforms.

This is the approach taken in https://github.com/pypa/setuptools/pull/5193.

### Add new UI for choosing between abi3 and abi3t builds.

There is nothing special about the free-threaded interpreter for these builds and it's possible to produce `abi3t` extensions and wheels using a GIL-enabled interpreter. But because we don't want to break existing `abi3` builds on GIL-enabled interpreters, I think we would need to have some way of opting in to abi3t builds. This probably requires a new `Extension` keyword argument.

### Alternative Solutions

I'm not sure there is an alternative for users of setuptools who want to do these builds.

### Additional context

I'm available on the PyPA discord if you want to have lower-latency chats about this topic.

### Code of Conduct

ngoldbaum:
## Summary of changes

Fixes #5205 

[PEP 803](https://peps.python.org/pep-0803/) proposes a free-threaded stable ABI. I'm working on [ecosystem support](https://github.com/Quansight-Labs/free-threaded-compatibility/issues/310) for a free-threaded stable ABI and [end-to-end tests](https://github.com/Quansight-Labs/stable-abi-testing) for PEP 803. PEP 803 was accepted in March and there will be support for it in Python 3.15.

This PR is set up to require a free-threaded build to produce abi3.abi3t wheels. It does this by making builds where `py_limited_api=True` is set on the free-threaded build on 3.15 or newer produce abi3t extensions and abi3.abi3t wheels.

In principle it's possible to do this on the GIL-enabled build as well but it'll require at least a new `Extension` keyword argument `py_target_abi3t=True` maybe?

As far as I can tell, `py_limited_api` is undocumented, so this doesn't add new docs, although I'd be happy to explore that if you'd like to see docs for this.

### Pull Request Checklist
- [x] Changes have tests

[PR docs]:


ngoldbaum:
I'll do another pass next week to get the tests passing here. In the meantime, I'd appreciate comments.

ngoldbaum:
In order to make the diff coverage happy, I'll probably need to:

* Add CI for Python 3.15
* Add free-threaded CI coverage for 3.14 and 3.15
* Add a new test extension that sets `py_limited_api = True` for an `Extension` object and then also sets `py_limited_api=cp315` in `setup.cfg`.

I'm going to punt on that for now but I hope to come back to that. I also think free-threaded CI coverage is probably needed independent of the PEP 803 work.

ngoldbaum:
Actually, on second thought, I think I might be able to use `unittest.mock` to fake a free-threaded build.

ngoldbaum:
Fun! It looks like a `more-itertools` release broke the type checking test.

ngoldbaum:
I looked at Python 3.15 testing a bit today and ran into https://github.com/ronaldoussoren/pyobjc/issues/671. Even if that isn't a problem on CI, the full pyobjc dependency tree takes quite a while to build, so any 3.15 testing will be Linux-only for now probably.

It probably does make sense to add an abi3t test extension that requires Python 3.15 or newer, but I'm going to punt on that for now.

Besides that issue, all code in this PR is tested.

setuptools/tests/test_build_ext.py
```diff
@@ -19,6 +20,10 @@
 from distutils.sysconfig import get_config_var
 
 IS_PYPY = '__pypy__' in sys.builtin_module_names
+# from a Mac running Python 3.14
+ABI3_EXT_SUFFIXES = ['cpython-314-darwin.so', 'abi3.so', 'so']
```
rgommers, line 24: Missing dots here, which is asymmetric with line 26. I think it should match `importlib.machinery.EXTENSION_SUFFIXES`, so adding `.`'s here looks like the fix.

setuptools/command/bdist_wheel.py
```diff
@@ -132,6 +132,19 @@ def safer_version(version: str) -> str:
...
 
+def stable_abi_tag(impl_name, impl_version):
+    abi_tag = None
+    if impl_name == "cp" and impl_version[0] == '3':
+        if sysconfig.get_config_var("Py_GIL_DISABLED"):
+            # per PEP 803 these are possible on older Python versions
+            # but in practice these builds need cp315 or newer
+            abi_tag = "abi3.abi3t"
+        else:
+            # This is wrong on on CPython 3.1, 3.0 ¯\_(ツ)_/¯
```
rgommers, line 143: typo: "on on"

rgommers: Easier to just delete this, the comment is irrelevant since only Python >=3.9 is supported

setuptools/tests/test_bdist_wheel.py
```diff
@@ -435,23 +436,31 @@ def test_universal_deprecated(dummy_dist, monkeypatch, tmp_path):
     name="extension.dist",
     version="0.1",
     description="A testing distribution \N{SNOWMAN}",
-    ext_modules=[Extension(name="extension", sources=["extension.c"])],
+    ext_modules=[
+        Extension(
+            name="extension",
+            sources=["extension.c"],
+            py_limited_api=True
```
rgommers, line 443: This change looks correct and is a bug fix for the existing test, which didn't quite make sense before - I think given the `test_limited_abi` test name, it was intended to test the Limited API indeed.

rgommers (commented):
Thanks for working on this @ngoldbaum. Overall this PR looks solid, I only found one minor issue in one of the added tests in `test_build_ext.py`.

ngoldbaum:
I marked this ready for review and updated the PR description. @jaraco @abravalheri I'd love your opinions on this.

ngoldbaum:
I think all the remaining CI failures are unrelated.

I had to update the test a bit to account for upstream changes in how abi3t works that happened just before the 3.15.0b1 release. I also took the opportunity to add a new test that verifies that limited API builds fail on 3.14t.

Also see https://github.com/pypa/setuptools/pull/5256 which updates the vendored version of `packaging`. The current version doesn't know about the `abi3t` tag.

@jaraco I'd appreciate your eyes on this. I'd like to enable downstream testing of abi3t for packages that depend on setuptools.

ngoldbaum:
Hi @jaraco Python 3.15rc1 is out and it would really help downstream libraries to be able to test this.

ngoldbaum:
Yay, CI is green! (except the continue-on-error job #5251 addresses).

dalcinl:
@ngoldbaum I think your idea of adding and additional `py_target_abi3t=False` argument (note: `False` as default) to `Extension` is the right way to go. *Explicit is better than implicit.*

Building for a target ABI should be decoupled from the runtime. The new abi3 vs abi3t situation is no really different than using for example Python 3.14 to build a `cp310-abi3` wheel targeting Python 3.10+, which is something that Python headers and the current setuptools machinery support just fine.

ngoldbaum:
Before I write more code here I'd
appreciate getting some input from a maintainer about how they'd like this to look.

dalcinl:
@ngoldbaum While you want for maintainer feedback, in the meantime maybe you could take a look at the following for future consideration?

Please take a look at this code that decides the extension suffix:
https://github.com/pypa/setuptools/blob/main/setuptools/command/build_ext.py#L79

Up to Python 3.14, the code above was fine, it would return the first and only abi3 suffix, for example:
```console
$ python3.14 -c 'from importlib.machinery import EXTENSION_SUFFIXES as es; print(es)'
['.cpython-314-x86_64-linux-gnu.so', '.abi3.so', '.so']
```

However, in Python 3.15, things look a bit different:
With a GIL-enabled Python
```console
$ python3.15 -c 'from importlib.machinery import EXTENSION_SUFFIXES as es; print(es)'
['.cpython-315-x86_64-linux-gnu.so', '.abi3-x86_64-linux-gnu.so', '.abi3.so', '.abi3t-x86_64-linux-gnu.so', '.abi3t.so', '.so']
```
With a free-threaded Python
```console
$ python3.15t -c 'from importlib.machinery import EXTENSION_SUFFIXES as es; print(es)'
['.cpython-315t-x86_64-linux-gnu.so', '.abi3t-x86_64-linux-gnu.so', '.abi3t.so', '.so']
```

Therefore, the `get_abi3_suffix()` code produces different results depending on you running on GIL or non-GIL Python. That cannot be right. As I commented before, the build-time interpreter should not influence the output, it is the user who should decide, for whatever mechanism, whether to build abi3 vs abi3t under Python 3.15+. This is something that this PR should eventually address, that is, the extension suffix should not depend on the GIL vs non-GIL build of the running, build-time interpreter.

rgommers:
> As I commented before, the build-time interpreter should not influence the output, it is the user who should decide, for whatever mechanism, whether to build abi3 vs abi3t under Python 3.15+

That seems unworkable. For a free-threaded interpreter, there's only one supported value (`abi3t`). For a GIL-enabled interpreter, `abi3` has to be the default for backwards compat reasons. Targeting `abi3t` from a GIL-enabled interpreter therefore has to be opt-in.

All build backends will do it like this I believe.

dalcinl:
> That seems unworkable.

I did not phrase things very well. I was not trying to make a point about defaults, and I understand the backward compatibility implication.

Let me restate my comments this way: I'm a module author, I wrote my code such that it can build `abi3t`. I setup my build backend to request for `abi3t` (I hope build backends will have explicit ways to support such a request, that's why I wrote https://github.com/pypa/setuptools/pull/5193#issuecomment-5384982769 ). Therefore, if I asked for `abi3t`, I should get an `abi3t` extension module, no matter whether I'm running the build with a GIL or non-GIL interpreter. IOW, being forced to run a free-threaded interpreter to get a `abi3t` build seems a bit ridiculous to me, a quite backwards to the "unifying" and universal nature of the `abi3t` builds for Py 3.15+.

> Targeting abi3t from a GIL-enabled interpreter therefore has to be opt-in.

Exactly! And that's what I was trying to point out. Currently, even with the current state of this PR, a GIL-enabled interpreter produces an `abi3` module tag. Opt-in mechanisms are missing. 

This PR is mostly about fixing the wheel tagging. The extension module suffix also needs some additional work.

Or maybe there is something I got fundamentally wrong, and a GIL-enabled interpreter cannot really produce an `abi3t` extension module that can run both on GIL-enabled and free-threaded interpreters?

rgommers:
Ah, that makes perfect sense now, thanks @dalcinl.  

> Or maybe there is something I got fundamentally wrong, and a GIL-enabled interpreter cannot really produce an `abi3t` extension module that can run both on GIL-enabled and free-threaded interpreters?

No, you got nothing wrong. I think that's still two: there should be UX to tell the build backend to produce `abi3t` on a GIL-enabled build also. So something like `limited_api=True, target_abi3t=True` or `limited_api='abi3t'`.

In `meson-python` we left that for step two; since it's opt-in anyway we can add that later in a backwards-compatible way. That's probably the case here too.

ngoldbaum:
Also just to clarify, I'm happy to add UI to support abi3t builds with a GIL-enabled interpreter, I just need a setuptools maintainer to give the OK and tell me what API they'd prefer.

PR #5202, opened 2026-03-29
fix: use per-extension subdirectory in build_temp to prevent parallel build race (#5196)

### LINKED ISSUE #5196: [BUG] Race condition due to reuse of build folder when building extensions in parallel
### setuptools version

setuptools=70.0.0

### Python version

Python 3.12

### OS

RHEL 9

_No response_

### Description

Having multiple extensions in `ext_modules` being compiled in parallel with `pip install --config-setting='--build-option=build_ext'  --config-setting='--build-option=-j24' ...` causes issues because they share the same build folder set at https://github.com/pypa/setuptools/blob/5a13876673a41e3cd21d4d6e587f53d0fb4fd8e5/setuptools/_distutils/command/build.py#L111-L112

So when they write anything into that folder they may overwrite or delete files used by other extensions

### Expected behavior

Building extensions in parallel has the same result as building serially

### How to Reproduce

- Add multiple extensions to `ext_modules` that use e.g. the same source files but compile them e.g. with different flags.
- Install with the parallel flag set

This was observed with [DeepSpeed](https://github.com/deepspeedai/DeepSpeed) so you could try installing that from source with "enough" cores but it might be too much to setup

### Output

Output varies and it might succeed depending on the amount of parallel actions going on and the available CPU cores. So it may sometimes succeed or fail seemingly randomly

bibekmhj:
## Summary
Fixes #5196

## Root cause
All extensions share the same `build_temp` directory when compiled
in parallel with `-j`. The `compiler.compile()` call uses
`output_dir=self.build_temp` for every extension, so `.o` files
with the same name overwrite each other.

## Fix
Derive a per-extension `output_dir` from `ext.name`:
```python
ext_build_temp = os.path.join(self.build_temp, *ext.name.split('.'))
```
`pkg.ext1` → `build/temp.../pkg/ext1/`
`pkg.ext2` → `build/temp.../pkg/ext2/`

## Testing
Added unit test verifying the two extensions receive different
`output_dir` values.

### DIFF
--- a/setuptools/_distutils/command/build_ext.py
+++ b/setuptools/_distutils/command/build_ext.py
@@ -560,10 +560,11 @@ class build_ext(Command):
         macros = ext.define_macros[:]
         for undef in ext.undef_macros:
             macros.append((undef,))
-
+            
+        ext_build_temp = os.path.join(self.build_temp, *ext.name.split('.'))
         objects = self.compiler.compile(
             sources,
-            output_dir=self.build_temp,
+            output_dir=ext_build_temp,
             macros=macros,
             include_dirs=ext.include_dirs,
             debug=self.debug,
--- a/setuptools/tests/test_build_ext.py
+++ b/setuptools/tests/test_build_ext.py
@@ -18,6 +18,8 @@ from .textwrap import DALS
 import distutils.command.build_ext as orig
 from distutils.sysconfig import get_config_var
 
+from unittest.mock import MagicMock
+
 IS_PYPY = '__pypy__' in sys.builtin_module_names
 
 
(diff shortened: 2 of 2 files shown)


Avasam:
Please note that https://setuptools.pypa.io/en/latest/development/developer-guide.html#code-conventions-and-other-practices
> Finally, the `setuptools/_distutils` directory should not be modified directly when contributing to the `setuptools` project. Instead, this directory is maintained as a separated project in https://github.com/pypa/distutils, and periodically merged into `setuptools`.

and https://github.com/pypa/setuptools/issues/5196#issuecomment-4053709619
> One thing that requires attention though is that changes in the `setuptools/_distutils` directory need to be addressed in the [`pypa/distutils`](https://github.com/pypa/distutils?rgh-link-date=2026-03-13T09%3A05%3A03.000Z) repository.

PR #5208, opened 2026-04-02
Work around typing issue from more_itertools

ngoldbaum:
## Summary of changes

Running the tests with the latest versions of the test dependencies will fail due to an issue in more_itertools: https://github.com/more-itertools/more-itertools/issues/1143.

For now, to unbreak CI, I'm adding a runtime cast to avoid the type checking issue.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

### DIFF
--- a/setuptools/config/expand.py
+++ b/setuptools/config/expand.py
@@ -122,9 +122,9 @@ def read_files(
     from more_itertools import always_iterable
 
     root_dir = os.path.abspath(root_dir or os.getcwd())
-    _filepaths = (os.path.join(root_dir, path) for path in
-                  # work around more_itertools/more_itertools#1143
-                  cast('Iterable[StrPath]', always_iterable(filepaths)))
+    # work around more_itertools/more_itertools#1143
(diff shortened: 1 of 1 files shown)


ngoldbaum:
The diffcov failure is spurious: the current diffcov configuration will never see top-level imports.

Unfortunately, when I try to fix it, I hit https://github.com/pytest-dev/pytest-cov/issues/740. The simplest thing is probably just to ignore the diffcov failure on this PR.

ngoldbaum:
https://github.com/more-itertools/more-itertools/pull/1144 and a more-itertools release might obviate the need for this.

Avasam (approved):
I think this is fine until if/when more-itertools figure out what they want from `always_iterable`. Overloads are not implemented in a standard way across all type-checkers (with mypy getting the short end of the stick, as always). And this doesn't affect setuptools' public API.

ngoldbaum:
The issue this PR is working around was fixed in `more-itertools` 11.0.2, so this isn't necessary.

PR #5232, opened 2026-05-21
Remove `typing.AnyStr` usage

sobolevn:
## Summary of changes

Hi 👋 
We are deprecating `AnyStr` with a warning since Python 3.16, this is a proactive PR to replace it with a modern alternative.
Context https://github.com/python/cpython/pull/150190

### Pull Request Checklist
- [ ] Changes have tests

[PR docs]:

This is not really a user-facing change (at least it should not be), so no changelog entries.

### DIFF
--- a/setuptools/_distutils/util.py
+++ b/setuptools/_distutils/util.py
@@ -17,7 +17,7 @@ import sys
 import sysconfig
 import tempfile
 from collections.abc import Callable, Iterable, Mapping
-from typing import TYPE_CHECKING, TypeVar
+from typing import TYPE_CHECKING, AnyStr
 
 from jaraco.functools import pass_none
 
@@ -30,8 +30,6 @@ if TYPE_CHECKING:
     from typing_extensions import TypeVarTuple, Unpack
 
     _Ts = TypeVarTuple("_Ts")
-
-    _StrBytes = TypeVar("_StrBytes", str, bytes)
 
 
 def get_host_platform() -> str:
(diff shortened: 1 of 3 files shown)


Avasam:
Thanks @sobolevn  ! Note that any change to the `_distutils` folder has to be done there: https://github.com/pypa/distutils or it'll get ovewritten next sync

Since this won't be immediately marked as deprecated, I'm tempted to recommend to this project's maintainers to use https://docs.astral.sh/ruff/rules/banned-api/ / https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_banned-api to prevent accidentally adding it back.

tai1105997-alt (approved):

jaraco:
Thanks @sobolevn! I pushed a small commit reverting the `_distutils/util.py` hunk: that tree is vendored from [pypa/distutils](https://github.com/pypa/distutils) and is overwritten on each sync, so the change there would be lost (and could conflict on the next sync). It's best made upstream in pypa/distutils. The setuptools-owned changes in `bdist_egg.py` and `glob.py` are great — merging those now.

jaraco:
Follow-up: I sent the `util.py` `AnyStr`→`TypeVar` change upstream as pypa/distutils#409 (crediting you as the origin), so it'll persist through the next distutils sync. 🙏

sobolevn:
@jaraco thank you! Sorry for the late response, it got lost in my notifications :(

PR #5240, opened 2026-06-07
fix: Remove incorrect _restore_install_lib override (closes #2728)

### LINKED ISSUE #2728: [BUG] v57.1.0 on Python v3.6 SSL Errors
### setuptools version

57.1.0

### Python version

3.6

### OS

macOS, Windows

The problem doesn't seem to occur on Linux. setuptools v57.0.0 works fine.

### Description

Running "python setup.py install" on my project fails to install the project's dependencies from PyPI.

### Expected behavior

Normal installation.

### How to Reproduce

python setup.py install

### Output

```Processing dependencies for sip==6.2.0.dev2107131757
Searching for toml
Reading https://pypi.org/simple/toml/
Download error on https://pypi.org/simple/toml/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748) -- Some packages may not be found!
Couldn't find index page for 'toml' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.org/simple/
Download error on https://pypi.org/simple/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748) -- Some packages may not be found!
No local packages or working download links found for toml
error: Could not find suitable distribution for Requirement.parse('toml')
```

### Code of Conduct

botbikamordehai2-sketch:
## What
In v57.1.0, the custom install command `install_with_pth` overrides `install_lib` incorrectly, causing installation paths to be corrupted and leading to SSL/certificate errors during dependency resolution on Python 3.6 (especially on macOS/Windows). The `_restore_install_lib` method attempts to undo a secondary effect of `extra_path` but compares a path suffix with Python code, which never matches, thus always resetting `install_lib` to `install_libbase`.

## Fix
Remove the `_restore_install_lib` method and its call in `finalize_options`. The `extra_path` mechanism does not actually require this correction; the original `finalize_options` from `install` already handles paths correctly.

Closes #2728

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,16 +43,6 @@ class install_with_pth(install):
 
     def finalize_options(self):
         install.finalize_options(self)
-        self._restore_install_lib()
-
-    def _restore_install_lib(self):
-        """
-        Undo secondary effect of `extra_path` adding to `install_lib`
-        """
-        suffix = os.path.relpath(self.install_lib, self.install_libbase)
-
-        if suffix.strip() == self._pth_contents.strip():
-            self.install_lib = self.install_libbase
 
 
 setup_params = dict(


setup.py
```diff
@@ -43,16 +43,6 @@ def initialize_options(self):
 
     def finalize_options(self):
         install.finalize_options(self)
```
Avasam, line 45: If this override now only calls the parent, why not remove it ?

PR #5244, opened 2026-06-28
Materialize product() passed to parametrize (pytest 9.1 compat)

jaraco:
## Problem

`setuptools/tests/test_config_discovery.py::TestDiscoverPackagesAndPyModules::test_purposefully_empty` passes an `itertools.product` (a one-shot iterator) directly as `argvalues` to `@pytest.mark.parametrize`.

pytest 9.1 deprecates non-Collection iterables for `argvalues` (`PytestRemovedIn10Warning`), which surfaces here as an **error during collection** — taking the whole `test_config_discovery.py` module down with it. The class-level decorator is exactly the hazard the deprecation targets: a bare iterator is exhausted after the first test method collects, silently skipping the rest.

See [pytest-dev/pytest#13409](https://github.com/pytest-dev/pytest/issues/13409) and the [deprecation note](https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators).

## Fix

Render the product eager via a small documented helper:

```python
greedy_product = compose(list, product)  # jaraco.functools.compose
```

so `argvalues` is a re-iterable `list`. The rationale lives in the helper's docstring.

`jaraco.functools` is already a setuptools test dependency, so no new requirement.

## Verification

- `test_purposefully_empty` now collects and passes (30 params).
- `ruff check` + `ruff format --check` clean.

Note: this only manifests with pytest >= 9.1; CI on an older pytest won't have flagged it yet, so this is a forward-compat fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/setuptools/tests/test_config_discovery.py
+++ b/setuptools/tests/test_config_discovery.py
@@ -6,6 +6,7 @@ from typing import cast
 
 import jaraco.path
 import pytest
+from jaraco.functools import compose
 from path import Path
 
 import setuptools  # noqa: F401 # force distutils.core to be patched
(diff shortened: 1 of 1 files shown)


jaraco:
mypy checks are failing with:

```
[2](https://github.com/pypa/setuptools/actions/runs/28333781829/job/83936383357?pr=5244#step:11:63)
[gw0] darwin -- Python 3.14.6 /Users/runner/work/setuptools/setuptools/.tox/py/bin/python
setuptools/tests/test_config_discovery.py:24: error: Need type annotation for "greedy_product"  [var-annotated]
setuptools/tests/test_config_discovery.py:24: error: Argument 1 to "compose" has incompatible type "type[list[_T]]"; expected "Callable[[product[tuple[Never]]], list[Never]]"  [arg-type]
setuptools/tests/test_config_discovery.py:176: error: Too many arguments  [call-arg]
```

jaraco:
Claude says python/mypy#15737 and python/mypy#13540 are precisely relevant.

PR #5248, opened 2026-07-04
Respect excluded subpackages in build_py

### LINKED ISSUE #3260: [BUG] `options.packages.find.exclude` not making effect when `include_package_data = True`
### setuptools version

62.1.0

### Python version

Python 3.10.4 (main, Mar 30 2022, 23:50:39) [GCC 10.2.1 20210110]

### OS

Tested in Alpine Linux 3.14, Debian bullseye and Ubuntu 20.04.4 LTS

```console
Python 3.10.4 (main, Mar 30 2022, 23:50:39) [GCC 10.2.1 20210110]

Package    Version
---------- -------
build      0.7.0
packaging  21.3
pep517     0.12.0
pip        22.0.4
pyparsing  3.0.8
setuptools 62.1.0
tomli      2.0.1
wheel      0.37.1
```

### Description

For some reason, when `include_package_data = True`, subpackages excluded via `options.packages.find.exclude` are still added to the wheel.

### Expected behavior

When listed via `options.packages.find.exclude` a subpackage should not be added to the wheel.

### How to Reproduce

```bash
> docker run --rm -it python:3.10-bullseye /bin/bash

mkdir /tmp/mypkg
cd /tmp/mypkg
mkdir -p mypkg/tests
touch mypkg/__init__.py
touch mypkg/resource.txt
touch mypkg/tests/test_mypkg.py
touch mypkg/tests/test_file.txt

cat <<EOS > pyproject.toml
[build-system]
requires = ['setuptools']
build-backend = 'setuptools.build_meta'
EOS

cat <<EOS > setup.cfg
[metadata]
name = mypkg
version = 42

[options]
include_package_data = True
packages = find:

[options.packages.find]
exclude = *.tests*
EOS

cat <<EOS > MANIFEST.in
global-include *.py *.txt
global-exclude *.py[cod]
prune dist
prune build
EOS

python -m pip install -U pip build setuptools wheel
python -VV
python -m pip list

rm -rf dist build mypkg.egg-info
python -m build --no-isolation
# `--no-isolation` is used to guarantee we have control over the
# versions of the build dependencies
unzip -l dist/*.whl
```

### Output

```console
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2022-04-11 17:49   mypkg/__init__.py
        0  2022-04-11 17:49   mypkg/resource.txt
        0  2022-04-11 17:49   mypkg/tests/__init__.py
        0  2022-04-11 17:49   mypkg/tests/test_file.txt
        0  2022-04-11 17:49   mypkg/tests/test_mypkg.py
      108  2022-04-11 17:49   mypkg-42.dist-info/METADATA
       92  2022-04-11 17:49   mypkg-42.dist-info/WHEEL
        6  2022-04-11 17:49   mypkg-42.dist-info/top_level.txt
      654  2022-04-11 17:49   mypkg-42.dist-info/RECORD
---------                     -------
      860                     9 files
```

The expected output should **not** include:
```diff
- mypkg/tests/__init__.py
- mypkg/tests/test_file.txt
- mypkg/tests/test_mypkg.py
```

itscloud0:
## Summary of changes

Fix `packages.find.exclude` so excluded nested Python subpackages are not copied into wheels through `include_package_data=True`.

Closes #3260.

## Problem

On current `main`, a project configured with:

- `include_package_data = True`
- `packages = find:`
- `exclude = *.tests*`

still ships `mypkg/tests/*` in the built wheel when those files are pulled in through `MANIFEST.in`.

The repo already had an xfailed regression for this case in `setuptools/tests/test_build_py.py`, and a direct repro on current `main` still produced a wheel containing `mypkg/tests/test_mypkg.py` and `mypkg/tests/test_file.txt`.

## Solution

When `build_py` walks manifest entries, stop promoting files from nested directories that already contain Python sources into the parent package's data files. That keeps explicitly excluded subpackages out of the wheel instead of reintroducing them through `include_package_data`.

The existing regression now asserts the intended behavior directly instead of warning and xfail-ing.

## Verification

- `uv run --extra test --extra cover pytest setuptools/tests/test_build_py.py -q`
- `uvx ruff check setuptools/command/build_py.py setuptools/tests/test_build_py.py`
- `uvx ruff format --check setuptools/command/build_py.py setuptools/tests/test_build_py.py`
- Reproduced the issue's toy `mypkg/tests` wheel build on `main`, then reran it after the patch and confirmed the wheel no longer contains `mypkg/tests/*`.

### DIFF
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -6,7 +6,7 @@ import operator
 import os
 import stat
 import textwrap
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Mapping
 from functools import partial
 from glob import glob
 from pathlib import Path
@@ -357,21 +357,30 @@ def _find_packages_exclude_patterns(dist: Distribution) -> tuple[str, ...]:
     with pyproject.open("rb") as file:
         data = tomllib.load(file)
 
-    patterns = (
-        data.get("tool", {})
-        .get("setuptools", {})
-        .get("packages", {})
-        .get("find", {})
-        .get("exclude", [])
-    )
-    return tuple(patterns)
+    setuptools_cfg = data.get("tool", {}).get("setuptools", {})
+    packages = setuptools_cfg.get("packages", {})
+    if not isinstance(packages, Mapping):
+        return ()
+
+    find = packages.get("find", {})
+    if not isinstance(find, Mapping):
+        return ()
+
+    patterns = find.get("exclude", [])
+    if isinstance(patterns, str):
+        return (patterns,)
+    if isinstance(patterns, Iterable):
+        return tuple(patterns)
+    return ()
 
 
 def _split_find_excludes(patterns: str) -> tuple[str, ...]:
     return tuple(filter(None, (line.strip() for line in patterns.splitlines())))
 
 
-def _is_explicitly_excluded(parent: str, filename: str, excluded: tuple[str, ...]) -> bool:
+def _is_explicitly_excluded(
+    parent: str, filename: str, excluded: tuple[str, ...]
+) -> bool:
     if not excluded:
         return False
 
@@ -470,7 +479,9 @@ class _IncludePackageDataAbuse:
         parts = list(itertools.takewhile(str.isidentifier, pkg.parts))
         if parts:
             importable = ".".join([parent, *parts])
-            if any(fnmatch.fnmatchcase(importable, pattern) for pattern in self._excluded):
+            if any(
+                fnmatch.fnmatchcase(importable, pattern) for pattern in self._excluded
+            ):
                 return None
             return importable
         return None
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -15,6 +15,7 @@ from typing import Any
 from more_itertools import unique_everseen
 
 from .._path import StrPath, StrPathT
+from ..compat.py310 import tomllib
 from ..dist import Distribution
 from ..warnings import SetuptoolsDeprecationWarning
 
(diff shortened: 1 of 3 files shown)


setuptools/command/build_py.py
```diff
@@ -338,6 +332,36 @@ def assert_relative(path):
...
 
 
+def _contains_python_sources(directory: str) -> bool:
+    return any(
+        all(
+            part.isidentifier()
+            for part in candidate.relative_to(directory).with_suffix("").parts
+        )
+        for candidate in Path(directory).glob("**/*.py")
+    )
```
abravalheri, line 344: I am a bit concerned about the amount of nested iteration introduced by this change.

`_find_package_data_owner` is called for each file discovered by the `_filter_build_files` function.

In turn, `_find_package_data_owner` contains a `while` loop.

Inside this loop, `_contains_python_sources` is called.

Inside `_contains_python_sources`, we have another two levels of iteration.

It is fine that `while directory and directory != previous and directory not in src_dirs:` is essentially the same as in the original code, but the addition of these extra loops appears to increase the overall complexity, specially because `Path.glob("**/*.py")` may repeatly invoke recursive filesystem scans.

Maybe this will impact performance on larger source trees...

abravalheri (commented):
Hi @itscloud0, thank you very much for the contribution.

I think I understand the motivation behind the PR. However, I think we should add another test to cover what I consider to be an interesting edge case.


Would it be possible to add a test along the following lines?


```text
pyproject.toml
MANIFEST.in
src/
└── mypkg/
    ├── __init__.py
    ├── data.txt
    └── subpkg/
        ├── sample1.json
        └── sample2.json
```


with:

```toml
[tool.setuptools]
include-package-data = true

[tool.setuptools.packages.find]
where = ["src"]
exclude = ["mypkg.subpkg", "mypkg.subpkg.*"]
```

and a `MANIFEST.in` that includes the contents of `mypkg/subpkg` (for the sake of the `sdist`).

My understanding of the reasoning behind this PR is that an explicitly excluded nested package should not be reintroduced into the final wheel via `include-package-data = true`.

If that interpretation is correct, I wouldn't expect the files under `mypkg/subpkg/` to appear in the wheel.

I realise there are different views on whether directories without Python files should be considered packages. Personally, I don't make any differentiation (which is consistent with the Python interpreter behaviour). But, for this example in particular, the distinction should be irrelevant since the directory is explicitly excluded (so the user expectation would be for the directory to not be present in the final wheel).

tai1105997-alt (approved):

PR #5262, opened 2026-07-15
feat: Warn on mismatched typing classifiers and py.typed markers

### LINKED ISSUE #5243: [FR] warn on packaging a `py.typed` marker but missing the `Typing :: *` classifier

A typed project or partial stubs-only package will ship a `py.typed` marker as per https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information

But it's easy to forget to also add the appropriate Trove Classifier for discoverability. The setuptools build backend already outputs warnings in some scenarios (for example: deprecated classifiers). It feels well within scope to warn about incorrect or missing classifiers.

warn on packaging a `py.typed` marker but missing the `Typing :: *` classifier

In this case, having either `Typing :: Stubs Only` or `Typing :: Typed` should be enough to avoid the warning I'm suggesting.

Inversely, having the classifier in a package that doesn't ship a `py.typed` file OR ends in `-stubs` (the package module name, not the distribution) should also be a warning.

### Alternative Solutions

This is different than https://github.com/pypa/setuptools/issues/1021 as this isn't auto-inserting anything, just showing a non-blocker warning.

### Additional context

_No response_

### Code of Conduct

pctablet505:
Fixes #5243

Added warnings during `build_py` to detect mismatches between `py.typed` markers and `Typing :: *` classifiers. Packages shipping `py.typed` without a typing classifier, declaring `Typing :: Typed` without `py.typed`, or using a `-stubs` package without `Typing :: Stubs Only` will now see a warning.

### DIFF
--- a/newsfragments/5243.feature.rst
+++ b/newsfragments/5243.feature.rst
@@ -1,0 +1,1 @@
+Added warnings to detect mismatches between ``py.typed`` markers and ``Typing :: *`` classifiers during builds.
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -16,7 +16,7 @@ from more_itertools import unique_everseen
 
 from .._path import StrPath, StrPathT
 from ..dist import Distribution
-from ..warnings import SetuptoolsDeprecationWarning
+from ..warnings import SetuptoolsDeprecationWarning, SetuptoolsWarning
 
 import distutils.command.build_py as orig
 import distutils.errors
@@ -78,6 +78,7 @@ class build_py(orig.build_py):
         if self.packages:
             self.build_packages()
             self.build_package_data()
+            self._check_typing_classifier()
 
         # Only compile actual .py files, using our base class' idea of what our
         # output files are.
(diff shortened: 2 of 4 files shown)


Avasam (approved):

abravalheri:
Hi @pctablet505 thank you very much for taking the time to prepare a contribution, and sorry for the delay in getting back to you.

It has tests and is easy to follow, so overall well put together PR, thank you.

Now on the conceptual side: to be very honest, I am not very enthusiastic about the whole `classifiers` ceremony/verbosity/thingy[^2]... and I usually try to debloat as much as I can the `pyproject.toml`.

As a result, most of the time my approach for classifiers is YAGNI[^1]. I understand other people may have a different approach and use them for other things; which IMHO "it is ok". Every developer has their own workflow...

But that also leads me to the conclusion that I wouldn't like the backend complaining at me for something that is not an error that needs to be handled, especially if I am not a huge fan of the advised action...

For this particular scenario, wouldn't the best approach be an opt-in linter kind of thingy?

(This only reflects my opinion, other maintainers may have a different one).

---

@Avasam, sorry I didn't reply to your issue before, but this is more or less where I stand right now. The TL;DR is: it's a valid concern, and it is appropriate for some workflows, but I don't see it as "universal" for all workflows, and I personally wouldn't like setuptools to yell at me for this specific reason 😅.

[^1]: Which implies that I assume `py.typed` as authoritative and classifiers as non-essential.
[^2]: And that depending on what you write there you have to keep chasing a moving target...

Avasam:
> wouldn't the best approach be [a] linter kind of thingy?

Conceptually, yes. But an actual linter is gonna have a really hard time with the dynamic nature of Python build backends.
Hence my proposal at the build backend level. (or at the level of tooling that does the upload, that works too, just more annoying to get the error late, even worst as a warning that still uploads).
For setuptools specifically, it may be possible as a plugin?

But unless you have it out of the box, there's also the issue of adoption,then you have the issue of adoption. There's a catch-22 because lots of projects have incorrect metadata displayed on their project (due to incorrect classifiers), leading to them being underused anyway as they're seen as unreliable. And no good or popular tool to validate it before upload.

---

> I wouldn't like the backend complaining at me for something that is not an error that needs to be handled

I prefer having all the tools tell me when something isn't quite right (with the ability to shut if off). Doesn't have to be hard errors. But we're definitely in opinion territory 😅

---
Not gonna die on that hill. Just trying to suggest fixes for an annoyance I have with incorrect metadata at an ecosystem level.

abravalheri:
> Not gonna die on that hill. Just trying to suggest fixes for an annoyance I have with incorrect metadata at an ecosystem level.

Half jokingly: down with the classifiers 🔥 (and down with double bookkeeping) 😝

pctablet505:
Closing this PR per maintainer feedback from @abravalheri that classifier validation is best handled at the linter level rather than inside the build backend. Thank you for the review and discussion!

Issue #5265, opened 2026-07-18
Fix mingw bare `-O` optimization flag under `-m32`


jaraco:
Part of the compilers decoupling epic #5264 (Phase 1).

The mingw compiler injects a bare `-O` optimization flag. `cc1` accepts this for 64-bit builds, but under `-m32` it rejects it:

```
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
```

(reported in #4873)

**Fix:** in `distutils/compilers/C/cygwin.py`, normalize the bare `-O` to `-O1` (or suppress it when a numbered `-O` is already present in the flags). Add a regression test and a newsfragment. Ships independently of the rest of Phase 1 — good warm-up.

Closes #4873.

jaraco:
Fixed by pypa/distutils#415 (merged).

Issue #5264, opened 2026-07-17
[debt] Remove distutils hack


jaraco:
Although distutils_hack is deprecated (#4137), there are still consumers that rely on the presence of distutils or access to its internals to conduct some business (see the [distutils deprecation tag](https://github.com/pypa/setuptools/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22distutils%20deprecation%22)).

One aspect of this deprecation is to make the compilers module independent. Some work has been done on that by moving the compilers into `distutils.compilers` with the intention of eventually releasing that functionality as a third-party `compliers.C` package. Before that can happen, those modules need to be decoupled from distutils (and any interaction with distutils made a formal interface). There is some work already drafted (and even some merged) in https://github.com/pypa/distutils/tree/debt/compilers-deps.

Let's finish the work. Let's provide solutions for each of the distutils deprecation issues and decouple the compilers functionality. Some backward incompatibility may be acceptable, but we'll aim to remain compatible where feasible.

## Plan

End-state: `compilers.C` fully decoupled from distutils and extractable as a standalone distribution under `jaraco/compilers.C`. `compilers` stays a namespace package so future `compilers.<lang>` siblings can be added. Removal of the top-level `distutils` hack is **gated** on these replacements shipping.

The work is broken into the sub-issues tracked above. Grouped into phases:

### Phase 1 — Decouple `compilers.C` from distutils internals (pypa/distutils)
- #5265 — Fix mingw bare `-O` under `-m32` (closes #4873)
- #5266 — Inline logging (drop `distutils._log`)
- #5267 — Vendor small build/filesystem helpers
- #5268 — Replace version & platform-compat shims
- #5269 — Move to stdlib `sysconfig` + injection shim (investigation)
- #5270 — Consolidate compiler exceptions at `compilers.errors`

### Phase 2 — Public API (pypa/distutils)
- #5271 — Public API for `compilers.C` (resolves #2806, #4540)

### Phase 3 — Migration docs (pypa/setuptools)
- #5272 — Fill migration-guide gaps for ccompiler / new_compiler / get_platform (resolves doc half of #2806)

### Phase 4 — Retire the hack (pypa/setuptools; gated on Phases 1–3)
- #5273 — Retire the distutils hack (see also #3625, #3622)

### Extraction
- #5274 — Stand up standalone `jaraco/compilers.C` distribution

[...]

PR #5277, opened 2026-07-19
Fix mingw bare `-O` optimization flag under `-m32`

### LINKED ISSUE #4873: mingw64 win32 compilation issue - does not accept "-O"
### setuptools version

76.0

### Python version

Python 3.12 and 3.13

### OS

Windows 11

This works fine for 64 bit compiling (win_amd64).  

### Description

When using the compiler=mingw64 option, the option for "-O" optimization is added automatically.  When using the 32 bit compiling mode (-m32), the compiler demands -Ox, with x being an integer or alternate characters, and halts abruptly with an errro.  This is not a problem with 64 bit compiling, and I was able to add an extra -O2 option by tweaking build_ext.

### How to Reproduce

Create a project that calls mingw64 with the option -m32 to generate 32 bit code.  Would be easier to just change -O to -O1, which is equivalent.

### Other detail

gcc -shared -O -Wall -IC:\progra~2\python\python313\include -IC:\progra~2\python\python313\Include -c pyrho.c -o build\temp.win32-cpython-313\Release\pyrho.o "-O2 -m32"
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
error: command 'C:\\Program Files\\mingw64\\bin\\gcc.EXE' failed with exit code 1

dchaudhari7177:
## Summary

`MinGW32Compiler` injected a bare `-O` into `compiler`, `compiler_so`, `compiler_cxx` and `compiler_so_cxx`. GCC documents `-O` and `-O1` as equivalent, but `cc1` rejects the bare form when targeting 32-bit:

```
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
```

so any `-m32` build fails, while 64-bit builds are unaffected. This emits `-O1` instead.

Closes #4873. Part of the Phase 1 warm-up in #5264 (filed as #5265).

## Approach

Of the two options suggested in #5265 — normalise the bare `-O` to `-O1`, or suppress it when a numbered `-O` is already present — I went with normalising, since it is the smaller change and is exactly equivalent for the 64-bit path that works today. Suppression would additionally have to decide which flag wins, which is a behaviour change for existing 64-bit users.

Scope is limited to `MinGW32Compiler`, matching the issue. `Compiler` (cygwin) also passes a bare `-O` at lines 81-84 and would hit the same `cc1` rejection under `-m32` — happy to fold that in if you'd like it in the same PR.

## Tests

Updated `test_set_executables` for the new flag, and added `test_no_bare_optimization_flag`.

The pre-existing tests in `test_mingw.py` are all gated on `is_mingw()`, so they skip everywhere except an actual mingw Python — including CI. The new test is gated only on cygwin, following `test_customize_compiler_with_msvc_python`, which already constructs a `MinGW32Compiler` on a stock Windows build. That way the regression is actually exercised rather than skipped.

Revert-verified: with the `cygwin.py` change backed out, the new test fails with `assert '-O' not in ['gcc', '-O', '-Wall']`.

## Local results (Windows 11, Python 3.12)

`pytest setuptools/_distutils/compilers/`

| | passed | failed | skipped | errors |
|---|---|---|---|---|
| clean tree | 8 | 1 | 7 | 19 |
| this branch | 9 | 1 | 7 | 19 |

The failure and errors are all in `test_unix.py` and reproduce identically on a clean tree — they are Windows-only and unrelated.

`ruff check` passes on both changed files. `ruff format --check` reports both files as unformatted, but does so on a clean tree as well (it wants double quotes throughout `cygwin.py`, including lines this PR does not touch), so I have not reformatted them.

### DIFF
--- a/newsfragments/4873.bugfix.rst
+++ b/newsfragments/4873.bugfix.rst
@@ -1,0 +1,1 @@
+The mingw compiler now passes ``-O1`` instead of a bare ``-O``. The two are equivalent to GCC, but ``cc1`` rejected the bare form when building 32-bit extensions with ``-m32``. -- by :user:`dchaudhari7177`
--- a/setuptools/_distutils/compilers/C/cygwin.py
+++ b/setuptools/_distutils/compilers/C/cygwin.py
@@ -256,11 +256,14 @@ class MinGW32Compiler(Compiler):
         if is_cygwincc(self.cc):
             raise Error('Cygwin gcc cannot be used with --compiler=mingw32')
 
+        # Use ``-O1`` rather than a bare ``-O``. The two are equivalent to GCC,
+        # but ``cc1`` rejects the bare form under ``-m32``.
+        # https://github.com/pypa/setuptools/issues/4873
         self.set_executables(
-            compiler=f'{self.cc} -O -Wall',
-            compiler_so=f'{self.cc} -shared -O -Wall',
-            compiler_so_cxx=f'{self.cxx} -shared -O -Wall',
-            compiler_cxx=f'{self.cxx} -O -Wall',
+            compiler=f'{self.cc} -O1 -Wall',
+            compiler_so=f'{self.cc} -shared -O1 -Wall',
+            compiler_so_cxx=f'{self.cxx} -shared -O1 -Wall',
+            compiler_cxx=f'{self.cxx} -O1 -Wall',
             linker_exe=f'{self.cc}',
             linker_so=f'{self.linker_dll} {shared_option}',
             linker_exe_cxx=f'{self.cxx}',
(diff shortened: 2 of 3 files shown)


jaraco:
Thanks for this, @dchaudhari7177! setuptools' `_distutils` is vendored from [pypa/distutils](https://github.com/pypa/distutils), so the fix needs to land there to survive the next vendor sync. I've carried your commit upstream (preserving your authorship) and extended it to also cover the Cygwin compiler, which has the identical bare-`-O` bug: pypa/distutils#415.

Closing this in favor of that PR — you'll be credited in the history and changelog. Much appreciated!

PR #5281, opened 2026-07-21
fix: use -O1 instead of bare -O for MinGW/Cygwin compilers

### LINKED ISSUE #4873: mingw64 win32 compilation issue - does not accept "-O"

### LINKED ISSUE #5265: Fix mingw bare `-O` optimization flag under `-m32`

Solaris-star:
## Summary

MinGW defaults injected bare `-O`. Under `-m32`, `cc1` rejects that:

```text
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
```

64-bit builds accept bare `-O`; 32-bit does not (#4873).

## Fix

Use `-O1` in MinGW32 and Cygwin compiler default command lines (equivalent optimization level where bare `-O` was accepted).

## Tests

- Updated `test_mingw` expected flags to `-O1`
- Added `test_default_optimize_flag_is_o1` (always runs; mocks non-cygwin gcc)

## News

`newsfragments/5265.bugfix.rst`

## Linked Issues

Closes #5265  
Fixes #4873

### DIFF
--- a/newsfragments/5265.bugfix.rst
+++ b/newsfragments/5265.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed MinGW/Cygwin compilers injecting bare ``-O``, which ``cc1`` rejects under ``-m32``; use ``-O1`` instead.
--- a/setuptools/_distutils/compilers/C/cygwin.py
+++ b/setuptools/_distutils/compilers/C/cygwin.py
@@ -78,10 +78,10 @@ class Compiler(unix.Compiler):
         shared_option = "-shared"
 
         self.set_executables(
-            compiler=f'{self.cc} -mcygwin -O -Wall',
-            compiler_so=f'{self.cc} -mcygwin -mdll -O -Wall',
-            compiler_cxx=f'{self.cxx} -mcygwin -O -Wall',
-            compiler_so_cxx=f'{self.cxx} -mcygwin -mdll -O -Wall',
+            compiler=f'{self.cc} -mcygwin -O1 -Wall',
+            compiler_so=f'{self.cc} -mcygwin -mdll -O1 -Wall',
+            compiler_cxx=f'{self.cxx} -mcygwin -O1 -Wall',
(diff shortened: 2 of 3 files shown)


jaraco:
Thanks for the contrib. Changes to distutils need to happen in [pypa/distutils](/pypa/distutils). Fortunately, that work has happened in pypa/distutils#415 and should be merged into Setuptools soon.

PR #5299, opened 2026-08-11
fix(build_py): preserve executable bit on .py package data in wheels

### LINKED ISSUE #5296: [BUG] setuptools 84.0.0 no longer preserves executable bit on  package_data  files in wheels
### setuptools version

84.0.0

### Python version

Python 3.12, 3.14

### OS

RHEL8, Ubuntu 24.04

_No response_

### Description

### Expected behavior

### How to Reproduce

```bash
mkdir pkg
[build-system]
requires = ["setuptools"]

[project]
name = "mypkg"
version = "0.0.1"

mypkg = ["scripts/*"]
EOF

mkdir -p pkg/mypkg/scripts
touch pkg/mypkg/__init__.py

cd pkg
cd ..

```

### Output

```console
```

aryansk:
## Summary of changes

When a `.py` file is declared as package data, it is also picked up as a module and copied by `build_module` with `preserve_mode=False`. The subsequent `build_package_data` copy with `preserve_mode=True` is then skipped as up-to-date (equal mtimes after the ns-precision `os.utime` change), so the executable bit is dropped from `build/lib` and the built wheel.

This re-applies the source file's permission bits explicitly after the package-data copy, so scripts shipped as `.py` package data keep their `+x` bit in wheels.

Repro (before): `python -m build --wheel` on a package with `readme`-style `.py` script in `package-data` produced a wheel entry with mode `0o644`; after the fix it is `0o755`.

Closes #5296

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/newsfragments/5296.bugfix.rst
+++ b/newsfragments/5296.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``build_py`` dropping the executable bit on ``.py`` files included as package data in built wheels -- by :user:`aryansk`
--- a/setuptools/command/build_py.py
+++ b/setuptools/command/build_py.py
@@ -173,6 +173,11 @@ class build_py(orig.build_py):
             self.mkpath(os.path.dirname(target))
             _outf, _copied = self.copy_file(srcfile, target)
             make_writable(target)
+            # The copy may have been skipped as up-to-date (e.g. when the
+            # destination was already created by ``build_module`` with
+            # ``preserve_mode=False``), so re-apply the source's permission
+            # bits explicitly to honor the executable bit on package data.
+            os.chmod(target, stat.S_IMODE(os.stat(srcfile).st_mode))
 
     def analyze_manifest(self) -> None:
         self.manifest_files: dict[str, list[str]] = {}
--- a/setuptools/tests/test_build_py.py
+++ b/setuptools/tests/test_build_py.py
@@ -2,6 +2,7 @@ import os
 import shutil
 import stat
 import warnings
+import zipfile
 from pathlib import Path
 from typing import ClassVar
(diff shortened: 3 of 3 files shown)


CAOShurong (changes requested):
Confirmed the approach and the root cause — but on the current head (e71db13) the line ordering causes a small behavioral deviation from 83.0.0 for read-only sources.

Method: WSL/ext4; the reporter's reproducer; setuptools 83/84 unpacked from PyPI, isolated via PYTHONPATH; this PR's `setuptools/command/build_py.py` swapped in as-is; wheel entry modes read from `external_attr`.

| build | source mode 0o755 | source mode 0o555 |
|---|---|---|
| 83.0.0 | 0o755 | 0o755 |
| 84.0.0 | 0o644 (bug) | 0o644 (bug) |
| this PR head | 0o755 | **0o555** |

The regression itself is fixed for normal sources, and I traced the mechanism to match the PR's description: `build_module` stages the package-data `.py` with `preserve_mode=False` (755 -> 644), then the mode-restoring `build_package_data` copy is skipped by the up-to-date check, so the explicit chmod is the right minimal fix.

The remaining issue is only ordering: here `make_writable(target)` runs **before** `os.chmod(target, S_IMODE(src))`, so a read-only-executable source (r-xr-xr-x, e.g. from a read-only checkout) ends up 0o555 in the wheel, whereas 83.0.0 produced 0o755 (`make_writable` runs last there and adds the write bit). Swapping the two lines — chmod first, then `make_writable` — restores the 83 behavior exactly for all source modes.

(FYI for maintainers: #5302 implements the same fix with the opposite ordering; the two PRs are duplicates of each other.)

Issue #426, opened 2015-09-07
Many new test failures with 18.3


ghost:
Originally reported by: **felixonmars (Bitbucket: [felixonmars](http://bitbucket.org/felixonmars), GitHub: [felixonmars](http://github.com/felixonmars))**

---

I am getting the following new test errors for setuptoos 18.3. I have tried to run 18.2 tests with exactly the same environment, and none of them failed. Any hints?

Platform: linux -- Python 3.4.3 -- py-1.4.30 -- pytest-2.7.2

```
==================================== ERRORS ====================================
____________________ ERROR at setup of Test.test_bdist_egg _____________________
file /build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_bdist_egg.py, line 29
      def test_bdist_egg(self, setup_context, user_override):
        fixture 'user_override' not found
        available fixtures: tmpdir, setup_context, capfd, recwarn, pytestconfig, monkeypatch, capsys
        use 'py.test --fixtures [testpath]' for help on them.

/build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_bdist_egg.py:29
 ERROR at setup of TestUserInstallTest.test_user_install_not_implied_user_site_enabled
file /build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_easy_install.py, line 166
          @wraps(func)
          def patched(*args, **keywargs):
        fixture 'user_override' not found
        available fixtures: tmpdir, setup_context, capfd, recwarn, pytestconfig, foo_package, install_target, monkeypatch, distutils_package, capsys
        use 'py.test --fixtures [testpath]' for help on them.

/build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_easy_install.py:166
 ERROR at setup of TestUserInstallTest.test_user_install_not_implied_user_site_disabled
file /build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_easy_install.py, line 172
          @wraps(func)
          def patched(*args, **keywargs):
        fixture 'user_override' not found
        available fixtures: tmpdir, setup_context, capfd, recwarn, pytestconfig, foo_package, install_target, monkeypatch, distutils_package, capsys
        use 'py.test --fixtures [testpath]' for help on them.

/build/python-setuptools/src/setuptools-18.3/setuptools/tests/test_easy_install.py:172
_________ ERROR at setup of TestUserInstallTest.test_multiproc_atexit __________
[...]

PR #5306, opened 2026-08-23
Restore hashability of Extension objects (broken since 84.0.0 dataclass conversion)

### LINKED ISSUE #5303: [BUG] 84.0.0: Extension is unhashable (dataclass conversion sets __hash__ = None) — breaks setup scripts using sets of extensions
### setuptools version

84.0.0 (also reproduced against current `main` of the vendored distutils)

### Python version

3.13.13 (mechanism is version-independent; also verified on 3.12.10)

### OS

Windows 11, but the mechanism is platform-independent

### Description

Since 84.0.0, `setuptools.extension.Extension` (via the vendored distutils) is unhashable:

```python
>>> from setuptools.extension import Extension
>>> hash(Extension('x', ['y.c']))
TypeError: unhashable type: 'Extension'
>>> Extension.__hash__
None
```

Cause: the dataclass conversion of `distutils.extension.Extension` (pypa/distutils#373, vendored in 84.0.0) uses the default `eq=True`, which generates `__eq__` and therefore sets `__hash__ = None`. 80.9.0 and earlier are hashable (identity equality, inherited hash).

Setup scripts that collect extensions in sets or dict keys break outright. Real-world instance: every pywin32 CI build currently fails with `TypeError: unhashable type: 'WinExt_pythonwin'` at their `setup.py`'s `{ext for ext, why in self.excluded_extensions}` (their workaround: mhammond/pywin32#2787).

There is also a quieter semantic change for code that never hashes: two distinct `Extension` instances with equal field values now compare equal, so `ext in some_list` can be true for an object that is not in the list.

### Expected behavior

`Extension` keeps the identity-based equality and hashability it has always had.

### How to Reproduce

```console
$ pip install setuptools==84.0.0
$ python -c "from setuptools.extension import Extension; hash(Extension('x', ['y.c']))"
```

### Output

```
TypeError: unhashable type: 'Extension'
```

Fix proposed at the source: pypa/distutils#426 (issue) and pypa/distutils#427 (PR, `eq=False` plus a regression test pinning both hashability and identity semantics). Filing here as well since this is where affected users will search, and the fix reaches setuptools only through the next distutils vendor sync.

aryansk:
Fixes #5303

## Problem

Since v84.0.0, `Extension` instances are unhashable:

```python
>>> from setuptools.extension import Extension
>>> hash(Extension('x', ['y.c']))
TypeError: unhashable type: 'Extension'
```

The dataclass conversion of the vendored `distutils.extension.Extension` (pypa/distutils#373) used the default `eq=True`, which generates `__eq__` and therefore sets `__hash__ = None`. Beyond the `TypeError`, this also silently changed equality semantics: two distinct `Extension` instances with equal field values now compare equal, so `ext in some_list` can be true for an object that was never appended.

## Change

Pass `eq=False` to `lenient_dataclass` in `setuptools/_distutils/extension.py`, restoring the identity-based equality and hashability `Extension` always had before the conversion. The lenient unknown-keyword behaviour is unchanged.

Note: the same conversion exists upstream in pypa/distutils, so the vendored copy will need this fix re-applied at the next sync unless it lands there first — happy to send that PR if preferred.

## Validation

- Reproduced on clean main (`hash()` → `TypeError`, `Extension.__hash__ is None`) with CPython 3.13; after the change: hashable, distinct instances unequal, usable as set members/dict keys.
- Lenient behaviour preserved: unknown keyword arguments still emit the deprecation warning.
- Added a regression test (`test_extension_is_hashable`); `setuptools/_distutils/tests/test_extension.py`: 4 passed, 2 skipped (optional mypy inference checks).
- News fragment added (`newsfragments/5303.bugfix.rst`).


abravalheri:
Thank you for the contribution @aryansk .


Please have a look on https://setuptools.pypa.io/en/latest/development/developer-guide.html#code-conventions-and-other-practices regarding the `_distutils` folder.


There is already https://github.com/pypa/distutils/pull/427 pending of review in the `pypa/distutils` repo.

aryansk:
Thank you for the pointer @abravalheri — you're right that `_distutils` changes belong upstream. I see pypa/distutils#427 already covers this exact regression (#426), so closing this draft as superseded. Thanks for the quick review.

anhduc12050820565-web (approved):

PR #5307, opened 2026-08-24
Fix MSBuild property for VS2019+ Current layout and instance-derived install dir

### LINKED ISSUE #5275: [BUG] setuptools.msvc.EnvironmentInfo MSBuild property result incorrect on VS2017+
### setuptools version

83.0.0

### Python version

3.14

### OS

Fedora

_No response_

### Description

I tried using the MSBuild property using Visual Studio 2026 and it did not return the correct path(s).

### Expected behavior

Correct paths returned:
```
['C:\\Program Files\\Microsoft Visual Studio\\18\\Enterprise\\MSBuild\\Current\\bin', 'C:\\Program Files\\Microsoft Visual Studio\\18\\Enterprise\\MSBuild\\Current\\bin\\Roslyn']
```

### How to Reproduce

1. import setuptools
2. ei = setuptools.msvc.EnvironmentInfo('amd64')
3. ei.MSBuild

### Output

Returns the below which are completely wrong for Visual Studio 2026.
```console
'C:\\Program Files (x86)\\Microsoft Visual Studio 18.7\\MSBuild\\18.7\\bin', 'C:\\Program Files (x86)\\Microsoft Visual Studio 18.7\\MSBuild\\18.7\\bin\\Roslyn'
```

CAOShurong:
Fixes #5275

## Summary

The `EnvironmentInfo.MSBuild` property has two independent defects on VS2019+ machines:

1. **Wrong layout since VS2019.** For every `vs_ver >= 15.0` the property builds paths as `MSBuild\<vs_ver>\bin`, but since Visual Studio 2019 (v16) the on-disk layout is the version-independent `MSBuild\Current\bin`. The issue's reported VS2026 output (`MSBuild\18.7\bin`) and my local VS2022 output (`MSBuild\17.1\bin`) are both nonexistent directories produced by this branch.
2. **Lossy version formatting.** `vs_ver` is a `major.minor` float, so a real 17.12 install formats as `"17.1"`. That string is used both as the registry lookup key (`self.ri.lookup(self.ri.vs, f'{self.vs_ver:0.1f}')`) — which finds nothing — and in the default path. `VSInstallDir` therefore resolves to a directory that does not exist.

## Fix

- Use `MSBuild\Current\bin` (+ `Roslyn`) for `vs_ver >= 16.0`; keep the historical per-version layouts for VS2015 (`C:\Program Files (x86)\MSBuild\14.0\bin`, which never had the VS-directory prefix) and VS2017 (`MSBuild\15.x\bin`).
- Filter returned candidates to directories that actually exist when at least one candidate exists, so consumers no longer receive dead paths.
- Prefer the exact `installationPath` recorded by the Visual Studio installer instance state (`SystemInfo.known_vs_paths`, already parsed from `C:\ProgramData\Microsoft\VisualStudio\Packages\_Instances\*\state.json`) in `SystemInfo.VSInstallDir`, matching on major version. This avoids the lossy float formatting entirely.

## Validation

On this Windows 11 machine with VS2022 Community 17.12 installed:

```
before:
  VSInstallDir: C:\Program Files (x86)\Microsoft Visual Studio 17.1   # does not exist
  MSBuild: ['...\Microsoft Visual Studio 17.1\MSBuild\17.1\bin',          # does not exist
            '...\Microsoft Visual Studio 17.1\MSBuild\17.1\bin\Roslyn']  # does not exist

after:
  VSInstallDir: C:\Program Files\Microsoft Visual Studio\2022\Community
  MSBuild: ['C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\bin',
            'C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\bin\Roslyn']
```

Both post-fix paths contain `MSBuild.exe` / Roslyn binaries.

Tests: new `setuptools/tests/test_msvc_msbuild.py` (8 cases) covering VS2015/VS2017/VS2019/VS2022-multiline/VS2026 layouts, empty-result behavior for old versions, existence filtering, plus an opt-in real-environment check asserting every returned path exists when a genuine VS install is present. Red/green verified: 5 of the new cases fail on unpatched `main`, all 8 pass with the fix; existing `setuptools/_distutils/compilers/C/tests/test_msvc.py` suite still green (15 passed, 1 skipped).

Signed-off-by: Shurong Cao <35021@users.noreply.github.com>


Avasam:
Also done in https://github.com/pypa/setuptools/pull/5276 and https://github.com/pypa/setuptools/pull/5278

PR #5312, opened 2026-08-25
docs: point dependency_links deprecation page to modern direct URL dependencies

### LINKED ISSUE #4859: [Docs] dependency_links removal documentation
### Summary

Thank you for your work on this package. I'll admit I'm pretty new to the python ecosystem, so I apologize if I'm off base here.

I was reading the steuptools documentation of `dependency_links` of the latest version. At the top, it has a warning that "Dependency links support has been dropped by pip starting with version 19.0 (released 2019-01-22)." I was confused about what to do next, since I needed to specify a link to a GitHub repo. I would have found it useful if there were a link to the current documentation on how to achieve the same result with current versions of pip.

I found this link that had a bit more up to date information: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies. It doesn't line up with what I found at https://packaging.python.org/en/latest/specifications/dependency-specifiers/#examples, though, which uses a `zip` address. I'm not sure what the best edit would be.

By the way, the issue template here says there's a "View on GitHub" link on every documentation page, but I was unable to find that on pages on the documentation site.

### OS / Environment

Irrelevant.

### Additional Information

I think adding to these docs can help developers understand how to solve a problem with less digging and be more up to date with the specs.

### Code of Conduct

CAOShurong:
Fixes #4859.

## Summary

The [dependency_links deprecation page](https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html) warns that pip dropped support in 19.0 (2019), but then continues with ~80 lines of working-looking `setup()` / `setup.cfg` recipes without ever telling the reader what to do instead. This is exactly the confusion reported in #4859: a reader lands on this page needing "a link to a GitHub repo as a dependency" and leaves without a modern answer.

## Change

Docs-only, no code:

1. Extend the warning box to state explicitly that everything below is historical and **no longer works** with current pip/setuptools.
2. Point readers to the modern replacement: PEP 508 direct references (`name @ URL`) declared in `dependencies`, linking to the existing ["Direct URL dependencies"](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies) section of the user guide (`:doc:` reference + section name; `docs/conf.py` does not enable `autosectionlabel`, so a bare section anchor `:ref:` is not available).

## Verification

- Sphinx HTML build of the modified page in an isolated app passes clean (roles `:pep:`/`:doc:` resolve; no new warnings).
- The referenced target exists at `docs/userguide/dependency_management.rst` L212 ("Direct URL dependencies") and documents both `git+https://...` and wheel-URL forms.


Avasam:
Also done in https://github.com/pypa/setuptools/pull/5305

PR #5313, opened 2026-08-25
docs: add common MANIFEST.in pitfalls section (recursive-include patterns, anchored vs global excludes)

### LINKED ISSUE #4158: [Docs] Add more examples to MANIFEST.in explanation
### Summary

I'm migrating https://github.com/pypa/packaging.python.org/issues/989 since the relevant document has moved here with #4108.

Posted by @shi-yan:

"""
this is a poor document regarding manifest.in usage.

https://packaging.python.org/guides/using-manifest-in/

for example it doesn't contain any example for using recursive-include

```
recursive-include folder
recursive-include parent/folder
```

don't work at all.

it has to be

```
recursive-include folder *.html
```
"""

### OS / Environment

_No response_

### Additional Information

Examples should be added.

### Code of Conduct

CAOShurong:
Fixes #4158

## Summary

The ["Controlling files in the distribution"](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) guide documents the `MANIFEST.in` command table, but gives no worked example for `recursive-include`. That has repeatedly tripped users up (see #4158, migrated from the packaging docs in 2023):

1. A bare `recursive-include <dir>` with no file pattern does **not** mean "include everything under this directory" — it is a template error (`'recursive-include' expects <dir> <pattern1> <pattern2> ...`), and the build fails. The "include everything" tool is `graft`.
2. `include`/`exclude` patterns are matched against paths **relative to the project root** (anchored), while `global-include`/`global-exclude` match file names anywhere — so `exclude *.pyc` silently matches nothing when users expect a recursive exclusion.

## What this PR adds

A short **"Common MANIFEST.in pitfalls"** section after the existing src-layout example:

```bash
recursive-include folder *.html   # OK: .html files anywhere under folder
recursive-include folder          # ERROR: at least one pattern is required
graft folder                      # OK: everything under folder
```

plus two bullet points contrasting anchored vs global pattern matching.

Docs-only change to `docs/userguide/miscellaneous.rst`; no code touched.

## Verification

Every claim in the new text was exercised against the current implementation:

- `recursive-include folder *.html` → includes only `.html` files at any depth under `folder/`
- `recursive-include folder` → raises `DistutilsTemplateError: 'recursive-include' expects <dir> <pattern1> <pattern2> ...`
- `graft folder` + `global-exclude *.css` → `.css` removed from the tree
- `graft folder` + `exclude *.css` → `.css` kept (anchored exclude matches nothing), confirming the anchored-vs-global distinction

Full Sphinx HTML build of `docs/` succeeds and the new section renders correctly.


Avasam:
Existing alternative docs PR: https://github.com/pypa/setuptools/pull/5261

PR #5180, opened 2026-02-10
Update stale docs references to pkg_resources as active module

veeceey:
## Summary

Addresses #5179. Following the deprecation of `pkg_resources` completed in #5173, several non-historical documentation pages and docstrings still reference `pkg_resources` as though it is an actively maintained module. This PR updates those remaining references:

- **docs/setuptools.rst**: Rewrite the "Transitional Note" section to reflect that `pkg_resources`-style namespace packages are deprecated, with a note directing users to PEP 420 implicit namespaces
- **docs/development/index.rst**: Remove mention of `pkg_resources` module familiarity and reference manual (the reference manual no longer exists in the main docs)
- **docs/userguide/distribution.rst**: Replace `pkg_resources.parse_version()` with `packaging.version.parse()` and update examples to use PEP 440-compliant version strings
- **docs/userguide/extension.rst**: Reference `importlib.metadata` as the primary API for EGG-INFO metadata access, with `pkg_resources` as the predecessor
- **docs/userguide/package_discovery.rst**: Fix `pkg_resource` typo (missing `s`) in two places, fix `pkgutils` typo (should be `pkgutil`), and add a `.. deprecated::` admonition to the Legacy Namespace Packages section clarifying these are no longer supported
- **setuptools/dist.py**: Replace `pkg_resources.require()` format reference with PEP 508 in the `install_requires` docstring

## Test plan

- [x] Verify the RST renders correctly (no broken cross-references or formatting) - confirmed via ReadTheDocs preview build
- [x] Confirm all `pkg_resources` references in non-historical docs now either point to the modern replacement or are clearly marked as deprecated
- [x] No functional code changes -- documentation and docstring only

PR #5314, opened 2026-08-25
docs: update remaining non-historical docs assuming pkg_resources is active

### LINKED ISSUE #5179: [Docs[ Non-historical documentations that assume a still active `pkg_resources`
Here's additional non-deprecated, non-historical documentations that assume a still active `pkg_resources`:

https://github.com/pypa/setuptools/blob/main/docs/setuptools.rst#transitional-note
> such as needing to import all namespace
packages during the initialization of the ``pkg_resources`` runtime, and also
the need for ``pkg_resources`` to be explicitly imported before any namespace
packages work at all. 

This entire section could be moved to your new blurb about transitioning.

https://github.com/pypa/setuptools/blob/main/docs/development/index.rst#development-on-setuptools
> This document assumes the reader has some passing familiarity with
*using* setuptools, the ``pkg_resources`` module, and pip.
> [...]
> as all of these subjects are
covered at length in the setuptools developer's guide and the
``pkg_resources`` reference manual.

https://github.com/pypa/setuptools/blob/main/docs/references/keywords.rst#user-content-keyword-namespace-packages
> The ``namespace_packages`` implementation relies on ``pkg_resources``. However, ``pkg_resources`` has some undesirable behaviours, 

https://github.com/pypa/setuptools/blob/main/docs/userguide/distribution.rst#specifying-your-projects-version
> If you want to be certain that your chosen numbering scheme works the way
you think it will, you can use the ``pkg_resources.parse_version()`` function
to compare different version numbers::

https://github.com/pypa/setuptools/blob/main/docs/userguide/extension.rst#adding-new-egg-info-files
> Some extensible applications or frameworks may want to allow third parties to
develop plugins with application or framework-specific metadata included in
the plugins' EGG-INFO directory, for easy access via the ``pkg_resources``
metadata API. 

https://github.com/pypa/setuptools/blob/main/docs/userguide/package_discovery.rst#legacy-namespace-packages
> Historically, there were two methods to create namespace packages. [...]

The historical mention of setuptools + pkgutils can stay. But I'm not sure if it should still give the full example on using them. At least it should be clarified that this is not only deprecated, but no longer supported at all. Either way `pkg_resource` is a typo and missing `s`

https://github.com/pypa/setuptools/blob/main/docs/userguide/development_mode.rst#limitations Not sure about this one. Maybe it could be rephrased ?
> Editable installations may not work with
:doc:`namespaces created with pkgutil or pkg_resources
<PyPUG:guides/packaging-namespace-packages>`.
Please use :pep:`420`-style implicit namespaces [#namespaces]_.

https://github.com/pypa/setuptools/blob/main/docs/userguide/pyproject_config.rst#setuptools-specific-configuration 
> **Obsolete** - only relevant for ``pkg_resources``, [...]

https://github.com/pypa/setuptools/blob/main/setuptools/dist.py
> install_requires' -- a string or sequence of strings specifying project
versions that the distribution requires when installed, in the format
used by 'pkg_resources.require()'

---

Additional notes (not in scope of this PR): https://github.com/pypa/setuptools/pull/5007#issuecomment-3872401178

_Originally posted by @Avasam in https://github.com/pypa/setuptools/issues/5173#issuecomment-3872304733_

CAOShurong:
Fixes #5179

## Summary

`pkg_resources` was removed from the codebase (in the `Remove pkg_resources` change), but several non-historical, non-deprecated documentation pages still describe it as an available module — some even in the present tense. This PR updates those remaining references.

## What changed

| File | Change |
|------|--------|
| `docs/setuptools.rst` | The pkg_resources "TRANSITIONAL NOTE" (about automatic `declare_namespace()`) described future removal; that removal has since happened. Marked the section historical with a deprecation banner pointing out that both the auto-declaration behaviour and `pkg_resources` itself have been removed. |
| `docs/development/index.rst` | Dropped "passing familiarity with the `pkg_resources` module" and the `pkg_resources` reference manual from the reader-assumptions paragraph. |
| `docs/references/keywords.rst` | `namespace_packages` warning said it "relies on `pkg_resources`" and that Setuptools "intends to obviate its usage in the future" — now states plainly that the keyword is deprecated and **no longer functional** because the `pkg_resources` runtime was removed. |
| `docs/userguide/distribution.rst` | The version-comparison example imported `pkg_resources.parse_version`. Replaced with `packaging.version.parse` from the `packaging` library (which setuptools itself depends on). All example outputs re-verified against a live interpreter; note `2.1-rc2` became `2.1rc2` because PEP 440 normalization rejects the hyphen form, and the invalid legacy example `0.6a9dev-r41475` (raises `InvalidVersion` under `packaging`) was dropped. |
| `docs/userguide/extension.rst` | EGG-INFO plugin metadata "for easy access via the `pkg_resources` metadata API" → now points at `importlib.metadata`. |
| `docs/userguide/package_discovery.rst` | Legacy namespace-packages section: fixed the `pkg_resource`/`pkgutils` typos noted in #5179; the `pkg_resources` style is no longer presented as "the method setuptools directly supports" but as removed (historical reference only); the intro now says support has been removed rather than merely deprecated. |
| `docs/userguide/development_mode.rst` | "Editable installations **may not work** with namespaces created with pkgutil or pkg_resources" → "do not work". |
| `docs/userguide/pyproject_config.rst` | `zip-safe`/`eager-resources` no longer cite the removed `pkg_resources`; `namespace-packages` marked deprecated and no longer functional. |

Deliberately untouched: everything under `docs/deprecated/`, changelog/history files, and source-code docstrings (the latter are tracked separately per the notes in #5007).

## Verification

- Full Sphinx HTML build of `docs/` passes; every touched page renders with the new wording (checked generated HTML for `keywords.html`, `package_discovery.html`, `distribution.html`, `extension.html`, `setuptools.html`). The only build warnings are pre-existing on `main`: the towncrier config lookup fails outside a towncrier-enabled checkout layout.
- `git diff --check` clean.
- Example outputs verified live:
  ```
  >>> parse("1.9.a.dev") == parse("1.9a0dev")
  True
  >>> parse("2.1rc2") < parse("2.1")
  True
  ```

## Newsfragment

`newsfragments/5179.doc.rst` added.


Avasam:
Also done in https://github.com/pypa/setuptools/pull/5180

PR #5316, opened 2026-08-27
fix: tolerate None __file__ in _distutils_hack frame detector

### LINKED ISSUE #5263: [BUG] Running pip 26.1.2 failed after installing legacy versions of setuptools
### setuptools version

setuptools<68.0.0

### Python version

Python 3.14.6

### OS

Windows

_No response_

### Description

I'm migrating my python 3.10 environment to 3.14.6. Some packages require a legacy version of setuptools, so I ran pip install "setuptools<68.0.0". After that, typing pip directly inside the terminal gave a AttributeError: 'NoneType' object has no attribute 'endswith'. setuptools 83.0.0 72.0.0 69.0.0 68.0.0 works fine. But setuptools==65.0.0 67.0.0 67.8.0 doesn't work, so the issue seems specific to <68.0.0.

### Expected behavior

pip should show the help message

### How to Reproduce

1. pip install "setuptools<68.0.0"
2. pip (however python -m pip ran as expected)

### Output

```console
$ pip install "setuptools<68.0.0"
Collecting setuptools<68.0.0
  Using cached setuptools-67.8.0-py3-none-any.whl.metadata (6.2 kB)
Using cached setuptools-67.8.0-py3-none-any.whl (1.1 MB)
    Found existing installation: setuptools 68.0.0
    Uninstalling setuptools-68.0.0:
      Successfully uninstalled setuptools-68.0.0
Successfully installed setuptools-67.8.0
$ pip
  File "%appdata%\Local\Python\bin\pip.exe.__script__.py", line 33, in <module>
    from pip._internal.cli.main import main
  File "<frozen importlib._bootstrap>", line 1371, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1333, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1267, in _find_spec
  File "%appdata%\Local\Python\pythoncore-3.14-64\Lib\site-packages\_distutils_hack\__init__.py", line 97, in find_spec
    return method()
  File "%appdata%\Local\Python\pythoncore-3.14-64\Lib\site-packages\_distutils_hack\__init__.py", line 145, in spec_for_pip
    if self.pip_imported_during_build():
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "%appdata%\Local\Python\pythoncore-3.14-64\Lib\site-packages\_distutils_hack\__init__.py", line 157, in pip_imported_during_build
    return any(
           ^^^
  File "%appdata%\Local\Python\pythoncore-3.14-64\Lib\site-packages\_distutils_hack\__init__.py", line 158, in <genexpr>
    cls.frame_file_is_setup(frame) for frame, line in traceback.walk_stack(None)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "%appdata%\Local\Python\pythoncore-3.14-64\Lib\site-packages\_distutils_hack\__init__.py", line 167, in frame_file_is_setup
    return frame.f_globals.get('__file__', '').endswith('setup.py')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'endswith'
$ python -m pip
Usage:
  %appdata%\Local\Python\pythoncore-3.14-64\python.exe -m pip <command> [options]

Commands:
  install                     Install packages.
  lock                        Generate a lock file.
  download                    Download packages.
  uninstall                   Uninstall packages.
  freeze                      Output installed packages in requirements format.
  inspect                     Inspect the python environment.
  list                        List installed packages.
  show                        Show information about installed packages.
  check                       Verify installed packages have compatible dependencies.
  config                      Manage local and global configuration.
  search                      Search PyPI for packages.
  cache                       Inspect and manage pip's wheel cache.
  index                       Inspect information available from package indexes.
  wheel                       Build wheels from your requirements.
  hash                        Compute hashes of package archives.
  completion                  A helper command used for command completion.
  debug                       Show information useful for debugging.
  help                        Show help for commands.

General Options:
  -h, --help                  Show help.
  --debug                     Let unhandled exceptions propagate outside the main subroutine, instead of logging them
                              to stderr.
  --isolated                  Run pip in an isolated mode, ignoring environment variables and user configuration.
  --require-virtualenv        Allow pip to only run in a virtual environment; exit with an error otherwise.
  --python <python>           Run pip with the specified Python interpreter.
  -v, --verbose               Give more output. Option is additive, and can be used up to 3 times.
  -V, --version               Show version and exit.
  -q, --quiet                 Give less output. Option is additive, and can be used up to 3 times (corresponding to
                              WARNING, ERROR, and CRITICAL logging levels).
  --log <path>                Path to a verbose appending log.
  --no-input                  Disable prompting for input.
  --keyring-provider <keyring_provider>
                              Enable the credential lookup via the keyring library if user input is allowed. Specify
                              which mechanism to use [auto, disabled, import, subprocess]. (default: auto)
  --proxy <proxy>             Specify a proxy in the form scheme://[user:passwd@]proxy.server:port.
  --retries <retries>         Maximum attempts to establish a new HTTP connection. (default: 5)
  --timeout <sec>             Set the socket timeout (default 15 seconds).
  --exists-action <action>    Default action when a path already exists: (s)witch, (i)gnore, (w)ipe, (b)ackup,
                              (a)bort.
  --trusted-host <hostname>   Mark this host or host:port pair as trusted, even though it does not have valid or any
                              HTTPS.
  --cert <path>               Path to PEM-encoded CA certificate bundle. If provided, overrides the default. See 'SSL
                              Certificate Verification' in pip documentation for more information.
  --client-cert <path>        Path to SSL client certificate, a single file containing the private key and the
                              certificate in PEM format.
  --cache-dir <dir>           Store the cache data in <dir>.
  --no-cache-dir              Disable the cache.
  --disable-pip-version-check
                              Don't periodically check PyPI to determine whether a new version of pip is available for
                              download. Implied with --no-index.
  --no-color                  Suppress colored output.
  --use-feature <feature>     Enable new functionality, that may be backward incompatible.
  --use-deprecated <feature>  Enable deprecated functionality, that will be removed in the future.
  --resume-retries <resume_retries>
                              Maximum attempts to resume or restart an incomplete download. (default: 5)
```

wahajahmed010:
## Summary

`_distutils_hack.DistutilsMetaFinder.frame_file_is_setup` can crash during pip startup on Python 3.14+ when a frame's `f_globals` carries a `__file__` set to `None`. The previous implementation used `frame.f_globals.get('__file__', '').endswith('setup.py')`, which only handles the *missing-key* case; when the key is present and the value is `None`, the `.endswith` call raises `AttributeError: 'NoneType' object has no attribute 'endswith'`. This is the same root cause behind the report in #5263: with a legacy setuptools installed, `pip` itself becomes unusable because the shim can't be imported.

The fix treats `None` and other non-string values the same as a missing `__file__` (i.e. "not a setup.py frame"). The frame walker in `pip_imported_during_build` therefore completes without raising, and the rest of `_distutils_hack` continues to work as before.

## How to reproduce (without my fix)

```python
class Frame:
    @property
    def f_globals(self):
        return {"__file__": None}

# On current main:
from _distutils_hack import DistutilsMetaFinder
DistutilsMetaFinder.frame_file_is_setup(Frame())
# AttributeError: 'NoneType' object has no attribute 'endswith'
```

With this branch loaded instead, the same call returns `False` cleanly.

## Pull Request Checklist
- [x] Changes have tests (added `test_frame_file_is_setup_handles_missing_file` in `setuptools/tests/test_distutils_adoption.py`, covering missing key, `None`, a real `setup.py` path, and a non-matching path)
- [x] News fragment added in `newsfragments/5263.bugfix.rst`

Closes #5263


abravalheri:
Hi @wahajahmed010 thank you very much for the contribution.

Would you be able to provide a real world reproducer that does not rely on a mock/stub for us to assess this issue?

You mentioned changes on Python 3.14+... Do you have a reference for that? I checked the release notes and I couldn't find any specific changes in `__file__` that allow it to be `None`, the [docs](https://docs.python.org/3/reference/datamodel.html#module.__file__) still say it has to be a string when present.

PR #5320, opened 2026-08-30
fix: skip distutils shim on Python 3.6 to avoid SSL errors (closes #2728)

### LINKED ISSUE #2728: [BUG] v57.1.0 on Python v3.6 SSL Errors

botbikamordehai2-sketch:
## What

The bundled `distutils-precedence.pth` shim in setuptools 57.1.0 causes SSL certificate verification errors on Python 3.6 on macOS and Windows when running `python setup.py install`. The issue is that the shim replaces the standard library `distutils` with the vendored copy, and on Python 3.6 that copy fails to load CA certificates correctly in some environments.

## Fix

If Python version is below 3.7 (i.e., 3.6), do not install the `.pth` shim. This restores the previous behavior for Python 3.6 users and avoids the SSL regression.

Closes #2728


setup.py
```diff
@@ -39,7 +41,12 @@ class install_with_pth(install):
 
     def initialize_options(self):
         install.initialize_options(self)
-        self.extra_path = self._pth_name, self._pth_contents
+        # Disable the distutils shim on Python 3.6 to avoid SSL regressions
+        # (https://github.com/pypa/setuptools/issues/2728)
+        if sys.version_info < (3, 7):
```
Avasam, line 46: This is dead code. Python 3.6 has long not been supported by setuptools

PR #5331, opened 2026-09-18
fix(archive_util): extract tars through PEP 706 data_filter

### LINKED ISSUE #5328: archive_util bypasses tarfile's PEP 706 extraction filters by calling private APIs
`setuptools.archive_util.unpack_tarfile` extracts through three private `tarfile` APIs, which means none of the hardening CPython added in PEP 706 ever runs.

```python
# unpack_tarfile
tarobj._extract_member(member, final_dst)

# _iter_open_tar
tar_obj.chown = lambda *args: None

# _resolve_tar_file_or_dir
tar_member_obj = tar_obj._getmember(linkpath)
```

### Why it matters

`_extract_member` is the internal worker that `extractall` calls *after* the extraction filter has vetted and rewritten the member. Calling it directly skips the filter entirely, so setuptools gets none of `data_filter`'s protections: absolute and upward link targets, device and FIFO members, setuid/setgid/sticky bits, and permission handling. `tar_obj.chown = lambda *args: None` is a hand-rolled substitute for one small piece of what the filter already does properly.

The containment check added in #5325 re-implements one of those protections by hand. That was the right scope for a security fix, but it means we now maintain our own partial copy of logic the stdlib maintains fully, and the remaining gaps stay open.

Being private API is a liability in its own right: signature or behavior can change in any release with no deprecation, and `_getmember` in particular has no stability guarantee.

### The complication

`unpack_tarfile` cannot simply become `tarobj.extractall(extract_dir, filter='data')`. The documented `progress_filter` callback takes `(source_path, dest_path)` and may return a *different* destination or `None` to skip, so callers can redirect individual members anywhere they like — including, legitimately, outside the nominal extraction directory. That is why #5325 applies its containment check to the preliminary destination, before `progress_filter` runs.

Reconciling the two is the actual work here. Rough options:

- Run each member through `tarfile.data_filter(member, extract_dir)` for validation and to pick up its sanitized `TarInfo`, then extract to the possibly-redirected destination. Keeps `progress_filter` semantics; still uses `_extract_member` for the write.
- Extract to a temporary directory with `extractall(filter='data')`, then move members into place according to `progress_filter`. Fully public API, at the cost of an extra copy and of changing when the callback is invoked relative to extraction.
- Reduce the `progress_filter` contract to filtering only, not redirection — a breaking change, but it would let the stdlib own extraction entirely.

`tarfile.data_filter` is available on all currently supported Pythons (`requires-python = ">=3.10"`) via the 3.10.12 / 3.11.4 security backports, though the exact minimum patch level is worth confirming before relying on it unconditionally.

Split out of #5325, where this was noted but deliberately left alone as out of scope for the security fix.

DawnofGenX:
## Problem

`unpack_archive`/`unpack_tarfile` extracted tar members through private `tarfile` APIs (`TarFile._getmember`, `TarFile._extract_member`) with a monkeypatched `chown` stub (`archive_util.py`, current `main`), bypassing the PEP 706 extraction filters that CPython added in 3.12 and backported to 3.10.12/3.11.4.

Consequences on interpreters without the filters (the gap #5328 calls out: 3.10.0–3.10.11, 3.11.0–3.11.3):

- **setuid/setgid/sticky bits survive extraction** — a regular member with mode `0o4755` is written to disk with `0o4755` intact.
- ownership stripping is done by stubbing `chown` rather than by the filter's null-uid/gid mechanism, and special-file handling depends on hand-rolled logic (`_resolve_tar_file_or_dir` + `LookupError`).

## Change (the issue's option 1, scoped)

Route every member through `tarfile.data_filter` (public, PEP 706) before `_extract_member`:

- permission bits sanitized, uid/gid nulled → `TarFile.chown` becomes a `-1/-1` no-op, so the monkeypatch is gone;
- absolute/escaping link targets and special files raise `FilterError` and the member is skipped, matching `extractall`'s per-member error isolation;
- no private `tarfile` API remains in the hot path. The old `_getmember(linkpath)` lookup is replaced by a last-match name index — the private lookup scanned members bottom-to-top, so a repeated name resolved to the *last* declaration, which is exactly what a dict built in iteration order keeps.

For interpreters where stdlib `tarfile` has no `data_filter`, the module falls back to the **vendored `backports.tarfile`** setuptools already ships (`setuptools/_vendor/backports/`, which provides `data_filter`/`FilterError`). Import is guarded by `hasattr`, so modern interpreters are untouched.

`progress_filter` semantics are unchanged: it still runs on the preliminary destination and may redirect/skip; containment is validated against `extract_dir` *before* the callback, matching #5325's stance ("the filter callback is trusted for what it returns"), so no behavior change for wheel-installing callers.

## Tests

- Existing suite: `test_archive_util.py` — 23 passed (the 1 xpassed existed on `main` before this change).
- 4 new regression tests: setuid stripping, special-file skipping, forward-symlink target resolution (guards the `_getmember` replacement), and progress-filter redirection still honored.
- **Regression proof:** reverting only `archive_util.py` (keeping tests) fails exactly `test_unpack_tarfile_strips_setuid_bits` — the setuid byte survives pre-fix and is sanitized post-fix.
- The fallback branch (vendored backport) was exercised directly: full `_iter_open_tar` drive against `backports.tarfile` on a mixed archive (setuid, fifo, forward symlink) → fifo skipped, setuid sanitized to `0o755`, forward symlink resolved and extracted as a regular file copy.
- `test_wheel.py` (26 passed), `test_distutils_adoption.py`, `test_dist_info.py` green; `ruff check`/`format` (0.15.x per `.pre-commit-config.yaml`) clean.

Closes #5328


DawnofGenX:
@mergifyio queue

sylvesterkaczmarek (commented):
Catching tarfile.FilterError here turns an unsafe archive member into a silent skip. With the stdlib data filter, those errors are fatal by default; silently continuing can make extraction/install succeed with files missing. I think this should propagate/translate the filter error instead, with an escaping-link or special-file regression.

DawnofGenX:
@sylvesterkaczmarek Good catch — fixed in 64d5b80. The `tarfile.FilterError` catch is gone: filter rejections are now translated to `UnsafeMember` (a `DistutilsError`, not an `UnrecognizedFormat`, so `unpack_archive` can't fall through to another driver) with a message naming the rejected member, e.g. `'evil-link' was rejected by the tar data filter: 'evil-link' is a link to an absolute path`. Extraction aborts loudly instead of succeeding with files missing.

Also added regression tests: a symlink member with an escaping (`../escaped`) and an absolute (`/tmp/escaped`) link target, and a FIFO member — each now asserts `pytest.raises(UnsafeMember)`. Members that merely fail link resolution but pass the filter (dangling-but-safe links) keep the historical skip behavior. The newsfragment wording is updated to match.

sylvesterkaczmarek (approved):
Rechecked 64d5b808. tarfile.FilterError is no longer silently swallowed: unsafe members are translated to UnsafeMember so unpack_archive cannot fall through and extraction fails loudly. The escaping/absolute symlink and FIFO regressions cover the failure mode I raised. My concern is resolved.

